PR #1, opened 2026-05-16
chore: gitignore .hal0ai/ dev runtime + commit uv.lock

thinmintdev:
## Summary
- `.hal0ai/` is the local HAL0_HOME override (etc/, var/, .venv/) — runtime state and a virtualenv
- `uv.lock` pins the resolved dep tree so CI and contributors install identical versions

## Open question
The `.hal0ai/` gitignore line was reverted locally — opening as draft so you can decide whether to keep it. The `uv.lock` half is the unambiguous part.

## Test plan
- [ ] CI lint + format green
- [ ] `uv sync` works against committed lockfile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ yarn-error.log*
 /etc-local/
 /usr-local/
 hal0-home/
+.hal0ai/
 
 # Secrets
 .env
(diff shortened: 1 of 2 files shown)


thinmintdev:
Rescoping: `.gitignore` half was superseded by 85aa3b3. Landing the `uv.lock` half directly on main.

PR #126, opened 2026-05-22
fix(toolbox-vulkan): install curl in runtime so HEALTHCHECK works

thinmintdev:
## Summary

- Add `curl` to the runtime stage of `packaging/toolbox/vulkan.Dockerfile`.
- The existing `HEALTHCHECK` directive runs `curl -fsS http://127.0.0.1:\$PORT/v1/models`, but the runtime image only installed `libcurl4` (the library), not the CLI binary.
- Result before this fix: every container reports `unhealthy` forever with `/bin/sh: 1: curl: not found` while `llama-server` itself is serving fine. Caught while bringing the b9279 rebuild up on the primary slot post-#124.

Pre-existing Dockerfile bug — not introduced by #124, just newly visible.

## Test plan

- [x] `apt-cache show curl` confirms ~1MB image-size impact, no other deps.
- [ ] CI's `toolbox.yml` rebuilds + publishes vulkan with the fix (same chain as #124).
- [ ] Once published, on the primary slot host: `docker inspect hal0-slot-primary --format '{{.State.Health.Status}}'` reports `healthy` after one healthcheck interval.

### DIFF
--- a/packaging/toolbox/vulkan.Dockerfile
+++ b/packaging/toolbox/vulkan.Dockerfile
@@ -102,6 +102,11 @@ ARG DEBIAN_FRONTEND=noninteractive
 #                         explicitly so the runtime contract is obvious)
 # - libgomp1            : OpenMP runtime — llama.cpp threads
 # - libcurl4            : runtime side of LLAMA_CURL=ON
+# - curl                : the CLI binary — needed by HEALTHCHECK below.
+#                         libcurl4 alone gives us the library; without
+#                         this the container reports `unhealthy` forever
+#                         ("/bin/sh: 1: curl: not found") even when the
+#                         server is serving fine.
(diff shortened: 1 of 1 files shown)


thinmintdev:
Closing — superseded by ADR-0006 (Lemonade migration). The vulkan toolbox layer this PR fixes is being retired in v0.2: the six per-modality toolboxes are replaced by a single Lemonade Server instance (`docs/internal/adr/0006-migrate-inference-to-lemonade.md`). Spending a 15-min toolbox rebuild + manifest-digest refresh cycle on a layer that's getting torn out is the wrong place for the polish budget right now.

If we end up needing one more polished cut on the current toolbox stack (e.g. a stability point release while Lemonade lands), this branch is trivial to revive — it's a single line in `packaging/toolbox/vulkan.Dockerfile` adding `curl` to the runtime APT install.

PR #250, opened 2026-05-23
fix(ui-v3): normalize slot shape in useSlots hook

thinmintdev:
## Root cause

Real `/api/slots` returns slot rows **without a `metrics` field**. The v3 React components (SnapshotStrip, SlotCard, PersonaPicker, slot-modals) read `s.metrics.toks` / `metrics.mem.toFixed()` / `metrics.precision` unguarded, throwing TypeError on first paint and unmounting the React tree — the flash-then-crash users hit on Dashboard + Slots after auth was disabled.

Earlier fix (PR #N — sed sweep of per-call defensive guards) corrupted `dashboard.jsx` and was reverted in d67f7ea. This lands the proper fix at the **hook layer**.

## Summary

- **`useSlots`**: adds `normalizeSlot()` adapter applied to list, detail, and status-union responses. Every metric field defaults to a safe value matching the `HAL0_DATA` mock contract in `ui/src/dash/data.jsx` (numbers → 0, nullable → null, strings → "").
- **`useModels`**: adds `normalizeModel()`. Defaults `labels: []` — `models.jsx:174` does `model.labels.map(...)` unguarded.
- **`useBackends`**: adds `normalizeBackend()`. Defaults `usedBy: []` + `state: 'unavailable'`.
- **`useLemonade`**: already defensive (`loaded: []` default); no change needed.
- **apiMock fixture**: the `**/api/**` catch-all glob was matching Vite source modules like `/src/api/hooks/useSlots.ts` and fulfilling them with `{}` (application/json), so the React tree never mounted under e2e. Tightened to absolute `/api/` + `/v1/` pathnames only. Unblocked the entire Phase B2 suite as a side effect.
- **New spec**: `ui/tests/e2e/specs/slots-empty-metrics.spec.ts` boots Dashboard + Slots against a backend that returns slots WITHOUT a `metrics` field; asserts no pageerror fires.

## Anti-scope respected

- Component markup unchanged (fix is at the hook layer)
- HAL0_DATA mock structure unchanged (`VITE_MOCK_LEMONADE=1` path unaffected — normalizers are pass-through on full-shape input)
- No new endpoints
- No other slice work touched

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm run build` clean
- [x] `npx playwright test` → 23 passed / 7 skipped (MCP + agent specs pre-skipped at source)
- [x] New `slots-empty-metrics` spec green: Dashboard + Slots both render with metrics-less slot rows, no JS errors
- [ ] Manual smoke against real backend with `VITE_MOCK_LEMONADE=0` (deferred — backend not on this worktree's host)

### DIFF
--- a/ui/src/api/hooks/useBackends.ts
+++ b/ui/src/api/hooks/useBackends.ts
@@ -28,17 +28,46 @@ export interface LemonadeSelf {
 const POLL_MS = 30_000
 const SNAPSHOT_POLL_MS = 5_000
 
+/**
+ * Normalize a backend row. Defaults `usedBy: []` and `state: 'unavailable'`
+ * so views that iterate `b.usedBy` or branch on state don't crash on
+ * partial backend data.
+ */
+export function normalizeBackend(raw: any): Backend {
+  const src = (raw && typeof raw === 'object') ? raw : {}
+  return {
+    id: typeof src.id === 'string' ? src.id : (typeof src.name === 'string' ? src.name : ''),
+    version: typeof src.version === 'string' ? src.version : (typeof src.ver === 'string' ? src.ver : ''),
+    state: typeof src.state === 'string' ? src.state : 'unavailable',
+    usedBy: Array.isArray(src.usedBy) ? src.usedBy.filter((x: any) => typeof x === 'string') : [],
+    recommended: !!src.recommended,
+    note: typeof src.note === 'string' ? src.note : undefined,
+    kind: typeof src.kind === 'string' ? src.kind : undefined,
+    device: typeof src.device === 'string' ? src.device : undefined,
+  }
+}
+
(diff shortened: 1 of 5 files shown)


thinmintdev:
Superseded by #249 — same fix (normalizeSlot helper applied in fetchSlotsUnion + useSlotDetail) landed first via a different branch. Closing as duplicate.

Cross-session collision: both this PR and #249 implement the deferred metrics-normalizer called out in the v3 cutover handoff. #249 has the slightly different implementation (explicit DEFAULT_METRICS const, model fallback to model_id/model_default) but the user-visible effect is the same — slots/hardware/dashboard no longer crash on partial backend payloads.

Apologies for the redundant work.

Issue #251, opened 2026-05-23
Lemonade: expose queue depth + coresident rollup via /v1/health


thinmintdev:
Follow-up from #221.

Lemonade Server v10.6.0 does not surface either of:

- `queued` — number of in-flight requests waiting on a busy backend.
- `coresident` — whether the NPU is currently holding the FLM trio
  resident (stt + embed + chat coresident pack).

Today both fields are inferred client-side or hardcoded. #221 wires the
UI to honor `null` (chips hide) so the dashboard never lies, but it
leaves the underlying signal absent.

## Options

1. **Patch upstream Lemonade** to add a `queued` integer + `coresident`
   boolean to `GET /v1/health`. Upstream PR.
2. **Compute server-side in hal0**: roll up `queued` from per-backend
   `/slots` (llama.cpp exposes pending-request count), and `coresident`
   from `useSlots` (any NPU slot serving + >=2 of {chat,stt,embed}
   ready on device=npu).

Option 2 is the path of least coordination - Lemonade releases ~weekly
and this would be opt-in anyway. Track here so hal0-api work has a
home.

Verified against `/v1/health` on hal0 LXC 10.6.0:

```json
{
  "all_models_loaded": [],
  "max_models": {"llm": 4, "embedding": 4, "image": 4, "reranking": 4, "transcription": 4, "tts": 4},
  "model_loaded": null,
  "status": "ok",
  "version": "10.6.0",
  "websocket_port": 9001
}
```

No queue field. `/v1/stats` is per-request perf (TTFT, tok/s) - also
no queue.

## Definition of done

- `/v1/health` (or `/api/health` rollup) returns numeric `queued` +
  boolean `coresident`.
- Footer + sidebar chips reappear (already wired in #221, gated on
  `!= null` / truthy).
- Doc update in the relevant API doc.

thinmintdev:
Obsolete. The lemond daemon this `/v1/health` extension targets was removed in the container-switchover epic #687 (`rg lemonade|lemond src/` = 0 hits). The footer 'queued'/'coresident' chips were deliberately removed with #221 (closed), the surviving RuntimeRollup omits both fields, and SchedulerCard is permanently gated ('lemond dispatcher is stateless; no real source'). Nothing actionable remains.

— _Verified during the v0.8.2b3 codebase assessment sweep (adversarial re-check against HEAD). Reopen if I missed a live sub-item._

PR #252, opened 2026-05-23
fix(ui-v3): footer chips honor backend null (closes #221)

### LINKED ISSUE #221: footer: queued chip hardcoded to 0 + NPU coresident derived from heuristic
**Surface:** Persistent footer bar (visible on every page) — the two right-side status chips.

**File:** `ui/src/components/Footer.vue:132` (queued), `:134-139` (npuCoresident).

**Current:**
- `queued` is `computed(() => 0)` with the comment "Lemonade has no queue endpoint yet (v0.2.x)" — the chip never appears because the value is always 0, but if Lemonade ever stalls behind a queue depth the operator gets no visual signal.
- `npuCoresident` is derived from `loadedCount >= 2` — a heuristic, not a real backend field. It will flash true any time two non-NPU models are loaded together.

**Real signal:**
- Extend `/v1/health` (or hal0-api's `/api/lemonade/state`) to return real `queued` depth.
- Add a real `coresident` boolean to the slot snapshot (or derive from per-slot `device == 'npu'` rolled up in hal0-api), and feed it to the chip instead of the count heuristic.

**Related:** #174 (Extras umbrella), #145 (metrics aggregator)

_Found by dashboard mock-data audit 2026-05-23._

thinmintdev:
Closes #221. Follow-up issue #251 filed for the backend side.

## Summary

Issue #221 flagged two footer chrome regressions in the v3 React dash:

1. The footer **queued** chip read `L.queued` from `useLemondRollup`,
   but the hook hardcoded `queued: 0` because Lemonade does not expose
   a queue depth on `/v1/health` (verified directly against hal0 LXC
   10.6.0).
2. The footer + sidebar **coresident** chips were literal strings — no
   backend signal behind them either.

Result: both chips lied about steady-state.

## Decision: Option A (hide on null), with mock continuity

The issue spec gave Options A (hide) or B (wire to backend). Backend
check confirmed Lemonade ships nothing usable — `/v1/health` returns
`status/version/loaded/max_models/websocket_port`, full stop; `/v1/stats`
is per-request perf. So Option A for now, with the backend follow-up
tracked at #251 (computing the rollup server-side in hal0-api is the
likely route, since Lemonade upstream releases weekly and would need
patching anyway).

## Changes

- **`ui/src/api/hooks/useLemonade.ts`** — `LemonadeHealth` now carries
  explicit `queued: number | null` and `coresident: boolean | null`.
  The hook coerces missing fields to `null` (not `0`/`true`) so callers
  can disambiguate "not surfaced" from "really zero". `useLemondRollup`
  forwards both.
- **`ui/src/dash/chrome.jsx`** — footer queued chip guards on
  `L.queued != null`; footer + sidebar coresident chips guard on
  `L.coresident` being truthy. Hidden when backend doesn't say.
- **`ui/src/api/mock.ts`** — `buildHealth()` round-trips
  `HAL0_DATA.lemond.{queued,coresident}` so the MOCK_LEMONADE-forced
  dev/demo build keeps showing the chips (anti-scope: demo continuity
  preserved without changing HAL0_DATA itself).
- **`ui/tests/e2e/fixtures/apiMock.ts`** — bonus drive-by: the
  `**/api/**` and `**/v1/**` catch-all globs were matching Vite's
  `/src/api/client.ts` module imports during E2E, breaking React mount
  on every spec on this branch. Re-anchored the patterns on the URL
  origin via regex (`/^https?:\/\/[^/]+\/api\//`). Restored 22 specs
  that were red on the base commit.
- **`ui/tests/e2e/specs/footer-chips-v3.spec.ts`** — new spec, 2 tests:
  chips render with default HAL0_DATA values; chips hide when the
  `queued`/`coresident` fields are clobbered to `undefined` via
  `addInitScript`.

## Test plan

- [x] `npm install` clean
- [x] `npm run build` clean (113 modules, 474 KB)
- [x] `npx tsc --noEmit` clean
- [x] `npm run test:e2e` — 23 passed / 0 failed / 7 skipped (was 0
      passed / 23 failed on base due to the fixture glob bug)
- [x] `ruff format --check && ruff check` — clean (no python touched)
- [x] Verified against live Lemonade on hal0 LXC:
      `curl 127.0.0.1:13305/api/v1/health` returns no queue/coresident
      fields; production build hides both chips

## Out of scope

- `mcp-main.jsx` has its own hardcoded "coresident" footer chip backed
  by `HAL0_DATA.lemond` (not the rollup hook). It's a v0.3 mock page,
  not the v3 live footer; left alone per the issue's anti-scope.
- HAL0_DATA mock keeps `queued: 0` + `coresident: true` so demo
  screenshots stay consistent.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/ui/src/api/hooks/useLemonade.ts
+++ b/ui/src/api/hooks/useLemonade.ts
@@ -22,6 +22,12 @@ export interface LemonadeHealth {
   max_loaded: number | null
   version: string | null
   throughput_mbps: number | null
+  // Lemonade /v1/health does not currently surface a queue depth or a
+  // coresident rollup (see hal0_lemonade_v1_load_schema). The hook
+  // preserves them as `null` so the UI can distinguish "not surfaced"
+  // from "really zero" and hide chips that have no real signal yet.
+  queued: number | null
+  coresident: boolean | null
 }
 
 export interface LemonadeStats {
@@ -47,6 +53,9 @@ export function useLemonadeHealth(): UseQueryResult<LemonadeHealth> {
         version: typeof body?.version === 'string' ? body.version : null,
         throughput_mbps:
           typeof body?.throughput_mbps === 'number' ? body.throughput_mbps : null,
+        // Explicit null (NOT 0) when the field is missing — see #221.
+        queued: typeof body?.queued === 'number' ? body.queued : null,
+        coresident: typeof body?.coresident === 'boolean' ? body.coresident : null,
       }
     },
     refetchInterval: POLL_HEALTH_MS,
@@ -86,8 +95,11 @@ export function useLemondRollup() {
     loaded: h?.loaded.length ?? 0,
     budget: h?.max_loaded ?? 4,
     throughput: h?.throughput_mbps ?? null,
-    queued: 0,
-    coresident: true,
+    // queued + coresident are null until Lemonade exposes them on
+    // /v1/health (#221, follow-up filed). Chips that consume these
+    // guard with `!= null` so they hide instead of lying with 0.
+    queued: h?.queued ?? null,
+    coresident: h?.coresident ?? null,
     lastTtft: stats.data?.time_to_first_token ?? null,
     lastTokPerSec: stats.data?.tokens_per_second ?? null,
   }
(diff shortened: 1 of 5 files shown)


thinmintdev:
Rebased onto `main` (was `feat/dash-v3-react` — retired when #235 merged v3 to main). Single commit replayed cleanly, no conflicts. Re-targeted base to `main`.

PR #366, opened 2026-05-28
fix(memory): REST /api/memory/add honors private:* namespace

### LINKED ISSUE #317: memory_add normalizes dataset to "shared" — private:* namespaces silently flattened
## Repro

```bash
# Add into a namespaced dataset.
curl -X POST http://127.0.0.1:8080/api/memory/add \
  -H "Content-Type: application/json" \
  -H "X-hal0-Agent: hermes-agent" \
  -H "X-hal0-Private: 1" \
  -d '{"text":"probe","tags":["probe"],"dataset":"private:hermes-agent"}'
# → {"id":"...","timestamp":"..."}

# Search the same dataset.
curl -X POST http://127.0.0.1:8080/api/memory/search \
  -H "Content-Type: application/json" \
  -H "X-hal0-Agent: hermes-agent" \
  -H "X-hal0-Private: 1" \
  -d '{"query":"probe","dataset":"private:hermes-agent"}'
# → {"items":[]}

# List the same dataset.
curl "http://127.0.0.1:8080/api/memory/list?dataset=private:hermes-agent"
# → {"items":[],"next_cursor":null}

# But list "shared" returns the item.
curl "http://127.0.0.1:8080/api/memory/list?dataset=shared"
# → {"items":[{"dataset":"shared",...,"text":"probe",...}]}
```

The REST shim in `src/hal0/api/routes/memory.py` passes `dataset` to `CogneeWrapper.add`. Somewhere downstream the namespace is dropped — items land in `"shared"` regardless. Search + list against the original namespace then return empty.

## Why it matters

- `hal0 agent bootstrap hermes` smoke `memory_roundtrip` test always fails after a clean install — operator can't tell whether the wiring is broken or this bug is.
- Per-agent private memory (ADR-0011 §4) is a no-op until this is fixed: every agent's "private" writes leak into the global `shared` bucket.
- `X-hal0-Private: 1` header is ignored as a write-side enforcement.

## Likely fix location

Inside `CogneeWrapper.add` or whatever it delegates to. Either the colon in `private:hermes-agent` is being stripped, or the wrapper has a hardcoded `"shared"` somewhere.

## Tracked because

Surfaced while landing #316 (hermes_provision regression repair). The `memory_roundtrip` smoke remained red after every other smoke went green — fixing this would close that last bootstrap smoke.

### LINKED ISSUE #367: CogneeWrapper collapses cross-client private:* writes on singleton wrapper
## Surfaced by

#317 / PR #366. REST `/api/memory/add` was fixed to resolve `private:<agent>` from `X-hal0-Agent` + `X-hal0-Private` headers and pass that dataset to the wrapper. The route is correct; the wrapper layer still collapses.

## Repro (post #366)

```bash
# REST writes resolve to private:hermes-agent.
curl -X POST http://127.0.0.1:8080/api/memory/add \
  -H 'X-hal0-Agent: hermes-agent' -H 'X-hal0-Private: 1' \
  -H 'Content-Type: application/json' \
  -d '{"text":"probe"}'

# But the actual Cognee dataset is still `shared` — `CogneeWrapper._effective_write_dataset`
# (src/hal0/memory/cognee_wrapper.py:869) collapses `private:*` to `shared` on a
# non-private wrapper. The singleton wrapper at app.state.memory_wrapper is created
# with client_id="anonymous", private_mode=False.
```

## Root cause

`CogneeWrapper` was designed for one-instance-per-(client_id, private_mode) — see `tests/memory/test_cognee_wrapper.py::test_private_writes_invisible_to_other_clients` which constructs `alice` + `bob` wrappers. The wrapper guards against a non-private instance writing to `private:<other>` so a wrong-client wrapper can't smuggle data.

But production runs ONE wrapper (`app.state.memory_wrapper`, anonymous, non-private). The REST and MCP layers are the identity gate. The wrapper's collapse guard is now defending against itself.

## Options (order: most-proper → blocker fix → spike)

a) Reshape `CogneeWrapper` so a single instance can serve multiple clients: take `client_id` + `private` per-call (audit log + scope filter both read from the per-call args, not the constructor). Touches every callsite + the audit log shape but is the truthful model.

b) Keep one wrapper but pass a `trusted_caller=True` kwarg to `add` / `search` / `list_items` from REST + MCP, signaling that the caller has already enforced identity. Wrapper-level collapse is skipped for trusted callers. Smaller diff, preserves the per-client wrapper code path for tests.

c) Stand up a per-request wrapper from a pool keyed by `client_id`. Simplest mental model but multiplies the Cognee-init cost.

## Acceptance

- Smoke `curl POST /api/memory/add` with private headers + `curl POST /api/memory/search` with same headers returns the just-written item.
- `hal0 agent bootstrap hermes` `memory_roundtrip` smoke passes.
- Existing `tests/memory/` private-isolation contract (Alice can't see Bob's private) holds.

## Tracked because

Closing the bootstrap smoke loop from #317 needs both halves. PR #366 ships the REST half so the route is no longer the bug, but the LXC smoke acceptance from #317 cannot succeed until this wrapper-side change lands too.

thinmintdev:
Closes #317, #367.

## Summary

This PR now closes both halves of the private-namespace bug in one merge:

- **#317 (REST contract)** — the REST shims under `/api/memory/{add,search,list}` honor `X-hal0-Agent` + `X-hal0-Private` headers and route through the shared `hal0.memory.namespace` resolver so they cannot drift from the MCP dispatcher again.
- **#367 (wrapper collapse)** — the singleton `CogneeWrapper` (`client_id="anonymous"`, `private_mode=False`) no longer flattens `private:<x>` writes to `shared`. The REST + MCP layers own identity resolution; the wrapper persists the resolved dataset string verbatim. Without this half, merging #317 alone created a regression window where the route promised private and the wrapper silently leaked to the shared bucket.

Folded the review's medium findings (path traversal, `private:` prefix in agent header, body `dataset="private:other"` from non-private callers) into the same change so the security envelope ships atomically.

## Commits

1. `refactor(memory)`: extract `_resolve_dataset` into `hal0.memory.namespace`; rewire MCP dispatcher. Pure refactor.
2. `fix(memory)`: REST handlers consume identity headers + route through the shared resolver; `source` injection rejected.
3. `fix(memory)`: wrapper-level private namespace + identity hardening (this turn). Closes #367 and addresses the medium findings from review.

## Changes by file

- `src/hal0/memory/cognee_wrapper.py` — `_effective_write_dataset` on a non-private wrapper persists `private:<x>` verbatim (was: collapsed to `shared`). Private-mode wrapper unchanged.
- `src/hal0/memory/namespace.py` — `resolve_write_dataset` raises `MemoryNamespaceError` on body `dataset="private:..."` when `private=False`. Non-private callers cannot address the private namespace by name.
- `src/hal0/api/routes/memory.py` — `_agent_id` validates `X-hal0-Agent` against `^[a-zA-Z0-9_-]{1,64}$` and rejects values starting with `private:`. New typed error `MemoryAgentIdInvalid` (code `memory.agent_id_invalid`, 400).

## Test plan

- [x] `tests/api/test_memory_rest_routes.py` — 19 cases (was 13). New: path-traversal in agent header, `private:` prefix in agent header, colon variant in agent header, body `private:other` without `X-hal0-Private`, `/list` private without agent header, cross-client read leakage.
- [x] `tests/memory/test_namespace_wrapper.py` (new) — fast (non-slow) unit + stubbed-cognee coverage of `_effective_write_dataset` across both postures; pins that `cognee.add` receives the resolved dataset string, not `"shared"`.
- [x] `tests/mcp/test_memory.py` still green (16/16) — the MCP `_resolve_dataset` shim re-raises the new `MemoryNamespaceError` as `MemorySchemaError`, no test churn.
- [x] Non-slow `tests/api/ tests/mcp/ tests/memory/` sweep — `478 passed, 3 skipped` locally.
- [x] `ruff check src tests` — clean.
- [x] `ruff format --check src tests` — clean.
- [ ] LXC smoke against a live `/api/memory` once merged — verifies the wrapper-level write path end-to-end (the `hal0 agent bootstrap hermes` `memory_roundtrip` smoke is the acceptance trigger from #317).

## Follow-ups

- An integration test that runs against a real `CogneeWrapper` (not the StubWrapper) would close the LXC-smoke gap in CI; the natural home is the bootstrap-smoke roundtrip and it's worth its own issue when that surface is next touched.

### DIFF
--- a/src/hal0/api/routes/memory.py
+++ b/src/hal0/api/routes/memory.py
@@ -25,8 +25,59 @@ from pydantic import ValidationError
 from hal0.api.middleware.error_codes import Hal0Error
 from hal0.config.loader import load_hal0_config, save_hal0_config
 from hal0.config.schema import GraphUpstreamConfig, MemoryGraphConfig
+from hal0.memory.namespace import (
+    DEFAULT_DATASET,
+    MemoryNamespaceError,
+    resolve_read_datasets,
+    resolve_write_dataset,
+)
 
 router = APIRouter()
+
+
+# ── ADR-0012 identity + ADR-0005 §3 namespace helpers ─────────────────────
+#
+# Post-ADR-0012 hal0-api is open on 0.0.0.0:8080; agent identity flows on
+# the ``X-hal0-Agent`` header (NOT Bearer — auth surface was removed).
+# Private-mode opt-in flows on ``X-hal0-Private`` to match the MCP mount
+# (:mod:`hal0.api.mcp_mount`); the same toggle gates the same namespace
+# promotion rule across both surfaces (issue #317).
+
+
+_AGENT_HEADER = "x-hal0-agent"
+_PRIVATE_HEADER = "x-hal0-private"
+_TRUTHY = frozenset({"1", "true", "yes", "on"})
+
+
+class MemoryNamespaceInvalid(Hal0Error):
+    """The caller's headers + body produced an unresolvable namespace.
+
+    Distinct from a body-shape error so the dashboard can paint a
+    different toast ("you asked for private without an agent identity")
+    vs a generic 400.
+    """
+
+    code = "memory.namespace_invalid"
+    status = 400
+
+
+def _agent_id(request: Request) -> str:
+    """Return the ``X-hal0-Agent`` value, or ``"anonymous"`` if absent.
+
+    Mirrors :func:`hal0.api.mcp_mount.client_id_resolver` for the REST
+    surface — both translate the absence of an identity header into the
+    same sentinel so audit + dataset resolution stay consistent.
+    """
+    raw = request.headers.get(_AGENT_HEADER)
+    if not raw:
+        return "anonymous"
+    return raw.strip() or "anonymous"
+
+
+def _is_private(request: Request) -> bool:
+    """Return whether the caller opted into ``--private`` mode."""
+    raw = request.headers.get(_PRIVATE_HEADER, "")
+    return raw.strip().lower() in _TRUTHY
 
 
 class MemoryGraphConfigInvalid(Hal0Error):
@@ -188,11 +239,22 @@ async def update_graph_config(request: Request) -> dict[str, Any]:
 
 @router.post("/add")
 async def memory_add(request: Request) -> dict[str, Any]:
-    """Add a memory item. Body: ``{text, dataset?, tags?, source?, metadata?}``.
+    """Add a memory item. Body: ``{text, dataset?, tags?, metadata?}``.
 
-    Returns ``{id, timestamp}`` from :meth:`CogneeWrapper.add`. The
-    ``dataset`` defaults to ``"shared"``; tags + metadata are
-    free-form lists / dict respectively.
+    Identity headers (issue #317):
+
+      - ``X-hal0-Agent``: post-ADR-0012 agent identity. Stamped onto
+        the wrapper's ``source`` field — server-injected so callers
+        cannot lie (ADR-0005 §5). Absent header → ``"anonymous"``.
+      - ``X-hal0-Private: 1``: opt into the private namespace.
+        Promotes ``dataset`` to ``private:<agent>`` regardless of the
+        body value (ADR-0005 §3).
+
+    The body's ``source`` field is REJECTED — clients supplying it is
+    treated as an attempt to impersonate, matching the MCP rule. Use
+    the ``X-hal0-Agent`` header to claim identity.
+
+    Returns ``{id, timestamp}`` from :meth:`CogneeWrapper.add`.
     """
     body = await _read_json_body(request)
     text = body.get("text")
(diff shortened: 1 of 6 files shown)


thinmintdev:
Folded the request-changes feedback into this same PR rather than splitting into #367 — merging the two halves separately would create a regression window where the REST contract promises private but the wrapper still collapses to shared.

What's in this turn (commit 47b451e):

1. **#367 (wrapper collapse)** — `CogneeWrapper._effective_write_dataset` on the non-private singleton wrapper now persists `private:<x>` verbatim instead of folding to `shared`. The REST/MCP layer owns identity resolution and stamps the resolved dataset; the wrapper's collapse guard was defending against itself given the production wrapper lives at `app.state.memory_wrapper` with `client_id="anonymous"`, `private_mode=False`.
2. **Medium finding — path traversal in `X-hal0-Agent`** — `_agent_id` now validates against `^[a-zA-Z0-9_-]{1,64}$` and returns 400 `memory.agent_id_invalid` on hostile values (e.g. `../etc/passwd`).
3. **Medium finding — `private:` prefix in `X-hal0-Agent`** — rejected before reaching the wrapper to prevent `private:private:<x>` double-prefix. The toggle (`X-hal0-Private: 1`) is the only path to the private namespace.
4. **Medium finding — body `dataset="private:other"` from non-private caller** — `resolve_write_dataset` raises `MemoryNamespaceError` → 400 `memory.namespace_invalid`. Symmetric with the header rule.

Test coverage added (6 new on top of the existing 13):

- `tests/api/test_memory_rest_routes.py`: path-traversal, `private:` prefix in agent header, colon variant, body `private:other` without toggle, `/list` private without agent header, cross-client read-leakage.
- `tests/memory/test_namespace_wrapper.py` (new file, non-slow): unit coverage of `_effective_write_dataset` across both postures + stubbed-cognee assertion that `cognee.add` receives the resolved dataset string.

Verification:
- `ruff check src tests` — clean.
- `ruff format --check src tests` — clean.
- `pytest tests/api/ tests/mcp/ tests/memory/ -m "not slow"` — 478 passed / 3 skipped.

Ready for another look.

PR #418, opened 2026-06-02
fix(agents): probe /api/status not /api/health for hermes readiness

thinmintdev:
## Problem

Hermes **0.14.0** gates every `/api/*` route behind an ephemeral per-process session token (dashboard `auth_middleware`), except routes on its `_PUBLIC_API_PATHS` allowlist. `/api/health` is **not** on that list, so the agent shim readiness probe received **401**, never saw a 2xx, never emitted `READY=1`, and systemd hit the start-timeout → **hermes agent restart loop**.

## Fix

Probe `/api/status` instead — it **is** on the public allowlist and returns 200 with the version/home/config-state body, which is exactly the liveness signal the shim needs.

- `src/hal0/cli/agent_shim.py`: `status_url` → `/api/status` (+ explanatory comment, + updated wait-for-ready comment)
- `tests/cli/test_agent_shim.py`: assertion updated
- `.gitignore`: ignore `.serena/` (Serena MCP local tooling, not source)

## Verification

- `pytest tests/cli/test_agent_shim.py` → **36 passed**
- `ruff check` + `ruff format --check` clean
- Already running live on LXC 105 (editable install); `hal0-agent@hermes.service` is `active (running)`, `NRestarts=0`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,6 @@ tests/release-gate-report.json
 # OS
 Thumbs.db
 .claude/worktrees/
+
+# Serena MCP project tooling (local index/cache, not source)
+.serena/
--- a/src/hal0/cli/agent_shim.py
+++ b/src/hal0/cli/agent_shim.py
@@ -93,11 +93,17 @@ class AgentConfig:
     def status_url(self) -> str:
         """URL the shim polls to confirm the agent's HTTP surface is up."""
 
-        # ``/api/health`` is upstream-stable across hermes ≥ 0.10 and is the
-        # cheapest endpoint we can hit to determine "ready". The Hermes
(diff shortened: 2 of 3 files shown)


thinmintdev:
Superseded — the unauthenticated-readiness-probe fix already landed on `main`. `src/hal0/cli/agent_shim.py` now probes `/health` (200, no auth) instead of `/api/health` (401 behind Hermes 0.14.0's session-token auth_middleware), with the crash-loop rationale documented inline at agent_shim.py:98-102. Same change shipped via the Hermes home-normalization lineage (commit 6e27a55). Closing as stale/redundant; the branch is also CONFLICTING against current main.

PR #469, opened 2026-06-04
feat(gateway): Bifrost normalization shim for Hermes → lemond

thinmintdev:
## What

A standalone, Hermes-first **LLM gateway** (`gateway/`) built on [Bifrost](https://github.com/maximhq/bifrost) that sits between hal0's agents and lemond. On every chat request it:

1. **Rewrites the model** to lemond's live primary slot (resolved from `/api/v1/health`, TTL-cached) — callers use one stable virtual name `lemonade/primary`, killing Lemonade's strict-name 404 tool-loops.
2. **Injects `chat_template_kwargs.enable_thinking=false`** (unless the caller opted in) — local reasoning models stop timing out.

Replaces hal0's per-client patches (e.g. the FLM wrap-patch) with one chokepoint. Mirrors DreamServer's LiteLLM role, but as a single fast Go binary — and, unlike DreamServer (one fixed GGUF), resolves the target **dynamically** because hal0 swaps its primary slot at runtime.

## Why a plugin (not a static alias)

Bifrost's `models:["*"]` passes the name through; it doesn't rewrite like LiteLLM's `"*"`. The dynamic-slot rewrite therefore has to live in a `PreLLMHook`. Confirmed against `bifrost/core v1.5.16` source that the openai/custom provider merges `ExtraParams` onto the wire (`MergeExtraParamsIntoJSON`), so the thinking flag actually reaches lemond.

## Validated locally on hal0-dev (no CT105 contact)

- ✅ `normalize/` pure logic — 5 unit tests green (model rewrite, thinking injection, lemond-health LLM-slot parse).
- ✅ ABI-matched `make build` via `go.work` over pinned `core/v1.5.16`: `bifrost-http` (167M) + `hal0_normalize.so` (29M).
- ✅ Live load: `bifrost-http` reports **`plugin status: hal0-normalize - active`**, binds `:8079`, no version-skew panic. `compat` (drop-params) plugin active by default.

## Held (draft) — needs CT105

- [ ] Live chat smoke against a loaded lemond (`scripts/smoke.sh`).
- [ ] `make deploy` to CT105 (binary/.so/config/unit; does **not** touch Hermes).
- [ ] Hermes cutover: `config.yaml` OpenRouter → `lemonade/primary` (Tier-2 behavior change).

## Files

`normalize/` (tested core) · `plugin/main.go` (Bifrost `.so` adapter + cached resolver) · `config/config.json` · `systemd/hal0-bifrost.service` (loopback `:8079`, `User=hal0`) · `Makefile` (ABI-matched build + held deploy) · `scripts/smoke.sh`.


thinmintdev:
## Session update — root-caused both blockers, fixes landed; one new blocker found

Research → plan → fix complete for the two validation blockers. Filed issues #470 #471 #472 #474.

### ✅ Issue #470 — Bifrost dropped `chat_template_kwargs` (FIXED, deployed)
Root cause (confirmed in pinned `core/v1.5.16`): Bifrost only merges `Params.ExtraParams` into the wire body when ctx flag `BifrostContextKeyPassthroughExtraParams == true` (`core/providers/utils/utils.go:1159`); the openai-base custom provider never sets it (vllm/sgl do). The plugin now sets it itself (`gateway/plugin/main.go`). Proven live via the equivalent header pre-fix. Commit 51ec84f.

### ✅ Issue #471 — resolver prefers iGPU chat slot + cache fallback (FIXED, deployed)
`ParseLoadedLLM` now de-prioritizes NPU/FLM slots (by `-FLM` convention + optional device/recipe/backend hints), keeps FLM as fallback for NPU-only setups; `resolvePrimary` serves last-known-good on `/health` failure (1.5s fast-fail) instead of passing through to a lemond 404. normalize/ tests 9/9 green. Deployed to CT105; bifrost reloads clean (`hal0-normalize - active`, :8079).

### ✅ Issue #472 — lemond control-plane wedge: two findings
- **(a) lemond backlog is unfixable in lemonade**: cpp-httplib `CPPHTTPLIB_LISTEN_BACKLOG=5` + 8-thread pool are compile-time; no flag/env/config. No control-plane/inference listener separation. → must be solved client-side + resolver hardening (#471).
- **(b) Hermes/Hindsight leak (premise corrected): FIXED.** It was NOT hal0-api in-process (deployed hal0 is `main`, Cognee, extraction OFF). It was the bundled Hindsight daemon in the Hermes venv respawn-churning + killed uncleanly (HF-token `PermissionError` + 5s SIGTERM racing 30s graceful shutdown). Fixed on CT105: `HF_TOKEN_PATH` → writable tokenless, kill-grace 35s (idempotent ExecStartPre reapply), `HINDSIGHT_API_LLM_TIMEOUT=25`, all via the systemd drop-in (hermes.env gets re-rendered). Hindsight side now: zero daemons/sockets/zombies.

### 🚧 Issue #474 — NEW blocker: hal0-api ↔ lemond control-plane wedge (pre-existing, was masked)
With Hindsight clean, lemond's control plane STILL wedges: `/health` times out, ~25 ownerless FIN-WAIT-2 corpses vs backlog 5, live ESTAB owned by `hal0 serve`. hal0-api polls the control plane and abandons connections when lemond is slow (35B serialized, `--parallel 1`), tying up lemond's 8 worker threads. Self-sustaining until lemond restart. Lives in deployed `main` (separate PR).

### State
- Gateway: fixed + deployed; bifrost up on :8079 with the new plugin. **Hermes still on OpenRouter — cutover NOT done (gated on a clean live smoke).**
- Live smoke pending: blocked by #474 + the 35B was evicted by the lemond restart needed during #472 recovery (only 4B FLM loaded now).

thinmintdev:
Superseded by in-hal0-api normalization (PR #483) + the chat_template_kwargs thinking fix (PR #484). Hermes now points directly at the hal0-api gateway (127.0.0.1:8080/v1, model hal0/primary) — the Bifrost shim (:8079) is retired (`systemctl disable --now hal0-bifrost` on CT105). resolve.go logic lives in src/hal0/normalize/resolver.py; thinking suppression in normalize/thinking.py. Branch kept for reference.

Issue #407, opened 2026-05-28
Build hal0-test-template CT 200 + pct-clone smoke harness for fresh-install testing


thinmintdev:
## What to build

A reusable Proxmox LXC template (CT 200, "hal0-test-template") and a smoke driver script that clones it, runs \`install.sh\`, exercises a few hal0 subcommands, runs the uninstall path, and destroys the clone. Each smoke run starts from byte-identical state.

### Why this exists

LXC 105 (\`hal0\`) is the live dev environment — full hardware passthrough, editable install, real models. It is the wrong target for "what does \`curl … | bash\` feel like for a new user?" testing because:

- Polluted with dev state (registry, slot toml, lemonade config, MCP installs, models).
- Can't easily reset between runs.
- Mixing dev + install/uninstall iteration loses both signals.

We need an ephemeral target that exercises the tarball install path **with the same hardware-access recipe as 105** so the iGPU / NPU show up to the installer. LXCs share the host kernel and gate \`/dev/*\` access via cgroup rules — both 105 and the test CT can hold cgroup rights to the iGPU / NPU simultaneously, with a brief contention window only when both try to load a model. For install + uninstall flow testing (no model warming required) there is no contention.

### Golden template (CT 200)

Built once. Carries:

- Vanilla distro (Debian 12 cloud or CachyOS — pick whichever matches LXC 105's base; verify by inspecting 105.conf)
- The strix-halo passthrough recipe from \`220.conf\` (privileged, apparmor unconfined, \`dev0\`–\`dev3\` cgroup lines from \`feedback / strix-halo-lxc-passthrough\` memory)
- \`/mnt/ai-models\` NFS bind-mount (read-only, from pve) so install doesn't have to re-download base models
- Standard \`halo\` user with sudo + ssh key
- Pre-installed: \`curl\`, \`ca-certificates\`, \`jq\`, \`git\`, \`rsync\`, \`htop\`
- Networking, sshd, locale, timezone
- **No hal0 artifacts at all** — that's the whole point

Once the CT is provisioned and verified, mark as template: \`pct template 200\`.

### First-boot provisioning

Choose one of:

- **Real cloud-init** — \`pct set 200 --cicustom user=local:snippets/hal0-test.yaml\` (works on Debian / Ubuntu cloud-init-enabled templates)
- **Proxmox hookscript** — native LXC mechanism, runs at pre-start / post-start. More reliable, more Proxmox-idiomatic.

Provisioning content (translate to whichever mechanism):

[...]

Issue #406, opened 2026-05-28
Reconcile install-mode taxonomy: editable /opt/hal0 vs tarball /usr/lib/hal0/current


thinmintdev:
## What to build

Declare hal0 supports two distinct install modes — **editable** (dev) and **tarball** (production) — codify auto-detection, and make `Updater.apply()` + the dashboard's Settings → Updates surface behave correctly in each.

### Why this exists

PR #386 fixed the Settings → Updates apply button (it now polls status and the extract step quarantines stale dirs). On the way to that fix we discovered LXC 105 carries BOTH layouts side-by-side:

| Path | Role | Used by |
|---|---|---|
| `/opt/hal0/` | git checkout + `pip install -e .` venv | `systemd hal0-api.service` |
| `/usr/lib/hal0/current → hal0-<v>/` | `Updater.apply()` extract target | nothing on this host |

Self-update on this LXC extracts the new release, swaps the `current` symlink, logs `swap_ok` — and the running editable venv stays on the old version because systemd points at `/opt/hal0`. The button "works" but the version stays stuck. End users on a clean `curl … | bash` install don't hit this because they only have the tarball layout. We need explicit, audited support for both modes.

### Two modes — definitions

| Mode | Source of truth | Update mechanism | Signing | Who has it |
|---|---|---|---|---|
| **editable** | `/opt/hal0/` git checkout + `pip install -e .` venv | `git pull && pip install -e . && systemctl restart hal0-api hal0-lemonade` | trust your git remote | contributors, LXC 105 |
| **tarball** | `/usr/lib/hal0/current → hal0-<v>/` symlink | `Updater.apply()` — atomic symlink swap with sigstore-verified tarball | full cosign chain | end users from `install.sh` |

### Auto-detection

`Updater` should detect mode at startup and cache the result. Heuristic (in priority order):

1. `/opt/hal0/.git/` exists AND the running interpreter's `*.dist-info` reports `Editable project location` → **editable**.
2. `/usr/lib/hal0/current` symlink exists and resolves under `/usr/lib/hal0/` → **tarball**.
3. Neither → raise `system.install_mode_undetected`; surface as a banner.

Don't gate on an env var or marker file unless the heuristic produces a false positive in the wild (file a follow-up if so).

### What changes per mode

**editable mode:**
[...]

PR #589, opened 2026-06-06
feat(install): FHS prod layout + updater re-pip so `hal0 update` actually takes effect (#495)

### LINKED ISSUE #495: installer: production path must produce the FHS /usr/lib/hal0/<ver>/current layout (corrects #406 premise)
## What to build
Make the **production** install path actually produce the FHS layout the updater and docs already assume: code at `/usr/lib/hal0/<ver>/` with a `current` symlink, and the venv re-pointable (drop `pip install -e` for prod). Today `install.sh` runs `pip install -e /opt/hal0` for **every** install — including the `curl | bash` path — so the `/usr/lib/hal0/current` tarball layout is **never produced**, the running daemon imports from `/opt/hal0/src/hal0`, and `Updater.apply()` writes to a path nothing imports from.

This **corrects the premise of #406**, which assumes "end users on a clean `curl | bash` only have the tarball layout." The W1 audit (`/home/halo/dev/hal0-audit/REPORT-w1-install.md`) found that is not true — the installer produces editable-at-`/opt/hal0` for everyone. This slice builds the missing tarball/FHS production path so #406's two-mode reconciliation has a real tarball mode to detect.

Decision context (locked, D1): FHS versioned tree is the production layout; an explicit editable `--dev`/`HAL0_HOME` mode remains for contributors and CT-105. The two must not silently mix.

## Acceptance criteria
- [ ] `install.sh` default (prod) writes `/usr/lib/hal0/<ver>/` + atomic `current` symlink; `/usr/local/bin/hal0` resolves through it
- [ ] Venv is re-pointable (no editable `.pth` into a git checkout on the prod path)
- [ ] `--dev`/`HAL0_HOME` produces the editable `/opt/hal0` mode with a marker; prod and dev refuse to co-exist with clear guidance
- [ ] Existing `/opt/hal0` installs (CT-105) keep working via the documented dev mode — migration path covered (see follow-up uninstall/migration slice)
- [ ] `paths.py` contract matches what the installer produces; ADR updated to state the production layout is FHS
- [ ] Verified on the CT-200 fresh-install smoke harness (#407): a clean install yields the FHS layout and `hal0 --version` reflects the installed version

## Blocked by
None - can start immediately (unblocks the updater + uninstall + #406 work)

thinmintdev:
## Parent
Closes #495 (corrects #406). Verified via the #407 install-smoke harness.

## What this does
Production installs now use the **FHS layout the updater already targeted**, and `hal0 update` actually changes the running version instead of being a silent no-op.

```
/usr/lib/hal0/hal0-<version>/   versioned code tree (full source incl. installer/, ui/)
/usr/lib/hal0/current -> ...    atomic-swap symlink
/usr/lib/hal0/venv             shared venv (survives update swaps)
```

### install.sh (prod path)
- copy source into the versioned dir, point `current` at it (atomic temp-symlink + `mv -T`)
- install hal0 **non-editable** into the shared venv (dev mode stays editable — protects the CT-105 `/opt/hal0` checkout)
- set `HAL0_UI_DIST=<current>/ui/dist` in `api.env` — the non-editable install moves `__file__` into site-packages, which breaks the dashboard's `ui/dist` walk-up resolution
- `WorkingDirectory` follows `current` so an update needs no unit rewrite
- warn (don't delete) when a legacy editable `/opt/hal0` install is present

### updater.apply()
- after the atomic symlink swap, **re-pip the swapped-in tree into the running venv** (`pip install --no-deps --force-reinstall <dir>`). The venv imports hal0 from its own site-packages, so the swap alone changed nothing — this is the fix for "`hal0 update` is a no-op on real installs."
- a failed re-pip rolls the symlink back so `current` and the venv stay consistent; editable/dev installs skip the step entirely

### hal0 uninstall (caught by the harness)
- resolved `uninstall.sh` via `parents[2]` of `hal0.__file__`, valid only for editable installs. Now tries the editable repo root **and** `<current>/installer/uninstall.sh` via `paths.usr_lib()`.

## Verification — #407 harness on a fresh Proxmox CT
`make harness-install ARGS="--from-tree ."` (template CT 201):
```
{"install_ok":true,"smoke_ok":true,"uninstall_ok":true,"residue":"clean"}
```
- FHS layout created correctly; non-editable install of hal0 + all deps
- api + Lemonade started; **dashboard served** (`hal0 v0.3.2a1 · slots=2`) — confirms `HAL0_UI_DIST` works under the non-editable layout
- uninstall leaves **zero residue** (paths, units, group all gone)

Run 1 caught the `hal0 uninstall` locator break; run 2 (post-fix) is fully green.

## Tests
- 3 new `apply()` cases: re-pip invoked in prod / skipped when editable / rollback on re-pip failure
- 1 new uninstall case: resolves the FHS path when not editable
- 54 updater+uninstall tests green; `ruff check` + `ruff format --check` clean; `bash -n` + shellcheck clean

## Acceptance criteria
- [x] Prod install produces the FHS layout (`/usr/lib/hal0/<versioned>` + `current` + shared venv)
- [x] `hal0 update` changes the running version (re-pip into the venv), not just the symlink
- [x] Dashboard serves under the non-editable install
- [x] `hal0 uninstall` works under FHS and leaves clean residue
- [x] Dev/editable mode unchanged (protects CT-105)
- [x] Verified e2e on a fresh CT via the #407 harness

## Follow-up dependency (not in this PR)
For the updater re-pip to work in real production, the **release tarball must ship an installable source tree** (pyproject + src + prebuilt `ui/dist`) so `pip install <dir>` succeeds offline. That's a `release.yml` packaging concern — flagging it so it's tracked before the next tagged release relies on `hal0 update`.

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -164,20 +164,33 @@ PY="${HAL0_PYTHON:-python3}"
 # the comment on TLS posture near the flag parser.
 API_BIND_HOST="0.0.0.0"
 
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 if [[ "${DEV_MODE}" -eq 1 ]]; then
+    # Dev: editable checkout, everything under one prefix. The updater
+    # refuses apply() in this mode (re-run `git pull && pip install -e .`).
     PREFIX="${HAL0_PREFIX:-${PWD}/.hal0ai}"
     ETC_DIR="${PREFIX}/etc/hal0"
     VAR_DIR="${PREFIX}/var/lib/hal0"
     UNIT_DIR="${PREFIX}/etc/systemd/system"
+    VENV_DIR="${PREFIX}/.venv"
+    CURRENT_LINK=""
     info "Dev mode — all paths under ${PREFIX}"
 else
-    PREFIX="${HAL0_PREFIX:-/opt/hal0}"
+    # Prod FHS (#495): code lives in a versioned dir with a `current`
+    # symlink; the venv is shared at ${FHS_ROOT}/venv so it survives
+    # `hal0 update`'s atomic symlink swaps (the updater re-pips `current`
+    # into this venv on apply).
+    HAL0_FHS_ROOT="${HAL0_PREFIX:-/usr/lib/hal0}"
+    VERSION="$(grep -m1 '^version' "${REPO_ROOT}/pyproject.toml" 2>/dev/null | sed -E 's/.*"([^"]+)".*/\1/')"
+    [[ -n "${VERSION}" ]] || VERSION="0.0.0"
+    PREFIX="${HAL0_FHS_ROOT}/hal0-${VERSION}"
+    CURRENT_LINK="${HAL0_FHS_ROOT}/current"
     ETC_DIR="/etc/hal0"
     VAR_DIR="/var/lib/hal0"
     UNIT_DIR="/etc/systemd/system"
+    VENV_DIR="${HAL0_FHS_ROOT}/venv"
+    info "FHS layout — code ${PREFIX}, current → ${CURRENT_LINK}, venv ${VENV_DIR}"
 fi
-VENV_DIR="${PREFIX}/.venv"
-REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 
 # ── Release verification gate ──────────────────────────────────────────────
 # Refuse to run as root against an UNVERIFIED release tree. The signed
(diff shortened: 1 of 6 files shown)


thinmintdev:
Correction on the follow-up note: the release tarball is **already pip-installable** — `release.yml`'s staging step (lines 120–141) ships `src/` + `pyproject.toml` + `installer/` + prebuilt `ui/dist` + manifest/LICENSE/README. So the updater's `pip install <extracted-dir>` works at update time with the existing release pipeline, and the UI is served from `<current>/ui/dist`. No `release.yml` change is needed before relying on `hal0 update`.

PR #637, opened 2026-06-07
fix(slots): guard NPU FLM trio shadows from independent load (shape Unit 0)

thinmintdev:
## What & why

Part of the shape-consolidation work (`docs/internal/brain-redesign/shape-consolidation-proposal-2026-06-07.md`), **Unit 0** — the trio-shadow handling, shipped as its own backend-only PR (smallest blast radius).

The NPU runs a **single FLM process** — the chat anchor (`device=npu type=llm`) — which also serves transcription/embedding via lemond-global `--asr/--embed`. The `stt`/`embed` slots are therefore **shadows** of that anchor, *not independently loadable*: a standalone `/v1/load` for whisper/embed on the busy single-tenant NPU returns **HTTP 500**.

> **Latent, not a live fire:** the current deployment ships `stt`/`embed` with `enabled=false`, so there's no active 500 today. This is defensive hardening so enabling them (or a stray transcription/embed request that misses the trio gate) can't 500. The adoption/reconcile/idle paths are already read-only; the only standalone load funnels through `SlotManager.load()` → `_spawn_locked`, which this PR guards.

## Changes (backend-only)

- **`is_npu_trio_shadow(cfg)`** helper (`manager.py`): `device==npu` AND `type ∈ {transcription, embedding}`; the anchor (`type=llm`) is excluded.
- **`load()` short-circuit** for shadows: skips both the spawn *and* `_await_ready` (which would probe the shadow's non-existent child port), transitions to `READY`. Trio inference is routed to the anchor's FLM process by the dispatcher, not to the shadow's port.
- **Delete dead `_FLM_TRIO_SLOTS = {agent, stt-npu, embed-npu}`** (`slots.py`): those names predate the deployment rename to `npu`/`stt`/`embed`, so the `coresident_group` trio badge never rendered in prod. Coresident grouping now keys on `device==npu`.

## Deliberately deferred

Shadow state-**display** derivation (driving `lemonade_state` from the anchor + `flm.args`) is *not* in this PR. Doing it naively over-reports `loaded` and conflicts with the existing deliberate `test_loaded_state_uses_chat_anchor_model`. The correct `flm.args`-gated version belongs with the lemond-config work in P2/P3.

## Tests

- New `tests/slots/test_npu_trio_shadow.py`: predicate; `load()` does not spawn for a shadow; anchor + GPU still spawn (the regression that would bite silently).
- New device-based coresident test pinning rename-independence (real `npu`/`stt`/`embed` names).
- Regression: slots + `test_v1_npu_trio_routing` + `test_flm_trio` suites — **169 passed**. `ruff check` + `ruff format --check` clean. No new mypy errors (12 pre-existing, not CI-gated).

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -37,6 +37,10 @@ hal0 slot unload <name>
 hal0 slot restart <name>
 hal0 slot swap <name> --model M
 hal0 slot logs <name> [--follow]             # SSE tail of journald
+hal0 slot create <name> --backend B          # create a new slot
+hal0 slot edit <name>                        # open slot TOML in $EDITOR
+hal0 slot show <name>                        # print slot config + state
+hal0 slot delete <name>                      # remove slot permanently
 ```
 
 `hal0 slot swap` is the most-used command in normal operation. It
@@ -46,11 +50,16 @@ through `unloading → starting → warming → ready`.
 `hal0 slot logs <name> --follow` is the journald tail wrapped over
 SSE; it shows the same stream the dashboard's Logs view tails.
 
+`slot add` / `slot remove` still work as deprecated aliases for
+`slot create` / `slot delete` but will be removed in a future release.
+
 ## Models
 
 ```sh
 hal0 model list
 hal0 model pull <ref>                        # HF streaming pull
+hal0 model register <ref>                    # register a local GGUF
+hal0 model show <ref>                        # print model metadata
 hal0 model rm <ref>
 hal0 model assign <ref> --slot S
 ```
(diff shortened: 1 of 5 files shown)


thinmintdev (commented):
REVIEWER VERDICT: APPROVE (posted as comment — GitHub blocks formal self-approval; PR author == reviewer account `thinmintdev`). gitops: treat as approved, safe to merge first as the keystone.

Both axes clear, CI all green (python 3.11/3.12 + ui + γ-suite).

**SPEC:** Adopts the dead session's shape-consolidation Unit 0 and un-breaks main's cli.mdx parity test. `docs/reference/cli.mdx` now documents the agent/capabilities/registry/memory/migrate/doctor groups the CLI grew but the doc lagged — flipping `test_cli_mdx_documents_every_command` green. Spot-checked new entries vs source: `slot create`/`show`/`delete` (slot_commands.py:316/581/518), `capabilities sync` (capabilities_commands.py:387), `memory graph status` (memory_commands.py → GET /api/memory/graph/status), `registry import` (registry_commands.py), `config hardware` (config_commands.py:179), `doctor toolbox-pull` (doctor_commands.py:152) — all real. The `slot add`/`remove` 'deprecated aliases' note matches reality (both `hidden=True` commands).

**STANDARDS:** Deletes the dead `_FLM_TRIO_SLOTS` frozenset (seed names never matched the prod rename, so the coresident badge never rendered) and re-keys `coresident_group` on `device==npu` — correct rename-independent fix. New `is_npu_trio_shadow()` + `load()` short-circuit defensively skips spawn AND the readiness probe (targets the shadow's non-existent child port) for stt/embed NPU shadows → READY; anchor (type=llm) excluded. Targeted tests present and meaningful (shadow predicate; load-does-not-spawn with anchor/GPU spawn regression; device-based coresident test pinning rename-independence). Backend-only blast radius; ruff-clean per CI.

Issue #652, opened 2026-06-08
EPIC: containerized GPU backend (podman + profiles) + dashboard rewire


thinmintdev:
## What to build
Move GPU LLM slots from Lemonade-forked baremetal binaries to **podman containers** built by the nightly toolbox fork, selected via **profiles** (image + bench-tuned flags), dispatched through hal0's existing remote-upstream proxy. Then rewire the dashboard to a hybrid (container + lemond) model.

Design: `hal0-container-runtime-design-2026-06-08.md`. Bench basis: `hal0-container-bench-2026-06-08.md`.

Decisions locked: container runtime (bench parity 52.8 vs 53.6 baremetal); profiles = flag-bundles on shared images; podman; slot OWNS container (1:1); phase-1 = GPU LLM slots only (lemond keeps embed/rerank/stt/tts; NPU/FLM untouched).

Per-slot optimal (bench): agent 35B MoE ace-saber = moe-rocmfp4 (MTP off) ~52.8 tok/s; chat 27B dense qwopus = dense-mtp-rocmfp4 (MTP on) ~24.4 tok/s.

## Child slices
Tracked as separate issues (see Blocked-by chains). This epic is the umbrella; do not close until all children merged + live cutover done.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

thinmintdev:
P2 hardening notes from #666 (spine) review — fold into a later hardening pass (non-blocking):
1. ContainerProvider.terminate: deregister upstream in a `finally` (orphan-on-error if daemon-reload/unlink raises after unload). Load path is clean.
2. Upstream registration is in-memory (matches existing _autoregister convention); persist across api-restart is P2.
3. container_spec() (ABC) vs _render_unit() can drift — single-source the argv.

thinmintdev:
P2 notes from #667 review: (1) /api/slots _container_state_enrichment does per-slot subprocess systemctl + httpx /health on a dashboard-polled hot path — bound/parallelize + reuse cp.is_active at scale. (2) gate 'crashed' test mocks reason; real crashed→inactive (detection is on enrichment path, tested). Non-blocking.

thinmintdev:
[...]

PR #665, opened 2026-06-08
slots: rename primary→chat, add hidden back-compat aliases

thinmintdev:
## Summary

- Renames seeded slot `primary` → `chat` and Hermes delegation slot reference `agent-hermes` → `agent` (`agent` was already in `NPU_SEEDED_SLOTS`, so no new TOML is created — the alias just redirects old references)
- Adds `SLOT_ALIASES = {"primary": "chat", "agent-hermes": "agent"}` in `manager.py` with `SlotManager._resolve_alias()` wired into all public methods
- **Aliases are hidden by construction**: `list()` / `iter_configs()` enumerate from `*.toml` on disk; no alias TOML exists → aliases never appear in `/api/slots`, dashboard, or catalogs
- `hal0/primary` virtual name moved to `VIRTUAL_ALIASES → hal0/chat` (back-compat for Hermes overrides)
- `hal0_chat_slot_alias_map()` injects `primary→chat_model_id` and `agent-hermes→agent_model_id` entries so requests using old slot names still reach the right model

## Risky refs explicitly left alone (noted here, not guessed)

- `GraphRouteLiteral = Literal["upstream","primary","agent"]` in `config/schema.py` — this is the memory graph routing enum (ADR-0014), **not** a slot name
- `primary_gpu` variables in `hardware.py` / `catalog.py` — GPU hardware variable, unrelated
- `memory/cognee_wrapper.py` `"primary"` route — same ADR-0014 memory enum

## Test plan

- [x] 20 new alias-specific tests in `tests/slots/test_slot_aliases.py`
- [x] Updated `tests/normalize/test_resolver.py` — hal0/primary alias, DEFAULT_CHAINS shape, LiveSlotResolver
- [x] All 284 tests pass in `tests/slots/ tests/dispatcher/ tests/normalize/`
- [x] `ruff check` + `ruff format --check` clean

## Live migration note

Existing installs have `primary.toml` on disk. The alias layer means old API calls (`model=primary`, `/api/slots/primary`) resolve transparently. A follow-up deploy step (issue #633 acceptance) should rename `primary.toml` → `chat.toml` on CT105 and restart the slot.

Closes #654, closes #633
Parent epic: #652

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #657, opened 2026-06-08
Dashboard: slot-card container variant (read-only)


thinmintdev:
## Parent
#652

## What to build
Consume a `slot.runtime` field in the dashboard. Branch slotIndicator() to a container path (running+health=serving, pulling/starting=warming, stopped=idle, crashed=error) and show an image-tag chip + container-state dot on SlotCard. Add container fields to the useSlots Slot interface. lemond slot cards render exactly as today.

## Acceptance criteria
- [ ] Container slots show image tag + container state; correct status dot
- [ ] lemond slots visually unchanged
- [ ] useSlots interface gains runtime/profile/image/container_status/health fields
- [ ] Playwright coverage for the container-slot card state

## Blocked by
- #656

🤖 Generated with [Claude Code](https://claude.com/claude-code)

thinmintdev:
UI-expert audit folded in (full: ~/dev/hal0-build/ui-audit.md). Scope additions for this slice:
- **DO FIRST (N1):** extract ONE runtime-aware `slotStatus(slot)→{phase,cls,label,isLive}` and refactor the 4 existing classifiers (slotIndicator slots.jsx:62, stateChipClass slot-modals.jsx:39, memory-map LIVE_STATES:15, phase slots.jsx:225) to consume it — else the container branch doubles a 4-way duplication to 8.
- Replace device chip + actual_backend/mismatch block (slots.jsx:294-309) with an **image-tag chip**; mismatch = running image ≠ profile image (backend-computed).
- **tok/s vs profile bench** on the metric tile (e.g. '48 / ~52').
- a11y: reduced-motion on pulse dots; enable-toggle role=switch/aria-checked/focus-visible (copy NpuSwitch:515).
- N2: inline-swap is now a COLD restart — confirm/toast + drive card to `starting` + '· cold restart' copy.
- N3/N5: action-row hit targets; a hybrid runtime micro-tag (container vs lemond).

PR #667, opened 2026-06-08
feat(dispatcher,slots): container readiness-gate + honest container state

### LINKED ISSUE #656: Container readiness-gate + honest container state in /api/slots
## Parent
#652

## What to build
Dispatcher preflight for remote/container slots: if the container is down/starting, return a loading state instead of a raw 502. Surface per-slot `container_status` (running|stopped|starting|crashed) + `container_health` (/health 200) + `state` in /api/slots, derived from podman + the container /health (not lemond /v1/health).

## Acceptance criteria
- [ ] Dispatcher returns loading (not 502) when a container slot is starting/down
- [ ] /api/slots reports container_status + container_health + state for container slots
- [ ] lemond slots' status path unchanged
- [ ] Tests for the gate + status mapping

## Blocked by
- #655

thinmintdev:
## Summary

- **Dispatcher preflight gate**: Container slots register as `kind="remote"` upstreams so the existing Lemonade slot gate (`call.slot_name`) never fired — forwarding a down container produced a raw 502 `ConnectError`. Added `container_slot_name` field to `UpstreamCall` (set by new `_container_slot_name_of()` helper, fires only when `Upstream.slot_name` is set — the marker `_register_container_upstream` now writes). `Dispatcher.forward()` calls new `_check_container_slot_ready()` before forwarding: probes `systemctl is-active` + `/health`, raises `SlotLoading` (503 + Retry-After) on failure instead of letting `ConnectError` surface as 502.
- **`SlotManager.container_readiness_check()`**: New public method returning `(ready: bool, reason: str)` from the two live probes — called by the dispatcher, keeps ContainerProvider knowledge inside SlotManager.
- **`/api/slots` container state fields**: New `_container_state_enrichment()` returns `container_status` (running|stopped|starting|crashed) and `container_health` (bool) for container-backed slots. Container slots are now skipped in `_lemonade_state_enrichment()` to avoid spurious `lemonade_state="idle"` from a lemond probe that has no knowledge of podman containers.

## Test plan

- [ ] `tests/dispatcher/test_container_gate.py` (9 tests): gate raises `SlotLoading` not 502 for inactive/starting/crashed; passes through when ready; Lemonade slots unaffected; no slot_manager → gate skipped
- [ ] `tests/api/test_slots_container_state.py` (7 tests): `container_status` running/stopped/starting/crashed; `container_health` True/False; Lemonade slot retains `lemonade_state` and has no `container_*` keys; `get_slot` also returns fields
- [ ] `tests/dispatcher/` (116 tests) — all pass
- [ ] `tests/slots/` (129 tests) — all pass
- [ ] `ruff check src/ tests/` — clean
- [ ] `ruff format --check src/ tests/` — 418 files already formatted

Parent epic: #652
Closes #656

### DIFF
--- a/src/hal0/api/routes/slots.py
+++ b/src/hal0/api/routes/slots.py
@@ -187,10 +187,17 @@ async def _lemonade_state_enrichment(request: Request) -> dict[str, dict[str, An
     ]
     trio_active = bool(npu_llm_enabled)
 
+    from hal0.slots.manager import SlotManager
+
     out: dict[str, dict[str, Any]] = {}
     for cfg in configs:
         name = str(cfg.get("name", ""))
         if not name:
+            continue
+        # Container slots use a separate enrichment path (_container_state_enrichment).
+        # Skip them here so they don't get a bogus lemonade_state="idle" from a
+        # lemond health probe that has no knowledge of the container.
+        if SlotManager._is_container_slot(cfg):
             continue
         enabled = cfg.get("enabled") is not False
         model_default = ""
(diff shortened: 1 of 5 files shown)


thinmintdev (commented):
VERDICT: APPROVE — merge-ready

The gate is correct and the lemond/dispatch paths are non-regressing. All four checks GREEN — full python matrix (3.11 8m30s / 3.12 9m15s) ran the FULL suite, clearing the cross-suite regression risk; ruff Lint + Format are green sub-steps (the #665 lesson held across router.py, manager.py, slots.py + new tests). γ-suite + ui green.

SPEC (vs #656 + design §3): ACs met.
- Dispatcher returns SlotLoading (503 + Retry-After) instead of raw 502 for down/starting container slots. VERIFIED end-to-end: _check_container_slot_ready raises SlotLoading(details={"retry_after_s":15,...}); error_codes.py middleware (line 81-83) promotes details["retry_after_s"] → real Retry-After header when status==503. The 503+Retry-After AC holds at runtime (the gate tests assert status+details but not the header itself — coverage nit, not a behavior gap).
- Gate separation is CORRECT — the crux of this PR. container_slot_name is deliberately separate from slot_name. Confirmed against branch source: _slot_name_of() has `if upstream.kind != "slot": return ""` at the top, so a container kind="remote" upstream (even with slot_name now set as the marker) yields call.slot_name="" → it does NOT enter the Lemonade _ensure_slot_loaded_backend_aware/_forward_with_serving branch (no auto-load, no SERVING wrap). _container_slot_name_of() returns the name only for kind=="remote" with slot_name set. The two markers are mutually exclusive; forward() ordering puts the lemonade branch first but they can never both be set.
- Genuine external remotes (OpenRouter/OpenAI, slot_name=None) → both markers empty → _forward_plain, no gate. Verified by test_container_slot_name_of_remote_without_slot_name + test_lemonade_slot_unaffected_by_container_gate.
- /api/slots: container_status (running|stopped|starting|crashed) + container_health for container slots via _container_state_enrichment; container slots SKIPPED in _lemonade_state_enrichment (no spurious lemonade_state="idle"). Verified by test_container_slot_has_no_lemonade_state + test_lemonade_slot_unaffected_no_container_keys.

GATE CORRECTNESS (all 4 states + edges):
- inactive → SlotLoading(reason="inactive"); active+unhealthy → "starting"; ready → passes through to _forward_plain; no slot_manager → gate skipped (test_container_gate_no_slot_manager_skips_gate). All covered.
- container_readiness_check guards config_missing + not_a_container_slot before probing. Sound.

REGRESSION RISK (lemond/dispatch): none. The slot_name branch in forward() is byte-unchanged; the container branch is additive and only reachable via the new mutually-exclusive marker. All-lemonade installs never set container_slot_name. Suite-confirmed.

NON-BLOCKING NOTES (tracer-acceptable; P2):
1. The "crashed" gate test (test_container_gate_crashed_raises_slot_loading) mocks reason="crashed", but container_readiness_check only ever returns inactive/starting/ready — a real crashed unit returns "inactive" (is_active False). Spec-fine: the GATE only owes down/starting→loading; real crashed DETECTION lives on the /api/slots enrichment path ("failed"→"crashed"), which IS genuinely tested (test_container_slot_status_crashed_when_failed). Pure test-naming fidelity nit.
2. /api/slots is a dashboard-polled hot read path; _container_state_enrichment now runs (per container slot, serially) a subprocess systemctl is-active + an httpx /health GET (up to 3s), plus a SECOND direct subprocess systemctl call on the inactive branch, alongside a second iter_configs(). Fine for the tracer single-container case; at scale this is subprocess fan-out + up-to-3s-per-slot stall on a starting container. Consider reusing cp.is_active output instead of re-shelling, and bounding/parallelizing in P2.

MERGE-READY: yes. Unblocks the UI card (#657). The two notes are P2 polish, not blockers.

PR #668, opened 2026-06-08
feat(ui): container SlotCard variant + N1 slotStatus unifier

thinmintdev:
## Summary

- **N1 (status vocabulary unifier)**: New `ui/src/dash/slot-status.js` with `slotPhase(slot)` + `slotIndicatorFromPhase()`. `slotIndicator()` delegates container slots to the container classifier; lemond slots go through the original code unchanged — all 16 `slot-indicator.spec` tests stay green with no modifications.
- **Container card branch**: `runtime==="container"` slots replace the device chip + backend-mismatch block with an IMAGE-TAG chip (last path segment of image ref, full ref on hover). A "container" runtime micro-tag (N5) distinguishes cold-swap behavior. Phase logic (button states) reads `container_status`/`container_health`.
- **N2 cold-swap confirm**: `InlineSwapPopover` shows "· cold restart" badge in the header for container slots and emits an info toast before firing the swap. Lemond hot-swap unchanged.
- **useSlots.ts**: `Slot` interface gains `runtime`, `profile`, `image`, `image_status`, `container_status`, `container_health`. `normalizeSlot` passes them through.
- **A11y**: `@media (prefers-reduced-motion: reduce)` kills pulse animation on `.dot.serving/.warming/.loading`. Enable-toggle gets `aria-label` + `role=switch`/`aria-checked` on the visible track. N3: `slot-actions` gets `touch-action:manipulation`.
- **Playwright**: 13-test `slot-card-container-v3.spec.ts` covering all container dot states + lemond regression guard + card-level chip assertions.

Closes #657

## Test plan

- [x] `slot-indicator.spec.ts` — all 16 lemond tests pass unchanged
- [x] `slot-card-container-v3.spec.ts` — all 13 new container tests pass
- [x] `slots-v3.spec.ts` — all 6 existing tests pass
- [x] `slot-lifecycle-controls-v3.spec.ts` — all 3 pass
- [x] `memory-map-v3.spec.ts` — all pass
- [x] Clean Vite build (130 modules, 9.1s, no errors)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #666, opened 2026-06-08
feat(slots): ContainerProvider — podman/docker container slot runtime

thinmintdev:
## Summary

- Add `SlotConfig.runtime` + `SlotConfig.profile` fields (default lemonade — existing slots unaffected)
- New `ContainerProvider` renders a self-contained `hal0-slot@<name>.service` systemd unit (`docker run` with `/dev/kfd`+`/dev/dri` passthrough, numeric video/render GIDs, `/mnt/ai-models:ro` at identical path, loopback `--publish=127.0.0.1:<port>:<port>`)
- `SlotManager` routes slots with `profile` set (or `runtime="container"`) to `ContainerProvider`; auto-registers a `kind="remote"` upstream in the live `UpstreamRegistry` so the dispatcher proxies traffic — no dispatcher code change needed
- 29 unit tests + live CT105 smoke at **49.5 tok/s** (ace-saber / moe-rocmfp4, port 8095)

## Implementation notes

- Container runtime auto-detects podman → docker (overridable via `HAL0_CONTAINER_RUNTIME`)
- llama-server requires space-separated args (`--host 0.0.0.0`, not `--host=0.0.0.0`) — caught in live smoke
- `UpstreamRegistry` injected into `SlotManager` via new `upstreams_registry` param (mirrors `pull_runner` injection pattern); `upsert` on load / `remove` on unload
- Lemonade slots are 100% unaffected — all five branch sites (`_spawn_locked`, `terminate`, `_await_ready`, `_is_active`, `_maybe_adopt_running_slot`) gated on `_is_container_slot(cfg)`

## Test plan

- [x] `pytest tests/providers/test_container.py` — 29 passed
- [x] `ruff check` + `ruff format` — clean on src/ and tests/
- [x] Live CT105 smoke: ace-saber answers `/v1/chat/completions` at **49.5 tok/s** (container direct on port 8095)
- [ ] Deploy to CT105 via `make deploy` + load a container slot end-to-end through hal0-api dispatcher (P2 migration task)

Closes #655
Parent: #652

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #661, opened 2026-06-08
Hermes → agent slot reasoning wiring (#635)


thinmintdev:
## Parent
#652

## What to build
Point Hermes at the renamed `agent` slot with the reasoning config per #635 (advanced-reasoning subagents). Verify Hermes routes to the containerized agent slot.

## Acceptance criteria
- [ ] Hermes config targets the agent slot (post-rename)
- [ ] Reasoning path verified end-to-end against the container slot
- [ ] No regression to existing Hermes routing

## Blocked by
- #655
- #654

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #672, opened 2026-06-08
feat(memory-map): container slot attribution via podman cgroup

### LINKED ISSUE #660: Memory-map: container slot attribution via podman cgroup
## Parent
#652

## What to build
Populate per-slot `mem_mb` for container slots from the podman container cgroup (generalize the existing docker-cgroup probe; name convention hal0-slot-<name>). The GTT-split fallback stays valid; NPU/FLM path unchanged.

## Acceptance criteria
- [ ] Container slots show resident+KV memory in the memory map
- [ ] GTT pool ceiling + fallback unchanged for non-container slots
- [ ] Test for podman cgroup mem read

## Blocked by
- #655

thinmintdev:
## Summary

- **Backend**: `_container_cgroup_mem_bytes(slot_name)` in `capacity.py` probes the live podman/docker cgroup (`memory.current`) for `hal0-slot-<name>`. `build_per_slot` tries this first for every non-NPU resident slot; returns 0 for lemond slots (no container exists) so the lemond file-size + KV estimate path is unchanged.
- **UI**: `slotLegendSub()` in `memory-map.jsx` shows image tag (truncated to last 32 chars) or profile slug for container slots instead of the uninformative `device=rocm` token. Both sidebar and expanded legend variants updated.
- **Tests**: 12 unit tests for the cgroup probe (happy path, no runtime, container absent, timeout, cgroupv1, docker fallback) + 2 new Playwright specs in `memory-map-v3.spec.ts`.

Closes #660
Parent: #652

## Test plan

- [x] `pytest tests/slots/test_container_cgroup_mem.py -q` — 12 passed
- [x] `ruff check + format --check` — clean
- [x] UI build (`vite build`) — no errors
- [x] `playwright test memory-map-v3.spec.ts` — 9 passed (includes 2 new container specs)

### DIFF
--- a/src/hal0/slots/capacity.py
+++ b/src/hal0/slots/capacity.py
@@ -32,6 +32,10 @@ from hal0.slots.state import SlotError
 if TYPE_CHECKING:
     from hal0.hardware.probe import HardwareInfo
 
+# Container-name prefix matches the convention in providers/container.py:
+# ``ExecStop = <runtime> stop -t 20 hal0-slot-<name>``.
+_CONTAINER_NAME_PREFIX = "hal0-slot-"
+
 
 # NOTE: We code against ``hal0.hardware.probe.HardwareInfo`` as the contract
 # even though the probe itself is currently a stub (raises NotImplementedError).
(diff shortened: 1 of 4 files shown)


thinmintdev (commented):
VERDICT: APPROVE-ON-GREEN — merge-ready once python matrix + γ-suite pass

ui already green; python (3.11/3.12) + γ-suite pending. Value-parity for lemond/NPU is clean and the structure is sound. One correctness caveat (lead, non-blocking) requires CT105 validation; the cost note is secondary.

SPEC (vs #660): ACs met structurally.
- _container_cgroup_mem_bytes: podman→docker runtime detect → inspect PID → /proc/<pid>/cgroup (v2 unified) → /sys/fs/cgroup/<rel>/memory.current. Error/timeout(1.5s)/no-runtime/PID≤0/cgroupv1(no "::")/unreadable all → 0. 12 unit tests cover every branch incl docker fallback.
- build_per_slot: probe runs only for RESIDENT, non-NPU slots — NPU/FLM branch sets its row and `continue`s BEFORE the probe (test_npu_slot_skips_cgroup_probe asserts the probe is never called). Lemond slots: probe returns 0 → falls through to the ORIGINAL file-size + KV registry estimate. **Lemond + NPU mem_mb values are byte-unchanged** — verified.
- UI: attributeFallback (GTT-split / NPU-divide) is UNTOUCHED and still gated on hasMemMetrics (used only when no slot reports mem_mb). slotLegendSub change is legend SUB-LABEL TEXT only (image tag/profile for container, device token for lemond) — does not touch attribution math. GTT pool ceiling logic unchanged.

CAVEATS:

1. **(LEAD — correctness, non-blocking, needs CT105 validation) The "memory.current includes weights + KV" assumption is the riskiest claim and is unverified on this hardware.** These are Strix Halo iGPU slots — model weights live in GTT (system RAM mapped via amdgpu/TTM). Whether GTT/DRM buffers are charged to the memory cgroup is kernel/driver-dependent and historically often NOT (the controller charges anon + page-cache, frequently not DRM buffers). --no-mmap (in every profile) makes the initial read anonymous+charged, but once llama.cpp copies tensors to HIP/GTT and frees the host copy, the resident ~22 GB may sit in GTT that memory.current does not count.
   Why this is sharper than a cost note: the `if cgroup_bytes > 0:` branch has NO sanity floor and `continue`s — ANY positive value wins, even 2 GB. So a partial-accounting cgroup (e.g. ~2 GB CPU-side runtime while the model is ~22 GB) REPLACES the correct ~22 GB registry estimate with a wrong ~2 GB, i.e. *under-reports container slots below what the old path showed* — only EXACTLY 0 falls back. That directly threatens AC1 ("show resident+KV memory"). All 12 tests mock the cgroup value, so neither they nor green CI can catch this.
   VALIDATION (maintainer, post-merge on CT105): load the ace-saber container slot, compare /api/slots mem_mb against the ~22 GB GGUF size. If it reads single-digit GB, the probe is missing GTT and the >0 branch is degrading attribution. Cheap hardening: floor the cgroup value against the file-size estimate — `mem = max(cgroup_mb, file_est_mb)` — or treat an implausibly-small cgroup (well below model size) like 0 so the fallback fires. Fits the epics deferred-live-validation pattern (#666 tok/s, #661/#669 e2e), so APPROVE-ON-GREEN, but please run the CT105 check before relying on the container numbers.

2. **(P2 — cost) Unconditional `inspect` subprocess per resident non-NPU slot per memory-map poll.** The probe is not gated on any container signal, so an all-lemond install forks a podman/docker `inspect` for every resident GPU slot on every dashboard poll (returns fast when the container is absent; the 1.5s timeout only bites if the runtime hangs). I confirmed theres no cheap container signal at this layer — the Slot object exposes only state/model_id/metadata/backend, not runtime/profile — so the authors "probe everything" is a defensible tradeoff, not an oversight. If `metadata` carries `profile`, gating on `meta.get("profile")` would skip the fork for lemond slots; otherwise plumb a runtime flag into the Slot. P2.

TRIVIA / OUT OF SCOPE:
- Expanded-legend variant now shows `· ≈` for approx lemond slots (old code was bare s.device) — trivially additive, cosmetic.
- memory-map.jsx imports isSlotLive from slot-status.js (#668-inherited parity item) — NOT touched by this PR; out of scope, remains a #668 follow-up.

MERGE-READY: yes, on green python + γ. Caveat #1 is a post-merge CT105 validation + likely a small follow-up floor, not a code blocker; #2 is P2. Closes the #652 container-runtime epic chain.

PR #862, opened 2026-06-16
fix(slots): derive slot ctx, never silently inherit llama-server 4096

thinmintdev:
## Root cause — the 2026-06-15 chat@4096 incident

Hermes reported "broken": the **chat** slot was serving at `--ctx-size 4096` instead of its configured window, so `hal0/primary` 400'd (`exceed_context_size_error n_ctx:4096`) on any conversation > 4096 tokens.

Two compounding defects were found (see auto-memory `hermes_chat_slot_ctx4096_compaction_2026-06-15`):

- **Defect A (this PR):** `ModelConfig.context_size` defaulted to **4096**, and `model_dump()` wrote that verbatim. The installer normally derives a real window from a model's *curated* arch max — but a **non-curated** model with empty registry metadata (`qwopus3.6-27b-v2`) had no native window to derive from, so it fell to the schema default 4096. When `context_size` was unset entirely, `container_spec` emitted **no** `--ctx-size` flag and llama-server silently booted at its own 4096 default.
- **Defect B (not this PR):** stale baked systemd unit — editing `chat.toml` doesn't re-render the unit; only `hal0 slot restart` re-reads. Tracked separately (proposed: extend the slot drift reconciler / `hal0 slot status` to compare running `--ctx-size` vs toml and WARN).

## What this PR changes (Defect A)

Make "unset" explicit and **derive a real window** — a slot can never again silently inherit 4096:

- **schema:** `context_size` default `4096` → `None`.
- **providers/container.`container_spec`** (load path): when the slot pins no ctx, derive the model's native window from the registry (`metadata.context_length` → `defaults.context_size`), **dense-capped at 32768**, else a safe **8192** floor.
- **slots/manager.`create_config`:** persist the derived window at create time so the TOML, dashboard, and running container all agree; the load-path derive remains a belt-and-suspenders fallback.
- **`_cfg_to_dict` + `config.loader._unflatten_slot_toml`:** elide a `None` `context_size` before the TOML writer (`write_toml_atomic` rejects `None`).

The dense cap means an *unconfigured* slot can't request an impractically large KV cache (operators set `context_size` explicitly for full windows, as `chat.toml` does at 131072).

## Tests

- New `TestContextSizeDerive`: native-dense-capped / model-defaults / **8192-floor-not-4096** / explicit-wins.
- Updated `ModelConfig` default expectation (`is None`) + loader round-trip None-elision.
- Targeted suites green: providers, slots, config, install, registry — **600+ passed**. (`test_slot_backend_control::...normalizes_gpu_form` is a pre-existing cross-test ordering flake — passes standalone, untouched here.)

## Diagnostic note (from the incident)
`/v1/models` advertises a uniform 32100 for the npu/utility/chat virtual aliases — **do not trust it**; probe the real served `n_ctx` via a >4096-token request.

### DIFF
--- a/src/hal0/config/loader.py
+++ b/src/hal0/config/loader.py
@@ -315,6 +315,11 @@ def _unflatten_slot_toml(cfg: SlotConfig) -> dict[str, Any]:
         },
         "model": data["model"],
     }
+    # context_size is Optional (unset → derived at load by the provider);
+    # None has no TOML representation, so elide any None in the model table.
+    model_tbl = out.get("model")
+    if isinstance(model_tbl, dict):
+        out["model"] = {k: v for k, v in model_tbl.items() if v is not None}
     extra = data.get("extra") or {}
     if isinstance(extra, dict):
         for k, v in extra.items():
--- a/src/hal0/config/schema.py
+++ b/src/hal0/config/schema.py
@@ -150,10 +150,15 @@ class ModelConfig(BaseModel):
         default="",
         description="Default model id from the registry.  Must exist in /var/lib/hal0/registry/.",
     )
-    context_size: int = Field(
-        default=4096,
+    context_size: int | None = Field(
+        default=None,
         ge=128,
-        description="Context window size in tokens.",
+        description=(
+            "Context window size in tokens. Unset (None) is NOT 4096: the "
+            "load path derives the model's native window (dense-capped) or a "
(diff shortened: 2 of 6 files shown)


thinmintdev:
Defect B (stale baked unit — toml edits don't re-render until `hal0 slot restart`) is now tracked separately in #863 with the proposed reconciler/`slot status` drift-WARN approach from this PR's body.

Issue #31, opened 2026-05-16
Add idle slot state for process-up-no-model containers


thinmintdev:
## What to build

A slot container can be running with no model loaded (e.g. `llama-server --model ""`) and still report `state: "ready"`. Callers route to it and get `400 "model name is missing"` from the upstream. A "ready" slot that cannot fulfil any inference request is a UX trap, worse than `offline` because routers will pick it.

Refine the slot state taxonomy so `ready` requires both the process being up AND at least one model advertised on the upstream's `/v1/models`. Add a new `idle` state for the process-up-no-model case. Update the UI badge to render `idle` distinctly from `ready`.

## Acceptance criteria

- [ ] Slot state vocabulary includes `idle` distinct from `ready` and `offline`
- [ ] A slot with `--model ""` (or equivalent) reports `idle`, not `ready`
- [ ] A slot with a model loaded and `/v1/models` non-empty reports `ready`
- [ ] Dashboard slot card distinguishes `idle` from `ready` visually
- [ ] State-transition unit tests cover the `idle ↔ ready` edges
- [ ] The state machine in `ARCHITECTURE.md` (or the closest doc) reflects the new state

## Blocked by

- #30

---

_Sourced from `tests/harness/FINDINGS.md` §15 (medium)._

thinmintdev:
Closed by #50 (squash commit 81eaba3 — auto-close regex missed the second issue number because both were on one `closes` token).

PR #905, opened 2026-06-19
Fix Operator Board black-screen (React #31) on add-task

thinmintdev:
## Problem
Adding a task to the Operator Board (kanban) black-screened the whole dashboard with:
> Minified React error #31 … object with keys {name, on_disk, counts}

## Root cause
`/api/board/profiles` and `/api/board/assignees` are proxied straight to Hermes (`board.py:320-321`). Hermes returns actors shaped `{name, on_disk, counts}` — keyed by `name`, **not** the `{id, label}` the board UI assumed. The `?? p` fallbacks in `board-view.jsx` (`label: p.label ?? p.id ?? p`, `pid = p.id ?? p`) then bottomed out at the **raw object**, which React can't render → error #31. With **no error boundary anywhere in the UI**, that single bad child unwound the entire tree → black screen.

The default e2e fixtures (`BOARD_PROFILES`/`BOARD_ASSIGNEES`) used the idealised `{id,label}` shape — which is exactly why the existing board specs never caught it.

## Fix
1. **Normalise at the hook boundary** — new dependency-free `boardActors.js` (mirrors `logRing.js` / `normaliseTask`) guarantees every assignee/profile exposes a primitive `id`+`label` derived from `name`. Fixes all render sites at the source; a raw object can never reach a JSX child.
2. **View-level error boundary** in `main.jsx` (`key={route}`) so any future unexpected payload degrades to a contained message instead of taking down the TopBar/Sidebar/Footer.

## Tests
- `boardActors.test.mjs` — unit, proves `{name,on_disk,counts}` → `{id,label}` and never leaks the raw object. RED-first verified (module missing → GREEN).
- `board-hermes-actor-shape-v3.spec.ts` — e2e regression feeding the **real** Hermes shape; asserts the board mounts, the actor `name` surfaces as a label, no `[object Object]`, and the error boundary never engages. RED verified by neutering the normaliser, then GREEN.
- Full board + dashboard e2e suite: **101 passed**. `tsc --noEmit` clean; production build clean.

Note: did **not** touch Hermes (per scope decision) and deliberately left the separate latent `kcard` `task.profile` vs `assignee` field-name display mismatch out of scope.

### DIFF
--- a/ui/src/api/hooks/__tests__/boardActors.test.mjs
+++ b/ui/src/api/hooks/__tests__/boardActors.test.mjs
@@ -1,0 +1,54 @@
+// Dependency-free test for the board assignee/profile normaliser.
+//
+// Run: node ui/src/api/hooks/__tests__/boardActors.test.mjs
+//
+// Bug of record: React error #31 ("Objects are not valid as a React child")
+// crashed the whole Operator Board to a black screen the moment a task was
+// added. /api/board/{profiles,assignees} are proxied straight to Hermes,
+// which returns actors shaped {name, on_disk, counts} — they carry `name`,
+// NOT the `id`/`label` the UI assumed. board-view's `p.id ?? p` /
+// `p.label ?? p.id ?? p` fallbacks then ended at the raw object, which React
+// cannot render. normaliseActor must guarantee primitive id+label so a raw
+// object can never reach a JSX child position.
+
+import { normaliseAssignee, normaliseProfile } from "../boardActors.js";
+
+let failures = 0;
+const ok = (cond, msg) => { if (!cond) { failures += 1; console.error("  ✗ " + msg); } else { console.log("  ✓ " + msg); } };
+const isPrimitive = (v) => v === null || (typeof v !== "object" && typeof v !== "function");
+
+// 1. THE BUG: a Hermes actor {name, on_disk, counts} must yield primitive id/label.
+{
+  const a = normaliseAssignee({ name: "scout", on_disk: true, counts: { ready: 1 } });
+  ok(a.id === "scout", "name → id");
+  ok(a.label === "scout", "name → label");
+  ok(isPrimitive(a.id) && isPrimitive(a.label), "id and label are primitives, never the raw object");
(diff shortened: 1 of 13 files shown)


thinmintdev:
Follow-up commit (099dbdb): also fixed the latent field-name mismatch noted above.

`normaliseTask()` emits the task field as **`assignee`**, but `kcard.jsx`, `lane.jsx` (by-profile grouping), and `board-view.jsx` (the Assignee filter) all read the nonexistent **`task.profile`** — so cards showed "unassigned", everything collapsed into one "unassigned" sublane, and the Assignee filter matched nothing. All three now read `task.assignee`. (`task-drawer.jsx:336` `r.profile` left alone — that's a *run* object with its own `profile`.)

New test `board-assignee-display-v3.spec.ts` (RED-first → GREEN). Full board+dashboard e2e now **103 passed**.

thinmintdev:
Folded the **new-task modal** into this PR (commit e8f0896).

The lane `+` was a stub that silently POSTed a hardcoded `{title:"New task"}` — a blank task that the dispatcher auto-advanced out of Ready (the reported "no modal, lands in Ready empty"). Now `+` opens an explicit `NewTaskModal`; nothing is POSTed until a title is entered and **Create** is hit. The modal's assignee dropdown is fed the normalised actor lists from this PR's `boardActors` fix, so it can't hit the `{name,on_disk,counts}` render crash.

Provenance: the `NewTaskModal` component + board-view wiring originated in a parallel session (uncommitted on the CT105 runtime); consolidated here so the crash fix, assignee-display fix, and modal ship as one coherent board PR. That session's separate `useBoardChat` SSE-protocol work is **left out of scope**.

Tests: +3 board-selector cases (opens-without-POST, title+Create POSTs `{title,body,status:lane}`, Cancel closes without POST) + `board-columns-render-v3` (board renders the real Hermes `{columns:[...]}` payload). Full board+dashboard e2e: **107 passed**.

Deployed to CT105 for testing (bundle `index-BUdMHwot.js`).

PR #961, opened 2026-06-23
feat(stacks): Focus layout from the design handoff

thinmintdev:
Replaces the Profiles-page-clone Stacks display with the **Focus layout** from the hal0 design-system handoff (`stacks_overhaul`). The active stack becomes a full-width hero with per-slot live state; the rest are a compact library grid below.

Implemented in the React codebase per the handoff README (recreate with the token set, don't ship the HTML).

### Layout
- **Header** — "RUNTIME" eyebrow + `Stacks` + Import / Snapshot / New stack.
- **Toolbar summary** — `N stacks · M active · K missing models`.
- **Hero (active stack)** — amber filament + glow, per-slot cells (live status dot + state + profile + model), tags, Edit / Export / Re-apply.
- **Library grid** — per-stack slot list + clone / edit / delete / export / Load.

### Load + pull flow (the capability ask)
- **LoadDialog** — confirms the load, lists each slot's model with present/not-found state, warns on missing models ("Load anyway" / "Pull missing first").
- **PullDialog** — pull each missing model, or "Queue all", before loading.

### Real wiring (not stubs)
- Load → apply commit + converge (create-on-apply from the prior PR).
- Pull → `POST /api/models/{id}/pull`.
- Export → portable `.hal0stack.json` download. Import / New / Snapshot reuse the existing flows.
- State from `useStacks` + live `useSlots` status + `useModels` (registry → "model not available → pull").

### Notes
- New `stacks.css` (`.stk-*`) matching the handoff; shims the 4 tokens it assumes that `dashboard.css` lacks (`--rad-xs`, `--ls-tight`, `--ls-title`, `--shadow-menu`).
- Fixes the prior **activity-pill / slot-count** overlap and the **editor status-dot** overlap (new header + hero markup).
- Only the **Focus** view is built — the handoff's other three (Cards/Pipeline/List) were prototype variants, per direction.
- `tsc --noEmit` + `vite build` clean. **Verified live on CT105** — hero, library, and LoadDialog all render against real data.

### DIFF
--- a/ui/src/dash/profiles.jsx
+++ b/ui/src/dash/profiles.jsx
@@ -75,115 +75,75 @@ function SlotPill({ name }) {
 
 // ── Profile card ──────────────────────────────────────────────────────────────
 
+// Info row in the stacks-card "slotlist" idiom: hued dot · label · value.
+function PfRow({ label, value, hue }) {
+  return (
+    <div className="stk-lib-slotrow">
+      <span className="pf-row-dot" style={hue ? { background: hue, boxShadow: `0 0 6px ${hue}` } : null} />
+      <span className="stk-csr-name mono">{label}</span>
+      <span className="stk-csr-model mono">{value}</span>
+    </div>
+  );
+}
+
+// Profile card — adopts the Stacks library-card shell (.stk-lib-*) so the
+// Profiles and Stacks grids read as one family. Same data + actions as before.
 function ProfileCard({ p, index, onEdit, onClone, onDelete }) {
   const meta = bk(backendOf(p));
   const isSeed = !!p.seed;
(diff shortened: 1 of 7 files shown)


thinmintdev:
Added: **Profiles cards now match the Stacks library cards.** Restyled `ProfileCard` onto the shared `.stk-lib-*` shell (name + intent header with a backend-hued chip + bench tok/s; quant/mtp/image/flags as compact rows; footer with seed/custom tag + clone/edit/delete icon buttons), and narrowed the grid to match (minmax 360→300, gap 14→10). e2e selectors in profiles-crud/profiles-page specs updated to the new card classes. Verified live on CT105 — the Profiles and Stacks grids now read as one family.

Issue #626, opened 2026-06-07
api: persist model-pull jobs to disk (reuse updater disk-mirror) so restart doesn't 404 the poll


thinmintdev:
## What to build

Updater jobs are disk-mirrored to `/var/lib/hal0/update-jobs` and survive restart; model-pull jobs live only in `app.state.model_pull_jobs`, so a restart mid-pull **404s the poll** — despite the docstring claiming "Mirror of the updater route shape." Extend the same disk-mirror pattern (same `make_job` primitive) to model-pull jobs.

Pointers: `updater.py:114`; `routes/models.py:1230`. Audit Q4.2.

## Acceptance criteria

- [ ] Model-pull jobs mirrored to disk like updater jobs
- [ ] Polling a model-pull job after a restart returns status, not 404

## Blocked by

None - can start immediately

PR #1176, opened 2026-07-10
feat(models): HuggingFace model update detection + one-click Update all

thinmintdev:
## What & why

Installed HF models can silently go stale when the source repo re-uploads a quant (same filename, new bytes). Today there's no signal and no easy refresh. This adds:

- an **"update available" indicator** on each stale model row, and
- a **"Update all · N"** top button that re-pulls every stale model in one click.

## How it works

The key realisation: the content SHA-256 we already capture at pull time (`Model.metadata["sha256"]`, the streamed digest) equals HuggingFace's advertised LFS `oid`. So "is this stale?" is just: **stored sha ≠ the repo's current `main` LFS oid** for that filename. No new dependency, no re-download to check — one metadata read against `/api/models/<repo>/tree/main` (the same endpoint inspect already uses).

## Backend (`src/hal0`)

- **`registry/updates.py`** (new): `check_updates()` compares each checkable installed model against its repo tree.
  - LFS-only + anchored-only: a row without a stored sha, or whose remote file isn't LFS-backed, reports `update_available=false` rather than guess — the dashboard never nags about a model it can't actually verify.
  - Per-repo tree cache (10-min TTL) so N models from one repo cost a single HF call; bounded concurrent fanout; every failure path degrades to "unknown", never 500s.
- **`api/routes/models.py`**:
  - `GET /api/models/updates` — freshness report (`?force=true` bypasses the cache).
  - `POST /api/models/updates/check` — force-refresh sibling.
  - `POST /api/models/update-all` — re-pull every stale model (skips any already mid-pull).
  - Extracted **`_spawn_hf_pull`** so the batch update and the single-model pull share identical job / persist / event / SSE-stream plumbing. New static-path routes are declared before `/{model_id}` so the matcher doesn't treat `updates` as a model id.

## Frontend (`ui`)

- **Hooks** (`useModels.ts`): `useModelUpdates` / `useCheckModelUpdates` / `useUpdateAllModels`, polled on a separate 5-min cadence so the main `/api/models` list stays a cheap disk read.
- **`models.jsx`**: `ModelRow` renders a warm `update` chip when stale; a `Update all · N` button appears in the page header only when something is stale, kicks the batch re-pull, and reuses the existing SSE + footer-event progress rails.
- Styling matches the existing pulling/warming accent convention.

## Tests

- `tests/registry/test_updates.py` — sha-vs-oid comparison, `sha256:` prefix normalisation, non-LFS / no-sha / unreachable degradation, per-repo fetch dedup.
- `tests/api/test_model_updates_routes.py` — report counts, `force` refresh, update-all kicks pulls / skips fresh / defers to in-flight jobs, and the `/updates` vs `/{model_id}` route-ordering guard.
- `ui/tests/e2e/specs/model-updates-v3.spec.ts` — indicator + button visibility and the update-all POST.
- Fixture fix: `apiMock` now mocks `/api/models/pulls` to return a list (the catch-all object was crashing `usePullsList`), which also repaired several pre-existing models-v3 e2e failures.

All backend (`ruff`, `mypy`, pytest) and UI (`tsc`, unit, models e2e) checks pass locally.

https://claude.ai/code/session_018JUyAmsZs6eFRJNij7UBZx

---
_Generated by [Claude Code](https://claude.ai/code/session_018JUyAmsZs6eFRJNij7UBZx)_

### DIFF
--- a/src/hal0/api/routes/models.py
+++ b/src/hal0/api/routes/models.py
@@ -1058,6 +1058,120 @@ def _hf_repo_for_model(registry: Any, model_id: str) -> str | None:
         return None
 
 
+# ── HF model updates (compare installed sha vs repo main) ─────────────────────
+#
+# These static-path routes are declared BEFORE ``/{model_id}`` so FastAPI's
+# in-order matcher resolves ``/updates`` / ``/update-all`` here instead of
+# treating "updates" as a model id (same ordering trick as ``/pulls``).
+
+
+@router.get("/updates")
+async def list_model_updates(request: Request, force: bool = False) -> dict[str, Any]:
+    """Report which installed HF models have a newer file on their repo.
+
+    Compares each checkable registry row's pull-time content sha256 against
+    the repo's current ``main`` LFS oid (see :mod:`hal0.registry.updates`).
+    ``?force=true`` bypasses the per-repo tree cache to re-probe HF.
+
+    Response::
+
+        {
+          "updates": [{"model_id", "hf_repo", "hf_filename",
+                       "update_available", "current_sha", "remote_sha",
+                       "reason"}, ...],
+          "count": <models checked>,
+          "available": <models with an update>
+        }
+
+    Never 500s the dashboard: an unreachable repo degrades that row to
+    ``update_available=false`` with ``reason="repo_unreachable"``.
+    """
+    from hal0.registry.updates import check_updates
+
+    registry = request.app.state.model_registry
+    infos = await check_updates(list(registry.list()), force=force)
+    updates = [i.as_dict() for i in infos]
+    available = sum(1 for u in updates if u["update_available"])
+    return {"updates": updates, "count": len(updates), "available": available}
+
+
+@router.post("/updates/check")
+async def check_model_updates(request: Request) -> dict[str, Any]:
+    """Force a fresh HF probe (cache-bypass) and return the update report.
+
+    POST sibling of ``GET /updates`` mirroring the app-updater's
+    ``/api/updates/check`` — the dashboard's "Check for updates" affordance
+    calls this to re-probe HF regardless of the tree cache.
+    """
+    return await list_model_updates(request, force=True)
+
+
+@router.post("/update-all", status_code=202)
+async def update_all_models(request: Request, background: BackgroundTasks) -> dict[str, Any]:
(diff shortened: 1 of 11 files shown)


thinmintdev (commented):

Four parallel implementations of this feature (#1176, #1178, #1179, #1181) were reviewed head-to-head; **#1178 was selected** because it is the only one that pins the update re-pull to the model's existing path. This PR will be closed as superseded once #1178 merges.

### Why not this one
- **Update can relocate models and orphan old bytes (medium-high).** `update_all_models` resolves capability via `_resolve_pull_capability(request, id, None)` → the row's `capabilities[0]`, and `run_pull` re-derives the destination through `_final_path_for_entry` — never the row's current `path`. A pre-layout model at `<store>/<id>/<file>` gets its update written to `<store>/chat/<id>/model.gguf`, the registry repoints, and the old multi-GB file is left untracked; path-pinned slot configs keep serving stale weights while the dashboard reports "updated". #1178 fixes exactly this with `dest_override`.
- **Subdirectory-hosted files are never checkable (low-medium, fails safe).** The tree fetch omits `?recursive=true`, so `UD-Q4_K_XL/model.gguf`-style filenames (standard unsloth/bartowski layout) never appear and are silently unverifiable. The basename-fallback unit test fabricates a tree shape the real API won't return.
- **Dead "Check for updates" surface:** `POST /api/models/updates/check` + `useCheckModelUpdates` exist but nothing in the UI consumes them.
- **`httpx.InvalidURL` escapes the catch** (it subclasses `Exception`, not `HTTPError`) — a hand-registered row with a malformed `hf_repo` can 500 `GET /api/models/updates`.
- Same-repo concurrent checks can stampede the tree fetch (the 1-call cache-dedupe test only passes under MockTransport scheduling).

### Genuinely good parts
The `_spawn_hf_pull` extraction (single + batch pulls sharing job/persist/event/SSE plumbing verbatim, including #626 restart-persist) is a refactor worth having regardless; the fail-safe posture ("never nag about a model we can't verify") is applied consistently; documentation density is excellent; and the γ-suite failure was verified to be main's pre-existing NPU-spec flake, not this PR (its own 3 new specs passed).

Scores vs siblings: correctness 7, design 8, tests 7, completeness 7 — edged out by #1178 (8/9/8/7) on the safety of the core action.

---

PR #1178, opened 2026-07-10
feat(models): HF update check + per-row indicator + Update all

thinmintdev:
## What

Lets operators update installed models when their HuggingFace files change upstream: each catalog row shows an **update ↑** indicator when a newer version exists, the detail pane gets an **Update** action, and a top **Update all (N)** button re-pulls everything stale in one click.

## How it works

**Detection (no schema migration).** A pulled model already records the sha256 of its bytes (`metadata.sha256` — the LFS object id for HF-hosted files). HF's tree API exposes the *current* LFS sha256 per file, so "update available" is simply: stored sha ≠ repo sha for the same `hf_repo`/`hf_filename`. Existing installs become update-aware without re-registration.

**Backend**
- `hal0/registry/update_check.py` (new): `fetch_remote_lfs_shas()` — one tree fetch per unique repo, concurrency-capped, fail-soft (unreachable/gated repo → "couldn't check", never a phantom update); `evaluate_model_update()` — the pure per-model verdict.
- `GET /api/models/updates/check` — runs the probe behind a 1-hour app-state cache (`?refresh=1` forces), single-flighted with a lock. Registered above the `/{model_id}` catch-all, like `/pulls`.
- `/api/models` rows now carry `update_available`, merged from the snapshot but **recomputed against the row's current sha** — so applying an update clears the badge on the next 30s poll without waiting out the check TTL. The list path never touches huggingface.co.
- `POST /api/models/{id}/update` — an update is a re-pull pinned to the row's **existing** path via a new `dest_override` threaded through `run_pull` (staging + checksum verify + atomic `os.replace` all unchanged). Pinning matters: a plain re-pull re-derives the destination through the capability-grouped layout and would relocate pre-layout models, orphaning their old bytes. Jobs are keyed by `model_id` as usual, so pull status/SSE/downloads-pane machinery works untouched.
- `_register_pulled` now also records `metadata.pulled_at` provenance.

**UI (Models view)**
- `useModelUpdatesCheck()` triggers the (server-cached) probe on mount and refreshes the catalog so badges appear immediately.
- `ModelRow` shows the amber `update ↑` chip; `ModelDetail` shows an "update available" chip plus the Update button (`useModelUpdateApply`).
- Header "Update all (N)" (`useModelUpdateAll`) fans out per-model updates with `Promise.allSettled` — one gated repo can't abort the batch. Progress surfaces in the existing Downloads pane.

## Safety / edge cases

- `update_available` is only ever true when **both** a local and remote sha exist and disagree — hand-registered rows (no recorded sha), non-LFS files, renamed/removed upstream files, and unreachable repos never flag.
- Update refuses rows without HF coordinates (422) and directory-shaped paths (FLM trees).
- A corrupt/tampered download fails HF's advertised-sha check *before* install, leaving the current file in place.

## Testing

- `tests/registry/test_update_check.py` — verdict matrix (differing/matching/unreachable/missing/no-local-sha), tree-fetch behaviour via `httpx.MockTransport` (LFS oid extraction, `sha256:` prefix strip, non-LFS skip, token header, failed-repo → `None`), and a full `run_pull(dest_override=…)` in-place replacement asserting bytes, path stability, and refreshed sha/`pulled_at`.
- `tests/api/test_models_updates.py` — check verdicts + TTL cache + `?refresh=1`, row-merge including the post-update self-heal, and the update route's 202/pinned-dest/in-flight-dedupe/404/422 envelopes.
- 18 new tests pass; registry + models suites (258 tests) green; UI `tsc --noEmit` and `vite build` clean. Full suite: 1281 passed, 1 pre-existing environment failure (`test_set_store_rejects_non_writable_path` expects a chmod-0555 dir to be unwritable, but CI-as-root bypasses permission bits — unrelated to this change).

## Notes / follow-ups

- `mmproj` sidecars are not re-pulled by an update (association is preserved); could be added by extending the update route with the row's sidecar filename.
- The check compares `main` — per-revision pinning is plumbed (`hf_download_url(revision=…)`) but not exposed.

https://claude.ai/code/session_01PWJUJweDSUspKGGxKrKqHD

---
_Generated by [Claude Code](https://claude.ai/code/session_01PWJUJweDSUspKGGxKrKqHD)_

### DIFF
--- a/src/hal0/api/routes/models.py
+++ b/src/hal0/api/routes/models.py
@@ -41,6 +41,7 @@ from hal0.registry.pull import (
 )
 from hal0.registry.pull import persist_pull_job as _persist_pull_job
 from hal0.registry.pull import pull_job_file as _pull_job_file
+from hal0.registry.update_check import evaluate_model_update, fetch_remote_lfs_shas
 
 # See slots.py for the writer-gate rationale.
 
@@ -234,9 +235,29 @@ async def list_models(request: Request) -> dict[str, Any]:
     data: list[dict[str, Any]] = []
     seen: set[str] = set()
     filtered = 0
+    # Last HF update-check snapshot (populated by /api/models/updates/check;
+    # never fetched on this hot path). The flag is recomputed against the
+    # row's CURRENT metadata.sha256 rather than replayed from the snapshot,
+    # so applying an update clears the badge on the next poll without
+    # waiting for the check TTL to expire.
+    update_checks: dict[str, Any] = {}
+    _upd_state = getattr(request.app.state, "model_update_state", None)
+    if isinstance(_upd_state, dict):
+        update_checks = _upd_state.get("models") or {}
     for entry in registry.list():
         dumped = _model_to_dict(entry)
         dumped["installed"] = True
+        chk = update_checks.get(entry.id)
+        if isinstance(chk, dict):
+            remote_sha = chk.get("remote_sha256")
+            local_sha = (entry.metadata or {}).get("sha256")
+            dumped["update_available"] = bool(
+                isinstance(remote_sha, str)
+                and remote_sha
+                and isinstance(local_sha, str)
+                and local_sha
+                and remote_sha != local_sha.lower()
+            )
         dumped.setdefault("object", "model")
         dumped.setdefault("created", now)
         dumped.setdefault("owned_by", "local")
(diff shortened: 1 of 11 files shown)


thinmintdev:
**γ-suite failure investigated — it was pre-existing main breakage, and this PR now fixes most of it.**

The γ-suite run on this PR failed 29 specs, but main's run at this branch's base commit ([66726e6](https://github.com/Hal0ai/hal0/actions/runs/29120417175)) fails the **identical 29 specs** — the suite has been red on every recent main push, so none of them were introduced here.

Root cause (reproduced locally): the e2e fixture's `/api/*` catch-all answers `GET /api/models/pulls` with `{}`, and `usePullsList` calls `jobs.some(...)` on that non-array — a render-time TypeError that unmounts the whole React tree, blanking every Models-page spec (and mobile-nav, which navigates there).

Pushed in c9e402a:
- `Array.isArray` guard in `usePullsList` so a mocked/misbehaving backend can't crash the dashboard — this alone takes the local γ-suite from 29 failed to 4.
- Mock-allowlist entry for the new `/api/models/updates/check` (networkFirst) so backend-less dev builds don't retry-spam a 500ing endpoint.

Local full γ-suite after the fix: **375 passed / 4 failed** — the remaining 4 are the NPU-occupancy-card specs (`slots-v3`, `inference-pane-v3`, `npu-occupancy-v3`) that also fail on main's baseline; they're `/slots`-page mock-data drift unrelated to this feature, left for a separate fix.

---

thinmintdev:
γ-suite on the merge commit (9bffb90): **381 passed / 3 failed** — down from main's baseline of 29.

The remaining 3 (`slots-v3:36`, `inference-pane-v3:230`, `npu-occupancy-v3:92`) are pre-existing spec drift from the #1174 NPU-card redesign: they assert `.cslot .cslot-strip` and `.cslot .slot-ctrls` / `.sctrl.stop`, but the redesigned `ComboSlot` renders `.cslot-top`/`.cslot-mx`/`.cslot-foot` with no strip and no inline lifecycle controls. They fail identically on main and need a spec rewrite against the new card design (including deciding whether the Stop→`POST /api/slots/npu/unload` assertion still has a UI affordance to attach to) — best done as a separate PR.

Also merged `origin/main` into the branch (9bffb90) to clear the conflict with #1177 — which, amusingly, had landed the same `Array.isArray` guard in `usePullsList` independently.

---

thinmintdev (commented):
## Code review — selected to merge among the four competing HF-update PRs (#1176, #1178, #1179, #1181)

**Verdict: MERGE.** Four parallel implementations of this feature were reviewed head-to-head. This one wins on the axis that matters most: it is the only one whose *update action* is actually safe.

### The differentiator: `dest_override` pinning is real and correct
Verified against main's pull machinery: a plain re-pull resolves capability from the registry row (`_resolve_pull_capability` → `capabilities[0]`, which nearly every row has) and re-derives the destination via `_final_path_for_entry` → `<pull_root>/<capability>/<id>/model.gguf`, never consulting the row's current `path`. So the siblings' "Update = re-pull" flow relocates any pre-capability-layout model (orphaning the old multi-GB file on disk), silently breaks ComfyUI-subdir checkpoints (new bytes land outside the tree ComfyUI scans while it keeps loading the stale file), and leaves path-pinned slot configs serving old weights. This PR's `dest_override` threading pins the re-pull to the existing path with the staging + sha-verify + atomic `os.replace` pipeline intact, and tests exercise it end-to-end through the real `run_pull`.

Other verified strengths: conservative verdict matrix (`update_available` only when both shas exist and differ — no phantom nags for hand-registered/non-LFS/unreachable rows, with machine-readable reason codes); correct single-flight lock; the `list_models` merge recomputes the flag against the row's *current* sha, so the badge self-clears on the next 30s poll after an update with zero invalidation plumbing.

### Findings (non-blocking, follow-up material)
- **Cross-filesystem `dest_override` (medium-low):** staging lives under `pull_root/.tmp`; a row whose `path` is on a different mount (e.g. `/mnt/ai-models/...`) makes `os.replace` raise EXDEV *after* the full download. Fails clean (job → failed), but wastes the transfer. Consider a same-directory temp file + `os.replace` fallback when the override crosses mounts.
- **HF tree pagination unhandled (low):** `/tree/main?recursive=true` paginates (~1000 entries); large repos silently drop files → `file_missing_or_not_lfs`. Fails safe (missed update, never a phantom one).
- **Stale-snapshot phantom badge (low):** if a model id is re-registered with different HF coords inside the 1h TTL, the cached verdict keyed by id can show a phantom badge until refresh/TTL. Self-heals.
- **No UI "Check now"** — the server supports `?refresh=1` but nothing sends it; worst-case staleness is 1h TTL + 30min poll.
- **mmproj sidecars** are neither checked nor re-pulled (shared gap across all four PRs); no slot reload after update, so a loaded model keeps serving the old inode until restart.

### Verification
CI (after the merge-of-main): python + ui green; the 3 γ-suite failures are byte-identical to main's own pre-existing NPU-spec failures (verified against main's push runs). Independently ran the new suites locally: 18/18 new tests, `tests/registry` + `tests/api` green, `tsc --noEmit` clean.

### What the siblings had that this lacks (candidate follow-ups)
- #1179's model-cards subsystem (cached README in the detail pane) — clean, well-tested, and orthogonal; portable as a follow-up.
- #1181's `hal0:notify` handoff to the #1177 notification bell (its event contract was verified correct) and its repairs to the 3 broken NPU γ-specs on main.
- A "Check for updates now" UI affordance (#1176 built the endpoint; nobody wired the button).

---

PR #1179, opened 2026-07-10
feat(models): HF update checks + Update all + cached model cards

thinmintdev:
## What

Lets users update their models when the HF repo re-uploads a newer build of the same file, and read each model's README/model card in the dashboard.

### Update checks (backend)

- **`hal0/registry/updates.py`** — `HEAD`s each HF-sourced row's `resolve/main` URL and compares the advertised `X-Linked-ETag` LFS sha256 against the sha256 the pull recorded in `metadata.sha256` (the same header the pull path already verifies against, WS-12). Remote ≠ local ⇒ update available. Rows that can't be compared honestly report `unknown` (no local sha — scanned/hand-registered files; non-LFS files) or `error` (HF unreachable), never a false positive.
- Probes are TTL-cached (30 min) per `(repo, filename)` coordinate; `update_available` is recomputed against the registry's *current* local sha at read time, so a completed re-pull flips a row to up-to-date with no cache invalidation.
- **`GET /api/models/updates`** — statuses + `available` list; `?refresh=true` re-probes and emits a `model.updates_available` event.
- **`POST /api/models/updates/apply`** — "Update all": re-pulls every outdated model (or an explicit `model_ids` subset) through the existing pull machinery — same job store, SSE progress, footer events, and atomic in-place `os.replace` swap. Active pulls and rows without HF coords are skipped with reasons.
- Mounted before the main models router so the literal `/api/models/updates` path wins over `GET /{model_id}`.

### Model cards (backend)

- **`hal0/registry/cards.py`** + **`GET /api/models/{id}/card`** — downloads the repo's `README.md` (`raw/main`, plain git blob) once and caches it under `/var/lib/hal0/model-cards/<id>.md`; subsequent reads are offline-fast. 512 KiB cap with a truncation marker; `?refresh=true` re-fetches with a stale-cache fallback when HF is down; the cache is dropped on model delete.

### Dashboard

- Amber **`update` chip** on catalog rows whose file has a newer build upstream (`data-testid="mdl-row-update-available"`).
- **`Update all · N`** button in the Models page header (`mdl-update-all`) → `POST /updates/apply`, jobs land in the existing Downloads pane.
- Detail pane: **update chip + one-click Update** action for the selected model.
- Collapsible **"Model card"** section in the detail sidebar — View README / Refresh, rendered from the server cache.
- New hooks (`useModelUpdates`, `useModelUpdatesApply`, `useModelCard`, `useModelCardRefresh`) fail soft against older backends.

## Tests

- `tests/registry/test_updates.py` (11) — sha compare matrix, TTL/refresh caching, coordinate dedup, error capture, post-re-pull recompute.
- `tests/registry/test_cards.py` (9) — fetch/persist/cache, refresh, stale fallback, truncation, 404s.
- `tests/api/test_model_updates_routes.py` (8) + `tests/api/test_model_card_route.py` (4) — route shapes, path-shadowing guard, apply skip reasons, delete cascade.
- `ui/tests/e2e/specs/models-updates-v3.spec.ts` (4) — row badge, Update all → POST, detail affordances, model card render.
- Fixture fix: mocked `/api/models/pulls` as an array — the `{}` catch-all crashed `usePullsList` and took the whole page down in mock mode, which un-breaks the pre-existing `models-v3.spec.ts` (1/12 → 12/12 passing).

`tests/registry` + `tests/api`: 1443 passed; the single failure (`test_settings_models_store::test_set_store_rejects_non_writable_path`) also fails on a clean `main` checkout — pre-existing, unrelated.

## CI status note

The γ-suite (chromium) check on this PR reports **3 failed / 385 passed** — the 3 failures are the NPU-occupancy specs (`slots-v3.spec.ts:36`, `npu-occupancy-v3.spec.ts:92`, `inference-pane-v3.spec.ts:230`). These same specs already fail on `main`: the base commit's run ([66726e6e, run 29120417175](https://github.com/Hal0ai/hal0/actions/runs/29120417175)) shows **29 failed / 349 passed**, a superset including these 3. This PR's `/api/models/pulls` fixture fix repairs the other 26; the remaining 3 are unrelated to this diff (NPU occupancy card / FLM slot card mocks) and are left for a separate fix.

https://claude.ai/code/session_01MJfPACpbJTUBsokg4wKZUw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ applying. Add those subsections to a version's section to surface them; see
 
 ## [Unreleased]
 
+### Added
+- HF model update checks: `GET /api/models/updates` probes each HF-sourced registry row's `resolve/main` URL (HEAD, `X-Linked-ETag` sha256 vs the sha recorded at pull time) and reports which installed models have a newer build upstream; results are TTL-cached and shared per (repo, file) coordinate.
+- `POST /api/models/updates/apply` — "Update all": re-pulls every outdated model (or an explicit `model_ids` subset) through the existing pull machinery (same jobs, SSE progress, footer events, atomic in-place swap).
+- Models dashboard: an amber `update` chip on catalog rows with a newer HF build, an `Update all · N` header button, and an update chip + one-click `Update` action in the detail pane.
+- Model cards: `GET /api/models/{id}/card` downloads the repo's `README.md` once and caches it under `/var/lib/hal0/model-cards/` (stale-cache fallback when HF is unreachable; cache dropped on model delete). The detail sidebar grows a collapsible "Model card" section with a View README toggle and a Refresh action.
+
 ## [0.9.6] — 2026-07-10
 
 ### Highlights
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -70,6 +70,7 @@ from hal0.api.routes import (
     images,
     installer,
     logs,
+    model_updates,
     models,
     npu,
     power,
@@ -1118,6 +1119,9 @@ def create_app() -> FastAPI:
     # Read-only ComfyUI "generation engine" status for the slots-page Image-Gen
     # tab (docker + systemd + ComfyUI HTTP), plus arbiter switchover controls.
     app.include_router(comfyui.router, prefix="/api/comfyui", tags=["comfyui"])
+    # HF update checks — registered BEFORE the main models router so the
+    # literal /api/models/updates path wins over its /{model_id} route.
+    app.include_router(model_updates.router, prefix="/api/models/updates", tags=["models"])
     app.include_router(models.router, prefix="/api/models", tags=["models"])
     # Issue #311: HuggingFace Hub discovery (search proxy). Sits next
     # to the models surface so the dashboard's "Search HF" button has a
--- a/src/hal0/api/routes/model_updates.py
+++ b/src/hal0/api/routes/model_updates.py
@@ -1,0 +1,188 @@
+"""Model update-check endpoints (mounted under /api/models/updates).
+
+The Models dashboard needs two things for HF-sourced rows:
+
+* ``GET  /api/models/updates``        — which installed models have a newer
+  build of their file sitting in the HF repo (sha probe, TTL-cached)?
+* ``POST /api/models/updates/apply``  — re-pull the outdated ones (all of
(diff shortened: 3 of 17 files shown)


thinmintdev (commented):

Four parallel implementations of this feature (#1176, #1178, #1179, #1181) were reviewed head-to-head; **#1178 was selected** because it is the only one that pins the update re-pull to the model's existing path. This PR will be closed as superseded once #1178 merges — but it contains the strongest candidate for a follow-up port (model cards, below).

### Why not this one
- **Update can relocate models and orphan old bytes (high).** Both the apply endpoint and the detail-pane Update button funnel into `run_pull`, which re-derives the destination via `_final_path_for_entry` from the row's first capability — never the row's current `path`. Pre-layout models get duplicated into the capability tree with the old multi-GB file orphaned; ComfyUI-subdir checkpoints are worse: `_resolve_pull_capability` hits the registry branch first and loses `comfyui_subdir`, so the update lands outside ComfyUI's tree (renamed `model.gguf`) while ComfyUI keeps loading the stale file — the update is silently ineffective. The module docstring's "atomic in-place swap" claim only holds for models already in the capability layout, and the route tests stub `_run_pull_with_events`, so nothing covers where bytes land.
- **Card truncation cap is measured in bytes but truncates in characters** (`cards.py`) — a CJK-heavy README can store ~4× the 512 KiB cap; the body is also fully buffered before capping.
- **`hf_token` is dropped when a client is injected** into `updates.py` (latent; production path unaffected).
- **Apply body parsing is gated on `Content-Length`** — a chunked POST with `model_ids` silently updates *everything* flagged.
- The `model.updates_available` event only fires on `?refresh=true`, which the UI never sends — the bell can't learn about updates from the normal poll path; probe errors are negatively cached for the full 30-min TTL.

### Genuinely good parts
The detection layer is the most technically sound of the four (reuses `_expected_sha256_from_response` — the exact header the pull verifier checks — so both sides of the comparison come from the same verified pipeline); the coordinate-keyed cache with read-time recompute is elegant and explicitly tested; the route-shadowing regression test is a nice touch; and the **model-cards subsystem** (`cards.py` + cached README panel with stale fallback, truncation, delete cascade) is a complete, well-tested orthogonal feature worth porting on top of #1178 — with the two card bugs above fixed in the port.

Scores vs siblings: correctness 6.5, design 8, tests 8, completeness 9 — the feature-richest of the four, but the core action's data-affecting flaw outweighs the extras.

---

PR #1181, opened 2026-07-10
feat(models): HF update checks — row badge, Update all, notification-bell handoff

thinmintdev:
## What

Lets users update installed models when their files change on HuggingFace:

- **`POST /api/models/check-updates`** — compares each installed HF-backed registry row's recorded content sha256 (written by `_register_pulled` on every pull) against the LFS sha256 HF currently advertises for the same repo/filename (the `X-Linked-ETag` on `resolve/main` — the exact digest the pull path verifies against). "Update available" means a re-pull would land different bytes.
- **Row indicator** — outdated rows in the Models catalog get an amber `update` chip; the detail pane shows an `update available` chip and a per-model **Update** button.
- **Update all** — an `Update all · N` header button re-pulls every outdated model through the existing pull pipeline (upsert + hash re-verify), with progress in the Downloads pane. One failed start doesn't block the rest.
- **Notification bell handoff** — `useModelUpdates` publishes a `hal0:notify` CustomEvent (stable id keyed on the outdated set, `kind: "update"`, `link: "#models"`) matching the topbar bell's dedupe + persistent-dismissal contract (bell lands separately in `claude/notification-bell-topnav-n8i58b`). The backend also emits `models.updates_available` on the EventBus when a live sweep finds updates.

## Design notes

- **No schema change.** Every pulled row already stores `metadata["sha256"]`, and HF advertises the current LFS sha on a cheap HEAD — detection is a pure comparison. Rows without HF coords are skipped; rows without a recorded local sha are reported but never flagged (hashing tens of GB on request isn't worth a badge).
- **Caching.** HF side cached per (repo, filename) for 15 min (`force: true` bypasses); the installed sha is read fresh from the registry on every call, so a badge clears as soon as a re-pull lands — no invalidation dance. The dashboard's 60s poll therefore costs one local registry read per cycle.
- **Applying an update IS a re-pull** — `POST /api/models/{id}/pull` already upserts, re-verifies the hash, and dedups queued/running jobs; no new download path.
- Per-file HF failures (404, gated, non-LFS) land on the row's `error` field; the endpoint stays 200.

## Tests

- `tests/api/test_models_update_check.py` — 9 cases (httpx `MockTransport`, same pattern as the inspect tests): sha diff/match, skip rules, missing-local-sha, TTL cache, `force`, self-healing after a re-pull, per-row errors, non-LFS files. All pass; `tests/api/test_models_routes.py` + `test_models_crud.py` + `tests/registry` (241) still green. `ruff` + `mypy` clean (no new mypy findings vs baseline).
- `ui/tests/e2e/specs/models-v3.spec.ts` — 4 new specs: badge only on outdated rows, Update all count + re-pull POSTs, empty state, and the `hal0:notify` contract shape. `tsc --noEmit` + `vite build` pass. Note: the full Playwright suite is flaky in my sandbox (pre-existing specs fail identically on a clean checkout — blank-page dev-server flake); a direct browser probe confirms the badge + button render correctly with a mocked outdated payload.

https://claude.ai/code/session_014gj5YALNZfAHb7qMvdCr8Q

---
_Generated by [Claude Code](https://claude.ai/code/session_014gj5YALNZfAHb7qMvdCr8Q)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ applying. Add those subsections to a version's section to surface them; see
 
 ## [Unreleased]
 
+### Added
+- Model update checks against HuggingFace: `POST /api/models/check-updates` compares each installed model's recorded sha256 against the LFS sha HF currently serves at `main` (per-file 15-min cache, per-row errors). The Models page shows an amber `update` badge on outdated rows, an `Update all · N` header button that re-pulls every outdated model through the existing pull pipeline, a per-model Update button in the detail pane, and publishes a `hal0:notify` event for the topbar notification bell.
+
 ## [0.9.6] — 2026-07-10
 
 ### Highlights
--- a/src/hal0/api/routes/models.py
+++ b/src/hal0/api/routes/models.py
@@ -34,6 +34,8 @@ from hal0.registry.pull import (
     PullInvalidSource,
     PullJob,
     PullJobNotFound,
+    _expected_sha256_from_response,
+    hf_download_url,
     list_persisted_jobs,
     make_job,
     run_flm_pull,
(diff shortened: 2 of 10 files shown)


thinmintdev (commented):

Four parallel implementations of this feature (#1176, #1178, #1179, #1181) were reviewed head-to-head; **#1178 was selected** because it is the only one that pins the update re-pull to the model's existing path. This PR will be closed as superseded once #1178 merges — but two pieces of it are prime follow-up ports (bell handoff, γ-spec repairs; below).

### Why not this one
- **Update can relocate models and orphan old bytes (high).** Both apply paths send a body-less `POST /api/models/{id}/pull`; `_resolve_pull_capability` returns the row's `capabilities[0]` and `_final_path_for_entry` routes to `<pull_root>/<capability>/<id>/model.gguf`, never consulting the row's current `path`. Pre-layout installs get duplicated with the old file orphaned; worst case is ComfyUI checkpoints — the registry branch returns `comfyui_subdir=None`, so an image model's update lands outside ComfyUI's tree renamed to `model.gguf`, and the image tab does render the Update path, so it's reachable. #1178 fixes exactly this with `dest_override`.
- **The bell handoff only works while the Models page is open** — `useModelUpdates` is mounted solely in `ModelsView`; a user who never visits `#models` never gets notified, which undercuts the headline feature.
- **Bell messages never self-clear** after Update-all succeeds, and a changed outdated set adds a second message while the stale one remains.
- **Errors are negatively cached for the full 15-min TTL** with no force/`Check now` affordance in the UI to recover.
- POST-only endpoint (no GET variant); Update-all is all-or-nothing with failures surfaced only as "Started X/N".

### Genuinely good parts
The fresh-local/TTL-remote cache split (badge self-clears after re-pull with zero invalidation logic) is the cleanest cache design of the four and explicitly tested; the sha comparison reuses the exact digest the pull verifier checks; the `hal0:notify` contract was verified to match the #1177 bell as landed (dedupe id, kind, link, persistent dismissal); the backend test suite is the best of the four (call-count cache proofs, self-clear test, per-row error capture). The drive-by repairs to the 3 NPU γ-specs that are red on main (`slots-v3`, `npu-occupancy-v3`, `inference-pane-v3`) are correct and main needs them — worth extracting into their own PR regardless of this one's fate.

Scores vs siblings: correctness 6.5, design 8, tests 8.5, completeness 8.

**Recommended follow-ups on top of #1178:** (1) port the γ-spec repairs, (2) port the `hal0:notify` bell handoff — moved to app-level polling so it fires without the Models page open, with a clear-on-empty-set rule.

---

PR #1182, opened 2026-07-10
fix: install/setup/update audit follow-ups — manifest pins on prod, honest NPU setup, doc + harness catch-up

thinmintdev:
Follow-up to the v0.9.6 install/setup/update audit. The full findings → fix mapping lives in `handoffs/install-setup-update-fixes-plan-2026-07-10.md` (committed here). Six tiers, one commit each:

## Functional fixes

- **Toolbox manifest pins were unreachable on every production install.** Nothing installs `/etc/hal0/manifest.json`, and the loader's only fallback resolved `manifest.json` relative to the source file — which on a non-editable pip install is the venv's `lib/`, where no manifest exists. `load_manifest()` returned `{}` on prod boxes, so the comfyui digest pin and `hal0 doctor toolbox-pull` silently fell back to tag pulls, and the 0.9.44 FLM digest never reached installed boxes. Fix: resolve `paths.usr_lib()/manifest.json` (`/usr/lib/hal0/current/manifest.json`) between the `/etc` operator override and the dev repo-root fallback — the `current` symlink is atomically swapped by install *and* update, so pins now track the running release with no migration. New unit tests cover the resolution order.
- **Dev installs mutated the host.** The hal0.bench provisioning block hardcoded `/etc/hal0/bench`, `/var/lib/hal0-bench`, and `/etc/systemd/system` even under `--dev` — the one block ignoring prefix isolation. Now uses `ETC_DIR`/`UNIT_DIR`/a prefix-aware state dir, and rewrites the bench units' hardcoded `/usr/lib/hal0/venv` ExecStart when `HAL0_PREFIX` or `--dev` changes the prefix.
- **`img` slot port typo** in the API tier path: `_SLOT_META` said 8186; ComfyUI's canonical port is 8188 everywhere else (seed TOML, schema port-range special case, the same file's `/services` repair mapping).
- **Setup's NPU step over-promised.** The TUI offered "embed + STT + TTS on the NPU", but apply routes only STT to the NPU on opt-in (embed → GPU, TTS → CPU; the trio passenger caps are dormant by design). Copy now describes the actual STT offload and points at the dashboard's NPU drawer for the real per-slot chat/ASR/embed trio toggles.

## Cleanup

- **Orphaned `installer/etc-hal0/slots/npu.toml` removed.** The seed loop copies `{flm,tts,rerank,utility,img}` — never `npu` — yet three places claimed otherwise (and seeding both would port-clash on 8088). Its `[npu]` trio how-to comment moved into `flm.toml`; the seed-shape test retargeted.
- **Docs caught up to v0.9.6:** `setup.mdx` no longer claims the installer passes `--no-slots` (it scaffolds model-less slot structure); `update-and-rollback.mdx` documents the third self-healing pass (stale slot-image retag) and drops dead cosign-hatch version references; `manage-slots.mdx` documents the slot-level image override; new `guides/benchmarks.mdx` covers the in-tree bench engine (CLI, units, politeness window, state root, sudo seam).
- **`tests/harness/installer-test.sh` de-rotted:** removed the Caddy/TLS/auth rows testing features deleted in v0.3.0 (ADR-0012), including a `--no-tls` flag that no longer exists; stopped asserting a `hal0-slot@.service` template the installer never ships (slot units render at runtime); corrected three stale narrative comments.
- **Hygiene:** `manifest.json` version `0.5.0-alpha.1`/`dev` → `0.9.6`/`stable` plus the self-contradicting qwen3tts digest note; `update-toolbox-digests.sh` no longer calls the existing `toolbox.yml` "never-built" and now syncs the manifest version from `pyproject.toml` so it can't rot again; FLM pin comment stopped narrating 0.9.43 history; `uv.lock` regenerated (still recorded hal0 0.9.4).

## Deliberately deferred (see the plan doc)

Wiring `PROFILE_BENCH` to the live bench store, setup writing `[npu]` trio toggles (dormant by explicit product decision), answers-file schema parity, and updater-applied release-manifest `toolbox_images` (superseded by the resolution fix above).

## Testing

- `tests/config` + `tests/updater`: 495 passed; `tests/install` + `tests/cli`: 462 passed; `tests/api` + `tests/providers`: 1516 passed.
- 3 failures are pre-existing root-sandbox artifacts (verified failing on the untouched tree): 2× moonshine ffmpeg 415 tests, 1× read-only-dir rejection test (root writes through `0o555`).
- `ruff check`/`format` clean on touched files; `bash -n` clean on both edited shell scripts.

https://claude.ai/code/session_017U4kurZCVgTnheuFbEKosa

---
_Generated by [Claude Code](https://claude.ai/code/session_017U4kurZCVgTnheuFbEKosa)_


thinmintdev (commented):
## Code review

**Verdict: MERGE.** Every load-bearing claim verified against the actual tree; no regressions found. The tier-1 manifest fix (the highest-risk piece) checks out end-to-end: resolution order is `/etc/hal0/manifest.json` override → `usr_lib()/manifest.json` → repo root (skipped under `HAL0_HOME`); the pin genuinely exists at `usr_lib()` on both paths (installer rsyncs the full repo into `${PREFIX}` and the release tarball explicitly packs `manifest.json`, both atomically swapping `current`); dev checkouts and the test fixtures are unaffected; the two new resolution-order tests pass against the branch. The npu.toml removal and the harness de-rot were confirmed dead surface (the seed loop never copied npu.toml; `install.sh` has zero occurrences of Caddy/`--no-tls`/`hal0-slot@.service`). Port 8188 matches the seed TOML, the schema's img special case, and the `/services` repair mapping.

### Non-blocking follow-ups
- **Dev bench state dir is provisioned but unwired.** The new comment in `install.sh` says "HAL0_BENCH_STATE points the engine at it", but nothing sets `HAL0_BENCH_STATE` — not the installer, not the bench units — and `bench/store.py` defaults to `/var/lib/hal0-bench`. Install-time host mutation is genuinely fixed, but a dev-install `hal0 bench run` still targets the host path while `${PREFIX}/var/lib/hal0-bench` sits unused. Suggest setting `HAL0_BENCH_STATE` in the dev-rendered units (or an env file) as a small follow-up.
- On a box with both a prod install and a git checkout (no `HAL0_HOME`), the checkout now reads the installed release's pins instead of repo-root — arguably correct, but worth a sentence in the `_find_manifest_path` docstring.
- `sed -i "s|/usr/lib/hal0/venv|${VENV_DIR}|g"` breaks on a `HAL0_PREFIX` containing `|`/`&`/`\` (pathological; nit).
- Top-level `README.md` still describes the retired `npu`/`stt-npu`/`embed-npu` seed-slot story — pre-existing rot this PR's doc tier could have swept.

CI: python + ui green; γ-suite failure is main's pre-existing NPU-spec trio (this diff touches no UI/e2e surface).

---

PR #1183, opened 2026-07-10
test(npu): align stale NPU occupancy γ-specs with the reworked card

thinmintdev:
## What

Extracts the γ-spec repairs from #1181 (per its review — "worth extracting into their own PR regardless of this one's fate"). Test-only; no product code.

Three Playwright specs still assert DOM the trio-drawer/NPU-card rework (#1172–#1174) removed, so the γ-suite has been **red on main — and on every open PR — since the merge** (verified by reproducing the identical failures locally on plain `main` and on two unrelated PR branches):

- **`slots-v3.spec.ts`** — asserted the per-slot card's inline TileStrip (`.cslot-strip`), removed in `86c73366`. Now asserts the card's current shape (`.cslot-top .nm`).
- **`inference-pane-v3.spec.ts`** — asserted per-card `.slot-ctrls`, which no longer exist. The per-card control is now the modality pill toggle (`.npu-switch`), with lifecycle moved to the card-header ▶/■/↻ buttons; both are asserted.
- **`npu-occupancy-v3.spec.ts`** — asserted a per-card `.sctrl.stop` Stop control. Stop now lives in the card header and targets the `flm` anchor slot, so the test seeds the slot as `flm` and asserts `POST /api/slots/flm/unload` fires.

## Verification

All 20 specs across the three files pass locally against the pinned Chromium (the three repaired ones were the only failures before).

https://claude.ai/code/session_014gj5YALNZfAHb7qMvdCr8Q

---
_Generated by [Claude Code](https://claude.ai/code/session_014gj5YALNZfAHb7qMvdCr8Q)_

### DIFF
--- a/ui/tests/e2e/specs/inference-pane-v3.spec.ts
+++ b/ui/tests/e2e/specs/inference-pane-v3.spec.ts
@@ -227,11 +227,14 @@ test.describe('NPU occupancy card (/slots · Inference tab)', () => {
     await expect(card.locator('.aie-grid .aie-col')).toHaveCount(8)
   })
 
-  test('lists a per-FLM-slot card with lifecycle controls', async ({ page }) => {
+  test('lists a per-FLM-slot card with its toggle + header lifecycle controls', async ({ page }) => {
     await page.goto('/#slots')
     const card = page.locator('.npu-card')
     await expect(card.locator('.cslot').first()).toBeVisible()
-    await expect(card.locator('.cslot .slot-ctrls').first()).toBeVisible()
+    // Per-card control is now the modality pill toggle (trio-drawer rework,
+    // #1172); slot lifecycle moved to the card-header ▶/■/↻ buttons.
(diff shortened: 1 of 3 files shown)


thinmintdev (commented):
## Code review

**Verdict: MERGE.** Test-only, and exactly the extraction recommended in #1181's review.

- Verified each new selector against main's UI source: `.npu-switch` (primitives/engine-panes), `button[title="Stop"]` in the card header (npu-pane.jsx:453, inference-pane.jsx:266), `.cslot-top .nm` (npu-pane.jsx:301 / npu.css:524), and the `flm` anchor slot naming (npu-pane.jsx:361-366). The repairs assert the reworked card's real controls — including re-wiring the Stop test to the actual `POST /api/slots/flm/unload` request — rather than weakening coverage.
- Decisive proof: this PR's γ-suite (chromium) check is **green** — the first fully green γ run in the repo since the #1172–#1174 card rework — while every other branch (including main's own push runs) fails exactly these three specs.

---

PR #1205, opened 2026-07-11
feat(ui): unify slots-page headings and slot-status indicators

thinmintdev:
## Summary

Two threads of slots-page UI polish, done together because they touch the same panes.

### 1. One heading style across all slots-page tabs
Every tab heading now reads as a single row — `[glyph]  BOLD NAME · dim-dot · meta segments` — instead of the previous mix of two-line title/sub blocks, `<h1>` view headers, and eyebrow strips.

- **Inference** — collapsed the standalone section label + two-line title + green "N loaded" pill into one heading row.
- **Image Gen (ComfyUI)** / **Profiles** — two-line title/sub → single `sec-label` row.
- **Endpoints / MCP** (shared `EnginePane`) — dropped the redundant eyebrow and folded the two-line title into the `sec-label` row. Kept the `.cpane-title` text/class so existing locators resolve; the eyebrow's counts still live in the summary strip below.
- **Stacks** — `RUNTIME` / `<h1>Stacks</h1>` view header → single `sec-label` row.
- Added generic in-header `.sec-label` rules so any `.engine-h` / `.cpane-h` / `.wcard-h` / `.vh` can host the same heading family.

Also on the NPU card: colorized the header lifecycle controls (**start** purple · **stop** red · **restart** yellow · **logs** grey) and renamed **Edit → Configure**. And added bottom margin to the telemetry/metrics header so it no longer sits flush against the tab bar.

### 2. One slot-status vocabulary everywhere
There were three status-dot systems that had drifted apart. Unified them onto one classifier and one color language:

| phase | color | animation |
|-------|-------|-----------|
| serving | green (`--ok`) | pulse |
| ready / idle | amber (`--hal0-accent`) | static |
| warming | orange (`--warn`) | pulse |
| error | red (`--err`) | static |
| offline | grey (`--fg-5`) | static |

- **Inference slot cards** — serving dot was hard-coded to the yellow `--comfy` accent; now green, with ready as amber. (Answers "serving = green, yellow = idle".)
- **Endpoints** — the real inconsistency: `epState()` keyed off the raw `state` string and flagged a healthy-but-quiet slot as **warming** while its container-aware slot card showed **ready**. Rewrote `epState()` to derive from the shared `slotIndicatorFromPhase()` classifier, and realigned `.ep-dot` colors to match.

**Intentional exception:** the NPU occupancy card dots stay purple — that's the NPU's device-identity color, not a drifted status system. Easy to fold into the shared vocabulary later if desired.

## Testing
- `npm run build` clean.
- Full Playwright e2e suite: **388 passed / 10 skipped**.
- Updated `inference-pane-v3.spec.ts` for the collapsed heading (asserts the single heading row and the removed epill).

### DIFF
--- a/ui/src/dash/comfyui-pane.jsx
+++ b/ui/src/dash/comfyui-pane.jsx
@@ -395,9 +395,12 @@ function CardHead({ engine, run, pct }) {
   return (
     <div className="engine-h wcard-h">
       <span className="engine-glyph"><Ci name="comfy" size={16} /></span>
-      <span className="col">
-        <span className="engine-title">ComfyUI</span>
-        <span className="engine-sub">image-gen engine · docker</span>
+      <span className="sec-label">
+        <b>ComfyUI</b>
+        <span className="dim">·</span>
+        <span className="meta">image-gen engine</span>
+        <span className="dim">·</span>
+        <span className="meta">docker</span>
       </span>
       <span className={'epill' + (hasRun ? ' generating' : '')}>
         <span className="dot" />
--- a/ui/src/dash/connections.css
+++ b/ui/src/dash/connections.css
@@ -132,11 +132,15 @@
 .eprow.expanded > .eprow-main { background: var(--bg-2); }
 .eprow.dim .eprow-main { opacity: 0.5; }
 
+/* ONE status vocabulary — matches .sdot (overhaul.css / engine-panes.css) so an
+   endpoint row and its slot card always show the same colour for the same phase:
+   serving = green + pulse · stale/ready = amber (idle-but-up) · warming = orange
+   pulse · error = red · offline = grey. */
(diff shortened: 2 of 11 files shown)


thinmintdev:
Added a follow-up commit (872d0f8) extending the slots-heading polish:

**Image Gen tab (comfyui-pane)**
- Moved the "inference slots paused while rendering" notice out of the card header into a banner **above the active render** (shown only while a render is running).
- Slid the "iGPU · exclusive" tag to the header right; moved the container controls (stop/restart/logs) up to the **header far right**; dropped the redundant "docker · :8188" meta.

**Endpoints tab (connections)**
- Sized `.cpane-title` to 12px so **"Local endpoints"** matches the inference/NPU engine-pane titles on the sibling tabs (was 15px).

Verified in a local `vite build` preview against the live backend at desktop + narrow widths.

PR #1338, opened 2026-07-22
feat(slots): LLAMA_SET_ROWS=1 toggle for multi-GPU row splitting

thinmintdev:
### Summary
Adds a slot-level `llama_set_rows` boolean that emits `Environment=LLAMA_SET_ROWS=1` in the container Quadlet unit. This enables row-level model splitting across multiple GPUs (`--gpu-split auto --set-row 1`).

### Changes
- **Schema** (`src/hal0/config/schema.py`): new `llama_set_rows: bool` field on SlotConfig, defaults to `False`
- **Container provider** (`src/hal0/providers/container.py`): pass through `slot_cfg.llama_set_rows` to `_llama_launch_plan`, merge into env dict alongside existing `[server].env`
- **Slot drawer UI** (`ui/src/dash/slot-modals.jsx`): 'Rows Split' PillToggle in Hardware grid after Ngl, with FieldInfoIcon popover

### Usage
Set `llama_set_rows = true` on a slot. On next restart, the container gets `LLAMA_SET_ROWS=1` in its environment. This tells llama-server's hip/cuda backend to split rows across GPUs rather than layers, which is the correct strategy for models that fit in VRAM row-by-row but not layer-by-layer.

### DIFF
--- a/src/hal0/config/schema.py
+++ b/src/hal0/config/schema.py
@@ -487,6 +487,16 @@ class SlotConfig(BaseModel):
             "resolve_chat_template and slot_flags_fold."
         ),
     )
+    llama_set_rows: bool = Field(
+        default=False,
+        description=(
+            "LLAMA_SET_ROWS=1 \u2014 splits a single model across multiple GPUs "
+            "at the row level (--gpu-split auto --set-row 1). Useful for "
+            "multi-GPU boxes where the model exceeds a single GPU's VRAM. "
+            "The env var is emitted as Environment=LLAMA_SET_ROWS=1 in the "
+            "container unit."
+        ),
+    )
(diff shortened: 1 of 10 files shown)


thinmintdev:
Closing for the 1.0 Preview cut after current-main review. This branch is conflicting, includes unrelated stale assertion changes and skipped E2E coverage, and the Rows Split toggle is omitted from drawer dirty-state protection. The feature is not required for Preview and LLAMA_SET_ROWS can already be supplied through typed server.env. Please reintroduce post-Preview from current main as a feature-only change with supported-runner/hardware evidence, dirty/save/restart coverage, and full CI. No current-main rework is being reverted.

PR #1339, opened 2026-07-22
fix(ui): normalize memory graph slot selector

thinmintdev:
## Summary
- normalize Hindsight memory graph slot aliases like hal0/utility to canonical available slots like utility
- submit the canonical slot value from the memory page selector/save path
- add focused unit coverage for alias normalization and unknown alias preservation

## Verification
- cd ui && npm run test:unit
- cd ui && npm run typecheck
- cd ui && npm run build

Note: build emits existing Vite chunk-size and Node DEP0205 warnings.


thinmintdev:
Closing for the 1.0 Preview cut. The supported backend contract uses bare extraction slot names (for example utility) and rejects routed model IDs such as hal0/utility, so this UI-only alias normalization currently has no demonstrated valid producer. Reopen only with a supported install/API reproduction, then canonicalize at the backend/config boundary and add GET/PUT plus both-panel UI coverage. No current-main behavior is being reverted.

PR #1350, opened 2026-07-26
fix: carryover correctness fixes (watchdog race · comfyui deadlock · bench · doctor json)

thinmintdev:
Five real bugs surfaced during 150 / lxc105 triage on the alpha.2 line that are **still present on `rework/descar`** — ported here with red→green regression tests.

| Fix | Bug | Regression test |
|---|---|---|
| **slots/watchdog** | `SlotWatchdog` stamped ERROR/OFFLINE (`force=True`) without the per-slot lock that `load`/`unload`/`restart` hold → races an in-flight big-model load into an illegal `error→ready` (slot wedged `error`, 409 on `/load`). All 3 stamp sites now route through `_stamp_locked` (take the lock + re-verify the fault). **Verified live on Strix Halo 150.** | `tests/slots/test_watchdog_load_race.py` (red-capable on descar) |
| **comfyui/fetch** | `_run_sequence` used `stdout=PIPE` never drained → `proc.wait()` deadlocks once a real `hf` download fills the ~64KB pipe → job hangs in `running` forever. → DEVNULL. | `tests/comfyui/test_fetch_no_deadlock.py` |
| **bench/parsers** | `accepted / draft_n` with `draft_n>0` but `accepted` None (partial timings) → TypeError drops the whole queue item. Guard both counts. | `test_parse_server_ab_drafts_without_accepted_does_not_crash` |
| **bench/store** | `current_cells` picked newest ok by `MAX(run_id)`; `run_id` is `<stamp>-<random hex>`, so same-second rows tie-break on the random suffix and surface an OLDER record than the planner's `newest_ok_by_cell`. Use `rowid_seq` (append order). | `tests/bench/test_store.py` |
| **cli/doctor** | `toolbox-pull --json` used `console.print_json` → ANSI escapes on a pipe → unparseable by `jq`. Emit plain `json.dumps`. | `test_toolbox_pull_json_output_is_plain_not_ansi_colorized` |

**Verification:** 654 slots/bench/comfyui/doctor tests pass · `ruff` clean · `check-sunset` green, scar 200 = baseline (neutral).

Not ported (flagged for maintainer judgment): the O25 model-mount defensive change (descar covers it via `assert_under_store(severity="warn")` + doctor `HAL0-MODEL-STORE-UNMOUNTED` guard by design) and a brain test-isolation conftest (descar has no such file).

### DIFF
--- a/src/hal0/bench/parsers.py
+++ b/src/hal0/bench/parsers.py
@@ -187,7 +187,10 @@ def _sa_run_to_rep(run: dict) -> Rep | None:
         return None
     draft_n = run.get("draft_n")
     accepted = run.get("draft_n_accepted")
-    accept = (accepted / draft_n) if draft_n else None
+    # Only compute acceptance when BOTH counts are present. A run can report
+    # draft_n>0 with draft_n_accepted null/absent (partial or older llama.cpp
+    # timings) — `None / draft_n` would TypeError and drop the whole run.
+    accept = (accepted / draft_n) if (draft_n and accepted is not None) else None
     return Rep(
         t_s=run.get("wall_s"),
         prefill_ts=run.get("prompt_per_second"),
--- a/src/hal0/bench/store.py
+++ b/src/hal0/bench/store.py
@@ -173,18 +173,22 @@ class Store:
                     ),
                 )
                 n += 1
-            # current_cells: newest ok record per cell_key. run_id sorts
-            # lexicographically by its leading UTC stamp, so MAX(run_id) among ok
-            # rows is the newest ok record for the key.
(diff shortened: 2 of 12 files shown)


thinmintdev:
Superseded: both commits landed on main via #1351 — c28b24e8 ≡ 4e82f190 (carryover correctness: watchdog race, comfyui deadlock, bench, doctor json) and c247e327 ≡ 335b954c (O25 heal: mount slot model-file dir outside configured roots), identical diffstats. rework/descar already contains main at 1c2ffead, so this PR carries no new content — its 7 python reds are staleness against a base predating the seeded-profile test repairs, not real failures. Closing.

PR #1354, opened 2026-07-26
fix(drawers): model owns mtp/reasoning/vision, slot owns hardware (spec-hw-slot-ownership §1)

thinmintdev:
## Summary

Closes the two-writer split in the drawers: `mtp` / `enable_thinking` / `vision` were writable from **both** the slot drawer and the model drawer, so a slot could silently disagree with its own bound model — the same failure mode already fixed on the hardware axis. Per spec-hw-slot-ownership §1 the model owns the logical tune (typed capabilities), the slot owns the physical layer.

- **Server**: `ModelDefaults` gains `enable_thinking`/`vision` (extra-blob persistence, mirroring the `capability_flags` idiom); `SlotConfig` drops all three. New `MODEL_OWNED_SLOT_KEYS` + `_deny_model_owned_keys` hard-reject wired into `reconcile_slot_updates` — symmetric to the existing `SLOT_HARDWARE_FLAGS` reject in the other direction. `container._effective_mtp` loses its slot-override tier; vision reads model defaults.
- **Migrator**: one-shot idempotent `hal0 slot migrate-tune` folds existing slot values onto the model. Per-model vote pool; ≥2 distinct votes on a field refuses the whole model and writes nothing (operator inspects raw TOML). Dry-run default, deploy-window gated.
- **UI**: slot drawer's Inference group removed; model drawer gains a Vision tri-state beside MTP/Reasoning; slot ctx relabeled "Context (slot override)". γ specs updated in the same commit as the UI change.
- **Stack path**: `_create_missing_slots` no longer writes vision/mtp into the create body — that path bypasses the guard (`SlotManager.create` doesn't route through `reconcile_slot_updates`), so stack-created slots were still being born with model-owned keys.

Notable discovery: the model drawer's Reasoning tri-state was already wired client-side but silently dropped server-side (no `ModelDefaults.enable_thinking`). It persists for the first time now.

## Risk grade

- [ ] low
- [x] med — user-facing change, new code path, migrator
- [ ] high

## Touched surfaces

- [x] API (`src/hal0/api/`)
- [ ] Auth / sessions
- [x] Slots / dispatch
- [x] Models / capabilities
- [ ] Installer
- [ ] Updater
- [ ] Board chat / MCP admin
- [x] Config / schema
- [x] UI (`ui/src/`, Playwright specs)
- [x] Docs (`docs/reference/cli.mdx` — CLI parity)
- [ ] CI / release

## §14.1 high-risk surfaces

- [ ] Unauthenticated board routes
- [ ] `AUTONOMOUS_WRITE_TOOLS`
- [ ] Installer / updater RCE-class

## Rollback

_Rollback:_ revert the merge. Pre-migration slot TOMLs still round-trip their old keys harmlessly via `extra="allow"`; the migrator is opt-in and dry-run by default, so nothing is rewritten until an operator runs it.

## Known follow-ups (documented in-code, deliberately out of scope)

- `stacks/apply.py` no longer projects the three fields onto a slot but has no model-registry dependency to fold them onto the model either — applying a stack silently drops them.
- `stacks/portable.py` snapshot/export still reads the keys from raw slot TOML; correct for pre-migration data, reports defaults once migrated.

## Test tiers run

- [x] α unit — `tests/config/`, `tests/slots/`, `tests/slot_view/`, `tests/registry/`, `tests/normalize/`, `tests/providers/`, `tests/stacks/`, `tests/db/` green (2400+)
- [x] β integration — `tests/api/test_slots_routes.py`, `test_chat_normalization.py` green
- [ ] γ release-gate — runs as the Playwright check on this PR

https://claude.ai/code/session_014PuHjo1FhvTL5uGS1VVNdk

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -130,6 +130,7 @@ hal0 setup --auto --no-pull --no-extensions
 | `slot capacity` | Per-slot resident memory + the `[slots].max_slots` budget. | `--json` |
 | `slot migrate-id-keying` | One-shot, operator-run: flip every slot artefact from name-keyed to id-keyed on disk. Destructive (backs itself up first); needs hal0 stopped. See the [migration runbook](/docs/operate/slot-id-keying-migration). | `--dry-run`, `--yes`/`-y`, `--stop-services` |
 | `slot migrate-hw` | One-shot, operator-run: unwind the flags-fold so hardware sticks to slots (folds model NGL/runner + profile/slot image pins onto each slot's typed HW grid). Dry-run by default; `--apply` writes (backs itself up first) and needs hal0 stopped. | `--apply`, `--yes`/`-y`, `--stop-services` |
+| `slot migrate-tune` | One-shot, operator-run: fold slot mtp/enable_thinking/vision onto the bound model's defaults (the model becomes the single authority), then drop the now-redundant slot keys. Divergent slot/model values REFUSE and are reported instead of silently picked. Dry-run by default; `--apply` writes (backs itself up first) and needs hal0 stopped. | `--apply`, `--yes`/`-y`, `--stop-services` |
 
 `slot create` flags:
 
--- a/graphify-out/GRAPH_REPORT.md
+++ b/graphify-out/GRAPH_REPORT.md
@@ -1,440 +1,478 @@
-# Graph Report - hal0  (2026-07-22)
+# Graph Report - agent-a0972b8b76969053f  (2026-07-26)
 
 ## Corpus Check
-- 1677 files · ~2,322,207 words
+- 1573 files · ~1,968,419 words
 - Verdict: corpus is large enough that graph structure adds value.
 
 ## Summary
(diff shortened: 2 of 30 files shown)


thinmintdev:
Superseded by #1353, which landed the same feature through descar while this was gating — model-owned MTP/Reasoning/Vision, slot-side hard-reject, migrator, and the drawer moves are all on main now. Closing rather than resolving the conflict; no point carrying a second implementation.

One piece of this PR is NOT covered by #1353 and is being salvaged separately: `api/routes/stacks.py::_create_missing_slots` still writes `vision` (unconditionally) and `mtp` into the slot-create body. #1353's guard (`reject_model_owned_slot_keys`) is wired only into `api/routes/slots.py`; `SlotManager.create()` never calls it, and the stacks path calls `slot_manager.create()` directly in-process — so stack-created slots are still born carrying model-owned keys, re-opening the two-writer split on that one path. Follow-up PR incoming.

Issue #1301, opened 2026-07-13
memory: hindsight→pgvector degrade ladder is inert; 'degraded' flag is misleading


thinmintdev:
## Summary
The boot-time degrade ladder (Hindsight unreachable → fall back to in-memory `PgVectorProvider`) never fires for a down daemon, and the `degraded` health signal is therefore misleading.

`provider_from_config` builds the Hindsight client inside a `try/except` that is *supposed* to catch an unreachable daemon and degrade. But `_build_hindsight_client` only calls `HindsightRestClient.from_env()`, which constructs an httpx client and does **no I/O** — so a daemon that is down does not raise at boot. The result is a live-but-broken `HindsightProvider` with `degraded=False`; failures surface only at first recall (as empty results), and `GET /api/status.memory_degraded` + `hal0 memory status` report healthy.

## Where
- `src/hal0/memory/__init__.py:45` `_build_hindsight_client` — no connectivity probe (self-flagged "P1-6" TODO in the docstring, `:48-54`).
- `src/hal0/memory/__init__.py:86-90` — the `except` that would degrade never triggers for a down daemon.
- Consumers of the misleading flag: `src/hal0/api/routes/health.py:55-65`, `src/hal0/cli/memory_commands.py:112-117`.

## Fix
Add a cheap `/health` (or `/stats`) probe in `_build_hindsight_client` so an unreachable daemon raises at boot and the existing degrade branch returns `PgVectorProvider` (with `degraded=True`). Keep it fast + timeout-bounded so boot isn't blocked.

_Filed from a full memory-system audit (2026-07-13)._

Issue #1300, opened 2026-07-13
memory: project:<id> isolation is a silent no-op in unified_bank mode


thinmintdev:
## Summary
Under the default config (`[memory].unified_bank=true`), writing to a `project:<id>` namespace silently collapses to the `shared` bank with **no compensating tag**, so project scoping does not exist — a `project:foo` write becomes an ordinary shared write, and a `project:foo` recall returns all of `shared`.

Contrast with `private:` which, on collapse, is preserved as a `visibility:private` + `agent:<id>` tag pair and enforced on read. `project:` gets no such marker.

## Where
- `src/hal0/memory/hindsight_provider.py:200` `_write_namespace` — unified mode returns `shared` for anything that isn't `agents`; `project:<id>` is not distinguished.
- `src/hal0/memory/hindsight_provider.py:252` `add` — only stamps `visibility:private` (for the private toggle) and `agent:<id>`; no `project:` tag.
- `src/hal0/memory/hindsight_provider.py:173` `_allowed_namespaces` — unified read collapses everything (except `agents`) to `shared`.

## Repro (unit-level)
Add with `dataset="project:foo"` in unified mode → retained to bank `shared` with no project tag → recall `dataset="project:bar"` returns the doc.

## Options
1. Stamp a `project:<id>` tag on collapse and filter reads by it (mirrors the `visibility:private` mechanism).
2. Or drop `project:<id>` from the accepted namespace set in unified mode so it fails loudly instead of silently degrading.

_Filed from a full memory-system audit (2026-07-13)._

Issue #1224, opened 2026-07-11
slot restart wedges on errored slots; slot load no-ops on loaded slots — config changes don't converge


thinmintdev:
Two related lifecycle gaps observed 2026-07-11 on CT105 (v0.9.6), both leaving a slot's systemd unit stale relative to its TOML:

**1. `POST /api/slots/{name}/restart` on a slot in `error` state hangs.** The CLI gets a ReadTimeout and the unit is never relaunched (no journald activity for the unit at all). `manager.restart()` = `unload()` → `load()`; the hang appears to be in the unload/terminate path when the unit has already failed. Meanwhile `slot load` recovers the same slot instantly.

**2. `slot load` on an already-loaded slot no-ops, so config edits never converge.** The unit file says "regenerated on every slot load", but a load on a slot in ready/idle short-circuits. Sequence that bites: `PUT /api/slots/ops/config {"port": 8091}` → `slot load ops` → returns the old snapshot, unit still `--port 8089`, and the subsequent implicit reload cycled through warming → error with nothing listening on either port. Recovery required `systemctl reset-failed` + a second `slot load` (from the error state, which does regenerate).

**Suggested direction:** make `restart` skip graceful drain for units already inactive/failed (`reset-failed` + straight to load), and make config-mutating PUTs either trigger a real unload→load convergence or mark the slot config-dirty so the next load regenerates unconditionally.

Repro:
1. Any slot in `error` state → `hal0 slot restart <name>` → ReadTimeout, no unit activity.
2. Loaded slot → `PUT /api/slots/<name>/config {"port": <new>}` → `hal0 slot load <name>` → unit keeps the old port.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

thinmintdev:
Implementation pointers (from the live diagnosis on CT105):

**Code paths**
- `SlotManager.restart()` (src/hal0/slots/manager.py) is `await self.unload(name)` → `await self.load(name)`. The wedge on errored slots is in the unload leg: `unload()` takes the per-slot asyncio lock, transitions → UNLOADING, then `await self.terminate(name)`. When the unit already exited (`hal0-slot@<name>.service` in `failed` state), terminate appears to wait on something that never completes — the REST caller gets a ReadTimeout and, critically, **no unit activity ever appears in journald** (`journalctl -u hal0-slot@ops` empty for the whole window), so `load()` is never reached.
[...]

PR #1359, opened 2026-07-27
fix(memory): gate /mcp/memory bulk delete + pin the ADMIN delete surface (#1302, #1024)

### LINKED ISSUE #1024: Memory: unguarded DELETE /api/memory/banks/{id} enables one-call data loss (RCA + hardening)
## Summary
On 2026-07-04 an agent misidentified the live Hindsight banks as orphaned and issued `DELETE /api/memory/banks/{bank_id}`, irreversibly cascade-deleting ~632 records across `private__hermes-agent` (275 facts), `private__hermes`, and `shared`. Recovered from a PBS backup.

## Root cause
`DELETE /api/memory/banks/{bank_id}` (`src/hal0/api/routes/memory_admin.py`, `_FORWARDS`) is a bare passthrough to Hindsight with **no confirmation, no soft-delete, no auth**. The entire `/api/memory/*` surface is unauthenticated and bound on `0.0.0.0:8080`, so any LAN host — or an agent's terminal/tool — can wipe a bank with one curl, bypassing the UI confirm dialog. Contributing UX bug: the dashboard mislabels the banks (colon vs double-underscore spelling; `list?bank` returns wrong data) which fed the misdiagnosis.

## Fixes (branch `fix/memory-tab-functional`, hotpatched on CT105)
- `DELETE /banks/{id}` now requires `?confirm=<bank_id>` (or body `{"confirm":...}`); mismatch → 400, no forward.
- `GET /api/memory/list?bank=` alias mapped to dataset (was silently returning default-dataset rows).

## Remaining hardening (this issue)
- Reintroduce write-auth (bearer or localhost-only) on **mutating** `/api/memory/*` routes; keep reads open if desired.
- Soft-delete / tombstone for banks (rename + TTL purge) so deletes are reversible.
- Audit-log every destructive memory op.

### LINKED ISSUE #1302: memory: delete surface hardening — no API auth + ungated /mcp/memory bulk delete
## Summary
Follow-up hardening after the provider-level fix for cross-agent private deletion (visibility gate added to `HindsightProvider.delete` on branch `worktree-turnstone-integration`). Two residual gaps remain that the provider fix does not cover:

1. **No auth on `/api/memory/*`.** Identity is a self-asserted `X-hal0-Agent` header (`src/hal0/api/routes/memory.py:97`, docstring `:50`); any LAN caller can claim any agent id. Body `source` is rejected so audit can't be desynced from the header, but the header itself is spoofable.
2. **`/mcp/memory` bulk delete is ungated.** `mcp__hal0-admin__memory_delete` gates `len(ids) > 1` through the approval queue (`src/hal0/mcp/admin.py:909`), but the standalone `/mcp/memory` dispatcher (`src/hal0/mcp/memory.py`, `make_dispatcher`) executes bulk delete directly with no gate — and its `memory_delete` docstring wrongly assumes approval already happened.

## Context / already fixed
The core exploit (agent A deletes agent B's `visibility:private` doc by guessing its id, since unified mode stores all private docs in one `shared` bank) is closed at the provider layer: `delete()` now resolves each target id's owner tag and withholds docs the caller can't see (fail-closed). See `tests/memory/test_unified_bank.py` (`test_unified_delete_*`).

## Ask
- Decide the auth posture for `/api/memory/*` (perimeter-only vs a token), and
- Make the standalone `/mcp/memory` bulk-delete honor the same approval gate as the admin surface (or document why it intentionally doesn't).

Related: `v0.5` (MCP admin + memory wiring).

_Filed from a full memory-system audit (2026-07-13)._

thinmintdev:
Closes #1302. Closes #1024.

## The real bug: ungated bulk delete on `/mcp/memory`

`hal0.mcp.admin.dispatch` classifies a `memory_delete` with >1 id as capital-D destructive and enqueues it on the `ApprovalQueue`. The standalone memory server (`mcp/memory.py::build_server`) wired its tools straight to `make_dispatcher`, so **the identical call on that mount executed immediately**. Demonstrated on the parent commit:

```
admin.is_gated(memory_delete, 3 ids) = True
/mcp/memory memory_delete -> {"status": "ok", "deleted": 3}
provider.delete calls      = [["a","b","c"]]
```

`_memory_delete`s own docstring asserted the opposite ("by the time we get here it has already been approved"). The MCP docs also already *claimed* the gate existed.

## The auth half was already closed — the issue titles are stale

Verified on the base commit: with `HAL0_REQUIRE_AUTH=1` every destructive memory route already behaved correctly (`anon=401 admin=503 client=403`). Both issues predate the KB-1 exposure table (2026-07-18).

What was genuinely missing is that **nothing held it in place**. No test asserted the classification, `match_rule("DELETE", "/api/memory/banks/{bank_id}")` resolved via the generic `"memory"` row, and #1024s own follow-up proposes "keep reads open if desired" — a widening whose natural expression silently downgrades the bank wipe to CLIENT.

## Changes

- `memory_delete` arity rule gets one home (`is_bulk_delete`); `admin.is_gated` delegates, `build_server` reads it
- `build_server` takes the same lifespan-scoped `ApprovalQueue`, returns `{status: pending_approval, approval_id}`, runs only via the bound executor on approve, and **fails closed with no queue**
- Every `DELETE` under `/api/memory` plus bulk `POST /api/memory/delete` pinned as explicit ADMIN rules *above* the generic prefix row, enumerated in `DESTRUCTIVE_MEMORY_ROUTES` next to `OPEN_ALLOWLIST`
- Corrected stale "auth was removed" comments in `routes/memory.py` and `mcp_mount.py`

No new privileged surface — `memory_delete` still maps to `POST /api/memory/delete`. Nothing added to any OPEN allowlist.

## Verification

- **59 new tests, all failing on the parent commit**
- `tests/memory tests/mcp tests/security` + targeted api/agents → **485 passed, 1 skipped**
- `tests/dispatcher tests/brain tests/activity tests/golden_paths` → **275 passed**
- Baseline arithmetic: 324 before → 383 after, exactly +59
- ruff / format / sunset (200 ≤ 200) / import smoke clean

Honest note: the bare anon-401 assertion passes on base too, since the perimeter was already right. The auth-side tests that genuinely fail are the pinning/ratchet/widening ones.

⚠️ `docs/reference/mcp-tools.mdx` and `docs/reference/api/rest-api.mdx` are **mirrored from hal0-web** — those edits are also ported in the hal0-web docs branch, or the next mirror sync will clobber them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,9 @@ applying. Add those subsections to a version's section to surface them; see
 
 ## [Unreleased]
 
-_Nothing yet._
+### Security
+
+- **Pin the destructive `/api/memory` routes as ADMIN, above the generic prefix rule** (closes the last open slice of #1024). #1024's incident was a single unauthenticated `DELETE /api/memory/banks/{bank_id}` that cascade-deleted ~632 live records; its echoed-`?confirm=` guard (#1028) and audit row (#1030) landed, but the classification itself was only ever ADMIN *by generic prefix* and nothing asserted it. That matters because #1024's own follow-up proposes "keep reads open if desired", and the natural expression of that — a `_prefix("/api/memory")` CLIENT rule — silently takes the bank wipe with it under first-match-wins. Two narrow rules (`any DELETE under /api/memory`, `POST /api/memory/delete`) now sit *above* the generic memory row, and `exposure.DESTRUCTIVE_MEMORY_ROUTES` enumerates the irreversible surface so a new memory delete route, or a reclassification, has to touch the constant in the same diff. `tests/security/test_memory_delete_auth.py` asserts the constant tracks the live route table, that each route resolves via a pinned rule (not the ADMIN fallback), that all of them survive a simulated reads-are-CLIENT widening, and — against the real `AuthEnforcementMiddleware`, armed — that an anonymous call gets `401`, a client/inference key gets `403`, and the operator's admin key clears the gate.
 
 ## [1.0.0-rc.1] — 2026-07-27 (R5 · the rework release)
 
--- a/docs/reference/api/rest-api.mdx
+++ b/docs/reference/api/rest-api.mdx
@@ -95,6 +95,14 @@ See [Services](/docs/operate/services) for the operator-facing guide (registry, 
 
 ### `/api/memory` — hardening
 
+- The whole `/api/memory` surface — reads included — is **ADMIN**-classed in
+  the route exposure table, so once auth is enabled every call needs an admin
+  credential (session cookie or admin bearer); a client/inference key gets
+  `403`. The irreversible subset (every `DELETE` under `/api/memory`, plus the
+  namespace `POST /api/memory/delete`) is additionally pinned by name in
+  `DESTRUCTIVE_MEMORY_ROUTES` and matched by rules that sit *above* the generic
+  prefix rule, so a later decision to open memory reads cannot take the bank
+  wipe with it.
 - Every destructive `/api/memory/*` operation — bank delete, and the
   memories/config/document/directive/operation/mental-model deletes, plus the
   namespace `POST /api/memory/delete` — records a durable audit row (actor +
--- a/src/hal0/security/exposure.py
+++ b/src/hal0/security/exposure.py
@@ -130,6 +130,7 @@ class _Rule:
 _GET: frozenset[str] = frozenset({"GET", "HEAD"})
 _POST: frozenset[str] = frozenset({"POST"})
 _PUT: frozenset[str] = frozenset({"PUT"})
+_DELETE: frozenset[str] = frozenset({"DELETE"})
 
 # Ordered, first-match-wins. See the module docstring for the full design
 # rationale and hal0-rework-plan.md §23.5 for the architecture this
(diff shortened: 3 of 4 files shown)


thinmintdev:

thinmintdev:

Rebased onto `main` and cut from **729 → 332** changed lines. Every surviving line is something `610150d7` does not implement.

**Dropped as duplicate:** the `/mcp/memory` bulk-delete gate itself. `610150d7` implements it at a *better* layer — inside `make_dispatcher` rather than a `build_server`-level wrapper — so there is one code path instead of two.

**Kept — `610150d7` does not touch `src/hal0/security/exposure.py` at all**, and does not list #1024 among the issues it closes:
- `_DELETE` + two narrow rules ordered **above** the generic `memory` row
- `DESTRUCTIVE_MEMORY_ROUTES` exported in `__all__`
- `tests/security/test_memory_delete_auth.py`

**Non-vacuity proven:** removing the two pinned rows and applying #1024s own proposed widening (`_prefix("/api/memory")` → CLIENT) downgrades all **8** destructive routes to `client`, including `DELETE /api/memory/banks/{bank_id}` — #1024s incident route.

### Honest caveat
This makes the classification correct and un-widenable. It does **not** close #1024 on a default box, because `require_auth_enabled()` ships **off** (operator decision 2026-07-19, finding O19).

### One gap deliberately not grafted
`610150d7`s `approval_queue=None` overloads two meanings — "admin already gated, run it" and "misconfigured memory server" — so a `build_server(approval_queue=None)` silently executes bulk deletes ungated. This branch failed closed there. There is exactly one production call site and it passes the queue, so this is a ratchet rather than a live bug, and fixing it properly means redesigning mains signature, not cherry-picking. Flagged for a follow-up.

### Also worth a separate one-line fix on main
`routes/memory.py`s new comment says ADR-0012 "removed auth and TLS platform wide". `hal0/api/auth.py::AuthEnforcementMiddleware` exists and `/api/memory` is ADMIN-classed, so that parenthetical is stale — what is true is that `require_auth_enabled()` defaults off.

**Verification:** ruff check ✅ · ruff format (985 files) ✅ · import smoke ✅ · sunset (195 ≤ 195) ✅ · `tests/security tests/mcp tests/memory tests/api` → **1892 passed, 1 skipped** · `tests/security` under fastapi 0.139 → 66 passed.

The fastapi version-skew hazard is **closed** — `tests/mcp tests/security` give identical results (226 passed, 1 skipped) on both 0.136.1 and 0.139.2. The earlier 485-vs-184 split was the `_IncludedRouter` bug that `610150d7`s `_iter_leaf_routes` fixes.

Issue #1372, opened 2026-07-30
fix(ui): slot drawer 'Clear override' never removes a persisted chat_template


thinmintdev:
## Summary

In the slot edit drawer (`#slots/:name` → **Model** group → **Template** row), clicking **Clear override** on a slot that has a persisted `chat_template` collapses the select back to the read-only "(from model)" display — but **Save sends nothing**. The override stays on disk and keeps being passed to `llama-server --chat-template` on the next restart, while the UI now claims the model default is in effect.

There is no other way to remove a per-slot chat-template override from the dashboard.

## Repro

1. Seed / configure a slot with `chat_template = "chatml"`.
2. Open `#slots/<name>`. The Template row renders the select pre-set to `chatml`.
3. Click **Clear override**. The row collapses to the model-default display.
4. Click **Save**.
5. Observe: **no** `PUT /api/slots/<name>/config` is issued at all. Re-opening the drawer shows `chatml` again (re-seeded from `slot.chat_template`).

Verified with a Playwright probe against the mocked API (`ui/tests/e2e/fixtures/apiMock.ts`): captured `PUT /api/slots/primary/config` bodies after the clear + Save = `[]`.

## Root cause

`ui/src/dash/slot-modals.jsx` gates the write on `overrideOpen`, which **Clear override** has just set to `false`:

```js
// onSaveClick
const chatTemplateChanged =
  overrideOpen && chatTemplate !== (slot.chat_template || "");
...
if (chatTemplateChanged) slotBody.chat_template = chatTemplate;
```

The same predicate also drives the unsaved-changes guard, so a cleared override is not even counted dirty:

```js
const dirty =
  ... ||
  (overrideOpen && chatTemplate !== (slot.chat_template || ""));
```

`onClick` for **Clear override** does `setChatTemplate(""); setOverrideOpen(false)` — which makes the clear provably unobservable to both the dirty check and the save body.

## Fix sketch (needs a contract decision, hence not a drive-by one-liner)

The predicate should compare the *effective* value rather than gate on `overrideOpen`:

```js
const effectiveTemplate = overrideOpen ? chatTemplate : "";
const chatTemplateChanged = effectiveTemplate !== (slot.chat_template || "");
```

[...]

PR #1373, opened 2026-07-30
test(ui): drawer field-wiring + rejected-write contracts, and fix the chat_template clear (#1371, #1372)

### LINKED ISSUE #1371: test(ui): field-wiring contract specs for the slot + model drawers
## Why

Several edit fields in the **slot drawer** (`#slots/:name`) and the **model drawer** (`#models` → *Edit options*) have no e2e assertion that the control's value actually reaches the wire with the right key, on the right route. The drawers batch some fields into `PUT /api/slots/{name}/config`, route others to their own endpoints (`POST /swap`, `PATCH /defaults`, `POST /api/models/{id}/duplicate`), and instant-apply a few. A silent wiring regression in any of those is invisible today.

Audited gaps (verified untested against `ui/tests/e2e/specs/`):

| # | Field | Expected wire | Prior coverage |
|---|---|---|---|
| 1 | Slot drawer · Model select | `POST /api/slots/{name}/swap` `{model_id}` | only the *card* picker (inference-pane-v3) and the inline popover (slots-wireup-v3) |
| 2 | Slot drawer · Parallel | `PUT /config {parallel}` | none |
| 3 | NPU · Embed toggle | `PUT /config {npu:{chat,asr,embed}}` | only the ASR lane |
| 4 | NPU · Chat model pick | `PUT /config {npu, model:{default}}` | none |
| 5 | NPU · pull-then-apply | `POST /api/models/{tag}/pull` → apply on `completed` | none |
| 6 | Model drawer · DuplicateModelDialog | `POST /api/models/{id}/duplicate` `{new_id, profile}` + error paths | none |

(`chat_template` save was on the original list but is already covered by `slot-drawer-profile-v3.spec.ts` C7k.)

## What

Two new contract spec files:

- `ui/tests/e2e/specs/slot-drawer-field-wiring-v3.spec.ts`
- `ui/tests/e2e/specs/model-drawer-duplicate-v3.spec.ts`

Both follow the established pattern: mocked `/api` via `tests/e2e/fixtures/apiMock.ts`, slot lists seeded through a `seedSlots` `addInitScript` that patches `window.HAL0_DATA`, mutations left to fall through to `page.route`, and exact request bodies asserted via `postDataJSON()`.

thinmintdev:
Closes #1371.

## What

Two new Playwright contract specs proving that every audited slot-drawer and model-drawer edit field lands on the wire with the right key, on the right route. **No source changes** — these lock in existing behaviour.

### `ui/tests/e2e/specs/slot-drawer-field-wiring-v3.spec.ts` (13 tests)

| ID | Field | Wire contract |
|----|-------|---------------|
| W1 | Model select | `POST /api/slots/{name}/swap {model_id}` — its own route, never folded into the batched `PUT /config`; re-picking the current model is a no-op |
| W2 | Model select, LIVE container | `ConfirmDialog` gate: cancel fires nothing and reverts the select, confirm fires the same body |
| W3 | Parallel | `PUT /config {parallel}` as an int; empty → `null` (inherit); untouched never rides a Save; sub-1 blocks Save with an inline error |
| W4 | NPU · Embed / NPU · Chat | instant-apply `PUT /config {npu:{chat,asr,embed}}` + `POST /restart`; the whole modality triple always rides (the backend replaces `[npu]` wholesale) |
| W5 | NPU chat-model pick (installed) | `PUT /config {npu, model:{default}}` — nested so the merge keeps sibling `[model]` keys |
| W6 | NPU chat-model pick (not installed) | `POST /api/models/{tag}/pull` **first**, no config write while the download is in flight, then the same apply once the SSE stream reports `completed` |
| W7 | Clear override | `test.fail()` tripwire for the bug below |

### `ui/tests/e2e/specs/model-drawer-duplicate-v3.spec.ts` (5 tests)

| ID | Case | Wire contract |
|----|------|---------------|
| D1 | picked device template | `POST /api/models/{id}/duplicate {new_id, profile}`, dialog closes, success toast |
| D2 | "— no template —" | `{new_id}` only — no null `profile` key |
| D3 | id suggestion | derives `<source id>-<device class>`, re-derives on template change, pins once hand-edited; one representative profile per device class in the picker |
| D4 | empty / same-as-source id | blocked inline, no POST, dialog stays open |
| D5 | server 409 | inline error + toast, dialog stays open for a retype |

Both follow the established pattern: mocked `/api` via `tests/e2e/fixtures/apiMock.ts`, slot lists seeded through a `seedSlots` `addInitScript` over `window.HAL0_DATA`, mutations left to fall through to `page.route`, exact request bodies asserted with `postDataJSON()`.

## Real bug found

**#1372 — slot drawer "Clear override" never removes a persisted `chat_template`.**

`onSaveClick` gates the write on `overrideOpen`, which the Clear button itself has just set to `false`:

```js
const chatTemplateChanged =
  overrideOpen && chatTemplate !== (slot.chat_template || "");
```

So clearing an override issues **no** `PUT /config` at all — the template stays on disk and keeps feeding `llama-server --chat-template`, while the drawer now displays the model default. The same predicate drives the dirty guard, so it isn't even counted as an unsaved change. There is no other way to remove a per-slot override from the dashboard.

Not fixed here: the correct wire value for a removal (`null` vs `""`) is a contract decision against `SlotManager.update_config`'s None-means-delete merge, so it wants its own PR. W7 asserts the desired contract under `test.fail()` — the suite stays green today and trips with an *unexpected pass* the moment #1372 lands, at which point the annotation gets dropped.

## Scope notes

- `chat_template` **save** was on the original gap list but is already covered by `slot-drawer-profile-v3.spec.ts` C7k — only the removal path was missing.
- Deliberately does not touch the drawer header toggle; that surface belongs to `slot-pin-toggle-v3.spec.ts` (#1368).

## Verification

```
npx playwright test slot-drawer-field-wiring-v3 model-drawer-duplicate-v3 --reporter=line
  18 passed (15.7s)

npx playwright test slot-edit-controls-v3 slot-drawer-profile-v3 model-drawer --reporter=line
  33 passed (27.5s)

npm run lint
  clean
```


thinmintdev:
**Scope grew — #1377 merged into this branch, so this PR now closes #1372 as well.**

The field-wiring audit that produced these specs found a real bug, and the fix is now stacked in here (it squashes to main together with the specs):

- `ui/src/dash/slot-modals.jsx` — the slot drawer's **Clear override** never removed a persisted `chat_template`. Both the save body and the dirty aggregate gated on `overrideOpen`, which the Clear button itself sets to `false` before either predicate runs, so the clear wrote nothing and the override survived on disk while the UI showed the model default. Collapsed the two duplicated predicates into one `chatTemplateDirty` derived from a single `effectiveChatTemplate`, so they can't drift apart again.
- The removal rides as `null`, not `""` — `reconcile_slot_updates` implements None-means-delete, while `""` persists an empty-string override.
- `tests/api/test_slot_config_validation.py` — two new route-level tests: `null` removes the key and leaves siblings (`[model].default`, `port`) intact; `""` is explicitly *not* a removal. No test covered None-deletion of any slot key before.
- W7 here drops its `test.fail()` annotation and tightens to require `null` exactly, plus a second case guarding the inverse (clear → re-pick the same template ⇒ no write).

Red-first evidence is in #1377: with the annotation off and the fix not yet applied, W7 captured `[]` PUTs against an expected one.

Also rebased onto main to pick up #1368 (Pinned/Unpinned header toggle) — re-verified, all 18 original specs still green against it.

Combined verification on the current head:
```
playwright: slot-drawer-field-wiring-v3 slot-drawer-profile-v3 slot-edit-controls-v3 model-drawer
            model-recipe-template-v3 model-edit-identity-v3 slots-v3 slots-wireup-v3 slot-pin-toggle-v3
  63 passed
pytest tests/api/test_slot_config_validation.py  →  13 passed
npm run lint  →  clean
```

Issue #1443, opened 2026-07-30
release: v1.0.0 GA — umbrella (version bump, CHANGELOG cut, tag, channel publish, live validation)


thinmintdev:
Release program umbrella (operator goal 2026-07-30; migrates Linear HAL0-8 + parts of HAL0-2/3). Gates:
- [ ] Fix queue drained (#1432/#1433/#1434 + fix-team PRs) and lxc105 redeployed + verified
- [ ] Version bump 1.0.0-rc.1 → 1.0.0 in pyproject.toml + ui/package.json; CHANGELOG [Unreleased] cut to [1.0.0] with Breaking/Migrations subsections (release.json extraction depends on them)
- [ ] Fresh-install validation on a clean LXC (install.sh, reboot-autostart, doctor)
- [ ] Upgrade-in-place validation on an existing install (halo150 / .150) via the real updater (cosign path)
- [ ] Tag v1.0.0 → release.yml (build/sign/verify/publish); channel pointer authorized
- [ ] Post-release: update channel check from a live box

thinmintdev:
## lxc105 redeploy + live verification — main @ `5be7f2a5` (2026-07-30)

Deployed with the **corrected** procedure — the previous hand-deploy skipped unit re-render and broke two slots. Now: build UI → rsync → `chown root:root` → pip install → record `hal0.previous` → atomic symlink swap → **`run_post_activation_migrations()`** (extracted by #1496) → **`rerender_slot_units()`** → `chmod 0600 api.env` → restart. Migrations returned `(1, 1)`; 1 unit re-rendered.

### Verified working live

| Fix | Evidence on the box |
|---|---|
| **#1417** id-keyed unit probe | `nano` **error → ready**, `agent` **error → offline**, `utility` crash-loop → **serving**. Slots that were healthy-but-reported-dead now report truthfully. |
| **#1367** pin guard | `POST /api/slots/utility/unload` → 409 `slot.pinned`, exact force hint; `pinned` lifted per entry (`agent`/`utility` true from the anchor set). |
| **#1466** api.env perms | `600 hal0:hal0` (was 0644), and the new **`✔ PASS  Secret file modes`** doctor row is live. |
| **#1420** retain-pipeline health | `hal0 memory status` now reports **`Writes: landing`, `failed=174`** — 174 retain failures that were previously invisible behind a healthy-looking `memory_degraded=false`. |
[...]

PR #1484, opened 2026-07-30
fix(memory): fail closed on an unaddressable namespace and unify the delete id contract (#1451, #1456)

### LINKED ISSUE #1451: mcp: memory namespace fail-open-empty collapses to a default shared sweep — approved foreign-bank memory_delete executes against the shared bank
**Severity: HIGH** · area: mcp/memory namespace (post-#1387) · surface: mcp

Audit finding title: Namespace filter's fail-open-empty collapses to a default shared sweep at the provider — an approved foreign-bank memory_delete executes against the shared bank

## Evidence

CONFIRMED on main (a0487038) and still present on github/main — `git diff main github/main --stat -- src/hal0/memory src/hal0/mcp` is empty, so the 4 unmerged upstream commits do not touch this code. Not filed in #1367-#1446 (`gh issue list --search "memory_delete OR fail-open OR namespace filter"` returns only the closed #1300/#1302/#448 and unrelated UI issues).

Static chain re-read: src/hal0/memory/namespace.py:141-142 `return [str(d) for d in requested if is_known_namespace(...)]` — an all-foreign list yields `[]`; the docstring at :128-133 calls this "fail-open-empty". src/hal0/memory/hindsight_provider.py:283/:289 (`list(requested or [_SHARED])`, `return out or [_SHARED]`), :292, and :353 (`_requested_scope`) all treat `[]` as falsy → default sweep. delete at :617 `self._allowed_namespaces(dataset or _SHARED, client_id)`. Same pattern at pgvector_provider.py:64.

Live-executed repro (python, repo src on path):
1. `resolve_read_datasets(["bogus-bank"], private=False, client_id="agent-x")` → `[]`; `["private:other-agent"]` → `[]`.
2. MCP dispatcher: `await memory._memory_delete(w, {"ids":["d1","d2"],"dataset":["bogus-bank"]}, client_id="agent-x", private=False)` → `{'deleted': 2}`, and the dataset handed to the wrapper is `[]` (captured: `[[]]`).
3. Real HindsightProvider + fake client, id `d1` living only in the shared bank:
   unified=True  → `_allowed_namespaces([]) == ['shared']`, delete → `{'deleted': 1}`, `delete_document` called with `bank_id='shared'`.
   unified=False → `_allowed_namespaces([]) == ['shared','private:agent-x']`, delete → `{'deleted': 1}`, bank `shared`.
   Unified mode's `_deletable_ids` gate does not save it: `_requested_scope([])` → `reqs=['shared']` → `(projects=set(), wants_unscoped=True)`, so ordinary unscoped shared docs are deletable.
4. Gating: src/hal0/mcp/admin.py:1247-1248 `if isinstance(requested, list): return True` — the call IS gated, i.e. the operator is shown args naming a nonexistent bank and approves; resolution to `[]`→shared happens after approval, inside the executor.
5. Live box: `/etc/hal0/hal0.toml` line 37 `unified_bank = true` (default is also True, src/hal0/memory/__init__.py:113) — the deployed mode is exactly the single-shared-bank case above.

Test gap re-verified: tests/mcp/test_memory.py:310-320 covers only partial filtering (["agents","not-a-real-bank"] → ["agents"]) and :323-330 the unknown-*string* rejection. Nothing asserts the all-foreign → `[]` case on search/recall/delete, and tests/memory/test_namespace_policy.py has no empty-result assertion at all.

Scope honesty: this is NOT a privilege escalation — the caller can already delete shared ids directly (single-id, no dataset, ungated). The confirmed harm is irreversible deletion of shared documents on a call the operator approved as targeting a different bank, plus reads (search/recall) silently returning shared results for a request scoped exclusively to namespaces the caller may not address. Related but distinct: the REST delete route (src/hal0/api/routes/memory.py:754) skips resolve_read_datasets entirely and forwards the list verbatim, which happens to fail *closed* there — the divergence the namespace module exists to prevent.

## Proposal

Make `[]` mean "no banks": have resolve_read_datasets raise MemoryNamespaceError (or return an explicit empty sentinel) when a non-empty requested list filters to empty, and/or make `_allowed_namespaces`/`_allowed` distinguish `[]` (empty result — sweep nothing) from `None` (default sweep) by replacing `list(requested or [_SHARED])` with an `is None` check. Add tests for the all-foreign list on search, recall, and delete at the dispatcher layer and on both providers; while there, route the REST delete through the same resolver so the two surfaces cannot drift again.

---

### LINKED ISSUE #1456: memory: id contract split — /api/memory/list returns fact ids, delete only accepts document ids; unified-mode delete removes nothing
**Severity: HIGH** · area: hindsight provider / REST + MCP contract · surface: memory

Audit finding title: Memory id contract split: /api/memory/list returns fact ids, delete only accepts document ids — unified-mode delete removes nothing

## Evidence

CONFIRMED on main HEAD a0487038 (== origin/main; hindsight_provider.py last touched by 2ae2e5e2). src/hal0/memory/provider.py:32-38 pins MemoryItem.id as 'the document_id (idempotent, recall-visible, delete-addressable) — NOT a per-fact id'. src/hal0/memory/hindsight_provider.py:540 (_list_fact_to_item) maps list items to `fact.get("id") or fact.get("document_id")` — the per-fact UUID — while recall's _fact_to_item (line 774) correctly prefers `fact.get("document_id") or fact.get("id")`. delete() (line 635) passes the caller's id straight to delete_document, which 404s on a fact id. Live re-verified 2026-07-30: `GET http://127.0.0.1:9177/v1/default/banks/shared/memories/list?limit=1` on hal0 returns id=6310e93a-8030-4b0b-8702-1d57c8a0b01f with a DIFFERENT document_id=19248371-e27e-44f4-b0fd-065f9870bc2b, and `GET http://127.0.0.1:8080/api/memory/list?limit=2` surfaces the fact id 6310e93a as `id` — so the advertised list→delete round trip 404-sweeps to {deleted: 0}. Worse in unified mode (live default: /etc/hal0/hal0.toml:37 `unified_bank = true`): _deletable_ids (hindsight_provider.py:590) matches caller ids against `fact.get("id") or fact.get("document_id")` — the same fact-id field — so even a CORRECT document_id never matches and is fail-closed withheld (line 631 `continue`); POST /api/memory/delete silently deletes nothing against a real Hindsight 0.8.x. Tests mask it: tests/memory/test_unified_bank.py:57-60 fakes list_memories with `"id": f["document_id"]` (fact id == document id, unlike the real shape). Route docstring src/hal0/api/routes/memory.py:735-738 ('Identity headers otherwise are not consulted') is stale vs the unified-mode ACL gate that consults client_id via _deletable_ids. Not fixed on main; not filed: swept gh issues #1367-#1446 — #1387 (single-id delete approval gating), #1302 (CLOSED, auth posture + /mcp/memory bulk-delete gate), #1420 (retain/degrade flag) are all different defects.

## Proposal

Make _list_fact_to_item expose document_id as `id` (matching recall and the ABC contract), or carry both fields; fix _deletable_ids to match on document_id; add a contract test whose fake list_memories has distinct fact id vs document_id (mirroring the real 0.8.x shape) covering list→delete and add→delete round trips in unified mode.

---

thinmintdev:
Two HIGH findings from the v1.0 polish audit (umbrella #1443). They meet in the same delete path, so they land together.

## #1451 — an all-foreign namespace list collapsed into a shared-bank sweep

`resolve_read_datasets` filtered a caller's `dataset` list against the spec §3 closed table and documented the result as **fail-open-empty**. That is correct for a partial drop (`["agents", "nope"]` → `["agents"]`) and catastrophic when the last entry goes, because `[]` is falsy and every consumer downstream read it back as "nothing requested":

- `hindsight_provider._allowed_namespaces`: `list(requested or [_SHARED])`, then `return out or [_SHARED]`
- `hindsight_provider._requested_scope`: same
- `hindsight_provider.delete`: `self._allowed_namespaces(dataset or _SHARED, ...)`
- `pgvector_provider._allowed`: same

So `dataset=["bogus-bank"]` resolved to `[]` at the front door and back to `["shared"]` inside the executor.

The operator-visible shape is what makes this HIGH rather than cosmetic. A bulk `memory_delete` is approval-gated on its **arguments** (`mcp/admin.py:1247` gates any list-valued dataset), so the operator was shown a call naming a bank that does not exist, approved it, and the executor deleted live documents out of `shared` — which, on the deployed default (`unified_bank = true`), is the only bank there is. Reads had the quieter half: a search or recall scoped exclusively to namespaces the caller may not address returned shared rows.

**Fix.** `[]` means *no banks*; only `None` may expand to the default.
- A non-empty request that resolves to nothing raises `MemoryNamespaceError` at the front door (400 REST / `mcp.memory_schema`), naming the rejected input.
- Both providers gain an explicit `is None` normalisation, so an empty list reaching a provider *without* the resolver sweeps nothing — defence in depth, because the REST route and the providers are both reachable that way.
- `POST /api/memory/delete` stops hand-rolling its list branch (`[str(d) for d in requested]`) and goes through `resolve_read_datasets` like the MCP surface, closing the exact two-surface drift `hal0.memory.namespace` exists to prevent.

Partial-drop degradation is unchanged and has its own negative control.

## #1456 — list returned fact ids, delete accepted document ids

`MemoryProvider` pins `MemoryItem.id` as "the document_id — idempotent, recall-visible, delete-addressable, **NOT** a per-fact id". `_fact_to_item` (recall) honoured it (`document_id or id`); `_list_fact_to_item` (list) inverted it (`id or document_id`). On a real Hindsight 0.8.x those are different UUIDs on the same item, so:

- the advertised list→delete round trip handed `delete_document` a fact id, 404-swept every bank, and returned `{"deleted": 0}` behind a 200;
- worse in unified mode (the deployed default): `_deletable_ids` matched caller ids against the same fact-id field, so even a **correct** document_id never matched and was fail-closed withheld — `POST /api/memory/delete` deleted nothing at all, for anyone.

**Fix.** `id` is the document_id on every surface. The per-fact id moves to `metadata.fact_id` rather than disappearing — it is the only handle on an individual extracted fact. `_deletable_ids` now returns a caller-id → document_id mapping, so a caller holding either handle resolves to the owning document while the engine is still addressed by document_id only. The route docstring's stale "Identity headers otherwise are not consulted" is corrected — unified mode *does* consult `client_id` through the ACL gate.

Widening the id match did **not** widen the ACL: an unresolvable id and another agent's `visibility:private` doc are still withheld, with explicit negative controls.

## Verification

Red-first, per defect.

`tests/security/test_memory_namespace_fail_closed.py` + `tests/memory/test_memory_id_contract.py` written against unmodified `main`: **24 failed, 6 passed**. Failures covered the resolver (4 parametrised all-foreign shapes), both providers in both bank modes, the three MCP handlers, and the REST route. After the fix: **30 passed**.

The id-contract file exists because every pre-existing test agreed with the bug — `tests/memory/test_unified_bank.py:57-60` fakes `list_memories` with `{"id": f["document_id"]}`, a shape the engine never emits. The new fake returns `id` and `document_id` distinctly, mirroring the live payload quoted in the issue.

Full suite: `HAL0_HOME=$(mktemp -d) uv run pytest -q` → **7866 passed, 16 skipped, 1 xfailed, 0 failed**. `ruff check` / `ruff format --check` / `scripts/check_sunset.py` clean.

> Note for the wave: the ~204 `tests/mcp` "env errors" everyone has been ignoring are `PermissionError: /etc/hal0/hal0.toml` and are fixed by setting `HAL0_HOME` to a temp dir. `tests/mcp` alone goes 33 errors → 213 passed. That bucket includes the namespace tests this PR touches.

## Out of scope

The `unified_bank=false` legacy multi-bank path keeps its no-gate delete (bank isolation already protects it); only the `[]` normalisation reached it.

Closes #1451
Closes #1456


thinmintdev:
Local full-suite verification (post-rebase, at `e2b2c403`):

```
7900 passed, 16 skipped, 1 xfailed, 48 warnings in 1006.01s (0:16:46)
```

`ruff check src tests` clean · `ruff format --check` clean (1006 files) · `scripts/check_sunset.py` clean (scar markers 194 <= baseline 194; no overdue sunset shims).

Red-first evidence, both new files against unmodified `main`: **24 failed, 6 passed** → **30 passed** after the fix.

PR #1485, opened 2026-07-30
fix(memory): guard the bank-memories wipe with the #1024 echoed confirm (#1457)

### LINKED ISSUE #1457: memory: DELETE /api/memory/banks/{id}/memories is an unguarded one-call bank wipe — the #1024 incident class survives on the sibling route
**Severity: HIGH** · area: memory_admin routes / delete-surface hardening · surface: memory

Audit finding title: DELETE /api/memory/banks/{id}/memories is an unguarded one-call bank wipe — the #1024 incident class survives on the sibling route

## Evidence

CONFIRMED on main HEAD a0487038 (== origin/main). src/hal0/api/routes/memory_admin.py:474 forwards ("DELETE", "/banks/{bank_id}/memories") verbatim through the generic _make_handler with audit only (_DELETE_ACTIONS 'memory.memories.delete', line 632) — no ?confirm echo, no blast-radius preview, no approval gate. Only DELETE /banks/{bank_id} got the post-#1024 guard: it is special-cased out of _FORWARDS into delete_bank (memory_admin.py:306-331) with echoed-id confirm + _delete_preview. The route IS listed in DESTRUCTIVE_MEMORY_ROUTES (src/hal0/security/exposure.py:375) and classifies ADMIN — but that classification is inert in the live default posture, re-verified 2026-07-30 on hal0: `GET /api/auth/status` → {"auth_required":false,"has_admin_key":false,"tier":"anon"}, and the ADMIN-class routes `GET /api/memory/banks/shared/stats` and `GET /api/memory/banks/shared/memories?limit=1` both answer 200 unauthenticated. Upstream semantics re-confirmed from Hindsight's live openapi (:9177): DELETE /v1/default/banks/{bank_id}/memories = 'Delete memory units for a memory bank … This is a destructive operation that cannot be undone.' Live blast radius: shared bank stats show total_nodes=1629, total_documents=315. So one anonymous `curl -X DELETE http://<host>:8080/api/memory/banks/shared/memories` wipes every memory unit with zero friction — the same irreversible-loss class #1024 documented, one path over. Not fixed on main; not filed: #1024 and #1302 are CLOSED and neither names this route; nothing in #1367-#1446 covers it.

## Proposal

Give the bank-memories wipe the same echoed-id confirm guard (and #1024-style dry-run preview) as DELETE /banks/{bank_id}; it is already in DESTRUCTIVE_MEMORY_ROUTES so only the handler needs special-casing out of the _FORWARDS table.

---

thinmintdev:
HIGH finding from the v1.0 polish audit (umbrella #1443). The #1024 incident class, surviving one route over.

## The gap

#1024 was `DELETE /api/memory/banks/{bank_id}` reached with a single unauthenticated `curl`, cascade-deleting ~632 live records. Its hardening — the echoed-`?confirm=<bank_id>` gate plus a dry-run blast-radius preview (#1028), and a `record_action` audit row (#1030) — was applied by lifting that **one** path out of the generic `_FORWARDS` passthrough table into a hand-written handler.

`DELETE /api/memory/banks/{bank_id}/memories` stayed in the table. Upstream documents it as *"Delete memory units for a memory bank … This is a destructive operation that cannot be undone."* It had the audit row and nothing else — no confirm echo, no preview, no gate. So the `record_action` entry described a wipe that had already happened, on a call anyone could make.

**Classification did not save it.** The route was already pinned ADMIN in `exposure.DESTRUCTIVE_MEMORY_ROUTES`, and in the shipped default posture that decides nothing: the audit re-verified `GET /api/auth/status` → `{"auth_required": false, "has_admin_key": false, "tier": "anon"}`, with ADMIN-class routes answering 200 unauthenticated, against a `shared` bank holding 1629 nodes across 315 documents. Classification is not a guard.

## The fix

`delete_bank_memories` is special-cased out of `_FORWARDS` for the same reason its sibling was: the generic handler cannot ask for an echo it knows nothing about. It now:

- requires `confirm=<bank_id>`, read from the query string **or** the JSON body — the same two places `delete_bank` reads it, so the pair cannot diverge on that detail;
- returns the same 400 `memory.confirm_required` payload with the same `_delete_preview` blast-radius counts, fail-soft when the stats probe is unreachable;
- forwards nothing until confirmed;
- keeps its `memory.memories.delete` audit row, and a *rejected* wipe produces none — a 400 is not a destructive op;
- still passes upstream filter params through, with `confirm` stripped so hal0's own gate never reaches the engine as an unrecognised filter.

Both handlers share one `_require_echoed_confirm`. Duplicating the check per route is precisely how they came to differ.

## The structural half

Adding a guard fixes one route. The reason this survived #1024 is that the passthrough table is the default and irreversibility is invisible in it.

New `exposure.CONFIRM_GUARDED_MEMORY_ROUTES` (a subset of `DESTRUCTIVE_MEMORY_ROUTES`) names the routes whose blast radius is a whole bank and which therefore **must** carry the echo. `tests/security/test_memory_bank_wipe_guard.py` drives every entry in that set against the live app — so naming a route there without wiring the guard fails, and a new bank-scoped delete dropped into `_FORWARDS` cannot inherit the passthrough silently. This mirrors what `tests/security/test_memory_delete_auth.py` already does for the ADMIN classification: state the claim as a constant, then prove it against the running route table.

## Verification

Red-first: the new file written against unmodified `main` (with only the constant added so the module imports) → **7 failed, 6 passed**. The 7 were the guard, both rejection shapes, the preview, the fail-soft preview, the confirm-token stripping, the audit split, and the parametrised `CONFIRM_GUARDED_MEMORY_ROUTES` row for the memories path. The 6 that already passed are the sibling bank-delete's existing guard — the negative control proving the harness itself works.

After: **106 passed** across `tests/security` + `tests/api/test_memory_admin_routes.py` (the pre-existing `delete_bank` coverage still green through the shared-helper refactor).

`ruff check` / `ruff format --check` / `scripts/check_sunset.py` clean. Full-suite result posted as a comment.

## Out of scope

The dashboard's memory console has no UI for this route, so there is no confirm dialog to add — the API gate is the whole surface today. If a UI lands, it should reuse the `preview` payload the 400 already returns.

Closes #1457


thinmintdev:
Local full-suite verification (post-rebase, at `18903b1a`):

```
7883 passed, 16 skipped, 1 xfailed, 48 warnings in 959.90s (0:15:59)
```

`ruff check src tests` clean · `ruff format --check` clean (1005 files) · `scripts/check_sunset.py` clean.

Red-first evidence: the new file against unmodified `main` (with only `CONFIRM_GUARDED_MEMORY_ROUTES` added so the module imports) → **7 failed, 6 passed**. The 6 pre-existing passes are the sibling bank-delete's guard — the negative control proving the harness works before the fix lands.

PR #1486, opened 2026-07-30
fix(security): make api.env owner-only and stop exposing service config as removable secrets (#1466, #1450)

### LINKED ISSUE #1450: settings: Secrets page exposes HAL0_* service config and auth keys as one-click, no-confirm removable secrets
**Severity: HIGH** · area: settings/secrets · surface: settings

Audit finding title: Settings ▸ Secrets exposes every api.env key — including HAL0_ADMIN_KEY/HAL0_CLIENT_KEY and service config — as one-click, no-confirm removable 'secrets'

## Evidence

CONFIRMED on main (HEAD a0487038) and live. src/hal0/api/routes/secrets.py: list_secrets (@router.get(""), ~L145) builds rows from list_env_keys(api_env) with no filter; src/hal0/api/_env_store.py:147 list_env_keys returns every uncommented KEY= line of /etc/hal0/api.env. Live re-run `ssh hal0 curl -s localhost:8080/api/secrets` returned 14 rows including HAL0_BIND_HOST, HAL0_PORT, HAL0_UI_DIST, HAL0_LOG_LEVEL, HAL0_MEMORY_ENABLED, HAL0_MCP_ALLOWED_HOSTS, HAL0_FLM_MODELS_DIR, HAL0_TURNSTONE_URL/CONSOLE_URL/TOKEN alongside real secrets (HF_TOKEN, OPENROUTER_API_KEY, MINIMAX_API_KEY, HERMES_SESSION_TOKEN). Auth keys: not yet present on this box (never rotated), but src/hal0/api/routes/auth.py:211 rotate_key → rotate_api_env_key and src/hal0/service_identity.py:29 _KEY_ENV = {admin: HAL0_ADMIN_KEY, client: HAL0_CLIENT_KEY} write them to the same api.env, after which they list as rows; _validate_name (^[A-Z][A-Z0-9_]{0,63}$, secrets.py ~L100) accepts them, and delete_secret (secrets.py ~L223) removes the api.env line atomically, pops os.environ live, and returns idempotent 204 — auth.py validates login against HAL0_ADMIN_KEY from env, so new sessions lock out. UI: ui/src/dash/settings/pages/integrations/SecretsPage.jsx Remove button onClick calls delSecret.mutate(s.name) directly — no ConfirmDialog anywhere in the file — and unknown keys are captioned 'Custom key · exported to hal0 services and slot containers as an env var'. Not already filed: enumerated all GH issue titles #1367–#1446, none cover this; closed #553 (auto-redact config values) is a different surface and adds no list filter or delete gate.

## Proposal

Filter the list route (or the page) to hide/lock reserved HAL0_* service-config and auth-key names (read-only rows with an explanatory chip), and add a type-to-confirm or at least a ConfirmDialog on Remove — deleting HAL0_ADMIN_KEY disarms auth (new sessions locked out while require_auth is on), deleting HAL0_PORT/HAL0_UI_DIST breaks the service on next restart.

---

### LINKED ISSUE #1466: security: secrets land in world-readable /etc/hal0/api.env — installer re-run and the perms engine both chmod it back to 0644
**Severity: HIGH** · area: secrets / key handling · surface: security

Audit finding title: Secrets land in world-readable /etc/hal0/api.env — installer re-run and the perms engine both chmod it back to 0644

## Evidence

CONFIRMED by re-verification 2026-07-30. Live lxc105: `stat -c '%a %U:%G' /etc/hal0/api.env` → `644 hal0:hal0`, and `grep -o '^[A-Z_0-9]*='` (names only, values never printed) shows the file currently holds HF_TOKEN, MINIMAX_API_KEY, OPENROUTER_API_KEY, HERMES_SESSION_TOKEN, HAL0_TURNSTONE_TOKEN. FOUR writers disagree on the mode, one more than originally filed: (1) /mnt/mintdev/repos/hal0/installer/install.sh:919 `chmod 0644 "${API_ENV_TMP}"` in the network-block refresh branch — and commit 233e305e (fix #1375, on main) made that refresh run on EVERY re-run over an existing api.env, so any upgrade/repair downgrades the file to 0644; (2) /mnt/mintdev/repos/hal0/src/hal0/api/_env_store.py:57,74 — dashboard Settings→Secrets writes 0o600; (3) /mnt/mintdev/repos/hal0/src/hal0/service_identity.py:35 `_API_ENV_MODE = 0o640` for key rotation, with routes/auth.py:169-171 promising 'mode forced to a never-world-readable 0640'; (4) NEW: /mnt/mintdev/repos/hal0/src/hal0/install/perms.py:175-177 — the perms-enforcement table pins api.env at `0o644` with an explicit `FIXME(phase4): api.env is 0644 (world-readable) but may carry tokens — candidate for 0640 root:hal0`, so the perms engine independently re-flattens any 0600/0640 tightening. install.sh:896-901 simultaneously warns 'api.env is 0644, world-readable' while recommending the dashboard Secrets path, which writes exactly that file. Not fixed on main (verified against HEAD a0487038 and the git log of all three originally-cited files). Not filed in GH #1367-#1446: `gh issue list -R Hal0ai/hal0 --state all --search` for 'api.env', 'world-readable', 'secret', '0644' returns nothing in range; #1106 (closed) covers only the install-time HF_TOKEN gather to /var/lib/hal0/secrets/hal0-api.env — a file that does not even exist on the live box (dir /var/lib/hal0/secrets is 755 root:root, no hal0-api.env inside), so every live secret sits solely in the 0644 file.

## Proposal

Single shared mode constant for api.env (0640, owner hal0 or root:hal0) honored by all FOUR writers: installer/install.sh:919 (preserve existing mode or hard-set 0640), src/hal0/api/_env_store.py (0600→0640 or keep 0600), src/hal0/service_identity.py (already 0640), and src/hal0/install/perms.py:177 (resolve the FIXME(phase4) — flip the PermRow to 0o640, otherwise the perms engine reverts any fix). Longer term, route dashboard-set tokens to the root-only /var/lib/hal0/secrets/hal0-api.env EnvironmentFile the installer already provisions for. Immediate mitigation on the live box: chmod 0640 /etc/hal0/api.env (requires write access; out of scope for this read-only verification pass).

---

thinmintdev:
Two HIGH findings from the v1.0 polish audit (umbrella #1443). They are the same mistake twice: `/etc/hal0/api.env` is two stores in one file — the operator's third-party credentials **and** hal0's own service configuration + auth keys — and everything that touched it treated it as one.

---

## #1466 — secrets sat in a world-readable file, and every repair path put them back

Live `lxc105`: `stat -c '%a %U:%G' /etc/hal0/api.env` → `644 hal0:hal0`, while the file held `HF_TOKEN`, `MINIMAX_API_KEY`, `OPENROUTER_API_KEY`, `HERMES_SESSION_TOKEN` and `HAL0_TURNSTONE_TOKEN`. Every local account could read them.

Four writers, three opinions — so the strictest always lost:

| # | Writer | Mode | Effect |
|---|---|---|---|
| 1 | `installer/install.sh:919` | `chmod 0644` | 233e305e (#1375) made the network-block refresh run on **every** re-run over an existing api.env, so any upgrade or repair re-published the secrets |
| 2 | `api/_env_store.py` | `0o600` | correct — and promptly undone by (1) or (4) |
| 3 | `service_identity.py` | `0o640` | key rotation, with `routes/auth.py` promising "mode forced to a never-world-readable 0640" |
| 4 | `install/perms.py:177` | `0o644` | `FIXME(phase4): api.env is 0644 (world-readable) but may carry tokens`. The engine whose *job* is converging the filesystem independently reverted every tightening |

Note the shape: (2) and (3) were already fixes. They did not hold, because (1) and (4) ran after them. Patching any single writer would have looked correct in a diff and changed nothing on a real box.

**Fix — one constant, four honourers.** `hal0.config.paths.API_ENV_MODE` (`0o600`), plus a single `paths.api_env()` resolver replacing the `paths.etc() / "api.env"` construction scattered across the tree. `_env_store`, `service_identity` and the `install/perms` PermRow all read it; `install.sh` sets `0600` on **both** the initial write and the refresh (a mismatch there means a re-run undoes the fresh install). Because the mode rides the temp file through the `os.replace`, the next dashboard write now *repairs* an already-widened file rather than preserving it — the upgrade path that actually matters.

`0600` rather than the FIXME's suggested `0640`: systemd reads `EnvironmentFile=` as the service manager, before dropping privileges, and the CLI reads it as root or as the service user — nothing needs the group bit, and a group-readable file becomes world-readable the moment a second account joins the group.

**Plus a backstop that does not trust the table.** New `hal0 doctor` row `check_secret_file_modes` fails **critical** on any group- or world-readable `api.env` / `openwebui.env`. Deliberately *not* generated from `install/perms.py` — that table is what got it wrong, and a check derived from it would have agreed with the bug. It asserts the property directly, so a fifth writer, a re-widened PermRow, or a hand `chmod` all surface.

---

## #1450 — the Secrets page offered service config and auth keys as one-click removals

`list_secrets` built its rows from `list_env_keys(api_env)` with no filter — every uncommented `KEY=` line. `delete_secret` removed anything matching `^[A-Z][A-Z0-9_]{0,63}$`: atomically, popping `os.environ` live, returning an idempotent 204.

So the dashboard rendered, as ordinary rows with a red **Remove** button and no confirmation dialog anywhere in the file:

- `HAL0_ADMIN_KEY` / `HAL0_CLIENT_KEY` — written there by `service_identity.rotate_api_env_key`, and what `routes/auth.py` validates every login against. Removing it locks every new session out while `require_auth` is on.
- `HAL0_PORT`, `HAL0_UI_DIST`, `HAL0_BIND_HOST`, `HAL0_MCP_ALLOWED_HOSTS`, … — removing them breaks the service on next restart.

All captioned *"Custom key · exported to hal0 services and slot containers as an env var"*.

**Fix.** `HAL0_` is a reserved namespace: owned by the installer, `hal0.toml` derivation and key rotation, not by this route.

- Reserved keys still **list** — an operator should be able to see what the service is configured with, and hiding them trades one lie for another — but carry `protected: true`.
- `POST`/`PUT`/`DELETE` on them is **403 `secret.protected`**, server-side. The UI is not the only caller; a dialog is a courtesy, this is the control. Refused deletes touch neither the file nor `os.environ`, and the refusal names the key but never its value.
- A prefix rule rather than an enumerated name list, on purpose: a list goes stale the moment a new `HAL0_*` var enters `install.sh`, and it is the *unlisted* one that then stays deletable. Nothing in `SECRET_PRESETS` has ever used this prefix, so no supported workflow is narrowed.
- The page renders protected rows locked with no mutating control at all, and every remaining **Remove** now goes through a type-the-name `ConfirmDialog` — the value is never stored anywhere else and never shown again after saving, so there is nothing to undo with.
- `mockFixtures.buildSecrets` gains the `protected` field and a `HAL0_ADMIN_KEY` row, so the forced-mock dashboard stops teaching the UI a payload shape the server never sends.

---

## Verification

Red-first, per defect.

`tests/security/test_api_env_mode.py` + `tests/security/test_secrets_protected_keys.py` against unmodified `main`: **red**, then **39 passed**. Coverage is per-writer rather than per-symptom — the constant, its group/world property stated independently of the number, each of the three Python writers (including "a pre-existing 0644 file is tightened by the next write"), the perms row, and `install.sh` checked textually since it cannot import Python (no `chmod 06[24]4` may remain near `API_ENV`; `0600` must appear on both branches).

`ui/tests/e2e/specs/secrets-remove-guard.spec.ts`: **5 passed** (locked rows expose no button and issue no DELETE; the confirm button stays disabled until the exact name is typed and on a near-miss; cancel deletes nothing; the wire DELETE fires only after confirmation). The pre-existing `hf-token-setting.spec.ts` still passes.

`ruff check` / `ruff format --check` / `scripts/check_sunset.py` clean; `bash -n installer/install.sh` clean. Full-suite result posted as a comment.

## Deployment note

Existing boxes keep `644` until something writes the file or `hal0 doctor perms --fix` runs — the new doctor row is what makes that visible. Immediate manual mitigation is `chmod 0600 /etc/hal0/api.env`.

Closes #1466
Closes #1450


thinmintdev:
Local full-suite verification (at `6eff1e7e`):

```
7909 passed, 16 skipped, 1 xfailed, 48 warnings in 1011.58s (0:16:51)
```

`ruff check src tests` clean · `ruff format --check` clean (1006 files) · `scripts/check_sunset.py` clean · `bash -n installer/install.sh` clean.

Red-first: the two new security files against unmodified `main` → red, then **39 passed**. `ui/tests/e2e/specs/secrets-remove-guard.spec.ts` → **5 passed** (plus the pre-existing `hf-token-setting.spec.ts`).

### Three pre-existing tests updated, deliberately

The first full-suite run surfaced exactly three failures, all of them tests that encoded the *old* contract rather than regressions. Listing them because "I changed the test" deserves the reason:

- `tests/api/test_auth_rotate.py::test_rotate_preserves_non_world_readable_mode` — asserted the local `0o640` and `not (mode & 0o007)`. Now asserts `paths.API_ENV_MODE` and `not (mode & 0o077)`: a group-readable file stops being "not world-readable" the moment a second account joins the group, which is the reason 0640 was not good enough.
- `tests/install/test_perms.py::test_flip_makes_etc_hal0_service_owned_and_setgid` — asserted `api.env` stays `0o644` under the comment *"mode warts are NOT touched by the flip"*. It was the last wart in that list; the assertion now pins `paths.API_ENV_MODE` **and** the literal `0o600`, so the constant cannot be widened without this failing too.
- `tests/cli/test_doctor_all.py::test_build_all_checks_composes_verify_plus_extras` — the row-composition list gained `secret-modes`. It also now monkeypatches `_SECRET_FILES` to empty: the new check stats real paths, and without that pin this composition test would pass or fail depending on the host's `/etc/hal0` modes. (On this dev box it correctly went red — `/etc/hal0/api.env` really is 0644 there, which is the finding.)

PR #1555, opened 2026-07-31
fix(slots): don't re-check default uniqueness on unrelated field edits

thinmintdev:
## Summary
- `check_default_uniqueness()` fired on every `update_config()` write whose merged config carried `default=true`, even when the caller's PATCH never touched `default`. A slot that already legitimately carries `default=true` (or has a pre-existing stale duplicate default from before this guard existed) could never have any other field edited — e.g. bumping `context_size` on the dashboard's slot edit drawer failed with `slot type 'llm' already has a default=true slot`.
- Gated the check on `changed_keys`: it now only fires when the write is the one actually setting `default=true`. Retroactive cleanup of a stale duplicate stays `default_slot_for`'s job at routing time (matches what the guard's own docstring already claimed but didn't implement).

## Test plan
- [x] `tests/slots/test_default_uniqueness.py` — added `test_update_config_unrelated_field_not_blocked_by_stale_peer_default`, all 8 tests pass
- [x] `tests/slots/` full suite — 658 passed

### DIFF
--- a/src/hal0/slots/config_write.py
+++ b/src/hal0/slots/config_write.py
@@ -309,7 +309,9 @@ def reconcile_and_guard_slot_config(
     """
     merged = reconcile_slot_updates(base, updates)
     check_npu_exclusivity(slot_name, merged, slots_dir=slots_dir)
-    check_default_uniqueness(slot_name, merged, slots_dir=slots_dir, changed_keys=set(updates.keys()))
+    check_default_uniqueness(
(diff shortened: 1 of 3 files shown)


thinmintdev:
Superseded: main commit 0c6b3a41 (#1560) already landed the same changed_keys gating in check_default_uniqueness plus equivalent test coverage. The merge conflicts here are the two implementations of the same fix colliding. Closing without merge.

PR #1631, opened 2026-08-08
feat(slots): on-demand capability slots — retire the vision lane, fix seed delete/recreate

thinmintdev:
## What

Per operator direction: capability slots are created **on demand** when their capability is enabled in settings — being a seed no longer grants special delete/create rules — and the dedicated `vision` slot lane is retired outright (vision is a model property, served by any llm slot whose model carries it).

## Changes

- **Delete guard**: the `cannot delete seeded slot` refusal is gone. Protection is the `pinned` lever alone (§21.10) — the default-pinned anchors (`agent`/`utility`/`npu`) still 409 without `force`, everything else deletes cleanly. Existence is now checked before pinned, so unknown names 404 instead of 409.
- **Seed tombstones**: deleting a static seed writes a tombstone; the boot-time seeding pass honours it instead of resurrecting the slot every start (the mechanism that made seeds effectively undeletable). Recreating the name clears the tombstone.
- **Recreate**: seed names are no longer reserved in `add_slot` — only the NPU trio names stay reserved (they collide with the occupancy pane's synthesised sub-cards).
- **Drift fix**: `SEEDED_SLOTS` is now a re-export of `STATIC_SEED_SLOTS`. The hand-copied list had drifted badly: `stt`/`vision` were protected-but-never-seeded ghosts; `brain`/`coder`/`flm`/`qwen3tts` were seeded-but-unprotected.
- **Vision lane retirement**: `vision.vision` capability child, catalog panel row, and orchestrator mappings removed; `capabilities.toml` drops a stray `[selections.vision]` on load; a boot migration deletes the untouched vision scaffold slot (a vision slot with a model bound is operator-owned and left alone — it keeps working as a plain llm slot). Model-side vision (`models_for_capability("vision")`, curated tags, `LoadedSlot.modalities`) is untouched.

## Verification

slots/capabilities/config/install suites: 1790 passed. New tests pin: seed deletable without force + tombstone honoured by seeding, recreate clears tombstone, anchor pin still enforced, NPU names still reserved, vision child retired, stray selections dropped, scaffold migration (remove/keep/idempotent).

### DIFF
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -1177,13 +1177,13 @@ async def _boot_slot_reconcile(app: FastAPI, ctx: BootState) -> None:
         log.warning("registry.type_tags_retirement_failed", error=str(exc))
 
     # Retire the pre-1.0 ``vision`` scaffold slot (untouched scaffolds only —
-    # vision is a model property now, served by any llm slot; see
-    # hal0.config.migrations.vision_slot_retirement).
+    # vision is a model property now, served by any llm slot). Routed through
+    # SlotManager.delete so identity row, port claim, and state go with the
+    # TOML on id-keyed boxes; see hal0.config.migrations.vision_slot_retirement.
     try:
-        from hal0.config import paths as _vision_paths
-        from hal0.config.migrations.vision_slot_retirement import migrate_vision_slot
+        from hal0.config.migrations.vision_slot_retirement import retire_vision_scaffold
 
-        if migrate_vision_slot(_vision_paths.slots_config_dir()):
+        if await retire_vision_scaffold(ctx.slot_manager):
             log.info("slot.vision_scaffold_retired")
     except Exception as exc:  # pragma: no cover — defensive
         log.warning("slot.vision_retirement_failed", error=str(exc))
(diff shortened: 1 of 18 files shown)


thinmintdev:
Codex findings all addressed in `dde694cd` (plus the #1632 residuals, which now live on this rebased branch):

- **Retire vision through the slot lifecycle (P2):** retirement now routes through `SlotManager.delete(force=True)` — identity row, port claim, and `slot_data_dir` state go with the TOML on id-keyed boxes.
- **Serialize tombstone RMW (P2):** `add/clear_seed_tombstone` now run under the cross-process `slot_write_lock`.
- **Honor NPU shadow deletions (P2):** inverted — tombstones are written for `STATIC_SEED_SLOTS` only; the trio is reconciled from the anchor by contract, so a trio tombstone was a dead promise.
- **Stack routing vs retirement (P1):** a saved stack's `child="vision"` row is skipped with a log instead of failing convergence (`_RETIRED_CHILDREN`).
- **Setup recreating vision (P2):** `vision` removed from `_SETUP_SLOTS` / `_SCAFFOLD_CAPS`.
- **Reconcile capability intent on delete (P2):** `SlotManager.delete` best-effort disables the slot's `capabilities.toml` selection via `disable_selections_for_slot`; re-enabling in settings recreates the slot.
- **#1632 residuals:** vision tag folds into capabilities only alongside an mmproj sidecar (P1); moe strips only when `architecture` is typed (P2); coder kept as the descriptive bench label (P2); bench capability union also reads migrated `defaults.mtp` (P2). Drawer field-help counts were fixed pre-merge.

Suites: install/stacks/slots/capabilities/registry/bench/config/cli/api — 5,390 passed locally.

PR #1638, opened 2026-08-08
fix(ui): slot drawer — widen info popups, consolidate lifecycle controls

thinmintdev:
## What changed

- **Field-info popups no longer collapse to one word per line.** `.field-info-pop` is absolutely positioned inside its ~19px icon wrapper, so shrink-to-fit resolved against that tiny containing block. `width: max-content` (capped at the existing 280px) sizes it to the text instead. Fixes every `(i)` popup app-wide (slot/model drawers, settings pages).
- **Slot drawer lifecycle controls consolidated.** Auto-Load moved from the Model section into the drawer header next to the Pinned toggle, same instant-apply style — the pair now reads as one story: Auto-Load = when the slot starts (boot), Pin = whether it may be stopped (eviction exemption + force-guarded unload/delete). Eviction priority moved into the Advanced disclosure. Net: two fewer always-visible rows; NPU slots (no Model section) gain the Auto-Load toggle.
- **Copy clarified.** Auto-Load and Pin tooltips now define each control against the other.

## Why

The drawer was getting tall, and Pinned vs Auto-load read as competing features when they're orthogonal (startup vs residency).

## Verification

- `slot-pin-toggle-v3` spec updated for the two-toggle header (testid selectors) and extended: P4 header Auto-Load fires `PUT /config { autoload }`; P5 priority hidden until Advanced opens, `PUT /config { priority }` on blur.
- 20/20 e2e green (`slot-pin-toggle-v3` + `slot-edit-controls-v3`), eslint clean, CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,21 @@ applying. Add those subsections to a version's section to surface them; see
 
 ## [Unreleased]
 
+### Fixed
+
+- Field-info popups (the `(i)` descriptions in the slot/model drawers and
+  settings pages) no longer collapse to one word per line — the popup was
+  shrink-to-fit against its tiny icon wrapper; it now sizes to its text up
+  to the 280px wrap width. The slot drawer's Auto-load and Pin descriptions
+  now also spell out the difference: Auto-load only controls boot start,
+  Pin only controls residency (eviction exemption + guarded unload/delete).
+- Slot drawer lifecycle controls consolidated: Auto-Load moved from the Model
+  section up into the drawer header next to the Pinned toggle (same
+  instant-apply style — the pair now reads as one story: Auto-Load = when it
+  starts, Pin = whether it may be stopped), and Eviction priority moved into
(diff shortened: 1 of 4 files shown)


thinmintdev:
Addressed all four Codex findings in 0efadd87: header now reserves space for the toggle group and ellipsizes long titles; info popups drop below the icon with a viewport cap at ≤720px; both header toggles carry aria-describedby with visually-hidden descriptions; Auto-Load and Eviction priority are hidden for NPU trio shadows (no process of their own).

thinmintdev:
Follow-up on the mobile clipping finding (comment 3741769657): confirmed still valid — `.field-info-pop`'s ≤720px placement is viewport-aware but not panel-aware, so an icon near an `.s-panel`'s (`overflow: hidden`) edge or last row can still clip. A correct fix needs the popup to escape its clipped ancestor (portal + `getBoundingClientRect`, or fixed positioning with a computed anchor) — bigger than this PR's copy/layout scope. Filed #1683 to track; not fixing inline here.

PR #1679, opened 2026-08-08
fix(agents): scope per-agent memory stats by client_id

### LINKED ISSUE #1654: memory: per-agent memory-stats chip reports the whole shared bank instead of that agent's private writes
## What's broken

`src/hal0/api/agents/memory_stats.py:169` calls `wrapper.list_items(dataset="private:<agent>", cursor=None, limit=500)` with no `client_id` — every other caller (`routes/memory.py:723`, `mcp/memory.py:454`) passes `client_id`.

In unified-bank mode (the shipped default — `provider_from_config` passes `unified_bank=True`), `HindsightProvider._allowed_namespaces` collapses `private:<agent>` → `[shared]`, and `_filter_visible` then resolves the caller via `_caller_agent(None)` → `"unknown"`. Every doc actually tagged `visibility:private` + `agent:<agent>` is therefore withheld (owner != "unknown"), while `_requested_scope("private:<agent>")` yields `wants_unscoped=True`, so every ordinary shared doc passes straight through.

The result is the exact inverse of the module's own documented contract ("Stats are scoped to the per-agent private namespace… the shared dataset is global and would muddy the chip"): `writes` becomes the shared-bank page count (capped at 500) and `last_write` is the newest shared doc from *any* agent.

**Fix:** pass `client_id=agent_id`.

## Repro

`GET /api/agents/hermes/memory/stats` on a unified-bank install where hermes has written 3 private items and the shared bank holds 500+ docs → `{"writes": 500, "last_write": "<newest shared doc ts>", "available": true}` instead of `writes: 3`.

## Severity

P2 — the per-agent activity chip on every agent card reports meaningless (shared-bank) numbers instead of that agent's own writes; existing tests miss it because the test double ignores `client_id` entirely.

## Found by

hal0 rc.2 pre-release review (functional review, "memory" section), verified by tracing `HindsightProvider`'s namespace/visibility resolution end to end.

thinmintdev:
## Summary
- `get_agent_memory_stats` (`memory_stats.py:169`) called `wrapper.list_items(dataset="private:<agent>", cursor=None, limit=500)` with no `client_id` — every other caller (`routes/memory.py`, `mcp/memory.py`) passes it.
- In unified-bank mode (the shipped default), `HindsightProvider._allowed_namespaces` collapses `private:<agent>` → `[shared]`, and `_filter_visible` resolves the caller via `_caller_agent(None)` → `"unknown"`. Every doc actually tagged `visibility:private` + `agent:<agent>` is filtered OUT (owner mismatch), while every ordinary shared doc passes straight through unscoped. The per-agent chip ends up reporting the whole shared-bank page count (capped at 500) instead of that agent's own writes — the exact inverse of the module's documented contract.
- Fix: pass `client_id=agent_id`, matching every other read path.
- The existing test double's `list_items()` didn't declare a `client_id` parameter at all, so the fix would `TypeError` against it — updated the fake to accept and record `client_id`, and added a regression test asserting it's threaded through on every call.

Closes #1654

## Test plan
- [x] Confirmed red: reverted the source fix (kept the updated fake) → new `test_memory_stats_passes_client_id_for_unified_bank_scoping` fails (`client_id` was `None`).
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/memory tests/agents/test_agent_memory_stats_endpoint.py -q` — 197 passed

### DIFF
--- a/src/hal0/api/agents/memory_stats.py
+++ b/src/hal0/api/agents/memory_stats.py
@@ -170,6 +170,7 @@ async def get_agent_memory_stats(agent_id: str, request: Request) -> dict[str, A
             dataset=namespace,
             cursor=None,
             limit=_LIST_PAGE_LIMIT,
+            client_id=agent_id,
         )
     except Exception as exc:
         # Wrapper failures shouldn't 500 the sidebar chip — surface as
--- a/src/hal0/memory/hindsight_provider.py
+++ b/src/hal0/memory/hindsight_provider.py
@@ -635,6 +635,25 @@ class HindsightProvider(MemoryProvider):
         return projects, len(projects) < len([r for r in reqs if r])
 
     @staticmethod
+    def _wants_private_only(requested: str | list[str] | None) -> bool:
+        """True when every requested namespace is ``private:<id>`` (#1654).
+
+        ``_is_visible_to`` only enforces OWNERSHIP of private docs — a
+        non-private doc always passes it, unconditionally, regardless of
+        what was requested. So a read for ``private:<agent>`` alone (no
+        ``shared``) currently returns identically to a ``shared`` read: the
+        caller's own private docs AND every ordinary public doc in the bank.
+        Callers that explicitly ask for the private namespace ONLY (e.g. the
+        per-agent memory-stats chip) want private docs alone; this is the
+        signal :meth:`list_items` uses to apply that extra filter. A mixed
+        request (``["shared", "private:x"]``) is NOT private-only — it wants
+        the ordinary union — so this returns ``False`` unless every entry is
+        a private namespace.
+        """
+        reqs = HindsightProvider._requested_list(requested)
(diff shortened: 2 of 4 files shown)


thinmintdev:
Confirmed — good catch. Passing \`client_id\` alone fixes caller resolution but \`_is_visible_to\` only enforces private ownership; it passes every non-private doc through unconditionally regardless of what was requested, so a \`private:<agent>\`-only read still returned every public shared doc too.

Fixed in 39d2fb02: added \`HindsightProvider._wants_private_only()\` — true only when every requested namespace is \`private:<id>\` (a mixed request like \`["shared", "private:x"]\` still wants the ordinary union, so it stays false there). \`list_items()\` now applies that as an extra filter (unified mode only; legacy multi-bank mode already gets this for free via bank isolation). Added a red-first regression test (\`test_unified_private_only_list_excludes_public_shared_docs\`) that reproduces the leak pre-fix and passes post-fix.

Issue #1510, opened 2026-07-30
stacks: snapshot→apply round-trip duplicates every slot on an id-keyed box


thinmintdev:
Snapshot records the slot **display name**; every apply-side consumer resolves the slot **by file stem**. On the shipped id-keyed layout those are different strings, so applying a snapshot of the live box creates a duplicate of every slot that already exists.

## Evidence

`snapshot_live_stack` sets `StackSlotEntry.slot` to the TOML’s embedded `name`, deliberately, not the stem:

`src/hal0/stacks/portable.py:316-333`
```
        # id-aware (P3-runtime-db inc4): on an id-keyed box *slot_name* is the
        # digit stem, not the display name — prefer the TOML’s own embedded
        # ``name`` ... and only fall back to the stem
        "name": g("name", slot_name),
```
`src/hal0/stacks/portable.py:362` → `display_name = sc["name"]`, and `:385-386` → `StackSlotEntry(slot=display_name, ...)`.

Every consumer then goes back to the filename:

- `src/hal0/stacks/apply.py:138-140`
  ```
  def _slot_path(self, slot_name: str) -> Path:
      base = self._slots_dir or paths.slots_config_dir()
      return base / f"{slot_name}.toml"
  ```
- `src/hal0/api/routes/stacks.py:156-162`
  ```
  def _slot_toml_exists(slot: str) -> bool:
      return (paths.slots_config_dir() / f"{slot}.toml").exists()

  def _missing_slot_names(cfg: StackConfig) -> list[str]:
      """Stack slots (with a model) whose TOML doesn't exist yet."""
      return [e.slot for e in cfg.slots if e.model and not _slot_toml_exists(e.slot)]
  ```

`list_slots()` even documents that portable export is one of the callers still stuck on stems:

`src/hal0/config/loader.py:317-326`
```
    NAME-KEYED public wrapper — preserved verbatim for the callers (doctor
    bundle, portable export, slot-flags migration) that still address slots by
    their on-disk stem. ... on an id-keyed box a stem may be a digit id
```

## Reproduction on the live box (lxc105)

`/etc/hal0/slots/` is fully id-keyed — `1.toml` … `16.toml`, zero name-keyed files:

```
$ ssh hal0 head -8 /etc/hal0/slots/1.toml
name = "agent"
type = "llm"
device = "gpu-vulkan"
...
$ ssh hal0 head -8 /etc/hal0/slots/13.toml
name = "rerank"
type = "reranking"
...
```

[...]

PR #1681, opened 2026-08-08
fix(slots): resolve the capability lane bilingually on an id-keyed box

### LINKED ISSUE #1643: slots: capability lane is name-keyed only — on an id-keyed box, disable leaves slot bound and routable, enable 5xxs
## What's broken

`SlotConfigStore._slot_path()` (`src/hal0/slot_config/__init__.py:471`) returns `base / f"{slot_name}.toml"`, and `CapabilityOrchestrator._ensure_slot_exists` (`orchestrator.py:716`) / `_ensure_slot_exists_npu` (`orchestrator.py:879`) probe that same literal name-keyed path. After the supported `hal0 slot migrate-id-keying` migration, the slot lives at `<id>.toml` with the display name in the body, so both probes miss it. Two consequences on the same box:

**(a) Disable never disables.** `_reconciled_slot` gets `raw_before is None` and returns `None`, so `apply_and_commit` writes `capabilities.toml` only — `[model].default` is never cleared, and `loaded_slot_from_config` keeps resolving the slot as active/routable, even though the dashboard shows the capability off.

**(b) Enable / model-change then 5xxs.** `SlotManager.create(slot_name, ...)`'s bilingual clobber guard uses `self._config_file_for(slot_name)` (`manager.py:616`), which *does* resolve `<id>.toml`, sees it exists, and raises `SlotConfigError("slot 'tts' already exists; use update to modify it")`. `_ensure_slot_exists` wraps that in `CapabilityApplyFailed` and the whole capability apply errors out.

This is the same id-keyed/name-keyed split #1633 fixed for `_iter_peer_configs` — the capability lane was missed. On-demand slot creation (#1631's headline mechanism) now depends on it.

## Repro

Repro'd against the real production classes (no mocks):

**(a)** `SlotConfigStore(capabilities_path=caps, slots_dir=slots)` over a slots dir holding only `12.toml` (`name = "embed"`, `[model].default = "nomic-embed:v1"`); `store.commit(store.apply(SlotSelection(slot='embed', child='embed', slot_name='embed', selection=CapabilitySelection(enabled=False, ...))))` → `12.toml` still has `model.default == 'nomic-embed:v1'` (expected `""`), and no `embed.toml` sibling is created.

**(b)** create slot `tts` → `migrate_slot_id_keying(...)` → new `SlotManager` → `paths.slots_config_dir()/'tts.toml'` does not exist (what the orchestrator probes); re-running `sm.create('tts', ...)` raises `SlotConfigError: slot 'tts' already exists; use update to modify it`.

## Severity

P1 — capability toggles silently no-op or 5xx for any operator running the supported id-keying migration.

## Found by

hal0 rc.2 pre-release review (functional review, "slots" section), verified with a live reproduction against the production `SlotConfigStore`/`SlotManager` classes.

### LINKED ISSUE #1664: slots: reconcile_trio_slots probes shadow TOMLs name-keyed — on an id-keyed box it retries create() every boot and skips normalization
## What's broken

`canon_path = slots_dir / f"{canon}.toml"` (`src/hal0/slots/npu/trio.py:139`) is a literal name-keyed probe. On an id-keyed box (post `hal0 slot migrate-id-keying`) the shadows live at `<id>.toml`, so `canon_path.exists()` is false on every boot: step 2 of the reconciler (device/profile/served_by/port/type normalization — the whole point of the routine) is skipped, and step 3 falls through to `mgr.create(canon, ...)`, which hits `SlotManager.create`'s bilingual clobber guard and raises `SlotConfigError("... already exists ...")`.

The per-shadow `except` swallows this into `log.warning("slot.reconcile_trio_shadow_failed")`, so it degrades quietly rather than crashing — but a shadow that has drifted (e.g. its `port` no longer matches a re-ported anchor, or a missing `type` gate) is never repaired, and `_is_npu_trio_request` gates on that `type`. Same root cause as the companion P1 capability-lane issue (#1643); listed separately because it's a distinct call site with a distinct symptom.

## Repro

On a box that ran `hal0 slot migrate-id-keying` with an FLM anchor: restart `hal0-api` and read the journal — `slot.reconcile_trio_shadow_failed slot=flm-stt error=slot 'flm-stt' already exists; use update to modify it` on every start, and `slot.trio_shadow_normalized` never appears even after hand-editing `<id>.toml` to drop `type`.

## Severity

P3 — degrades quietly (warning log only) but silently disables shadow-slot normalization/repair on every id-keyed box.

## Found by

hal0 rc.2 pre-release review (functional review, "slots" section), verified by tracing `canon_path` against `SlotManager`'s bilingual `_config_file_for`/`_peek_key` resolver.

thinmintdev:
## What's broken

`hal0 slot migrate-id-keying` is a supported migration: it moves a slot's TOML to `slots/<id>.toml` and keeps the display name in the body. Three capability-lane probes still addressed the file as the literal `slots/<name>.toml`, so on a migrated box the whole capability enable/disable lane broke:

**(a) Disable never disabled (#1643).** `SlotConfigStore._slot_path()` returned `<name>.toml`, so `apply()` read `raw_before is None`, `_reconciled_slot` returned `None`, and `commit` wrote `capabilities.toml` only. `[model].default` was never cleared, so `loaded_slot_from_config` kept resolving the slot as active and **routable** while the dashboard rendered the capability as off.

**(b) Enable / model change 503'd (#1643).** `CapabilityOrchestrator._ensure_slot_exists` / `_ensure_slot_exists_npu` probed the same literal path, missed, and fell through to `SlotManager.create` — whose own (already bilingual) clobber guard resolved `<id>.toml`, saw it, and raised `slot 'tts' already exists; use update to modify it`. The orchestrator wrapped that in `CapabilityApplyFailed` and the apply returned `capability.apply_failed`.

**(c) Trio shadows never normalized (#1664).** `reconcile_trio_slots` probed `slots_dir / f"{canon}.toml"`, so step 2 (device/profile/served_by/port/type normalization — the point of the routine) was skipped on every boot and step 3 re-attempted the same rejected `create`, swallowed into a `slot.reconcile_trio_shadow_failed` warning. Filed separately because it's a distinct call site, but it is the **same root-cause helper**, so it's fixed here.

## The fix

All three resolve through the one bilingual seam the stacks lane already uses — `hal0.slots.layout.resolve_slot_stem` (#1510). No parallel lookup was added.

The trio reconciler's legacy→canon rename additionally delegates to the canonical `SlotManager.rename` when the shadow is id-keyed, so the identity row moves with the label. A bare file move would have stranded the row under the legacy name and left the shadow unresolvable by its new one — a rename is a relabel of a stable id, not a re-key, so the stem stays put.

**Name-keyed boxes are byte-identical.** `resolve_slot_stem` is stem-first: `<name>.toml` still resolves with a single `exists()` and never reads a TOML.

## Verification

New `tests/slots/test_capability_lane_id_keyed.py` (9 tests, red before / green after). It builds a genuinely id-keyed `HAL0_HOME` — driving the real `migrate_slot_id_keying` where a lifecycle is involved — and exercises the production `SlotConfigStore` / `CapabilityOrchestrator` / `SlotManager` classes, no mocks:

- disable → `<id>.toml` `[model].default` cleared, no `<name>.toml` sibling fabricated
- disable → `loaded_slot_from_config` returns `None` (unbound **and** unroutable)
- enable → rebinds the model in `<id>.toml`, no 5xx
- `_ensure_slot_exists` / `_ensure_slot_exists_npu` are no-ops after migration (pre-fix: `CapabilityApplyFailed`)
- a genuinely missing slot is still created (the resolver must not turn creation into a silent no-op)
- trio reconcile normalizes a drifted id-keyed shadow in place and still creates the missing one
- trio legacy rename keeps the id-keyed stem, preserves the operator's model, and moves the identity row (`get_by_name("stt-npu") is None`, `status("flm-stt")` resolves)
- name-keyed layout regression guard

Suites run locally, all green:

```
tests/slots tests/slot_config tests/capabilities tests/stacks tests/dispatcher → 1161 passed
tests/slots/test_capability_lane_id_keyed.py                                   → 9 passed
```

`ruff format` / `ruff check src tests` clean; `mypy` on the touched files reports only the 2 pre-existing `orchestrator.py` errors present on `main`.

Closes #1643
Closes #1664

### DIFF
--- a/src/hal0/capabilities/orchestrator.py
+++ b/src/hal0/capabilities/orchestrator.py
@@ -56,6 +56,7 @@ from hal0.model_meta import canonical_device
 from hal0.profiles import ProfileCatalog, ResolvedProfile
 from hal0.registry.store import ModelRegistry
 from hal0.slot_config import SlotConfigStore, SlotSelection
+from hal0.slots.layout import resolve_slot_stem
 from hal0.slots.manager import SlotManager
 
 log = logging.getLogger(__name__)
@@ -712,9 +713,14 @@ class CapabilityOrchestrator:
         slot, so the SlotManager would raise SlotNotFound on the first
         load. We synthesise a minimal config from the selection and let
         ``SlotManager.create()`` do the persist + state initialisation.
+
+        The "does it already exist?" probe goes through the bilingual layout
+        seam (#1643): a literal ``<name>.toml`` probe missed every slot on an
+        id-keyed box, so this fell through to ``SlotManager.create``, whose
+        own (bilingual) clobber guard then raised ``slot 'tts' already
+        exists`` and turned every enable / model change into a 503.
         """
-        cfg_path = paths.slots_config_dir() / f"{slot_name}.toml"
-        if cfg_path.exists():
+        if resolve_slot_stem(paths.slots_config_dir(), slot_name) is not None:
             return
 
         port = self._next_free_slot_port()
@@ -874,10 +880,11 @@ class CapabilityOrchestrator:
         ``device=npu``, ``provider=flm``, and always stamps the trio
         ``type`` so dispatch gating activates. No-op when the slot TOML
         already exists (its existing fields, including ``type``, survive —
-        ``update_config`` is a shallow top-level merge).
+        ``update_config`` is a shallow top-level merge) — resolved through the
+        bilingual layout seam so an id-keyed ``<id>.toml`` counts as existing
+        (#1643).
         """
-        cfg_path = paths.slots_config_dir() / f"{slot_name}.toml"
-        if cfg_path.exists():
+        if resolve_slot_stem(paths.slots_config_dir(), slot_name) is not None:
             return
         port = self._next_free_slot_port()
         cfg_dict: dict[str, Any] = {
--- a/src/hal0/slot_config/__init__.py
+++ b/src/hal0/slot_config/__init__.py
@@ -468,8 +468,26 @@ class SlotConfigStore:
         return self._capabilities_path or capabilities_toml_path()
 
     def _slot_path(self, slot_name: str) -> Path:
+        """The on-disk config path for a slot addressed by DISPLAY name (#1643).
+
+        A capability selection — like a stack entry (#1510) — addresses its
+        slot by the portable display name, but the stem is this box's storage
+        detail: after ``hal0 slot migrate-id-keying`` the same slot lives at
+        ``<id>.toml`` with the name in the body. Resolving through the ONE
+        bilingual seam (:func:`hal0.slots.layout.resolve_slot_stem`) is what
+        makes a capability **disable** actually clear ``[model].default``
+        instead of reading ``None`` for a slot that is plainly on disk (and
+        leaving it bound and routable).
+
+        Falls back to the name itself when nothing resolves — a name-keyed box
+        never reads a TOML here (the stem-first ``exists()`` wins), and a slot
+        that genuinely does not exist yet keeps the name-keyed path the
+        creation lane expects.
+        """
+        from hal0.slots.layout import resolve_slot_stem
+
         base = self._slots_dir or paths.slots_config_dir()
-        return base / f"{slot_name}.toml"
+        return base / f"{resolve_slot_stem(base, slot_name) or slot_name}.toml"
 
     # ── apply (compute-only) ─────────────────────────────────────────────────
 
--- a/src/hal0/slots/npu/trio.py
+++ b/src/hal0/slots/npu/trio.py
@@ -55,6 +55,50 @@ def _log_renamed(old: str, canon: str) -> None:
     log.info("slot.trio_shadow_renamed", extra={"from": old, "to": canon})
 
 
+def _slot_table(raw: dict[str, Any]) -> dict[str, Any]:
+    """The table a slot's SCALAR fields actually live in.
+
+    Two on-disk shapes are supported: flat (scalars at the root, what the
+    runtime writes) and nested (scalars under ``[slot]``, the haloai shape the
+    id-keying migration preserves verbatim). ``_flatten_slot_toml`` treats
+    ``[slot]`` as authoritative and drops root scalars into ``extra`` when it
+    exists, so a writer that normalizes at the root of a nested file produces a
+    change the runtime never sees — and, worse, one that looks converged on the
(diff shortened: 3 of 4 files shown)


thinmintdev:
### Codex triage

**1. `trio.py` — "Handle active legacy shadows before calling rename" — ✅ fixed (1c6dfe69).**
Confirmed: `load` short-circuits a trio shadow straight to READY (`manager.py:1530-1538`), so on a live box it persists exactly the state `SlotManager.rename`'s offline guard rejects (`manager.py:1010-1015`) — the reconcile would have failed into the per-shadow `except` on every boot. Took the suggested route: a shadow-specific `_relabel_shadow_in_place` that skips the guard. The guard exists because a *name-keyed* slot's systemd unit still carries its name; an id-keyed shadow moves no file and touches no unit, so the label lives only in the TOML body and the identity row, and the helper moves both. Regression test: `test_trio_reconcile_relabels_a_non_offline_id_keyed_shadow` loads the shadow first, asserts it is not OFFLINE, then asserts the relabel and the identity-row move still land.

**2. `trio.py` — "Normalize fields in the authoritative slot table" — ✅ fixed (1c6dfe69).**
Confirmed against `_flatten_slot_toml` (`loader.py:388-401`): with `[slot]` present, only its keys are hoisted and root scalars go to `extra`. Both the compare and the write now resolve the file's authoritative scalar table first, so the repair converges instead of duplicating. Regression test: `test_trio_reconcile_normalizes_inside_the_nested_slot_table` asserts through the real `load_slot_config`, which is what actually catches a root-scalar write.

**3. `slot_config/__init__.py` — "Update fields inside legacy slot tables" — ✅ valid, ➡️ filed as #1685, out of scope here.**
Same root-vs-nested projection bug, but it lives in the shared `merge_slot_config` primitive, not in this PR's change: `SlotManager.update_config` and `hal0.stacks.apply` hit it identically on a name-keyed box today, so it predates #1643/#1664 and is not a regression from this diff. Note `[model]` is a sibling *table* in both shapes, so the P1 disable path this PR fixes is unaffected — the silent field is the scalar set (`device`/`provider`/`profile`). Fixing it properly means making the shared primitive shape-aware with a round-trip test, which is a wider blast radius (three writers) than an rc.2 bug fix should carry. #1685 has the repro and the suggested fix, and notes that this PR's local `_slot_table` accessor should collapse into the shared helper when it lands.

PR #1686, opened 2026-08-08
fix(updater): strip None before TOML-dumping migrated config

### LINKED ISSUE #1652: settings: config-migration writer dumps None into tomli_w — TypeError once a real (non-v2) migration runs
## What's broken

`_maybe_run_config_migrations` (`src/hal0/updater/updater.py:1131`) does `raw = cfg.model_dump(mode="python")` → `run_migrations` → `write_toml_atomic(toml_path, new_raw)`. Unlike `save_hal0_config` (which uses `exclude_none=True` precisely because None has no TOML representation and `tomli_w` raises `TypeError` on it), this path keeps `None` values. `SecurityConfig.require_auth` and `trust_forwarded_for` both default to `None`, so **any** config that reaches this writer raises `TypeError: Object of type 'NoneType' is not TOML serializable`.

`TypeError` is not a `Hal0Error`, so it escapes `Updater.commit()`'s `except Hal0Error` handler (no staged-tree cleanup, the commit job just dies) and is only contained by `install.sh`'s `hal0_migration_step` warn-wrapper.

Today this is masked by arithmetic, not by design: the only registered migration is v2, and the profile-catalog gate either stamps v2 first (source==target → noop) or caps `ceiling=1` (target==source → noop), so the write is never actually reached. The moment a v3 migration lands — or any caller passes a `min_data_version` above the on-disk version with no ceiling — this becomes a hard failure of every `hal0 update`.

**Fix:** `model_dump(mode="python", exclude_none=True)`, or route through `save_hal0_config`.

## Repro

Direct proof of the writer defect today:
```python
write_toml_atomic(p, Hal0Config().model_dump(mode='python'))
# TypeError: Object of type 'NoneType' is not TOML serializable
```

## Severity

P2 — currently masked by version-arithmetic coincidence; will hard-fail every `hal0 update` the moment a v3 config migration is registered.

## Found by

hal0 rc.2 pre-release review (functional review, "settings" section), verified against the real `write_toml_atomic`/`Hal0Config` writer.

thinmintdev:
## Summary
- `_maybe_run_config_migrations` (`src/hal0/updater/updater.py`) dumped `cfg.model_dump(mode="python")` straight into `write_toml_atomic`, unlike `save_hal0_config`, which uses `exclude_none=True` because `None` has no TOML representation and `tomli_w` raises `TypeError` on it.
- `SecurityConfig.require_auth` and `trust_forwarded_for` both default to `None`, so any config reaching this writer with a real migration pending raised `TypeError: Object of type 'NoneType' is not TOML serializable` — a plain `TypeError`, not a `Hal0Error`, so it escapes `Updater.commit()`'s exception handler entirely (no staged-tree cleanup).
- Today this is masked by arithmetic, not design: the only registered migration is v2, and the profile-catalog gate keeps `source == target` or `ceiling == source`, so this writer is never actually reached. The moment a v3 migration lands, this becomes a hard failure of every `hal0 update`.

## Fix
Add `exclude_none=True` to match `save_hal0_config`. Pydantic re-supplies the default on load, so dropping `None` on write is safe for any field whose default is `None`.

## Test plan
- [x] Red-first: `test_a_pending_migration_does_not_typeerror_on_none_fields` simulates a v3 migration by monkeypatching the migrations registry (no real v3 exists yet) — fails with the exact `TypeError` from the issue on main, passes after the fix.
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/config tests/updater tests/api/test_settings_apply.py -x -q` — 1028 passed, 9 skipped.

Closes #1652

### DIFF
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -1128,7 +1128,16 @@ def _maybe_run_config_migrations(
         )
         return (source, source)
 
-    raw = cfg.model_dump(mode="python")
+    # ``exclude_none=True`` matches ``save_hal0_config``: None has no TOML
+    # representation and ``tomli_w`` raises TypeError on it (#1652).
+    # SecurityConfig.require_auth / trust_forwarded_for both default to
+    # None, so every config hit this before exclude_none was added here —
+    # masked only by version arithmetic (the only migration is v2, and the
+    # profile-catalog gate keeps source==target or ceiling==source, so this
+    # writer was never actually reached). Pydantic re-supplies the default
+    # on load, so dropping None on write is safe for any field whose
+    # default is None, same as save_hal0_config.
+    raw = cfg.model_dump(mode="python", exclude_none=True)
     new_raw, new_version = run_migrations(raw, target_version=target)
     write_toml_atomic(toml_path, new_raw)
     log.info(
--- a/tests/updater/test_migration_writer_none.py
+++ b/tests/updater/test_migration_writer_none.py
@@ -1,0 +1,66 @@
+"""#1652: ``_maybe_run_config_migrations`` dumps ``None`` into ``tomli_w``.
+
+``_maybe_run_config_migrations`` (``src/hal0/updater/updater.py``) does
+``cfg.model_dump(mode="python")`` -- unlike ``save_hal0_config``, which uses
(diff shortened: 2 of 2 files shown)


thinmintdev:
@codex review

PR #1692, opened 2026-08-08
fix(updater): follow redirects on release-manifest fetch

### LINKED ISSUE #1688: updater: release-manifest fetch rejects GitHub release asset URLs (HTTP 302 not followed)
The manifest fetch in src/hal0/updater/updater.py (~line 518) creates its httpx.AsyncClient without follow_redirects, so any HAL0_RELEASES_URL pointing at a GitHub release asset — the documented interim hosting until releases.hal0.dev exists (scripts/release-prototype/RELEASE_PIPELINE_NOTES.md) — fails with 'release manifest fetch returned HTTP 302'. The sibling tarball download (~line 802) already uses follow_redirects=True.

Hit live on 2026-08-08 while running the v1.0.0-rc.1 → rc.2 update test on the CT150 box:

    Error: PUT /api/updates/channel → HTTP 500: could not fetch release manifest:
    release manifest fetch returned HTTP 302 from
    https://github.com/Hal0ai/hal0/releases/download/v1.0.0-rc.2/preview.json?channel=preview

Also worth checking while in there: the channel query parameter appended to a custom HAL0_RELEASES_URL (the '?channel=preview' suffix above) — GitHub asset URLs tolerate it, but verify it is intentional for custom URLs.

Security note for the fix: following redirects for the manifest bytes is safe because the manifest is verified against its cosign sibling bundle before any field is trusted; the same reasoning already applies to the redirect-following tarball client. Please confirm the sibling-bundle fetch path has the same redirect behavior fixed.

Found by the rc.2 deployment test during the pre-release review.

thinmintdev:
## Summary
- `_fetch_release_manifest_bytes` (src/hal0/updater/updater.py) built its `httpx.AsyncClient` without `follow_redirects=True`, so a `HAL0_RELEASES_URL` pointing at a GitHub release asset — the documented interim hosting until releases.hal0.dev exists — failed with `release manifest fetch returned HTTP 302`. GitHub 302s asset downloads to `objects.githubusercontent.com`.
- The sibling tarball/bundle client (`_download`, used for both the tarball and the manifest's sibling `.bundle`) already sets `follow_redirects=True`; the manifest client is now aligned with it.
- Checked `src/hal0/api/routes/updater.py`: it delegates to `releases_url`/`fetch_release_manifest` from the updater module rather than duplicating an httpx client, so no separate fix was needed there.
- Checked the `?channel=` query param appended to custom `HAL0_RELEASES_URL` values (`releases_url()`): confirmed intentional and already covered by `test_releases_url_appends_channel_for_http_override`.
- Confirmed the sibling `.bundle` fetch still verifies the *exact* fetched manifest bytes after following redirects (the bytes written to disk for cosign verification are the ones `_fetch_release_manifest_bytes` returned, i.e. post-redirect content) — added a regression test for this via `httpx.MockTransport`.

## Test plan
- [x] New tests in `tests/updater/test_manifest_fetch_redirects.py`, red-verified against pre-fix code (reproduces the exact issue error, then goes green after the fix)
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/updater tests/release tests/api/test_updater_routes.py -q` → 387 passed
- [x] `ruff check` / `ruff format --check` clean on changed files

Closes #1688

### DIFF
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -515,7 +515,7 @@ async def _fetch_release_manifest_bytes(channel: str = "stable") -> bytes:
     import httpx
 
     try:
-        async with httpx.AsyncClient(timeout=10.0) as client:
+        async with httpx.AsyncClient(timeout=10.0, follow_redirects=True) as client:
             resp = await client.get(url)
     except httpx.HTTPError as exc:
         raise OSError(f"release manifest fetch failed for {url}: {exc}") from exc
--- a/tests/updater/test_manifest_fetch_redirects.py
+++ b/tests/updater/test_manifest_fetch_redirects.py
@@ -1,0 +1,144 @@
+"""Regression tests for GitHub-#1688: manifest/bundle fetch must follow 302s.
+
+``HAL0_RELEASES_URL`` is documented as tolerating a GitHub release-asset URL
+(the interim hosting until releases.hal0.dev exists — see
+scripts/release-prototype/RELEASE_PIPELINE_NOTES.md). GitHub redirects asset
+downloads to ``objects.githubusercontent.com`` with an HTTP 302, so every
+client that fetches manifest or sibling-bundle bytes over http(s) must follow
+redirects. These tests use an ``httpx.MockTransport`` so no real network call
+is made.
+"""
+
+from __future__ import annotations
+
+import functools
+import json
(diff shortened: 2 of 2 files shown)


thinmintdev:
@codex review

PR #1693, opened 2026-08-08
fix(installer): honour .seed-tombstones in install.sh seed loop

### LINKED ISSUE #1650: installer: install.sh seed loop ignores .seed-tombstones — a deliberately deleted static seed resurrects on repair/upgrade re-run
## What's broken

#1631's contract is that deleting a static seed slot writes `.seed-tombstones` (`hal0.install.static_seeds.add_seed_tombstone`, called from `SlotManager.delete`), and "the boot-time seeding pass honours the deletion instead of resurrecting the slot." Only the Python boot path honours it: `seed_static_slots` checks `name in tombstoned` (`static_seeds.py:95`).

The bash seed loop in `installer/install.sh:1712` (`for seed_slot in flm tts rerank utility img agent brain qwen3tts coder embed`) gates solely on `slot_name_exists()` (filename + `name =` grep over `*.toml`) and never reads `.seed-tombstones`, so it re-copies the seed. `install.sh` is a documented repair/upgrade-in-place path (`updater.py:1152`, `privileged.py:110` — "re-run the installer (`sudo bash install.sh`)"), so a routine repair/upgrade silently undoes a deliberate slot deletion — the headline fix of #1631, for the one code path an operator is actually told to re-run.

## Repro

`hal0 slot delete coder` (or Delete from the dashboard) → `/etc/hal0/slots/.seed-tombstones` gains `coder`, `coder.toml` is gone, `hal0-api` restarts leave it gone. Then `sudo bash installer/install.sh` on the same box → `slot_name_exists coder` is false → `cp installer/etc-hal0/slots/coder.toml /etc/hal0/slots/coder.toml`, and the deleted slot is back as a card on the dashboard.

## Severity

P2 — undoes the intended, tested behavior of a shipped delete flow via the documented repair/upgrade path.

## Found by

hal0 rc.2 pre-release review — flagged independently by both the "slots" functional-review section and the Codex-comment sweep on PR #1631 (same root cause, same repro), consolidated into one issue here.

thinmintdev:
## Summary
- `install.sh`'s curated-slot seed loop gated only on `slot_name_exists()` and never read `.seed-tombstones`, so re-running the installer (the documented repair/upgrade-in-place path — `updater.py`, `privileged.py`) resurrected a static seed slot the operator had deliberately deleted via `hal0 slot delete`.
- Add `slot_seed_tombstoned()`, mirroring `hal0.install.static_seeds.read_seed_tombstones` (same file, `${ETC_DIR}/slots/.seed-tombstones`), and check it before the existing `slot_name_exists()` guard in the seed loop.

## Test plan
- [x] New `tests/installer/test_seed_tombstone_guard.py` extracts the real `slot_seed_tombstoned` function from `install.sh` (same technique as `test_seed_slot_guard_id_keyed.py`) and drives it against fixture trees: tombstoned name recognised, non-tombstoned live, missing file/dir don't error, substring names don't match, call site gates before `cp`, filename matches the Python side.
- [x] `uvx --from shellcheck-py shellcheck installer/install.sh` — no new warnings (pre-existing SC1091/SC2034 only).
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/installer tests/install tests/slots/test_static_seeds.py -q` — 259 passed.

Closes #1650

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1707,12 +1707,31 @@ slot_name_exists() {
     return 1
 }
 
+# slot_seed_tombstoned — mirrors hal0.install.static_seeds.read_seed_tombstones:
+# `hal0 slot delete <name>` on a STATIC_SEED_SLOTS member writes its name into
+# ${ETC_DIR}/slots/.seed-tombstones (one name per line, SlotManager.delete →
+# add_seed_tombstone). #1631's contract is that this file records a
+# deliberate operator DELETION and seeding must honour it instead of
+# resurrecting the slot. Only the Python boot-time closer
+# (hal0.install.static_seeds.seed_static_slots) checked it — this bash loop
+# gated on slot_name_exists alone, so `sudo bash install.sh` (the documented
+# repair/upgrade-in-place path, updater.py / privileged.py) silently undid a
+# deliberate deletion on every re-run. #1650.
+slot_seed_tombstoned() {
+    local want="$1" f="${ETC_DIR}/slots/.seed-tombstones"
+    [[ -f "${f}" ]] || return 1
+    grep -qxF "${want}" "${f}"
+}
+
 SEEDED_NEW_SLOTS=()
 SEEDED_EXISTING_SLOTS=0
 for seed_slot in flm tts rerank utility img agent brain qwen3tts coder embed; do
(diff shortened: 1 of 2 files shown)


thinmintdev:
@codex review

PR #1694, opened 2026-08-08
fix(models): lazy_quant checks ROCmFPX resolver before generic regex

### LINKED ISSUE #1659: models: lazy_quant skips the ROCmFPX family resolver — pulled/scanned rows show a raw preset instead of the family label
## What's broken

`lazy_quant` (`src/hal0/services/models_service.py:236`) resolves only through `quant_from_filename` (the generic `(?<![a-z0-9])(i?q\d[_a-z0-9]*|bf16|f16|f32)(?![a-z0-9])` regex), never `quant_from_rocmfpx_filename`. `detect()` deliberately checks the ROCmFPX resolver first (`registry/detect.py:377-381`) precisely because these repacks carry a custom `general.file_type` the LLAMA_FTYPE table can't map.

This matters more than the "lazy backfill for pre-existing registries" docstring implies, because `Model.quant` is almost never stored: `registry/pull.py` never sets it on the `Model` (only into the `meta.json` sidecar, and even that with `quant=None`), and `registry/discover.register_candidate` omits it too. `add_from_path` is the only writer. So effectively every pulled and every auto-scanned row falls through to `lazy_quant`.

The generic regex is greedy across `[_a-z0-9]*`, so it returns the whole preset token rather than the family: `Q8_0_ROCMFPX` instead of `ROCmFP8`, `Q4_0_ROCMFP4_FAST` instead of `ROCmFP4`. hal0's own brain models are ROCmFPX repacks (`install/brain_model.py:57`, `BRAIN_HF_REPO = "Hal0ai/hal0-brain-sft-ROCmFPX-GGUF"`).

## Repro

Pull `Hal0ai/hal0-brain-sft-ROCmFPX-GGUF` file `hal0-brain-sft-Q8_0_ROCMFPX.gguf`. The row has `quant=None`, so `/api/models` backfills via `lazy_quant` → `"Q8_0_ROCMFPX"`, rendered verbatim in the Models list quant chip and detail pane. Register the identical file via `POST /api/models/add-from-path` instead and `detect()` stores `"ROCmFP8"` — same bytes, two different labels depending on how the model got registered.

## Severity

P3 — cosmetic/labeling only, but affects hal0's own bundled brain model and any pulled ROCmFPX repack.

## Found by

hal0 rc.2 pre-release review (functional review, "models" section).

thinmintdev:
## Summary
- `lazy_quant()` (`src/hal0/services/models_service.py`) now tries `quant_from_rocmfpx_filename()` before falling back to the generic `quant_from_filename()` regex — same resolver order `detect()` already uses on the eager path.
- Fixes the label mismatch where a pulled/auto-scanned ROCmFPX repack (quant backfilled via `lazy_quant`) showed the raw preset token (`Q8_0_ROCMFPX`) while the identical file registered via `add-from-path` (which goes through `detect()`) showed the family label (`ROCmFP8`).

## Why
`Model.quant` is almost never stored at write time (`registry/pull.py` and `registry/discover.register_candidate` both omit it), so nearly every pulled/scanned row falls through to `lazy_quant` on the `/api/models` hot path. hal0's own bundled brain model (`Hal0ai/hal0-brain-sft-ROCmFPX-GGUF`) is a ROCmFPX repack, so this was hitting the common path, not an edge case.

Closes #1659

## Verification
- Added `test_lazy_quant_resolves_rocmfpx_family_from_path`, `test_lazy_quant_resolves_rocmfpx_family_from_hf_filename`, `test_lazy_quant_still_resolves_generic_quant` to `tests/api/test_models_routes.py`.
- Confirmed red first: reverted the fix, new ROCmFPX tests failed (`Q8_0_ROCMFPX` != `ROCmFP8`, `Q4_0_ROCMFP4_FAST` != `ROCmFP4`), generic-quant test still passed.
- Restored fix: `tests/api/test_models_routes.py` + `tests/registry/test_detect.py` — 53 passed.

### DIFF
--- a/src/hal0/services/models_service.py
+++ b/src/hal0/services/models_service.py
@@ -237,16 +237,22 @@ def lazy_quant(dumped: dict[str, Any]) -> str | None:
     """Best-effort quant label for a serialised row missing ``quant``.
 
     Filename-token only (cheap enough for the list endpoint's 30s poll):
-    the on-disk basename first, then the HF variant filename. Rows whose
-    quant is only knowable from the GGUF header (hash-named blobs) get it
-    at registration via detect() instead.
+    the on-disk basename first, then the HF variant filename. Checks the
+    ROCmFPX family resolver before the generic quant regex — same order
+    as the eager path in ``detect()`` — so ROCmFPX repacks (hal0's own
+    brain model included) resolve to their family label (``ROCmFP8``)
+    instead of the raw preset token the generic regex would greedily
+    match (``Q8_0_ROCMFPX``). Rows whose quant is only knowable from the
+    GGUF header (hash-named blobs) get it at registration via detect()
+    instead.
     """
-    from hal0.registry.detect import quant_from_filename
+    from hal0.registry.detect import quant_from_filename, quant_from_rocmfpx_filename
 
     for key in ("path", "hf_filename"):
         raw = dumped.get(key)
         if isinstance(raw, str) and raw:
-            quant = quant_from_filename(Path(raw).name)
+            name = Path(raw).name
+            quant = quant_from_rocmfpx_filename(name) or quant_from_filename(name)
             if quant:
                 return quant
     return None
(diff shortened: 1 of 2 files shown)


thinmintdev:
@codex review

PR #1697, opened 2026-08-08
fix(memory): resume list_items cursor at raw offset, not visible count

### LINKED ISSUE #1667: memory: list_items cursor advances by raw offset corrected with visible-item count — pages silently skip rows under ACL filtering
## What's broken

`consumed = bank_offset - max(0, len(items) - limit)` (`src/hal0/memory/hindsight_provider.py:950`) subtracts the number of *visible* items that overflowed the page from `bank_offset`, which counts *raw* rows consumed. The two only coincide when `_filter_visible` drops nothing.

In unified-bank mode, the shared bank holds other agents' `visibility:private` docs and every project's `project:<id>` docs, so a page routinely over-fetches (`fetch = max(limit*2, 50)`) and filters heavily; the correction then under-rewinds and the next cursor starts past raw rows that still held unreturned visible items — those items become unreachable through pagination. The comment directly above the line claims the opposite invariant ("The offset counts raw rows consumed, not visible ones, so the walk stays stable regardless of what the ACL filtered").

## Repro

Unified mode, `limit=50`. One raw page of 100 rows yields 60 visible (40 belong to another agent / another project). `items=60`, `bank_offset=100`, `consumed = 100 - 10 = 90`. The 10 visible items not returned lived among raw rows 0–100; resuming at raw offset 90 skips whichever of them sat below row 90. `GET /api/memory/list?limit=50` then `?cursor=shared@90` never returns them. Confirmed with a fake-bank repro that filters mid-page — page 2 comes back empty and 10 items are permanently unreachable via pagination.

## Severity

P3 — silent data loss in pagination under ACL filtering (unified-bank mode only); items aren't deleted, just unreachable through the list/search cursor walk.

## Found by

hal0 rc.2 pre-release review (functional review, "memory" section), verified with a constructed repro reproducing the exact cursor math (consumed=90).

thinmintdev:
## Summary
- `list_items`'s cursor-rewind math subtracted the number of *visible* overflow items from the *raw* bank offset (`consumed = bank_offset - max(0, len(items) - limit)`), which only equals the correct raw rewind when the ACL filter drops nothing from the tail of the last fetched chunk.
- In unified-bank mode the shared bank interleaves other agents' `visibility:private` docs and out-of-scope `project:<id>` docs, so a real page filters unevenly — the old correction under-rewound and the next cursor started past raw rows that still held unreturned visible items, making them permanently unreachable via pagination.
- Fix: track each visible item's raw offset directly as it's collected (the raw row index immediately after the row that produced it), and resume the next page at the raw offset right after the last item actually kept from that bank.

## Test plan
- [x] New repro test (`test_list_items_cursor_survives_acl_filtering_mid_page`): 100 raw rows, 40 ACL-withheld unevenly, walks the cursor to completion and asserts every one of the 60 visible items is returned exactly once. Fails on `main` (loses `f84`, `f87`, `f88`, `f89`), passes after the fix.
- [x] Negative control (`test_list_items_negative_control_still_withholds_foreign_private_docs`): asserts another agent's private docs never surface on any page across the same paginated walk — the fail-closed ACL behavior is unchanged by the cursor-math fix.
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/memory tests/security -x -q` → 364 passed.
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest -q -k "memory or hindsight"` → 751 passed.

Closes #1667

### DIFF
--- a/src/hal0/memory/hindsight_provider.py
+++ b/src/hal0/memory/hindsight_provider.py
@@ -932,6 +932,17 @@ class HindsightProvider(MemoryProvider):
         * Over-fetch per bank so a heavily-filtered page can still fill, and
           only advertise a ``next_cursor`` when the bank actually reported more
           rows behind the offset we consumed.
+
+        Cursor math (#1667): the resume offset for a bank cut mid-page is
+        derived from the RAW row that produced the last item actually
+        returned, not from ``raw_offset - overflow_count``. The two only
+        agree when the ACL filter drops nothing from the tail of the last
+        fetched chunk — in unified-bank mode the shared bank interleaves
+        other agents' private docs and out-of-scope project docs, so a page
+        routinely filters unevenly, and the visible-count correction
+        under-rewinds and strands unreturned visible rows behind the next
+        cursor forever. Tracking each visible item's raw offset directly
+        keeps the walk exact regardless of what the ACL withheld.
         """
         banks = [namespace_to_bank(ns) for ns in self._allowed_namespaces(dataset, client_id)]
         start_bank, offset = _decode_cursor(cursor, banks)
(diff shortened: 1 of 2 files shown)


thinmintdev:
@codex review

PR #1698, opened 2026-08-08
fix(slots): tombstone the vacated name when renaming a static seed slot

### LINKED ISSUE #1651: slots: renaming a static seed slot causes the next boot to re-seed a duplicate under the old name
## What's broken

`SlotManager.rename` (`src/hal0/slots/manager.py:988`) relabels the identity row and moves/rewrites the TOML, but writes no seed tombstone for the vacated name and never consults one for the new name.

The startup seeding pass (`api/__init__.py:1464`) skips a name only when it is in `ctx.slot_manager.identity_names()` (which now carries the NEW label), tombstoned, or `<name>.toml` exists on disk. After renaming a `STATIC_SEED_SLOTS` member, all three are false for the old name, so `seed_static_slots` copies the curated seed back and `fold_identity()` registers it — the operator ends up with two slots (their renamed one plus a fresh grey seed tile) after a plain service restart.

Mirror hazard: renaming *into* a tombstoned seed name leaves the stale tombstone in place, unlike `create()`, which calls `clear_seed_tombstone` (`manager.py:2322`).

## Repro

`hal0 slot rename brain steward` (slot must be OFFLINE) → `/etc/hal0/slots/brain.toml` gone, identity row now `steward`. `systemctl restart hal0-api` → boot seeding logs `slots.startup_seed names=['brain']` and re-materialises `brain.toml` from `installer/etc-hal0/slots/brain.toml`; `GET /api/slots` now lists both `steward` and `brain`.

## Severity

P2 — a routine rename of a static-seed slot silently duplicates it on the next restart.

## Found by

hal0 rc.2 pre-release review (functional review, "slots" section), verified with a live rename + restart reproduction.

thinmintdev:
## Summary
- `SlotManager.rename()` (`src/hal0/slots/manager.py:988`) relabelled the identity row and moved/rewrote the TOML but wrote no seed tombstone for the vacated name, and never consulted/cleared one for the new name.
- The startup seeding pass skips a name only when it's in `identity_names()`, tombstoned, or `<name>.toml` exists. After renaming a `STATIC_SEED_SLOTS` member, all three were false for the old name, so `seed_static_slots` copied the curated seed back on the next boot — duplicating the slot.
- Fix: on rename, tombstone the vacated name if it's a static seed (mirrors `delete()`), and clear any tombstone on the new name (mirrors `create()`) so renaming *into* a previously-deleted seed name doesn't leave a stale tombstone blocking it.

## Test plan
- [x] Red-first: added `test_renaming_a_static_seed_tombstones_the_old_name` and `test_renaming_into_a_tombstoned_seed_name_clears_it` to `tests/slots/test_id_keying.py` (the suite that wires a real `SlotIdentityStore`/`PortAuthority` through `rename()`) — both failed against pre-fix `manager.py`, confirmed the repro.
- [x] Implemented the fix, tests now pass.
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/slots tests/install tests/api/test_startup_slot_seed.py -q` — 744 + 325 passed (full slots suite + install/seed suites), no regressions.

Closes #1651

### DIFF
--- a/src/hal0/slots/manager.py
+++ b/src/hal0/slots/manager.py
@@ -994,6 +994,12 @@ class SlotManager:
         also moves those two artefacts and re-keys the in-memory caches. To
         avoid orphaning a live ``hal0-slot@<oldname>`` unit, rename is refused
         unless the slot is OFFLINE.
+
+        If ``slot_name`` is a static seed (:data:`hal0.install.static_seeds.
+        STATIC_SEED_SLOTS`), the vacated name is tombstoned so the boot-time
+        seeding pass does not resurrect a duplicate under the old label
+        (#1651); renaming into a tombstoned ``new_name`` clears it, mirroring
+        :meth:`create`.
         """
         if self._identity is None:
             raise SlotConfigError(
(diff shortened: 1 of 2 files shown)


thinmintdev:
@codex review

PR #1702, opened 2026-08-08
fix(memory): dedupe both sides of the project-scope width check

### LINKED ISSUE #1668: memory: duplicate project namespace in a read/delete scope silently widens it to unscoped shared docs
## What's broken

`_requested_scope` (`src/hal0/memory/hindsight_provider.py:635`) computes `wants_unscoped` as `len(projects) < len([r for r in reqs if r])`, comparing a **set** of project namespaces against a **list** of requests. A repeated project entry makes the set smaller than the list, so `wants_unscoped` flips True and `_is_in_scope` starts admitting every doc that carries no project tag.

`namespace.resolve_read_datasets` does not de-duplicate the caller's list (`kept` preserves duplicates), so the malformed scope reaches the provider intact. Affects `search`/`recall`/`list` and — via `_deletable_ids` — `delete`, where a delete the caller scoped to one project can reach unscoped shared documents.

## Repro

`POST /api/memory/search {"query":"x","dataset":["project:apollo","project:apollo"]}` returns unscoped shared docs alongside apollo's, where `{"dataset":["project:apollo"]}` (no duplicate) correctly returns only apollo's. Same input to `POST /api/memory/delete` makes unscoped shared doc ids deletable under a project-scoped call.

## Severity

P3 — a client-supplied duplicate in the dataset list (accidental or malicious) silently widens read/delete scope beyond the intended project; requires a specific malformed-but-plausible request shape.

## Found by

hal0 rc.2 pre-release review (functional review, "memory" section), verified with a live repro confirming the set-vs-list comparison flips `wants_unscoped`.

thinmintdev:
## Summary
- `_requested_scope` (`hindsight_provider.py:635`) computed `wants_unscoped` as `len(projects) < len([r for r in reqs if r])` — comparing a **set** of project namespaces against a **list** of requests that can carry duplicates.
- A repeated project entry (e.g. `dataset=["project:apollo", "project:apollo"]`) shrinks the set relative to the list even though every entry names the same project, flipping `wants_unscoped` True and admitting unscoped shared docs into a read/delete that only ever asked for one project.
- Fix: dedupe both sides before comparing (`len(projects) < len({r for r in reqs if r})`). Affects `search`/`recall`/`list` and, via `_deletable_ids`, `delete` — a delete scoped to one project with a duplicate could otherwise reach an unscoped shared doc.

## Test plan
- [x] `test_duplicate_project_entry_does_not_widen_recall_to_unscoped` — fails on `main` (leaks `"everyone knows"`), passes after the fix.
- [x] `test_duplicate_project_entry_does_not_widen_list_to_unscoped` — same shape for `list_items`.
- [x] `test_duplicate_project_entry_does_not_widen_delete_to_unscoped` — negative control: a duplicate-carrying project-scoped delete must not reach an unscoped shared doc (fail-closed, #1451 lineage); asserts `deleted == 0` and the shared doc survives.
- [x] `test_distinct_multi_project_entries_do_not_widen_to_unscoped` — guards against a naive "just dedupe the list" fix that could flip the comparison the other way for legitimate multi-project reads with no duplicates.
- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/memory tests/security -x -q` → 366 passed.

Closes #1668

### DIFF
--- a/src/hal0/memory/hindsight_provider.py
+++ b/src/hal0/memory/hindsight_provider.py
@@ -629,10 +629,18 @@ class HindsightProvider(MemoryProvider):
         An empty request scopes to nothing at all, so neither is True
         (#1451: ``[]`` used to read back as ``[shared]`` here too, which is
         what let the unified-mode delete gate wave shared docs through).
+
+        Both sides of the comparison must be deduplicated (#1668): a caller
+        repeating the same project entry (e.g. a client-side retry building
+        ``["project:apollo", "project:apollo"]``) must not make ``projects``
+        (a set) look smaller than the raw request count and flip
+        ``wants_unscoped`` True — that silently admitted unscoped shared docs
+        into a call that only ever named one project.
         """
         reqs = HindsightProvider._requested_list(requested)
         projects = {ds for ds in reqs if isinstance(ds, str) and ds.startswith(_PROJECT)}
-        return projects, len(projects) < len([r for r in reqs if r])
+        non_empty = {r for r in reqs if r}
+        return projects, len(projects) < len(non_empty)
 
     @staticmethod
     def _is_in_scope(item: dict[str, Any], projects: set[str], wants_unscoped: bool) -> bool:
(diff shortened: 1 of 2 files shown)


thinmintdev:
@codex review

Issue #1545, opened 2026-07-30
repo: every PR conflicts on CHANGELOG.md — consider changelog fragments


thinmintdev:
Measured cost, not a theory: during the v1.0 fix program **55 PRs merged in one day**, and very nearly every one hit a `CHANGELOG.md` conflict at least once — several hit it three or four times as `main` moved during their CI runs.

The cause is structural: every PR inserts a bullet at the top of the same `### Fixed` block in `## [Unreleased]`, so any two concurrent PRs conflict by construction. The resolution is always the same mechanical "keep both sides", which means it carries no information and is pure overhead.

Three real costs observed today:

1. **Time.** Conflict resolution became a significant share of the orchestration work — repeatedly re-merging branches that had no semantic disagreement whatsoever.
2. **It masks genuine conflicts.** A PR showing "3 conflicting files" where 2 are noise (CHANGELOG + `graphify-out/`, see #1503) is exactly the condition under which someone resolves the third carelessly.
3. **It is dangerous to bulk-resolve.** I corrupted two branches today by staging conflict markers while batch-resolving; caught by a post-resolution marker scan, but the class shouldn't exist.

Side effect already visible in the tree: union-resolving repeatedly has left `## [Unreleased]` with **5x `### Fixed`, 2x `### Added`, 2x `### Security`**. That in turn exposed #1499 (the release-notes extractor took only the last block per heading and silently dropped the rest) — a duplicate-heading bug that only mattered *because* the union pattern generates duplicate headings.

## Proposal

Adopt changelog fragments (towncrier or an equivalent, including a plain `changelog.d/*.md` convention with a small assembler):

- each PR adds **its own file** (`changelog.d/1234.fixed.md`), so two PRs can never touch the same line;
- the release cut assembles fragments into a single `## [x.y.z]` section with one heading per category — which also makes duplicate headings structurally impossible;
- CI can require a fragment on any PR touching `src/` or `ui/src/`, replacing the current "remember to add a CHANGELOG entry" convention with something enforceable.

Related: #1503 (`graphify-out/` conflicts on nearly every PR — same "regenerable/append-only artifact as merge participant" shape), #1499 (duplicate-heading extraction bug, now fixed).

Found by the v1.0 polish/fix program (release umbrella #1443). Not release-blocking; it is a workflow tax that will keep compounding.

thinmintdev:
### Fresh corroboration from tonight's queue (2026-07-30, ~19:50–21:30 UTC)

[...]

PR #1846, opened 2026-08-12
fix(ui): overview Services widget reads svc.services, not svc.data.services

### LINKED ISSUE #1836: ui(overview): the Services widget always says 'no companion services' — RDServicesCard reads svc.data.services but useServices() returns { services }
## What happens

The Overview dashboard's "Services" widget is unconditionally dead: it renders **"no companion services"** on every load regardless of how many companion services exist or are up.

Cause is a one-line property-access bug at `ui/src/dash/dashboard-redesign.jsx:849-850`:

```js
const svc = useServices()
const services = svc.data?.services ?? []     // svc has no `data` key
```

`useServices()` (`ui/src/api/hooks/useServices.ts:85-97`) returns `{ services, mdns, pending }`. `svc.data` is therefore always `undefined`, `ordered` is always `[]`, and the empty-state branch always wins. `svc.pending` resolves correctly to `false`, so the widget shows the false assertion rather than "source pending".

Correct sibling usage for contrast — `ui/src/dash/services.jsx:245`: `const { services, mdns, pending } = useServices()`.

## Decisive evidence

Live, from inside the page origin, at the same moment the widget says none exist:

```
Services card innerText = "SERVICES\n⇄\nno companion services"
await fetch('/api/services') -> {status:200, n:4, ids:["openwebui:true","comfyui:false","hermes:true","hindsight:true"]}
```

Three of four companion services up, served 200 to the page itself. Not data-starved, not auth, not CORS, not a cache.

Present in the bundle actually being served:
```
$ curl -s http://127.0.0.1:8080/ | grep -o 'assets/[A-Za-z0-9._-]*\.js'  ->  assets/index-DdgCtbtz.js
$ grep -o ".\{460\}no companion services" /usr/lib/hal0/hal0-1.0.0-rc.5/ui/dist/assets/index-DdgCtbtz.js
  ...function NC({swap:t,onGo:s}){var o;const n=Hb(),r=((o=n.data)==null?void 0:o.services)??[]...
```

Not opt-in — it is the default widget in cell `c2` (`ui/src/api/hooks/useDashLayout.ts:62,91`), so every install lands on a dashboard containing this false statement.

## Scope corrections vs the field report

- **Narrower**: only the Overview widget. The dedicated Services page (`#services`) destructures correctly and renders all four cards with correct state — verified live (`OpenWebUI UP up 17h … ComfyUI DOWN …`). That is the workaround, and it is why this is not a blocker.
- **Wider**: not state-dependent. No service configuration on any box makes this widget render a card.
- The widget consumes `/api/services` (4 entries incl. hindsight), **not** `/api/services/health` (3 entries) as the report cited. Doesn't change the verdict — the correct endpoint returns more data.

## Fix

One line: `svc.data?.services` → `svc.services`, or destructure like `ui/src/dash/services.jsx:245`. No test covers `RDServicesCard` — `ui/src/dash/__tests__/` has no services-widget test, which is why it survived four release candidates.

## Environment

hal0 1.0.0rc5, ct152 (`10.0.1.152`), fresh install, CPU-only unprivileged LXC. Read-only run: GETs and one browser session, box left exactly as found.

## Severity — major

The default landing page ships a false statement about the operator's own services. Mitigating: long-standing, not a GA regression, and the dedicated page is correct.

## Other hardware

Universal — client-side JavaScript with no dependency on backend data, hardware probing, backend selection, profile flags, or install path.

Cross-checked read-only against ct105-prod (`10.0.1.142`): the identical always-undefined access is present in the shipped bundles for **0.9.1, 0.9.5, 1.0.0-rc.1 and 1.0.0-rc.5**:
```
$ grep -o ".\{200\}no companion services" /usr/lib/hal0/hal0-*/ui/dist/assets/index-*.js
```
Introduced with the widget in `8789a07f` ("feat(ui): dashboard redesign — fixed-band layout with swap-in-place widgets (#1061)") — the bug has been present since the widget's first commit and never worked. GPU boxes, privileged containers and upgraded boxes all hit it; an upgraded box has always had it, so no user will perceive it as new in 1.0.0.

---
Found by the 1.0.0-rc.5 release-validation run. Full report: `tests/release-validation/reports/v1.0.0-rc.5.md` (branch `h0/rc5-release-testing`). Finding key: `ui-overview-services-widget-empty`.

thinmintdev:
## What changed

`RDServicesCard` (the Overview dashboard's Services widget) read
`svc.data?.services`, but `useServices()` (`ui/src/api/hooks/useServices.ts`)
returns `{ services, mdns, pending }` — there is no `data` key. `svc.data` was
therefore always `undefined`, the derived services list was always `[]`, and
the widget permanently rendered "no companion services" regardless of how
many companion services existed or were up (`svc.pending` correctly resolved
to `false`, so the widget showed the false empty-state assertion instead of
"source pending").

Extracted the widget's list derivation into a pure `services-widget-pure.js`
module:
- `orderServices(services, order, limit)` — ordering/limit logic, unchanged
  behavior from the inline version.
- `servicesForWidget(svc, order, limit)` — the actual extraction boundary
  that had the bug (`svc.data?.services` → `svc.services`), now covered by a
  unit test without needing to render the component (this repo has no React
  Testing Library setup; existing `dash/__tests__/*.test.ts` files are all
  pure-function tests).

`RDServicesCard` now calls `servicesForWidget(svc, SERVICE_ORDER)` instead of
inlining the buggy access.

## Why

Root cause, not just the symptom: the fix is the same one-line property
access noted in the issue, but pulling the derivation into a pure,
independently-testable function is what lets a regression test actually
exercise the exact code path that broke, and prevents the next widget refactor
from silently reintroducing a similar hook-shape mismatch.

## Verified

- `npm run test:unit` — 13 files / 126 tests pass, including the new
  `services-widget-pure.test.ts` (5 tests, one of which explicitly documents
  the `svc.data` bug and asserts the fixed code path does not go through it).
- Confirmed the new test's regression assertion against the pre-fix access
  pattern (`svc.data?.services`) evaluates to `[]` regardless of populated
  `svc.services` data — i.e. it reproduces the reported symptom.
- `npm run typecheck` (`tsc --noEmit`) — clean.
- `npm run lint` (`eslint .`) — clean.
- No Python touched, so `ruff format --check` is a no-op for this change.

## Left out of scope

- The dedicated Services page (`ui/src/dash/services.jsx`) already
  destructures `useServices()` correctly and was not affected.
- No CHANGELOG edit, per the current fix-wave convention (consolidated
  CHANGELOG PR lands separately, tracked in #1545).

Closes #1836


thinmintdev:
## Review — the fix is right, the test is not a regression test

**The one-line fix is correct.** `useServices()` (`ui/src/api/hooks/useServices.ts:85-101`) returns a flattened `{ services, mdns, pending }`; `svc.data` never existed. `servicesForWidget(svc, SERVICE_ORDER)` → `orderServices(svc?.services, order, 4)` reproduces the previous ordering/limit semantics exactly (named-first, then the rest, `.slice(0, 4)`), so nothing else about the widget moves. I checked the sibling hooks this file consumes (`useStatsPower`, `useThroughputHistory`, `useRequestsRollup`, `useHardware`, `useConfigUrls`) — those genuinely do expose `.data`, so `useServices` was the odd one out and there is no second instance of this class of bug in `dashboard-redesign.jsx`. The only other consumer, `dash/services.jsx:245`, already destructures correctly. Scope is clean: three files, no CHANGELOG hunk, no drive-by refactors.

### Blocking — the new test passes with the bug restored

I verified this rather than reasoning about it. On this branch I reverted **only** the call site in `dashboard-redesign.jsx` back to the buggy form, leaving `services-widget-pure.js` and its test in place:

```
  const services = svc.data?.services ?? []
  const ordered = [ ...SERVICE_ORDER.map(...) ... ].slice(0, 4)

$ npx vitest run
 Test Files  13 passed (13)
      Tests  126 passed (126)
```

The defect is back, the widget is dead again, and the suite is green. The PR body's claim that the new test "exercise[s] the exact code path that broke" and "prevents the next widget refactor from silently reintroducing a similar hook-shape mismatch" does not hold — `servicesForWidget` is new code, and nothing asserts that `RDServicesCard` calls it. The untested surface is exactly the one the issue named: *"No test covers `RDServicesCard` ... which is why it survived four release candidates."* That is still true after this PR.

The second test case is a tautology, not a detector:

```js
const buggyServices = (svc as { data?: { services?: unknown } }).data?.services
expect(orderServices(buggyServices, SERVICE_ORDER)).toEqual([])
```

`svc` is a literal defined three lines up with no `data` key, so this asserts `undefined ?? [] === []` against a value the test itself constructed. It documents the bug; it cannot detect its return. The remaining assertion duplicates the first case.

**Concrete remedy — the harness already exists.** The PR body says "this repo has no React Testing Library setup", which is true but skips the Playwright + `apiMock` suite that already renders this exact dashboard (`ui/tests/e2e/specs/dashboard-redesign-v3.spec.ts`) and runs in CI as γ-suite (chromium). `services` is the default widget for cell `c2` (`useDashLayout.ts:91`), and `services-v3.spec.ts` already has a reusable `SERVICES_PAYLOAD` and the route-override pattern. A case that routes `/api/services` to that payload and asserts the Services card renders `.rd-svc-grid` cells rather than "no companion services" fails against the old code and passes against this one. Alternatively, a static source check in the existing dependency-free style (cf. `react-hooks-order.test.mjs`, which already AST-walks all of `src/dash`) asserting that no `useServices()` consumer reads `.data` would cover the whole class.

Keep the pure module either way — it matches the `selection-pure.js` / `slot-status.js` convention and does no harm. It just should not be presented as the regression guard, because it is not one.

### Nit

`services-widget-pure.js` is untyped JS and, with `checkJs: false` and an `include` of only `src/**/*.ts{,x}`, is outside `tsc`'s view entirely. Making it `.ts` and typing the parameter as `{ services?: ManagedService[] }` would at least pin the shape at the module boundary; it can't close the hole at the `.jsx` call site, which is the underlying reason a hook-shape mismatch was invisible to `npm run typecheck` for four release candidates. Out of scope for this PR, but worth a note.

### Verification I ran

- `npx vitest run` on the branch: 13 files / 126 tests pass.
- `npx tsc --noEmit`: clean. `npx eslint .`: clean.
- Reverted-call-site experiment above: 126/126 still pass.
- CI on the PR: all eight checks green, including γ-suite (chromium).

### Codex bot

No review-bot comments, review-thread comments, or reviews are present on this PR (`pulls/1846/comments` → 0, `issues/1846/comments` → 0). Nothing to triage — and the silence is not an approval, so it does not offset the finding above.

**Verdict: changes requested** — add a check that actually fails when the call site regresses. The production change itself is good to go as written.

thinmintdev:
## Review response — blocking finding accepted, verified, fixed

**The blocking finding was correct.** I reproduced it before changing anything: reverting only the `RDServicesCard` call site back to `svc.data?.services` (leaving `services-widget-pure` and its test in place) left the vitest suite fully green. The regression lived at the untested call site, and the second unit case really was a tautology — `undefined ?? [] === []` against a fixture the test constructs three lines earlier. Both PR-body claims it supported ("exercises the exact code path that broke") were wrong. Pushed as `f85cf30e`.

Also correct: the PR body's premise that this repo has no way to render the dashboard in a test. The Playwright + apiMock γ-suite already does.

### What changed

**1. A regression test that actually fails without the fix** — `ui/tests/e2e/specs/overview-services-widget-v3.spec.ts`

`services` is the default widget for dashboard cell `c2` (`useDashLayout.ts` `CELL_DEFS`), so a plain `/#dashboard` visit mounts `RDServicesCard`. The spec route-overrides `/api/services` with a real payload (shape borrowed from `services-v3.spec.ts`) and asserts the widget renders `.rd-svc-cell` rows, not the empty state.

Proof it fails against the unfixed call site — I reverted `dashboard-redesign.jsx:851` to the original inline `svc.data?.services` and ran only this spec:

```
  1 failed
    [chromium] › overview-services-widget-v3.spec.ts:99:3 › renders the services from /api/services, not the empty state

    Error: expect(locator).toHaveCount(expected)
    Expected: 0
    Received: 1
    - waiting for locator('.rd-card').filter({ has: locator('.rd-card-title').filter({ hasText: /^Services$/ }) }).locator('.rd-empty')
        14 × locator resolved to 1 element
           - unexpected value "1"

      108 |     // The defect: the widget always fell through to this empty state.
    > 109 |     await expect(card.locator('.rd-empty')).toHaveCount(0)
  1 passed (8.2s)
```

With the fix restored: `2 passed`. A second case pins the genuine empty-list path (`no companion services` when the backend reports none) so the assertion isn't one-sided — that's the one that passed in both runs.

**2. The tautology is gone.** The unit case now feeds an input carrying *both* a populated `data.services` decoy and the real `services`, so any `.data`-first implementation returns the decoys and fails. Verified by temporarily making `servicesForWidget` read `svc.data?.services ?? svc.services`:

```
  - "openwebui", "comfyui", "hermes", "hindsight"
  + "decoy-a", "decoy-b"
  Tests  1 failed | 5 passed (6)
```

**3. Nit 1 (untyped `.js`) — taken.** `services-widget-pure.js` → `.ts`, parameters typed against `ManagedService` / a `ServicesHookResult` interface, so the module boundary is inside tsc's view. As you noted this cannot close the `.jsx` call-site hole; the e2e spec is what covers that, and the file comment says so.

**4. One thing outside the findings, flagged for your call.** While wiring the test I found `npm run test:unit` is not run by *any* workflow — `ci.yml`'s `ui` job runs lint, typecheck, build only, and the `.test.mjs` files aren't run either (vitest's `include` is `src/**/*.test.ts`). So the entire vitest suite gated nothing, which is part of why a decorative unit test read as coverage. I added a `Unit tests` step to `ci.yml`'s `ui` job (~0.4s for 13 files). It's a CI-config hunk in a fix PR — happy to split it out if you'd rather keep this PR to three files.

**Nit 2** was positive verification, no action.

### Verified

- `npx playwright test overview-services-widget-v3 services-v3 dashboard-redesign-v3 --project=chromium` → 21 passed
- `npx vitest run` → 13 files / 127 tests passed
- `npx tsc --noEmit` → clean; `npx eslint .` → clean
- No Python touched, so `ruff format --check` is unaffected
- No CHANGELOG hunk; automerge not enabled

Issue #1793, opened 2026-08-09
fix(agent): hermes provision smoke_tests phase reports status=ok while recording failures — and runs before any model exists


thinmintdev:
Found during v1.0.0-rc.4 fresh-install validation (CT151), confirmed twice (original provision snapshot + `hal0 agent reprovision hermes` rerun).

## What happens

The original provision (completed 20:55:25Z, before any model was loaded on the box) recorded in `provision.json`:

```
"smoke_tests": { "status": "ok", "failures": [
  "chat_completions: chat/completions: timed out",
  "memory_roundtrip: memory_search returned no items for just-written marker" ] }
```

2 of 6 smoke checks failed; the phase still reports `ok`, `hal0 agent status hermes` shows every phase green, and the dashboard presents a fully healthy hermes. The failures were real: no chat model existed yet (the box's first model was registered ~25 min after provision), so the smoke was structurally guaranteed to fail its chat and memory checks — and then mask them.

## Two defects

1. **Reporting:** a phase with recorded failures must not be `status=ok` (or the failures should at least surface in the status table — the Detail column carries them but the Status column says ok and nothing rolls up).
2. **Sequencing:** the provision smoke runs at install time when no model can exist yet. Either defer the chat/memory smoke until a model is available, or report "skipped: no model loaded" instead of a timeout that gets swallowed.

Verified during the same validation: after models were loaded, a reprovision's smoke passes — the checks themselves are sound.

Found by automated release-validation agents; confirmed via provision.json snapshots before filing.

PR #1849, opened 2026-08-12
fix(agents): resolve chat anchor by-id, ungate memory_roundtrip

### LINKED ISSUE #1831: installer: the rc.5 smoke-test chat preflight checks hal0/agent against GET /v1/models, which never lists virtuals — chat_completions and memory_roundtrip are skipped on every install
## What happens

rc.5's new smoke-test preflight `_chat_model_ready` (`src/hal0/agents/hermes_provision.py:4717`, added by `56ad690e` / #1798, `git tag --contains` → v1.0.0-rc.5 only) cross-checks hermes' `model.default` against the ids returned by `GET /v1/models` (the **list** route).

That route deliberately never advertises the `hal0/*` virtual names (`src/hal0/api/routes/v1.py` ~938: "Canonical virtual names (hal0/agent, hal0/utility, hal0/npu) are no longer advertised here", behaviour since #1153) and additionally suppresses chat slots' raw model ids in favour of their bare alias. The shipped default under live-resolve is `model.default: hal0/agent` (`hermes_provision.py:1465`), so **the anchor can never be a member of that list on any box**; under `HAL0_HERMES_LIVE_RESOLVE=0` the pinned physical model id is likewise suppressed.

Consequence: `chat_completions` and `memory_roundtrip` are skipped on every hal0-only install with the reason `'hal0/agent' not loaded on the gateway` — which is **false**. The two probes that in rc.4 actually caught real failures (chat timeout, memory no-marker — regressions.yaml `hermes-smoke-ok-over-failures`, #1793) now never run at all.

Secondary: `memory_roundtrip` has no chat dependency whatsoever (pure MCP `memory_add`/`memory_search`; it ran and passed on ct105-prod) yet is gated behind the chat preflight, so it would be over-skipped even with a correct anchor check.

## Decisive evidence

```
$ python3 -c "import json;d=json.load(open('/var/lib/hal0/state/agents/hermes/provision.json'));print(json.dumps(d['phases']['smoke_tests'],indent=1))"
"failures": [], "skipped_count": 2, "status": "ok"
chat_completions / memory_roundtrip = {"passed": null, "skipped": true,
   "detail": "skipped: no chat model loaded ('hal0/agent' not loaded on the gateway)"}

$ curl -s http://127.0.0.1:8080/v1/models   ids -> ['brain','utility','zzskep','zzfpxmodel','zzcrashmodel','zzbroken','rcrerank']
   'hal0/agent' in list: False | 'agent' in list: False | 'hal0-brain-sft-f16' in list: False

$ curl -s -w '%{http_code}' http://127.0.0.1:8080/v1/models/hal0/agent
200 {"id":"hal0/agent",...,"name":"brain · hal0-brain-sft-f16","context_length":32768}
```

The **by-id** route resolves virtuals (`_resolve_virtual_model_entry`, `v1.py:1005`); the list route does not. The preflight uses the list route. The anchor is routable; the recorded reason is false.

Installed source matches the repo: `/usr/lib/hal0/venv/.../hal0/agents/hermes_provision.py:4745` → `if model_name not in contexts: return False, f"'{model_name}' not loaded on the gateway"`.

## Why it shipped

The unit tests never use the real anchor. `tests/agents/test_hermes_provision.py:1951` `_write_ready_chat_config` writes `model.default: m1` and mocks `fetch_model_contexts=lambda: {"m1": 8192}`; `TestChatModelReady` (line 2039) exercises only synthetic ids. No test asserts the shipped `hal0/agent` default against a realistic `/v1/models` payload.

## What should happen

Resolve the anchor through `GET /v1/models/{id}` (which already handles virtuals) or the LiveSlotResolver instead of list membership; add a regression test pinning the real shipped default `hal0/agent` against a realistic `/v1/models` payload; ungate `memory_roundtrip` (it has no chat dependency).

## Not claimed (refuted during verification)

- This did **not** mask the Hermes 64K-floor break. `_smoke_chat_completions` (4544) does a raw urllib POST to `model.base_url/chat/completions` and `_smoke_memory_roundtrip` (4605) calls MCP directly; neither goes through the hermes CLI, where that gate lives.
- The skip is **disclosed**: `hal0 agent status` renders `smoke_tests | ok | 2 skipped`. The defect is lost verification coverage plus a false reason string, not a hidden failure.
- The `status=ok`-over-skips policy is by design and is not being filed. The #1793 warn-on-failure path is still reachable by the other four probes.

## Environment

hal0 1.0.0rc5, ct152 (`10.0.1.152`), fresh install, CPU-only unprivileged LXC.

## Severity — major

Two release-gating probes are permanently dead on the shipped configuration, with a false explanation in the record.

## Other hardware

Hits every box type. The exclusion of `hal0/*` virtuals from `GET /v1/models` is unconditional in `_aggregate_models`, and the shipped `model.default: hal0/agent` is written regardless of backend. Read-only cross-check on ct105-prod (GPU, privileged, upgraded, same 1.0.0rc5): `hal0/agent` is absent from `/v1/models` even with the `agent` slot model-bound and GPU slots serving — so a **GPU fresh install skips both probes identically**. Prod only avoids the skip because an operator repointed `model.default` at the passthrough id `MiniMax-M3`. Nothing here depends on CPU-only or the unprivileged container: the preflight short-circuits before any chat round trip.

---
Found by the 1.0.0-rc.5 release-validation run. Full report: `tests/release-validation/reports/v1.0.0-rc.5.md` (branch `h0/rc5-release-testing`). Finding key: `smoke-tests-permanently-skip-chat-probes`.

thinmintdev:
## What changed

`_chat_model_ready` (the rc.5 smoke-test chat preflight) cross-checked
`model.default` against `GET /v1/models` — the **list** route, which
deliberately never advertises hal0's canonical virtual names (`hal0/agent`,
`hal0/utility`, `hal0/npu`, #1153) and folds a chat slot's raw model id into
its alias entry. The shipped default is `model.default: hal0/agent`, so the
anchor could never be a member of that list on any box. Result:
`chat_completions` and `memory_roundtrip` were skipped on every hal0 install
with a false `'hal0/agent' not loaded on the gateway` reason.

Fix, matching the issue's recommendation:

- Add `_fetch_model_route_ready(model_id)`, which hits `GET
  /v1/models/{id}` (the **by-id** route — resolves virtuals via the
  LiveSlotResolver, and is exactly what `_smoke_chat_completions` itself
  dispatches against) instead of checking list membership.
- `_chat_model_ready` now calls this through a new `InstallIO.
  fetch_model_route_ready` seam instead of `fetch_model_contexts`.
- Ungate `memory_roundtrip` from `_MODEL_DEPENDENT_PROBES`: it drives the
  MCP `memory_add`/`memory_search` tools directly and has no chat
  dependency, so gating it behind the chat anchor over-skipped it even when
  memory was fully working (secondary defect called out in the issue).

## Why

Two release-gating probes were permanently dead on the shipped
configuration (rc.4's chat-timeout and memory-no-marker regressions,
regressions.yaml `hermes-smoke-ok-over-failures` / #1793, would no longer
be caught). Root cause is the list-vs-by-id route mismatch, not a bug in
the probes themselves — same seam `_smoke_chat_completions` already uses
correctly.

## How verified

- New regression test `TestChatModelReady::
  test_shipped_virtual_default_ready_via_by_id_route_not_list` pins the
  real shipped default (`model.default: hal0/agent`) against a realistic
  `/v1/models` list payload that never contains it, and confirms readiness
  comes from the by-id route. Failed before the fix (list-membership
  check), passes after.
- Updated the three existing tests that depended on the old
  `fetch_model_contexts`-based gate (`TestChatModelReady`,
  `test_smoke_tests_phase_runs_each_probe_collecting_results`,
  `test_smoke_tests_phase_records_failures_as_warn_without_blocking`,
  `test_smoke_tests_phase_skips_model_dependent_probes_without_chat_model` —
  the last now asserts `memory_roundtrip` runs even when the chat anchor is
  not ready).
- `uv run pytest tests/agents/ -q` → 653 passed, 1 xfailed (pre-existing,
  unrelated).
- `uv run pytest -q` (full suite) → passed.
- `uv run ruff check` / `uv run ruff format --check` on both changed files
  → clean.

## Out of scope

- `CHANGELOG.md` is intentionally untouched (per wave convention — one
  consolidated CHANGELOG PR lands at the end).
- Did not touch `_smoke_chat_completions` or `_smoke_memory_roundtrip`
  themselves — both were already correct (issue's "Not claimed" section);
  only the preflight anchor check and the probe gating set changed.

Closes #1831

### DIFF
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -3256,6 +3256,37 @@ def _fetch_model_contexts() -> dict[str, int]:
     return out
 
 
+def _fetch_model_route_ready(model_id: str) -> bool:
+    """Ask ``GET /v1/models/{id}`` whether ``model_id`` actually routes.
+
+    Unlike :func:`_fetch_model_contexts` (backed by the **list** route,
+    ``GET /v1/models``), this hits the **by-id** route, which resolves hal0's
+    canonical virtual names (``hal0/agent``, ``hal0/utility``, ``hal0/npu``)
+    via the LiveSlotResolver (see ``_resolve_virtual_model_entry`` in
+    ``hal0.api.routes.v1``). The list route deliberately never advertises
+    those virtuals (#1153) and also folds a chat slot's raw model id into
+    its alias entry, so checking list *membership* against the shipped
+    ``model.default: hal0/agent`` can never succeed even when the model is
+    fully routable (#1831). 404/other client errors and network failures
+    both mean "not ready"; only a clean response counts.
+    """
+    from urllib.error import HTTPError, URLError
+    from urllib.parse import quote
+    from urllib.request import Request, urlopen
+
+    req = Request(
+        f"{HAL0_API_URL}/v1/models/{quote(model_id, safe='/')}",
+        headers={"Accept": "application/json"},
+    )
+    try:
+        with urlopen(req, timeout=3.0) as resp:
+            return 200 <= resp.status < 300
+    except HTTPError:
+        return False
+    except (URLError, OSError, TimeoutError):
+        return False
+
+
 def _slot_kind(slot: dict[str, Any]) -> str:
     """Best-effort capability classifier — handles a few schema variants.
 
@@ -4527,12 +4558,14 @@ def _phase_voice_wire(ctx: _StepCtx) -> PhaseResult:
 # burying it in the Detail column. self_report surfaces the same rollup in
 # the bootstrap-completion memory item.
 #
-# chat_completions and memory_roundtrip both need a routable chat model to
-# mean anything; at install time — before any model has ever been loaded —
-# that's structurally impossible, so a preflight (`_chat_model_ready`)
-# decides ONCE whether the route is live and both probes report
-# `skipped: no chat model loaded` instead of burning their timeout on a
-# doomed request that then gets recorded as an opaque failure.
+# chat_completions needs a routable chat model to mean anything; at install
+# time — before any model has ever been loaded — that's structurally
+# impossible, so a preflight (`_chat_model_ready`) decides ONCE whether the
+# route is live and the probe reports `skipped: no chat model loaded`
+# instead of burning its timeout on a doomed request that then gets
+# recorded as an opaque failure. memory_roundtrip has no chat dependency
+# whatsoever (it drives the MCP memory_add/memory_search tools directly) and
+# always runs (#1831).
 
 
 def _wrapper_bin() -> Path:
@@ -4711,7 +4744,11 @@ def _smoke_hermes_doctor(_state: BootstrapState, io: InstallIO) -> tuple[bool, s
 #: Probes that only mean something once a chat-capable model is actually
 #: routable. Gated behind :func:`_chat_model_ready` so an install-time smoke
 #: run (no model has ever been loaded yet) reports `skipped`, not a timeout.
-_MODEL_DEPENDENT_PROBES = frozenset({"chat_completions", "memory_roundtrip"})
+#: memory_roundtrip does NOT belong here — it exercises the MCP
+#: memory_add/memory_search tools directly and has no chat dependency
+#: whatsoever (#1831); gating it behind the chat anchor over-skipped it even
+#: when memory itself was fully working.
+_MODEL_DEPENDENT_PROBES = frozenset({"chat_completions"})
 
 
 def _chat_model_ready(state: BootstrapState, io: InstallIO) -> tuple[bool, str]:
@@ -4719,12 +4756,25 @@ def _chat_model_ready(state: BootstrapState, io: InstallIO) -> tuple[bool, str]:
 
     Reads ``model.default`` from config.yaml — the exact field
     :func:`_smoke_chat_completions` targets — and cross-checks it against the
-    live gateway model list via ``io.fetch_model_contexts`` (the same
-    ``/v1/models`` seam ``model_automap`` already uses), instead of paying a
-    full chat-completion round trip just to find out nothing is loaded. This
-    is what lets smoke_tests tell "the route is genuinely broken" (a real
-    failure) apart from "no chat model exists yet" (#1793) — never raises;
-    any config/parse/network problem reports not-ready with the reason.
+    live gateway via ``io.fetch_model_route_ready`` (``GET
+    /v1/models/{id}``), instead of paying a full chat-completion round trip
+    just to find out nothing is loaded.
+
+    This resolves the anchor through the **by-id** route rather than
+    checking membership in the **list** route's payload (``GET
+    /v1/models``): the list route deliberately never advertises hal0's
+    canonical virtual names (``hal0/agent`` et al. — see
+    ``_aggregate_models`` in ``hal0.api.routes.v1``, #1153) and also folds a
+    chat slot's raw model id into its alias entry, so list membership can
+    never hold for the shipped ``model.default: hal0/agent`` even when the
+    model is fully routable. The by-id route resolves virtuals via the
+    LiveSlotResolver (``_resolve_virtual_model_entry``), matching exactly
+    what :func:`_smoke_chat_completions` itself dispatches against (#1831).
+
+    This is what lets smoke_tests tell "the route is genuinely broken" (a
+    real failure) apart from "no chat model exists yet" (#1793) — never
+    raises; any config/parse/network problem reports not-ready with the
+    reason.
     """
     import yaml  # type: ignore[import-untyped]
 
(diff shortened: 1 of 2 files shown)


thinmintdev:
## Post-merge review — #1849 (fixes #1831)

Verified against `origin/main` sources, not the PR description. Verdict: **the headline defect is genuinely fixed, but the fix is partial** — one path the issue explicitly named is still broken. No blocking scope problems.

### 1. Does it fix the reported defect? Yes, on the shipped path

`_chat_model_ready` now resolves the anchor through `io.fetch_model_route_ready` → `GET /v1/models/{id}`. Confirmed against the route: `src/hal0/api/routes/v1.py:968` is `@public_router.get("/models/{model_id:path}")` (so the `hal0/agent` slash survives — and `quote(model_id, safe='/')` correctly keeps it unescaped), and `_resolve_virtual_model_entry` (`v1.py:1005`) resolves `hal0/*` via the `LiveSlotResolver`, including the `fallback=True` (nothing warm yet) path. So the issue's exact repro — `model.default: hal0/agent`, absent from the list payload, 200 on the by-id route — is addressed at the seam that produced it.

The `#1793` semantics are not weakened, which I checked specifically: `hal0_slot_alias_models` advertises **cold as well as warm** slots ("Both warm and cold slots are advertised"), so the old list check was never a liveness check either. Both routes gate on *model-bound*, not *loaded*. No new "burn the 60s timeout at install time" exposure from the anchor change.

### 2. Still broken: `HAL0_HERMES_LIVE_RESOLVE=0` — the same false skip

The issue called this out explicitly ("under `HAL0_HERMES_LIVE_RESOLVE=0` the pinned physical model id is likewise suppressed"), and the fix does not cover it.

Under `HAL0_HERMES_LIVE_RESOLVE=0`, `_config_pairs` (`hermes_provision.py:1448`, `:1465`) writes:

```
model.default  = primary["model_id"]        # a chat slot's RAW physical model id
model.base_url = primary["backend_url"]     # the slot's own llama-server, NOT the gateway
```

Now trace that id through the new preflight:

- `get_model` calls `_aggregate_models(show_all=True)`, which suppresses every chat slot's raw model id in favour of its alias row — `if mid in seen or mid in chat_model_ids: continue`, and `hal0_chat_slot_model_ids` "does NOT filter on loaded state" (`src/hal0/api/__init__.py:593`). So the physical id is never in `data`.
- `_resolve_virtual_model_entry` returns `None` immediately for anything not starting with `hal0/` (`v1.py:1023`).
- → `NoRouteFound` → 404 → `_fetch_model_route_ready` returns `False` → **`'<physical-id>' not loaded on the gateway`, `chat_completions` skipped** — the identical false reason string #1831 was filed over.

And it is doubly wrong there, because in that mode `_smoke_chat_completions` POSTs to the slot's `backend_url` directly and never touches the gateway catalog at all. Which makes the new docstring claim — "matching exactly what `_smoke_chat_completions` itself dispatches against" — false precisely in the case where the anchor check fails.

**Root cause vs symptom:** the underlying cause is that the preflight uses a *discovery catalog* to answer a *routability* question. The by-id route is a strictly better catalog, so this is a real improvement, not a symptom patch — but it is still a catalog check, and the non-default path stays dead. Please file a follow-up: either gate the probe against `model.base_url` (the endpoint the probe actually uses) rather than a hardcoded `HAL0_API_URL`, or treat a 404 for a non-virtual id as *inconclusive* → run the probe rather than skip it. The regression-register entry `smoke-preflight-skips-chat-probes` (`tests/release-validation/regressions.yaml:426`) will pass on the default config and quietly keep hiding this.

### 3. Does the new test fail without the fix? Yes — verified by running it

Checked out `main`'s test file against the pre-fix `hermes_provision.py` and ran it:

```
7 failed, 116 passed
TypeError: InstallIO.__init__() got an unexpected keyword argument 'fetch_model_route_ready'
```

and with the fix in place: `123 passed in 2.85s`.

Caveat on strength: `test_shipped_virtual_default_ready_via_by_id_route_not_list` fails pre-fix on the *seam rename*, not on behaviour. It does still pin the behaviour (reverting `_chat_model_ready`'s body to list membership while keeping the seam fails it, since `hal0/agent ∉ {"brain","utility"}`), so it earns its keep — but the `fetch_model_contexts=realistic_list_contexts` argument is inert, and nothing exercises `_fetch_model_route_ready` itself. There is also no counterpart to the existing `assert io.fetch_model_contexts is hp._fetch_model_contexts` (`tests/agents/test_hermes_provision_context.py:33`), so a default-wiring typo on the new seam would ship silently. Worth one line.

### 4. Regressions from the ungating

`_MODEL_DEPENDENT_PROBES` dropping `memory_roundtrip` is correct — `_smoke_memory_roundtrip` goes through `io.mcp_memory_call` → `POST /api/memory/{add,search}` with a 5s timeout, no chat path. One behaviour change worth knowing: on a fresh install the probe now always runs, so a memory layer that is not yet answering flips `smoke_tests` from `ok` to `warn` where it previously reported a skip. `_phase_namespace_register` explicitly tolerates the memory layer being unavailable mid-install, so that state is considered reachable by design. Low risk (same hal0-api process the `admin_tools_list` probe already requires), and arguably the more honest output — just don't be surprised by it in the next validation run.

No other caller of `fetch_model_contexts` is affected (`model_automap` / `render_live_context` still use it, unchanged).

### 5. Scope

Clean. Two files, no `CHANGELOG.md` hunk, no drive-by refactor, the comment rewrites all track the behaviour change.

### 6. Codex bot

The bot posted **nothing** on this PR — zero reviews, zero review comments, `reviewDecision` empty. That is silence (usage limit or a missed trigger), not an approval, and nothing here was auto-triaged before merge.

PR #1850, opened 2026-08-12
fix/v1 models cache eviction

### LINKED ISSUE #1837: api: GET /v1/models never evicts a deleted slot's model id — the composite cache merges instead of replacing, advertising ids that hard-404
## What happens

`GET /v1/models` never evicts an id from its `hal0`-owned composite catalogue for the lifetime of the `hal0-api` process. Any model id that was bound to an llm slot when the composite cache was primed stays advertised forever, even after the slot and the registry entry are gone.

`_prime_hal0_composite_cache` (`src/hal0/api/__init__.py:621-627`) **merges** the fresh slot-config read into the existing bucket instead of replacing it (docstring line 607: "Merges rather than replaces"), so the bucket grows monotonically. Only two prime sites exist (startup line 1252, slot→ready event line 686), both through this helper.

When the id is also absent from the registry, requests against it hard-404 with `dispatch.no_route` — an OpenAI-compatible client that builds its picker from `/v1/models` offers an entry that can never work.

## Decisive evidence (ct152, hal0 1.0.0rc5, uptime ~17 h, `NRestarts=0`)

```
curl -s http://10.0.1.152:8080/v1/models -> ids ["brain","utility","zzfpxmodel","zzcrashmodel","zzbroken"]
hal0 slot list  -> 11 slots; none named or bound to zzbroken/zzfpxmodel/zzcrashmodel
hal0 model list -> 4 models; none of the zz ids

curl -w '[%{http_code}]' .../v1/models/zzbroken -> 200 {"id":"zzbroken",...,"downloaded":false}
curl               .../api/models/zzbroken      -> 404 model.not_found
POST /v1/chat/completions {"model":"zzbroken"}  -> 404 dispatch.no_route
   "model 'zzbroken' not found in registry, no upstream advertised it, and legacy slot resolution failed"
```

Provenance of the ghosts, from `journalctl -u hal0-api`: `slot=zzskeptic model_id=zzbroken`, `slot=zzs7 model_id=zzbroken`, `slot=zzfpxslot model=zzfpxmodel`, `slot=zzcrash model_id=zzcrashmodel`. All four slots are gone; the ids are not.

## The mechanism in the original report is wrong — do not use its repro

The catalogue is **not** built from orphaned files on disk, and `hal0 model add X` + `hal0 model rm X -f` does **not** reproduce it. `_fetch_hal0_composite_models` (`src/hal0/api/__init__.py:203-233`) iterates *slot configs* with `type == "llm"`; the registry is never a row source (`_catalog_row`, `v1.py:870-897`, uses it only to set `downloaded`/labels/checkpoint). The orphan `/var/lib/hal0/models/zzbroken/zzbroken.gguf` is a red herring.

Corroboration that the registry is not a row source (read-only, ct105-prod): 62 registered models, only 5 hal0-owned ids in `/v1/models` (all slot aliases). 57 registered models correctly absent.

## What should happen

Seed the FLM multiplex tags (`embed-gemma:300m`, `whisper-v3:turbo` — the actual reason the merge exists, per the docstring) inside `_fetch_hal0_composite_models`, and let the prime **replace** rather than merge. `list_models`' own docstring already describes the composite as "every ready chat-capable slot's model id", which is not what it returns.

Cleared by `systemctl restart hal0-api` (the cache is in-process `AppContext` state re-primed from live slot configs at startup). Not verified on the box — other lanes were mid-run — but it follows from `model_cache` being an in-memory dict with no persistence.

## Environment

hal0 1.0.0rc5, ct152 (`10.0.1.152`), fresh install, CPU-only unprivileged LXC.

## Severity — minor

Recoverable by a restart; the milder and much more common variant (rebinding a chat slot from model A to B leaves A advertised) resolves through the by-design agent-anchor substitution rather than 404ing. Only the delete-the-model case produces the hard 404.

Distinct from known-issue `dispatcher-unknown-model-passthrough` (that is a *registered* model answered by the agent anchor; here the id is not registered at all and the anchor is offline).

## Other hardware

Fully hardware-independent — application-level cache logic in the API process. GPU boxes, privileged containers and upgraded boxes all hit it identically; ct105-prod is clean only because its `hal0-api` was restarted recently and no slot has been deleted since. An upgraded box is marginally *less* exposed right after an upgrade, since the package upgrade restarts `hal0-api`. Exposure scales with uptime and with how often the operator deletes slots or rebinds slot models.

---
Found by the 1.0.0-rc.5 release-validation run. Full report: `tests/release-validation/reports/v1.0.0-rc.5.md` (branch `h0/rc5-release-testing`). Finding key: `v1-models-advertises-orphan-zzbroken`.

thinmintdev:
## Summary

- `GET /v1/models` never evicted a deleted slot's model id from its `hal0`-owned composite catalogue: `_prime_hal0_composite_cache` **merged** the fresh slot-config read into `model_cache["hal0"]` instead of replacing it, so any id ever bound to an `llm` slot stayed advertised for the life of the `hal0-api` process — including after the slot (and its registry entry) were deleted. Clients that pick a model from `/v1/models` could pick a ghost id that hard-404s (`dispatch.no_route`) on every request.
- Root cause: the merge existed only so a fresh (replacing) prime wouldn't drop the FLM multiplex tags (`embed-gemma:300m`, `whisper-v3:turbo`) seeded separately by `_seed_multiplex_models`. Fixed at the root by folding that tag derivation directly into `_fetch_hal0_composite_models` (it already reads the same slot configs), so every fetch produces the complete, correct set — including multiplex tags — and `_prime_hal0_composite_cache` can safely `model_cache["hal0"] = list(models)` on every call instead of merging forward.
- `_seed_multiplex_models` is left in place (still covered by its own unit tests, still called on the startup path) but is now effectively a no-op there since the prime it follows already includes the multiplex tags; it remains useful for callers that populate `model_cache["hal0"]` some other way (e.g. an explicit `upstreams.toml` override, where composite priming is skipped entirely).

## Why

Reported by the v1.0.0-rc.5 release-validation run (issue #1837, finding `v1-models-advertises-orphan-zzbroken`): a client building its model picker from `/v1/models` could offer an id that can never work. The issue's own repro attempt (orphan file on disk + `hal0 model add/rm`) was a red herring — the composite catalogue is built from *live slot configs*, not the registry or disk — so I reproduced the real mechanism directly: prime with a slot bound to a model id, delete the slot, re-prime, and assert the id survives (it did, before this fix).

## Verification

- New regression test `test_priming_evicts_a_model_id_whose_slot_was_deleted` in `tests/api/test_upstream_dedup.py` — fails on `main` (asserts `['ghost-model'] == []`, gets `['ghost-model']`), passes after the fix.
- `uv run --extra dev pytest tests/api -q` — 1731 passed, 1 skipped (pre-existing journalctl skip), including all `test_upstream_dedup.py`, `test_seed_multiplex_models.py`, and `test_model_cache_refresh.py` cases (multiplex seeding and the ready-event re-prime path both still behave correctly).
- `uv run --extra dev ruff check` / `ruff format --check` on the touched files — clean.

## Out of scope

- `CHANGELOG.md` — left untouched per the fix-wave convention (tracked separately, #1545).
- The milder, more common variant noted in the issue (rebinding a chat slot from model A to B leaves A advertised) resolves through the existing by-design agent-anchor substitution and isn't a 404 — not touched here.

Closes #1837

### DIFF
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -247,9 +247,8 @@ async def _fetch_hal0_composite_models(
         is_flm = "flm" in (cfg.get("provider", ""), cfg.get("backend", ""))
         if not is_flm or not name:
             continue
-        # Container-era schema is the [npu] table (what FLMProvider builds
-        # the --asr/--embed argv from); [defaults] load_* is the pre-#733
-        # legacy shape, kept so older tomls keep seeding.
+        # Two config shapes accepted here — see _seed_multiplex_models for
+        # why (container-era [npu] table vs. pre-#733 [defaults] load_*).
         npu_table = cfg.get("npu") or {}
         defaults = cfg.get("defaults") or {}
         load_embed = npu_table.get("embed") or defaults.get("load_embed")
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -191,6 +191,16 @@ async def _fetch_hal0_composite_models(
     place without an operator having to wire a matching upstreams.toml
     entry per slot.
 
+    Also folds in the FLM multiplex tags (``embed-gemma:300m`` /
+    ``whisper-v3:turbo``, see :func:`_seed_multiplex_models`) for any FLM
+    slot whose config opts into them — those tags never appear in an FLM
+    slot's own ``model.default`` (they ride along on the same process) so
+    they have to be derived here, freshly, on every fetch. Doing it inside
+    this fetch (rather than as a separate merge step downstream) is what
+    lets callers safely REPLACE their cache with this return value instead
+    of merging it forward (#1837 — a merge-forward is how a deleted slot's
+    model id kept getting advertised after the slot was gone).
+
     The returned list is sorted + deduplicated and cached for
     ``ttl_seconds`` to keep the cold-start fan-out cheap while still
     picking up new slots within a handful of seconds.
@@ -231,6 +241,25 @@ async def _fetch_hal0_composite_models(
             continue
         seen.add(model_id)
         models.append(model_id)
+
+    for cfg in cfgs:
+        name = cfg.get("name", "")
+        is_flm = "flm" in (cfg.get("provider", ""), cfg.get("backend", ""))
+        if not is_flm or not name:
+            continue
+        # Container-era schema is the [npu] table (what FLMProvider builds
+        # the --asr/--embed argv from); [defaults] load_* is the pre-#733
+        # legacy shape, kept so older tomls keep seeding.
+        npu_table = cfg.get("npu") or {}
+        defaults = cfg.get("defaults") or {}
+        load_embed = npu_table.get("embed") or defaults.get("load_embed")
+        load_asr = npu_table.get("asr") or defaults.get("load_asr")
+        if load_embed and _FLM_EMBED_TAG not in seen:
+            seen.add(_FLM_EMBED_TAG)
+            models.append(_FLM_EMBED_TAG)
+        if load_asr and _FLM_ASR_TAG not in seen:
+            seen.add(_FLM_ASR_TAG)
+            models.append(_FLM_ASR_TAG)
 
     models.sort()
     _HAL0_MODEL_CACHE[_HAL0_COMPOSITE_CACHE_KEY] = (monotonic_now + ttl_seconds, list(models))
@@ -604,10 +633,17 @@ async def _prime_hal0_composite_cache(
     means the first request after startup doesn't have to pay the
     slot-iteration cost.
 
-    Merges rather than replaces: any tag already in the bucket that the
-    fresh catalogue read doesn't reproduce (e.g. FLM multiplex tags seeded
-    by :func:`_seed_multiplex_models`) is preserved, mirroring
-    ``_fetch_and_cache``'s merge behaviour for ordinary upstreams.
+    REPLACES ``model_cache["hal0"]`` wholesale on every call — it must NOT
+    merge the fresh catalogue read with whatever was there before. A slot
+    that's since been deleted (or rebound to a different model) must have
+    its old id fall out of the bucket, not accumulate forever (#1837: a
+    merge-forward left deleted slots' model ids advertised via
+    ``/v1/models`` for the life of the ``hal0-api`` process — ids that hard
+    404 on dispatch since neither a slot nor a registry entry backs them
+    anymore). FLM multiplex tags (``embed-gemma:300m`` / ``whisper-v3:turbo``)
+    are freshly re-derived on every call too — see
+    :func:`_fetch_hal0_composite_models` — so replacing here doesn't drop
+    them.
 
     Skipped if an explicit ``upstreams.toml`` entry already claims the
     name ``hal0`` — operator overrides win, and their entry's model cache
@@ -619,12 +655,7 @@ async def _prime_hal0_composite_cache(
         log.info("slots.hal0_composite_skipped", reason="explicit_upstream_registered")
         return
     models = await _fetch_hal0_composite_models(slot_manager)
-    existing = model_cache.get("hal0", [])
-    merged = list(models)
-    for tag in existing:
-        if tag not in merged:
-            merged.append(tag)
-    model_cache["hal0"] = merged
+    model_cache["hal0"] = list(models)
 
 
 # ── FLM multiplex model seeding ────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


thinmintdev:
## Post-merge review — PR #1850 (fixes #1837)

Reviewed the merged diff (`4187e9b`) against the issue's repro, ran the tests both with and
against the fix, and verified both Codex bot findings independently.

**Verdict: the change is correct and necessary, but it does not close #1837, and it introduces a
narrow regression. Recommend a follow-up before this finding is marked closed in the rc.5
regression register.**

---

### What's good

- **The test is genuinely red without the fix.** Verified by hand, not by trusting the PR body:
  `git checkout 4187e9b^ -- src/hal0/api/__init__.py` and re-running gives
  `AssertionError: assert ['ghost-model'] == []` on
  `test_priming_evicts_a_model_id_whose_slot_was_deleted`. With the fix, `tests/api/test_upstream_dedup.py`,
  `test_seed_multiplex_models.py`, `test_model_cache_refresh.py`, `test_v1_models_filters.py`,
  `test_v1_slot_alias_models.py`, `test_models_catalogue.py`, `test_virtual_models.py` are all green (50 passed).
- **The root-cause framing is right, and the fix is at the cause, not the symptom.** The merge existed
  only to preserve the multiplex tags; folding that derivation into `_fetch_hal0_composite_models`
  and letting the prime replace is the correct shape, and it is exactly what the issue prescribed.
  I confirmed the relocated derivation works for both config shapes (`backend="flm"` + `[npu]`,
  and legacy `provider="flm"` + `[defaults] load_*`) with an ad-hoc probe against the real function.
- **No scope creep.** Two files, no `CHANGELOG.md` hunk, no drive-by refactor. Clean.

---

### 1. (blocking for closing #1837) The reported repro is still reproducible — deleting a slot never re-primes

**Codex's first P2 comment is correct. Confirmed by reading the code, not just agreeing with the bot.**

`DELETE /api/slots/{name}` (`src/hal0/api/routes/slots.py:1085-1098`) calls `SlotManager.delete` and
nothing else. `SlotManager.delete` (`src/hal0/slots/manager.py:2538`) calls `unload()` for a running
slot — a `ready → offline` transition — and emits no `ready` event. The only runtime re-prime,
`_refresh_model_cache_on_ready` (`src/hal0/api/__init__.py:691-716`), gates on `data.get("to") != "ready"`
and `continue`s otherwise. `GET /v1/models` reads `model_cache["hal0"]` directly
(`src/hal0/api/routes/v1.py:909`) and never calls the fetch itself.

Net effect after this PR: delete the slot bound to `zzbroken` and `zzbroken` **stays advertised**
until some *unrelated* slot happens to transition to ready, or `hal0-api` restarts. That is the
ct152 scenario in the issue verbatim. The PR turned "never evicted" into "evicted opportunistically,
on an edge that has nothing to do with the deletion" — strictly better, not closed.

The correct place is the delete path: call `_hal0_model_cache_clear()` + re-prime on a successful
`delete_slot` (and, for the rebind variant, on `PUT /{name}/config` when `model.default` changes), or
subscribe to a slot-deleted event. Cheapest alternative if you'd rather not add a write-path hook:
have `list_models` call `_fetch_hal0_composite_models` directly — it already has a 5s TTL, which is
exactly the "self-healing within a handful of seconds" property this cache was designed for, and it
would make the whole prime-vs-merge question moot for `/v1/models`.

### 2. (blocking) Regression: replace-on-failure blanks the composite catalogue and it does not self-heal

**Codex's second P2 comment is also correct, and the more likely trigger is worse than the one it described.**

`_fetch_hal0_composite_models` converts an `iter_configs()` exception into `cfgs = []`
(`src/hal0/api/__init__.py:214-218`), caches that empty result under the 5s TTL, and returns it.
`_prime_hal0_composite_cache` now does `model_cache["hal0"] = list(models)` unconditionally — so a
transient enumeration failure during a ready-event refresh **wipes the bucket**. The old merge made
this failure mode harmless. Reproduced:

```
AFTER TRANSIENT FAILURE: []          # was ['embed-gemma:300m', 'minicpm', 'whisper-v3:turbo']
```

The *partial* case is more likely than a total raise, because `iter_configs` swallows per-slot
`SlotConfigError` and `continue`s (`manager.py:2188-2195`). One slot TOML that is momentarily
unparseable — an editor writing it non-atomically, a half-written config — silently shortens `cfgs`,
and the replace drops that healthy slot's model id:

```
AFTER PARTIAL SKIP: ['m-a']          # 'm-b' dropped because b's TOML failed to parse this once
```

Because `/v1/models` reads the cache rather than the fetch, there is **no self-heal**: the bucket
stays wrong until another ready event or a restart. Visible blast radius is bounded (chat-slot ids
are suppressed from the raw catalog as duplicates of their alias rows, so what actually disappears
is the FLM multiplex tags and the dashboard's synthetic `hal0` tile contents) — but "bounded and
permanent" is the same shape of bug #1837 was.

Fix: distinguish "enumeration failed" from "genuinely empty". Have the fetch raise (or return
`None` / a sentinel) on failure, don't poison the TTL cache with the empty result, and have the
prime leave the previous bucket alone. Note `_refresh_model_cache_on_ready` already wraps the prime
in a `try/except` that logs `model_cache.hal0_composite_refresh_failed` — it just never fires,
because the fetch swallows the error before it can.

### 3. (nit) The FLM derivation is now duplicated, and the only tests cover the dead copy

The `is_flm` / `npu_table` / `defaults` block in `_fetch_hal0_composite_models` (lines 245-261) is a
verbatim copy of the one in `_seed_multiplex_models` (lines 748-765). By the PR's own account the
seed copy is now a no-op on the startup path — yet `tests/api/test_seed_multiplex_models.py` (the
entire 17-case suite that guards this logic) exercises **only that dead copy**. Nothing asserts that
the live path — `_prime_hal0_composite_cache` → `_fetch_hal0_composite_models` — emits
`embed-gemma:300m` / `whisper-v3:turbo`.

That's a false-confidence trap with a specific failure signature: if the new copy drifts (a renamed
config key, say), the tags would still be present at startup — `_seed_multiplex_models` backfills at
`__init__.py:1290` — and then **silently vanish on the first slot-ready event**, with a fully green
test suite. Either extract the derivation into one shared helper both call, or add a prime-path
case asserting the tags. I verified by hand that the copy is currently correct for both config
shapes, so this is a maintenance hazard rather than a live bug.

---

### Codex bot triage

Both inline comments were real and both were left unaddressed at merge. Neither is a false
positive — I confirmed #1 by reading the delete path and the event filter, and #2 by executing it.
The bot's silence-is-not-approval caveat cuts the other way here: it spoke, and it was right.

### Suggested follow-up

One issue covering (1) and (2) — they're the same seam and the same fix session. #1837 should not
be counted as closed in `tests/release-validation/regressions.yaml` until the delete path evicts;
its regression probe as written would pass today for the wrong reason.

Issue #1424, opened 2026-07-30
fix(slots): map a failed systemctl restart to a typed error instead of 500 system.internal (raw CalledProcessError)


thinmintdev:
## Summary

Three defects on the slot lifecycle write path, all reproduced live on lxc105 (git a0487038) against the `utility` slot:

1. **`POST /api/slots/{name}/load` and `/restart` return `HTTP 200` with `state: "offline"` and an empty `metadata.message`.** The caller has no way to tell a successful load from a failed one except by re-reading the state.
2. **`systemctl restart` failing surfaces as `HTTP 500 {"code":"system.internal"}`** — the `CalledProcessError` from the sudo seam escapes to Starlette's error middleware instead of being mapped to a typed 409/503 with the systemd reason.
3. **No `reset-failed` before `restart` on the lazy-load path.** The quadlet ships `StartLimitIntervalSec=300` / `StartLimitBurst=5`; a slot that gets re-loaded at request cadence trips `start-limit-hit`, after which every `restart` returns 1 for up to five minutes and the slot is unloadable via the API.

### (1) 200 with state=offline

```
$ time curl -sS -XPOST http://localhost:8080/api/slots/utility/load
{"name":"utility","state":"offline","port":8081,"model_id":"fastcontext-4b-rocmfp4",
 "backend":"vulkan","metadata":{"updated_at":1785414651.17,"message":"","adopted":true,
 "backend":"vulkan","health_ok":false,"provider":"llama-server"},...}
HTTP 200
real 0m21.909s
```

Same for restart:

```
$ curl -sS -XPOST http://localhost:8080/api/slots/utility/restart -w ' HTTP:%{http_code} time:%{time_total}\n'
{... "state":"offline" ... "health_ok":false, "message":"" ...} HTTP:200 time:71.558
```

### (2) 500 on a failed restart

```
$ curl -sS -XPOST http://localhost:8080/api/slots/utility/load -w '\nHTTP:%{http_code}\n'
{"error":{"code":"system.internal","message":"internal server error","details":{}}}
HTTP:500
```

journal:

```
[error] hal0.unhandled  error="Command '['sudo', '-n', '/usr/lib/hal0/bin/hal0-systemctl',
                                'restart', '15']' returned non-zero exit status 1."
╭─ Traceback (most recent call last) ─╮
│ starlette/middleware/errors.py:165  │  exc = CalledProcessError(1, ['sudo', '-n', …
```

### (3) start-limit wedge

```
[...]

PR #1851, opened 2026-08-12
fix(cli): give slot lifecycle verbs a client timeout with server headroom

### LINKED ISSUE #1832: cli(slots): every mutating slot verb caps its HTTP read at 10s against a 180s server budget — load/unload/restart print ReadTimeout and exit 1 on success
## What happens

Every mutating `hal0 slot` verb (`load`, `unload`, `restart`, and by shared code path `swap`/`create`/`rename`) issues its POST through `api_post`, whose signature default is `timeout: float = 10.0` (`src/hal0/cli/_shared.py:169-172` → `_api_request` → `httpx.Client(timeout=10.0)`). There is no `--timeout` flag, no `--wait`, and no env override (grep for `HAL0_API_TIMEOUT`/`API_TIMEOUT` in `src/` returns nothing).

The server-side handler blocks for the whole operation: `POST /{name}/load` awaits `sm.load()` → `_await_ready()` → `wait_ready()`, which polls `/health` for up to `_HEALTH_TIMEOUT_S = 180.0` (`src/hal0/providers/container.py:589`).

**10 s client cap against a 180 s server budget — an 18x mismatch.** Any slot lifecycle operation over ~10 s prints a ReadTimeout and exits 1 while the operation continues server-side and completes normally. The CLI emits no hint that the operation is still running and no instruction to poll; the failure is indistinguishable from a genuine failure timeout.

## Repro (ct152, hal0 1.0.0rc5)

```
$ hal0 slot unload brain >/dev/null 2>&1; sleep 5; date -u +%H:%M:%S; time hal0 slot load brain; echo "exit=$?"
02:56:38
Error: POST http://127.0.0.1:8080/api/slots/brain/load failed: ReadTimeout: timed out
exit=1
real    0m10.458s
```
Polling afterwards: `warming` 02:56:55 → `ready` at **02:57:51** (~73 s after the command); `curl -o /dev/null -w %{http_code} http://127.0.0.1:8089/health` → `200`. **The load fully succeeded.**

`restart` hits the identical cap (`real 0m10.442s`, slot converged `serving` at ~60 s), and so does `unload`. 10.44/10.46 s = the 10.0 s httpx read timeout plus ~0.4 s CLI startup — not load-dependent jitter.

## What should happen

`hal0 slot load` waits for the operation, as `docs/guides/manage-slots.mdx` states verbatim: "Loads the slot's container and waits for it to go healthy." It cannot; it waits 10 s.

## Suggested fix

Pass an explicit long timeout at the slot lifecycle call sites (`src/hal0/cli/slot_commands.py:263/281/297`, and `319/351/512` share the helper), or better, give `api_post` a `timeout=None` option for lifecycle verbs and have `slot load/restart/unload` poll `/api/slots/<name>` with a progress line until terminal state. **Fixing only `load` leaves `unload` and `restart` broken.**

Whoever fixes #1424 should take this at the same time — the two produce byte-identical user-visible strings today, and until this is fixed an operator cannot tell #1424's real failure from this false one.

## Distinct from #1424

#1424 is three server-side semantics defects (load/restart returning 200 with `state: offline`, `CalledProcessError` escaping as 500, no `reset-failed` before restart); its subject is genuine failures being undiagnosable. This is the inverse — a **success** reported as a client-side timeout — and its cause is a client-side constant #1424 never mentions. No other open issue matches (`gh issue list --search "ReadTimeout"` → only closed #1224/#1791/#420).

## It corrupts hal0's own tooling

- `scripts/release-test.sh:178,197,212` gates pass/fail on `ssh_exec "${REMOTE_HAL0_BIN} slot load ${SLOT}"`'s exit status — the project's release smoke test reports a failed backend row for a load that in fact succeeded.
- `installer/install.sh:2006` tells the operator to run `hal0 model pull <id> && hal0 slot load agent --model <id>` — a `&&` chain followed by a command that exits 1 on success.
- `src/hal0/api/routes/slots.py:1245`'s own docstring reasons about the operator "wait[ing] out the 180s health timeout" — the blocking endpoint assumes a patient caller, and the CLI is not one.

## Environment

hal0 1.0.0rc5, ct152 (`10.0.1.152`), fresh install, CPU-only unprivileged LXC. brain slot found `serving` and left `serving`.

## Severity — major

Every scripted or human slot lifecycle operation on a non-trivial model reports failure on success.

## Other hardware

Yes, a GPU box hits this, and not speculatively:

- ct105-prod (`10.0.1.142`, same 1.0.0rc5) carries the identical constant: `grep -n "def api_post" -A 4 /usr/lib/hal0/venv/lib/python3*/site-packages/hal0/cli/_shared.py` → `timeout: float = 10.0`.
- #1424's own repro was captured on the GPU box and records `POST /api/slots/utility/load` → `real 0m21.909s` and `restart` → `time:71.558` — 2–7x over the cap, on GPU hardware, with a 4B model. ct105's real slots are 27B/35B and load slower still.
- Privileged vs unprivileged, fresh vs upgraded: irrelevant, same client constant.

The only configuration that escapes it is one where every lifecycle op finishes under ~10 s (tiny models, runner image local, gguf warm in page cache). CPU-only ct152 is a convenient place to observe this, not a precondition.

Not adjudicated anywhere: no ADR covers it, no `known-issues.yaml` entry matches.

---
Found by the 1.0.0-rc.5 release-validation run. Full report: `tests/release-validation/reports/v1.0.0-rc.5.md` (branch `h0/rc5-release-testing`). Finding key: `slot-load-readtimeout-on-success`.

thinmintdev:
## What

`hal0 slot load/unload/restart/swap` now pass an explicit
`SLOT_LIFECYCLE_TIMEOUT_S = 220.0` client-side read timeout at their
`api_post` call sites, instead of falling through to `api_post`'s
`timeout: float = 10.0` default.

## Why

Every one of these verbs blocks the HTTP response server-side until the
slot's state machine converges:

- `load`: `SlotManager.load()` → `_await_ready()` polls `/health` for up
  to `providers.container._HEALTH_TIMEOUT_S = 180.0`.
- `unload`: bounded terminate (`_terminate_timeout_s`).
- `restart`: unload then a full load — worse than a single load.
- `swap`: unloads the live slot then loads the new model — same shape
  as restart.

`api_post`'s default 10s read timeout is an 18x mismatch against the
180s server budget alone, and worse for restart/swap. Any lifecycle op
over ~10s raised `httpx.ReadTimeout`, printed `Error: ... ReadTimeout:
timed out`, and exited 1 — while the operation continued unattended
server-side and completed successfully. This is a **success reported
as a failure**, not a genuine timeout: confirmed on ct152 (CPU-only)
where `load` converged to `ready` ~73s after the CLI had already
printed the error and exited, and on the GPU box (#1424's own repro)
where `load`/`restart` took 22–72s.

This corrupts scripted callers too:
`scripts/release-test.sh` gates pass/fail on the exit code, and
`installer/install.sh`'s documented `hal0 model pull <id> && hal0 slot
load agent --model <id>` chain fails on its own success.

`rename` (`api_post(.../rename)`) and `create` (`api_post("/api/slots")`)
were deliberately left at the 10s default — neither blocks on the
health poll server-side (`create_slot` only writes config; the caller
follows up with a separate `load`), so raising their timeout would just
mask a genuinely slow/broken request.

## How verified

- Added `tests/cli/test_slot_commands.py`: monkeypatches `api_post` to
  capture call kwargs and asserts each of load/unload/restart/swap
  passes `timeout >= 180.0`. Confirmed the test fails against
  unpatched `slot_commands.py` (no `timeout` kwarg captured) before the
  fix, and passes after.
- `uv run pytest tests/cli/ -q` → 721 passed.
- `uv run ruff check src tests` / `ruff format --check src tests` → clean.

## Out of scope

- #1424 (server-side load/restart semantics — 200 with `state: offline`,
  `CalledProcessError` escaping as 500, no `reset-failed` before
  restart) is a separate, distinct defect per the issue.
- No `--wait`/progress-poll flag or `HAL0_API_TIMEOUT` env override was
  added — the issue's "better" option was left for a follow-up if
  wanted; a bounded, generous client timeout fully resolves the
  reported symptom.

Closes #1832

### DIFF
--- a/src/hal0/cli/slot_commands.py
+++ b/src/hal0/cli/slot_commands.py
@@ -40,6 +40,18 @@ from hal0.hardware.stats import SLOT_PORT_RANGE_END, SLOT_PORT_RANGE_START
 app = typer.Typer(help="Manage inference slots.")
 console = Console()
 
+# The server-side state machine blocks the HTTP response until the slot
+# converges: SlotManager.load -> _await_ready polls /health for up to
+# providers.container._HEALTH_TIMEOUT_S (180s). restart and swap both
+# unload the live slot before loading the replacement, so their worst-case
+# budget stretches past a single load's. api_post's default 10s read
+# timeout is far under any of these budgets — every mutating lifecycle
+# verb would raise ReadTimeout and exit 1 on an operation that in fact
+# succeeded server-side (#1832). Give these call sites a client timeout
+# with real headroom over the server's own worst case instead of the
+# blanket default.
+SLOT_LIFECYCLE_TIMEOUT_S = 220.0
+
 
 class SlotProvider(StrEnum):
     """Providers valid for a slot (mirrors PLAN.md §1 provider list).
@@ -260,7 +272,7 @@ def slot_load(
         raise typer.Exit(1)
     try:
         body = {"model_id": model} if model else {}
-        snap = api_post(f"/api/slots/{name}/load", json=body)
+        snap = api_post(f"/api/slots/{name}/load", json=body, timeout=SLOT_LIFECYCLE_TIMEOUT_S)
     except CliApiError as exc:
         die(str(exc))
         return
@@ -278,7 +290,7 @@ def slot_unload(
     if _api_unreachable(url):
         raise typer.Exit(1)
     try:
-        snap = api_post(f"/api/slots/{name}/unload")
+        snap = api_post(f"/api/slots/{name}/unload", timeout=SLOT_LIFECYCLE_TIMEOUT_S)
     except CliApiError as exc:
         die(str(exc))
         return
@@ -294,7 +306,7 @@ def slot_restart(
     if _api_unreachable(url):
         raise typer.Exit(1)
     try:
-        snap = api_post(f"/api/slots/{name}/restart")
+        snap = api_post(f"/api/slots/{name}/restart", timeout=SLOT_LIFECYCLE_TIMEOUT_S)
     except CliApiError as exc:
         die(str(exc))
         return
@@ -348,7 +360,9 @@ def slot_swap(
     if _api_unreachable(url):
         raise typer.Exit(1)
     try:
-        snap = api_post(f"/api/slots/{name}/swap", json={"model_id": model})
+        snap = api_post(
+            f"/api/slots/{name}/swap", json={"model_id": model}, timeout=SLOT_LIFECYCLE_TIMEOUT_S
+        )
     except CliApiError as exc:
         die(str(exc))
         return
--- a/tests/cli/test_slot_commands.py
+++ b/tests/cli/test_slot_commands.py
@@ -1,0 +1,81 @@
+"""Tests for the mutating ``hal0 slot`` verbs' HTTP timeout (#1832).
+
+The server-side lifecycle handlers (load/unload/restart/swap) block the
+HTTP response until the slot's state machine converges — up to 180s for
+a cold load (``providers.container._HEALTH_TIMEOUT_S``), and longer still
+for restart/swap, which unload then load. ``api_post``'s default read
+timeout is 10s, an 18x mismatch: any lifecycle op over ~10s used to raise
+``ReadTimeout`` and exit 1 while the operation continued and completed
+successfully server-side. These call sites must pass an explicit timeout
+with headroom over the server's worst-case budget.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+from typer.testing import CliRunner
+
+from hal0.cli import slot_commands
+
+runner = CliRunner()
+
+
+@pytest.fixture
+def captured(monkeypatch: pytest.MonkeyPatch) -> dict[str, Any]:
+    captured: dict[str, Any] = {}
+
+    def fake_unreachable(_url: str) -> bool:
+        return False
+
+    def fake_post(path: str, *, json: dict[str, Any] | None = None, **kw: Any) -> dict[str, Any]:
+        captured["method"] = "POST"
+        captured["path"] = path
+        captured["body"] = json or {}
+        captured["kwargs"] = kw
+        return {"state": "serving", "model_id": (json or {}).get("model_id")}
+
+    monkeypatch.setattr(slot_commands, "_api_unreachable", fake_unreachable)
+    monkeypatch.setattr(slot_commands, "api_post", fake_post)
+    return captured
+
+
+def _timeout_ge_server_budget(kwargs: dict[str, Any]) -> None:
+    """The CLI's client-side read timeout must clear the server's own
+    180s health-poll budget with real headroom — not just barely exceed
+    the old 10s default."""
+    assert "timeout" in kwargs, "lifecycle call site passed no explicit timeout kwarg"
+    assert kwargs["timeout"] >= 180.0, (
+        f"timeout={kwargs['timeout']} is under the server's 180s health-poll budget"
+    )
+
+
+def test_slot_load_passes_lifecycle_timeout(captured: dict[str, Any]) -> None:
+    result = runner.invoke(slot_commands.app, ["load", "primary"])
+    assert result.exit_code == 0, result.output
+    assert captured["path"] == "/api/slots/primary/load"
+    _timeout_ge_server_budget(captured["kwargs"])
+
+
+def test_slot_unload_passes_lifecycle_timeout(captured: dict[str, Any]) -> None:
+    result = runner.invoke(slot_commands.app, ["unload", "primary"])
+    assert result.exit_code == 0, result.output
+    assert captured["path"] == "/api/slots/primary/unload"
+    _timeout_ge_server_budget(captured["kwargs"])
+
+
+def test_slot_restart_passes_lifecycle_timeout(captured: dict[str, Any]) -> None:
+    result = runner.invoke(slot_commands.app, ["restart", "primary"])
+    assert result.exit_code == 0, result.output
+    assert captured["path"] == "/api/slots/primary/restart"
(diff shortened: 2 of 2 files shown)


thinmintdev (commented):
## Review — PR #1851 (post-merge)

Verified against `github/main` @ `9c57a83a`, plus a live run of the new test against unfixed code.

### What holds up

- **It does fix #1832's repro.** The four blocking verbs (`load`/`unload`/`restart`/`swap`) now pass an explicit timeout; ct152's measured 73s load and ~60s restart both land well inside 220s.
- **The test is genuinely red without the fix.** I ran the merged `tests/cli/test_slot_commands.py` against a pre-fix `slot_commands.py`: all four fail on `AssertionError: lifecycle call site passed no explicit timeout kwarg` (`kwargs = {}`). Not a tautological test.
- **`create` / `rename` correctly left at 10s.** Confirmed `create_slot` (`src/hal0/api/routes/slots.py`) only writes TOML/drop-in/env/state and explicitly does not start the slot; `rename` likewise never touches the health poll. The PR's stated rationale is accurate.
- **Scope is clean.** Two files, no CHANGELOG hunk, no drive-by refactor, test-first commit split.

### 1. (blocking) 220s does not actually clear the server's worst case — Codex P1 #1 is correct

The constant's comment claims headroom "over the server's own worst case". The arithmetic does not support that:

- `SlotManager.restart` (`src/hal0/slots/manager.py:1813`) is `await self.unload(name)` then `await self.load(name)`. `unload` is bounded by `_terminate_timeout_s = 30.0` (`manager.py:296`), `load` by `_HEALTH_TIMEOUT_S = 180.0` (`providers/container.py:589`). Worst case is **210s before any overhead** — 220s is ~4.7% of margin, not headroom.
- With the default `preload_evict_enabled`, `load` first runs `preload_evict.admit`, which executes `await host.unload(candidate.name)` **sequentially per evicted candidate** inside the load path. Two evicted slots = +60s, i.e. **>240s for a plain `load`**, before restart/swap add their own unload.

So on a memory-tight box — precisely where restart/swap are slowest — the ReadTimeout-on-success this PR exists to kill still fires. Symptom survives; the next validation run re-finds it. This wants either a value derived from the server constants (e.g. `_HEALTH_TIMEOUT_S + k*_terminate_timeout_s`) or the issue's own suggested `timeout=None` on these blocking endpoints.

### 2. (blocking) `hal0 model run` still carries the identical defect — Codex P1 #2 is correct

`src/hal0/cli/model_commands.py:586` posts to the same blocking `/api/slots/{target}/load` at `api_post`'s 10s default. It is worse than the verbs this PR fixed:

- It is a **documented** path — `docs/guides/pull-and-register-models.mdx:211`: `hal0 model run qwen3-4b --slot agent --timeout 120`.
- The command advertises a `--timeout` contract for its readiness poll, and the 10s POST silently overrides it: control never reaches the poll loop.
- On the false failure it prints `Check compatibility with: hal0 slot show <target>` — actively misleading diagnostics on a load that succeeded.

#1832 says in bold "Fixing only `load` leaves `unload` and `restart` broken". The same logic applies here: the defect is a client constant, and one blocking call site was left holding it. `tests/cli/test_model_first_run_commands.py::test_model_run_loads_and_waits_ready` already stubs this call, so the fix is a one-line change plus one assertion.

### 3. (blocking) `hal0 update --restart-slots` — same defect, worst instance, not flagged by anyone

`src/hal0/cli/update_commands.py:477` posts `/api/updates/restart-slots` at the 10s default. `restart_drifted_slots` (`src/hal0/api/routes/updater.py:1100`) loops `await sm.restart(name)` over **every** drifted slot in one request. Server budget is N × 210s against a 10s client cap — the CLI reports failure on a sweep that goes on to bounce every slot unattended, which is materially worse than the reported bug.

### 4. (nit) Non-CLI clients of the same blocking endpoints

`src/hal0/mcp/admin.py:1565` `_call_rest(timeout_s: float = 30.0)` serves the `slot_load` / `slot_restart` / `model_swap` MCP tools, with no per-tool override. Agents driving slot lifecycle get the same success-reported-as-failure at 30s. Out of this PR's stated scope; worth a follow-up so the fix isn't CLI-only.

Taken together, 2–4 are why I read this as a call-site patch rather than a root-cause fix: the cause is "generic short client defaults applied to endpoints that block on the slot state machine", and it is still present at three of the seven call sites.

### 5. (nit) `timeout=220.0` widens connect/write/pool, not just read

`_api_request` does `httpx.Client(timeout=timeout)`, which sets all four httpx timeouts. `_api_unreachable`'s 1s HEAD preflight covers a dead daemon, but a half-open or wedged daemon now stalls the CLI for 220s with no output. `scripts/release-test.sh` has no per-command timeout (only `ConnectTimeout=10` on ssh), so a genuinely wedged slot now stalls the release smoke test ~220s per op. `httpx.Timeout(220.0, connect=5.0)` would keep the read budget without widening connect.

### 6. (nit) A ~4-minute silent wait has no progress output

Deferring the poll-with-progress option is disclosed in the PR body and is a defensible call, but the failure mode it leaves is "operator assumes the CLI hung and Ctrl-C's a load that was 150s into converging". Worth the follow-up the body mentions.

### Codex triage

Both bot comments verified against the source and **upheld** (P1 #1 → finding 1, P1 #2 → finding 2). Neither was addressed before merge. The bot's top-level review body was the boilerplate-only variant, so its non-comment on everything else is not a signal either way.

**Verdict: changes-requested** — already merged, so this converts to required follow-up work. Findings 1–3 all reproduce #1832's exact user-visible symptom on paths this PR left untouched or under-budgeted.

thinmintdev:
Filed the verified residuals from the review above (findings 1–4, including both upheld Codex P1s) as #1857 so they aren't lost now that this is merged.

Issue #1827, opened 2026-08-11
hermes: fresh install cannot chat — the installer-pulled brain model resolves to a 32,768 ctx, below Hermes' hard 64,000 floor


thinmintdev:
## What happens

On a **fresh 1.0.0-rc.5 install** the bundled Hermes agent refuses to start with the shipped config:

```
# sudo -u hal0 HOME=/var/lib/hal0 hermes -z "Reply with exactly: HERMES-RC-OK"
hermes -z: agent failed: Model hal0/agent has a context window of 32,768 tokens,
which is below the minimum 64,000 required by Hermes Agent.
real 0m1.623s
```

Same error for `-m hal0/brain` and `-m hal0/utility`. A bogus model id gives HTTP 404 instead, which proves the 32,768 is a real resolved value, not a generic fallback.

## What should happen

`hermes -z "..."` on a fresh install reaches the model and answers.

## Root cause (not the #1788/#1802 advertising clamp — that clamp is honest)

`_CTX_DENSE_CAP = 32768` in `src/hal0/providers/container.py:1199`. A model with no explicit `defaults.context_size` resolves to `min(native, 32768)`, so the installer-pulled brain row (curated `context_length=131072`, `defaults.context_size=None`) both launches **and** advertises 32768 even though `installer/etc-hal0/slots/{brain,agent,utility}.toml` all set `context_size = 65536`.

Nothing on the fresh-install path stamps a model-level context size — `registry/pull`, `hardware/recommend._resolve_primary_ctx` (which itself caps dense at 32768, `src/hal0/hardware/recommend.py:58`), `install/orchestrate` — and the shipped `/var/lib/hal0/.hermes/config.yaml` sets no `model.context_length`.

Hermes' floor is an upstream hard gate:
- `…/venvs/hermes/lib/python3.12/site-packages/agent/model_metadata.py:185` → `MINIMUM_CONTEXT_LENGTH = 64_000`
- `…/agent/agent_init.py:1842` raises when `_ctx < MINIMUM_CONTEXT_LENGTH`

rc.5 did not create the mis-sizing, it stopped hiding it: before #1802 (`8f18f2cf`, tagged v1.0.0-rc.5 only) `/v1/models` echoed the raw 65536 slot ceiling, so Hermes' gate passed while llama-server was already running `--ctx-size 32768`.

## Decisive evidence

```
# curl -s http://127.0.0.1:8080/v1/models       -> brain 32768, utility 32768; no `agent` entry
[...]

Issue #1617, opened 2026-08-07
fresh install: hindsight extraction always exceeds the seeded model's 8192-ctx fallback, so every retain fails out of the box


thinmintdev:
Found on a pristine fresh-install test (CT151, ubuntu-26.04, main + #1616). The fresh box serves exactly one model — the seeded `hal0-brain-sft-q8-rocmfpx` — which ships **no context metadata and no defaults.context_size**, so `_resolve_context_size` lands on the 8192 safe fallback. Hindsight's extraction prompts exceed that, so **every retain fails** (`Context size has been exceeded`, HTTP 500 from the slot) and the failed-operations counter climbs from the first hermes auto-retain onward — silent, because retain is async.

Live remediation that proved the diagnosis: `PUT /api/models/hal0-brain-sft-q8-rocmfpx {"defaults":{"context_size":32768}}` + slot restart, then retrying the failed operations via the new `memory_operation_retry` MCP tool → all retains completed.

Suggestions:
1. The seeded brain model's registry entry should declare `defaults.context_size` explicitly (it is the extraction fallback target on a fresh box).
2. Doctor row: assert the effective ctx of whatever model serves `hal0/utility` (or its fallback) is ≥ a floor hindsight extraction needs (~16k), FAIL otherwise — same silent-failure family as #1543/#1613.
3. Upstream (hindsight): extraction chunking could target the serving model's advertised ctx instead of a fixed chunk size.

Refs #1543, #1613.

Issue #1108, opened 2026-07-06
installer-setup WS-E: safe slot activation — enable-on-pull-success + clamp context to hw budget


thinmintdev:
Part of the guided **Hal0 Installer Setup** redesign (Wave 2 / WS-E, decision Q7). Fixes the `oom` artifact.

## What to build
When the guided step assigns a model, two hazards: (a) slots are seeded `enabled=true` while the pull is queued separately → the slot can start before the model exists; (b) `context_size` trusts `curated.context_length` rather than a hardware budget → OOM on first warm. Create the slot disabled, flip `enabled=true` only on pull completion, and clamp `context_size` to a computed VRAM/RAM-aware budget.

## Acceptance criteria
- [ ] Slot created disabled; enabled only after its pull completes (no start-before-model race)
- [ ] A failed pull leaves a clearly-marked disabled slot (no crash-loop)
- [ ] `context_size` clamped to a hardware budget; first-warm OOM no longer reproduces
- [ ] Tests cover the enable-on-success transition + the clamp

## Blocked by
- #1097 (hardware.json budget for the context clamp)

## Answer-file coverage (added 2026-07-05)

Part of the `hal0-setup.yaml` headless answer file (core loader #1115, spec `handoffs/hal0-setup-answers-spec-2026-07-05.md`). The loader warn-ignores this key until wired here:

- [ ] `slots[].context_size` and `slots[].enabled_on_pull` (spec §5) are honored from the answer file — `auto` clamps to the hw budget; explicit values pass through.

PR #1852, opened 2026-08-12
fix(providers): honor a configured slot ceiling over the blanket dense cap

thinmintdev:
## What changed

`providers.container._resolve_context_size` no longer applies the blanket
`_CTX_DENSE_CAP = 32768` *underneath* a slot's on-disk `[model].context_size`.
For the DERIVED window path (the model declares no `defaults.context_size`, so
the window comes from the GGUF arch max / curated catalogue) the cap is now the
slot's own configured ceiling when one is written — `_dense_cap_for()`.

Also: a garbage hand-edited ceiling (`context_size = "64k"`) is now treated as
absent rather than taking the launch path out with a `ValueError` — the slot
table reaches this resolver as a raw dict with no schema validation in between,
and I now read it one step earlier. An *implausible* ceiling is screened the
same way, on the same `[_CTX_MIN_PLAUSIBLE, _CTX_MAX_PLAUSIBLE]` range the model
path uses: now that the ceiling REPLACES the dense cap on the derived path, a
hand-written (or pre-screen, still-readable) `99999999` would hand a
262144-native model exactly the impractical KV cache `_CTX_DENSE_CAP` exists to
prevent (#1108's first-warm OOM), and `0`/`-1` are a slot that never warms
rather than a ceiling of zero.

### Second commit — the advertised window has to reach the same backfill

The read-only surfaces Hermes actually gates on (`/v1/models`,
`api.__init__._slot_ctx_size`, the SlotView aggregator) go through
`resolve_effective_context_size`, which builds its `model_info` from
`entry.model_dump()`. `_model_key` is **not** a field on
`registry.model.Model` — only the launch path stamps it — so `_native_ctx`
reached the read-only path with no model id and the #1617 curated backfill
could never fire there. Measured on this branch before the second commit, on
ct152 against its real registry entry with `metadata` emptied to the
older-build shape: **advertised 8192, served 65536**. Hermes would still have
refused, just with a different number. Fixed by stamping `_model_key` in the
wrapper (and letting `_native_ctx` fall back to the `id` key a `model_dump()`
does carry).

The `slot.context_size_no_longer_overrides` warning is also now
derived-path-aware. On that path the ceiling already won as far as it can and
the remaining limit is the model's native window; the old hint ("set the window
on the model to restore it") invited the operator to write a
`defaults.context_size` above what the GGUF actually supports, which is not
dense-capped.

## Why (root cause)

`Refs #1827` (deliberately NOT `Closes` — see "What this does not fix") — on a fresh 1.0.0-rc.5 install the bundled Hermes agent could
not start at all:

```
Model hal0/agent has a context window of 32,768 tokens, which is below the
minimum 64,000 required by Hermes Agent.
```

The installer-pulled brain model carries no `defaults.context_size`, so its
window came from the derived path: `min(native 131072, 32768)` = 32768. The
shipped seeds `installer/etc-hal0/slots/{agent,brain,utility}.toml` all ask for
`context_size = 65536`, but under the v1.0 ownership flip the slot value is a
ceiling that can only clamp DOWN — so 65536 could never be reached and all
three slots launched (and, after #1802's honest advertising, reported) 32768.
That is below Hermes' hard `MINIMUM_CONTEXT_LENGTH = 64_000`, which raises
rather than degrades.

The root defect is that `_CTX_DENSE_CAP` is an **unconfigured-slot** guard —
it exists so a slot nobody tuned can't request a 262144-token KV cache — but it
was being applied to *configured* slots too. A ceiling on disk is written by the
installer's hardware budget (`install.orchestrate._clamp_context_size`), by a
shipped seed, or by an operator; it is by definition a configuration signal.
The shipped configuration and the resolver disagreed, and the resolver silently
won. This is hardware-independent, so a fresh GPU install hit it identically.

Fixing it at the resolver rather than by stamping `defaults.context_size` on the
pulled model means it self-heals on upgraded boxes too, with no migration step,
and it does not require lying to Hermes: llama-server actually gets
`--ctx-size 65536`, so what `/v1/models` advertises is what is served.

## Invariants preserved

- Never above the model's native window (a big ceiling can't conjure context).
- Never above the ceiling itself.
- A slot with **no** ceiling on disk is unchanged — still `min(native, 32768)`,
  which is what `slots.capacity._DEFAULT_CTX_TOKENS` budgets against.
- A ceiling at or below 32768 still wins outright (#1108's VRAM clamp intact).
- An explicit `defaults.context_size` on the model is still authoritative and
  is still only clamped down by the ceiling, never raised by it.
- The `safe_fallback` 8192 path is unchanged — a ceiling is not evidence about
  the model.

### On-box (ct152, 10.0.1.152, rc.5, restored to its rc.5 baseline afterwards)

Branch `container.py` dropped into `/usr/lib/hal0/venv/.../hal0/providers/`,
`hal0-api` restarted:

| | rc.5 baseline | this branch |
|---|---|---|
| `/v1/models` → `brain`, `utility` | `32768` | **`65536`** |
| wrapper vs the real registry row | `32768` | `65536` |
| wrapper vs the same row with `metadata={}` (older-build shape) | **`8192`** | **`65536`** |
| rendered quadlet after a slot restart | `--ctx-size 32768` | `--ctx-size 65536` |
| llama-server `/props` → `n_ctx` | — | **`65536`** |
| `hermes -z "reply with OK"` | timed out (see below) | **exit 0, answered `OK`** |

So the advertised number, the rendered argv and the window llama-server reports
back all agree at 65536, above Hermes' 64,000 floor, on a real box. The rc.5
baseline `hermes -z` run timed out rather than emitting the context-window
refusal string, so that one cell is not a clean negative control — the refusal
itself is documented on ct151/ct150 in #1827. The box was restored to its rc.5
baseline afterwards (file checksum verified, temp files removed, `utility`
quadlet re-rendered back to `--ctx-size 32768`).

### Tests

- New regression tests, red before the fix and green after:
  - `tests/providers/test_container.py::TestDenseCapDoesNotUndercutAConfiguredSlot`
    (11 cases: the fresh-install shape, end-to-end `--ctx-size` through
    `container_spec`, the #1617 curated-backfill shape, and each invariant above)
  - `tests/providers/test_container.py::TestAdvertisedWindowMatchesTheServedWindow`
    — the second commit's coverage, driven through the PUBLIC wrapper against a
    real pydantic `Model` (no hand-stamped `_model_key`, which is a shape the
    read-only path can never produce), plus an equality test that the advertised
    number and the `--ctx-size` `container_spec` emits are the same number.
    Verified red on the unfixed branch: `assert '65536' == '8192'`.
  - `tests/install/test_static_seeds.py::test_shipped_seed_resolves_above_the_hermes_floor`
    — walks the REAL shipped `agent`/`brain`/`utility` TOMLs through the REAL
    resolver against each `BRAIN_MODEL_IDS` pick, so either half of the contract
    drifting back under 64,000 fails in CI instead of on a box.
- `uv run --extra dev pytest tests/providers tests/install tests/slots tests/api`
  → **3285 passed, 1 skipped** (no hermes
  needed by the new tests).
- `uv run ruff check src tests` → clean; `uv run ruff format --check src tests` → clean.

## What this does NOT fix — why `Refs`, not `Closes`

#1827 must stay open. The issue thread's ct150 comment documents a second,
independent path to the same refusal that this change does nothing for: a box
where the model already carries `defaults.context_size = 96000` and the agent
slot's own ceiling is `4096`, so the effective window is 4096 and Hermes refuses
anyway. A ceiling-aware dense cap cannot help when the ceiling is the thing that
is too low. Behind it sit two defects — in-place upgrades never reconcile a slot
seed the operator never chose, and nothing preflights the anchor window before
handing it to Hermes.

Filed as **#1867**, which #1827 should be closed against together with this PR.

For the same reason the claim that this "self-heals upgraded boxes with no
migration step" is scoped: it does, but only where the existing ceiling is
already >= 64K (or absent). A box carrying an old 4096 ceiling stays broken
until #1867 lands.

## Deliberately out of scope

- **Filed as #1868:** `hardware/recommend._resolve_primary_ctx`'s own `_DENSE_CTX_CAP = 32768` is
  untouched. It writes a *ceiling* for the recommended primary slot, and with
  this change that ceiling is now honored as written — but the Hermes-reachable
  slots are the static seeds, not the primary recommendation. Raising the
  recommender's dense policy is a separate tuning call.
- **Filed as #1868:** static seeds are still copied verbatim by `install.sh` without a
  hardware-budget clamp. A tiny box therefore now gets the 65536 the seed always
  asked for (and that pre-1.0 hal0 actually served) rather than an accidental
  32768. Hardware-clamping the static seeds at install time is a real follow-up,
  but it is a different change with a different blast radius.
- `agent` being absent from `/v1/models` (model-less llm slots are omitted by
  design) is incidental to this issue and left alone.
- No CHANGELOG hunk, per the wave rule (#1545).

### DIFF
--- a/src/hal0/providers/container.py
+++ b/src/hal0/providers/container.py
@@ -1195,6 +1195,12 @@ def _profile_runtime_family(slot_cfg: dict[str, Any]) -> str | None:
 # GGUF arch max, cap dense models so an unconfigured slot can't request an
 # impractically large KV cache, and otherwise use a safe floor. Mirrors
 # hal0.hardware.recommend's installer-side policy.
+#
+# _CTX_DENSE_CAP is an UNCONFIGURED-slot guard, not a hardware fact — see
+# _dense_cap_for. A slot that pins its own [model].context_size is by
+# definition configured, and applying this blanket constant *underneath* that
+# number is what shipped every fresh install at 32768 despite three seeds
+# asking for 65536 (#1827).
 _CTX_SAFE_FALLBACK = 8192
 _CTX_DENSE_CAP = 32768
 
@@ -1300,6 +1306,63 @@ def _native_ctx(model_info: dict[str, Any]) -> int | None:
     return None
 
 
+def _slot_ceiling_int(slot_ceiling: Any) -> int | None:
+    """The slot's ``[model].context_size`` as an int, or None when unusable.
+
+    The slot table reaches :func:`_resolve_context_size` as a raw dict straight
+    off the TOML — no schema validation between disk and here — so a
+    hand-edited ``context_size = "64k"`` used to take the launch path out with a
+    ``ValueError``. Unusable means ABSENT: the slot simply expresses no ceiling,
+    exactly like a slot that never wrote one.
+    """
+    if slot_ceiling is None:
+        return None
+    try:
+        return int(slot_ceiling)
+    except (TypeError, ValueError):
+        log.warning(
+            "slot.context_size_unparseable_ceiling",
+            extra={
+                "raw": repr(slot_ceiling),
+                "hint": "the slot's [model].context_size is not a number; ignoring it "
+                "and letting the model's own window decide. Fix it in the slot drawer.",
+            },
+        )
+        return None
+
+
+def _dense_cap_for(slot_ceiling: int | None) -> int:
+    """The cap to apply to a DERIVED (native/curated) window for this slot.
+
+    :data:`_CTX_DENSE_CAP` exists so an **unconfigured** slot cannot ask for an
+    impractically large KV cache off the back of a 131072/262144-token arch max.
+    It is a blanket constant: hardware-independent, box-independent and —
+    the bug — configuration-independent.
+
+    #1827: a slot with an on-disk ``[model].context_size`` is *configured*. That
+    number was written either by the installer as a hardware budget
+    (:func:`hal0.install.orchestrate._clamp_context_size`), by a shipped seed, or
+    by an operator in the slot drawer. Applying the 32768 constant underneath it
+    meant a fresh install resolved every installer-pulled model to 32768 no
+    matter what the slot asked for — ``installer/etc-hal0/slots/{agent,brain,
+    utility}.toml`` all ask for 65536, and all three silently launched (and,
+    post-#1802, honestly advertised) 32768. That is below the bundled Hermes
+    agent's hard 64,000 floor, so ``hermes -z`` could not start on a box
+    straight out of ``install.sh``.
+
+    So: when a ceiling is on disk, it *is* the cap. The result stays bounded
+    twice over — never above the model's own native window, and never above the
+    ceiling itself (the caller's clamp) — so a slot can only reach a window that
+    BOTH the model supports and the box was explicitly configured for. With no
+    ceiling on disk nothing is configured and the blanket cap still applies,
+    which keeps :data:`hal0.slots.capacity._DEFAULT_CTX_TOKENS` (the
+    unpinned-slot budget estimate) correct as written.
+    """
+    if slot_ceiling is None:
+        return _CTX_DENSE_CAP
+    return max(_CTX_DENSE_CAP, slot_ceiling)
+
+
 def _resolve_context_size(
     slot_ceiling: int | None,
     model_info: dict[str, Any],
@@ -1318,10 +1381,11 @@ def _resolve_context_size(
        but an unparseable or implausible value is ignored rather than launched
        (#1414; see :data:`_CTX_MIN_PLAUSIBLE`).
     2. ``model_info["metadata"]["context_length"]`` — the GGUF-derived native
-       window (:func:`_native_ctx`), dense-capped at :data:`_CTX_DENSE_CAP`.
-       Still a model fact, but a *derived* one, so an explicit choice outranks
-       it — the reverse of the pre-1.0 order, which let a 262144-token arch max
-       shadow a deliberate ``defaults.context_size``.
+       window (:func:`_native_ctx`), dense-capped at :func:`_dense_cap_for`
+       (:data:`_CTX_DENSE_CAP`, or the slot's own configured ceiling when that
+       is higher — #1827). Still a model fact, but a *derived* one, so an
+       explicit choice outranks it — the reverse of the pre-1.0 order, which let
+       a 262144-token arch max shadow a deliberate ``defaults.context_size``.
     3. :data:`_CTX_SAFE_FALLBACK` (8192) — never llama-server's silent 4096.
 
     ``slot_ceiling`` (the on-disk ``[model].context_size``) is NO LONGER an
@@ -1337,29 +1401,37 @@ def _resolve_context_size(
 
     Net effect on a box that already has a slot-level ``context_size`` on disk:
     where the clamp was below the model's window (the installer's case) the
-    effective context is UNCHANGED; where the slot was silently inflating the
-    model's window (the dual-ownership bug) it drops to what the model asked
-    for. The effective context can therefore only stay the same or shrink —
-    never grow — so no box gains a KV cache it cannot hold. Every case where
-    the ceiling actually changes the answer is logged, so the change is visible
-    in ``journalctl -u hal0-api`` rather than silent.
+    effective context is UNCHANGED; where the slot was silently inflating a
+    model's EXPLICIT window (the dual-ownership bug) it drops to what the model
+    asked for. The one case where the ceiling can RAISE the answer is the
+    *derived* window (path 2): there the ceiling replaces the blanket dense cap
+    instead of being applied under it (:func:`_dense_cap_for`, #1827), so the
+    slot gets the window its own config asked for — still bounded by the
+    model's native window, still never above the ceiling, and never reached at
+    all on a slot with no ceiling written. Every case where the ceiling changes
+    the answer is logged, so the change is visible in ``journalctl -u hal0-api``
+    rather than silent.
 
     Guarantees a non-None int.
     """
+    ceiling = _slot_ceiling_int(slot_ceiling)
+
     declared = _model_declared_ctx(model_info)
     if declared is not None:
         resolved, source = declared, "model.defaults.context_size"
     else:
         native = _native_ctx(model_info)
         if native:
-            resolved, source = min(native, _CTX_DENSE_CAP), "model.metadata.context_length"
+            resolved, source = (
+                min(native, _dense_cap_for(ceiling)),
+                "model.metadata.context_length",
+            )
         else:
             resolved, source = _CTX_SAFE_FALLBACK, "safe_fallback"
 
-    if slot_ceiling is None:
+    if ceiling is None:
         return resolved
 
-    ceiling = int(slot_ceiling)
     if ceiling < resolved:
         log.info(
             "slot.context_size_clamped_by_slot",
--- a/src/hal0/providers/container.py
+++ b/src/hal0/providers/container.py
@@ -1293,7 +1293,7 @@ def _native_ctx(model_info: dict[str, Any]) -> int | None:
             return int(md["context_length"])
         except (TypeError, ValueError):
             pass
-    model_id = model_info.get("_model_key")
+    model_id = model_info.get("_model_key") or model_info.get("id")
     if model_id:
         try:
             from hal0.registry.curated import get_curated
@@ -1314,11 +1314,24 @@ def _slot_ceiling_int(slot_ceiling: Any) -> int | None:
     hand-edited ``context_size = "64k"`` used to take the launch path out with a
     ``ValueError``. Unusable means ABSENT: the slot simply expresses no ceiling,
     exactly like a slot that never wrote one.
+
+    Implausible means unusable too, on the same
+    ``[_CTX_MIN_PLAUSIBLE, _CTX_MAX_PLAUSIBLE]`` range the MODEL path screens
+    against (:func:`_model_declared_ctx`). This matters more now than it did
+    when the ceiling could only clamp down: under #1827 the ceiling REPLACES
+    the dense cap on the derived path, so a hand-written ``context_size =
+    99999999`` (or a row persisted before the API's
+    ``model.context_size_out_of_range`` screen existed) would hand a 262144-
+    native model the very impractically-large KV cache :data:`_CTX_DENSE_CAP`
+    exists to prevent — #1108's first-warm OOM. ``0`` and ``-1`` are screened
+    here for the same reason: they are not a ceiling of zero, they are a slot
+    that never wrote one, and passing them through to ``--ctx-size`` is a slot
+    that never warms.
     """
     if slot_ceiling is None:
         return None
     try:
-        return int(slot_ceiling)
+        value = int(slot_ceiling)
     except (TypeError, ValueError):
         log.warning(
             "slot.context_size_unparseable_ceiling",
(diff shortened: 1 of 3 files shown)


thinmintdev:
## Review — fix at the right seam, one verified hole and one closure-hygiene ask

I checked this against #1827's repro rather than the PR description: fetched the branch, ran the
new tests at HEAD (21 + 18 green), then ran them again at the pre-fix commit `533d8ddb` to confirm
they are real regression tests — **8 failures in `TestDenseCapDoesNotUndercutAConfiguredSlot` and
9 in `test_shipped_seed_resolves_above_the_hermes_floor` without the fix**. They are honest tests,
not narration.

The diagnosis holds up. `min(native 131072, _CTX_DENSE_CAP 32768)` with the seed's 65536 sitting
above it as a clamp-down-only ceiling makes 65536 unreachable by construction, which is exactly
what ct151/ct152 observed. Moving the cap to `max(_CTX_DENSE_CAP, ceiling)` on the **derived** path
only is the root-cause seam, not a symptom patch: it is one function, it leaves the explicit
`defaults.context_size` precedence alone, and it needs no migration. Scope is clean — 3 files, no
CHANGELOG hunk, no drive-by refactor, CI green.

Two things I want addressed before merge.

---

### 1. BLOCKING — the surface Hermes actually gates on can't reach the curated backfill, and the new test hides it

`test_curated_backfill_path_also_clears_the_floor` asserts the #1617 shape (registry row with empty
`metadata`, window recovered from the curated catalogue) clears 64,000. It passes only because the
test hand-stamps `_model_key`:

```python
mi = _model_info(_model_key="hal0-brain-sft-q8-rocmfpx", metadata={})
```

The launch path does stamp it (`slots/manager.py:3514`). **`resolve_effective_context_size` — the
read-only wrapper behind `/v1/models`, `_slot_ctx_size` and the SlotView aggregator — does not.**
It builds `model_info` from `entry.model_dump()`, and `_model_key` is not a field on
`registry.model.Model`:

```
['architecture','backends','capabilities','capability_flags','default','defaults','hf_filename',
 'hf_repo','id','license','metadata','mmproj','modalities_override','name','path','quant',
 'size_bytes','tags']
```

Verified on this branch:

```
_model_key present: False | metadata: {}
advertised (empty metadata, ceiling 65536): 8192
advertised (metadata present):             65536
```

So on a box whose brain row carries no `metadata.context_length` — precisely the population #1617's
catalogue fallback was written for — llama-server launches at **65536** while `/v1/models`
advertises **8192**, and Hermes (which gates on the advertised number, that being the whole point of
#1802) still refuses, just with a different number in the error. The new test states that case is
covered; it isn't, on the only surface that matters for #1827.

One line closes it, in `resolve_effective_context_size`:

```python
model_info.setdefault("_model_key", model_id)
```

with a test that goes through `resolve_effective_context_size` (not `_resolve_context_size`) for the
empty-metadata row. If you'd rather not widen the change, then please delete or rename
`test_curated_backfill_path_also_clears_the_floor` so it doesn't claim coverage the product doesn't
have — but the one-liner seems clearly better.

### 2. BLOCKING (cheap) — `Closes #1827` closes an issue with a documented, still-failing repro

The issue thread carries a second reproduction the PR body doesn't mention: on **ct150** the model
already has `defaults.context_size = 96000` and Hermes still refuses, because the *agent slot's own
ceiling is 4096*. That comment states plainly that the ceiling-aware seam "helps fresh installs,
does nothing here". This change does not fix that box, so:

- the PR body's "it also self-heals **upgraded** boxes (the regression path called out in the
  report) with no migration step" is true only when the existing ceiling is already ≥64K. Please
  narrow that sentence — as written it reads as a claim the ct150 comment disproves.
- `Closes #1827` will mark a GA blocker resolved while one of its recorded reproductions still
  fails, and I can't find any open issue covering it (nothing in the current open list, and
  `fix/smoke-chat-preflight-anchor` is #1831, a different anchor problem).

Either downgrade to `Refs #1827` and leave the issue open for the slot-ceiling half, or file the
follow-up first (the issue's seam 2: a preflight that refuses to hand Hermes an anchor below its
floor with an actionable message, plus the "in-place upgrades never reconcile a slot seed the
operator never chose" defect from the same comment) and link it here.

---

### Non-blocking

3. **`_slot_ceiling_int` screens unparseable but not implausible.** The model path has
   `_CTX_MIN_PLAUSIBLE`/`_CTX_MAX_PLAUSIBLE` and the API 400s `model.context_size_out_of_range`;
   the slot ceiling now has equal power to *raise* the derived window and screens neither. Verified
   on this branch with a 262144-native model:

   ```
   ceiling 1000000  -> 262144      (rc.5: 32768)
   ceiling 99999999 -> 262144      and ModelConfig(context_size=99999999) validates — ge=128, no upper bound
   ceiling 0        -> 0           (pre-existing; schema ge=128 only screens NEW writes, legacy TOMLs bypass it)
   ceiling -1       -> -1          (pre-existing)
   ```

   A typo'd or legacy pre-1.0 override ceiling now yields the model's full arch max — the exact
   "impractically large KV cache off a 262144 arch max" the `_CTX_DENSE_CAP` docstring says the
   constant exists to prevent, and #1108's first-warm OOM. You already added the sanitizer; applying
   `_CTX_MAX_PLAUSIBLE` (and treating ≤0 as absent) inside it is a two-line symmetry.

4. **File the two "deliberately out of scope" items as issues.** A squashed PR body is not a
   tracker. The first one is the real memory risk of this change: static seeds are copied verbatim
   with no hardware budget, so every small box's brain slot now warms at 65536 instead of 32768
   — on the same CPU-only LXC class the issue was found on.

5. **No on-box verification for a GA blocker.** The issue explicitly provisioned
   `pct rollback 151 rc5-installed` for verifying this fix. Unit tests can't show that brain
   actually *warms* at 65536 on the 8-core CPU box, that `--ctx-size 65536` reaches the unit, and
   that `hermes -z` answers. Please run that before it goes in the GA candidate — and if it's
   deferred to the next validation sweep, say so in the body.

6. **Nit — the `slot.context_size_no_longer_overrides` warning is now misleading on the derived
   path.** With a ceiling above the model's *native* window (e.g. ceiling 131072, native 40960) it
   still fires "this slot's larger [model].context_size no longer wins. Set the window on the model
   to restore it." But it now *does* win up to native, and an operator following that hint would set
   `defaults.context_size` — which is not dense-capped and plausible to 2**24 — above the model's
   real window. Worth a derived-path-specific message ("clamped to the model's native window").

Nothing from the Codex bot on this PR (no reviews, no review comments) — that's silence, not
approval, so I did not treat it as a signal either way.

Core change is sound and I'd approve it on (1) plus a decision on (2).

thinmintdev:
## Review addressed — both blocking findings confirmed and fixed

Thanks; both blocking findings reproduced exactly as described. Details below,
one section per finding.

---

### [blocking] `resolve_effective_context_size` cannot reach the curated backfill — CONFIRMED, fixed

Reproduced before touching anything, with a real `registry.model.Model`:

```
$ uv run python -c '...'
curated ctx: 131072
'_model_key' in model_dump(): False    'id' in model_dump(): True    metadata: {}
wrapper (read-only surface): 8192
launch-shape (_model_key stamped): 65536
```

So the branch as reviewed advertised 8192 while llama-server launched 65536 —
exactly your reading. `_model_key` is not a field on `Model` (pydantic would not
accept a leading-underscore one either); only `slots/manager.py` and
`_best_effort_model_info` stamp it.

**Fix** (`src/hal0/providers/container.py`): `model_info.setdefault("_model_key", model_id)`
in the wrapper, as you proposed, plus `_native_ctx` now falls back to the `id`
key that a `model_dump()` *does* carry, so a caller that builds its dict some
third way is covered too.

**Test through the wrapper, not the private resolver.** New
`TestAdvertisedWindowMatchesTheServedWindow` in `tests/providers/test_container.py`,
built from a real `Model(id=..., path=...)` — no hand-stamped `_model_key`
anywhere. Proven red against the unfixed `container.py`:

```
FAILED ...::test_curated_backfill_is_reachable_from_the_read_only_surface
FAILED ...::test_advertised_equals_what_llama_server_is_launched_with
E       AssertionError: assert '65536' == '8192'
2 failed, 3 passed
```

The third test in the class is the `advertised == served` equality: one side
through `resolve_effective_context_size`, the other through
`container_spec(cfg, _best_effort_model_info(cfg))`, both driven off the same
registry entry.

You were also right that `test_curated_backfill_path_also_clears_the_floor` was
false coverage. It now builds its dict from a real `Model.model_dump()` and
stamps `_model_key` explicitly with a comment saying it is standing in for what
`slots/manager` does at launch — so it documents the launch shape instead of
pretending to cover the read-only one, which the new class now really does.

---

### [blocking] `Closes #1827` on a blocker with a live repro — CONFIRMED, downgraded + follow-up filed

Re-read the ct150 comment; it is exactly as you describe (`defaults.context_size = 96000`,
agent slot ceiling `4096`, Hermes refuses, and the comment says in terms that
the ceiling-aware seam "does nothing here"). Also confirmed #1831 is a different
anchor defect and no open issue covered this.

- PR body now says **`Refs #1827`**, with a new "What this does NOT fix — why
  `Refs`, not `Closes`" section.
- Filed **#1867** — "an upgraded box keeps an old slot ceiling below Hermes'
  floor, and nothing preflights the anchor window" — covering both halves you
  named: the preflight anchor floor check, and in-place upgrades never
  reconciling a slot seed the operator never chose. Linked from the PR body and
  from #1827.
- The "self-heals upgraded boxes with no migration step" claim is now scoped in
  the body to "only where the existing ceiling is already >= 64K (or absent)".

---

### [nit] `_slot_ceiling_int` screens unparseable but not implausible — agreed, fixed

Verified your numbers on a 262144-native model. `_slot_ceiling_int` now applies
`_CTX_MIN_PLAUSIBLE`/`_CTX_MAX_PLAUSIBLE` and treats a non-plausible value as
absent, with its own `slot.context_size_implausible_ceiling` warning, matching
what `_model_declared_ctx` does on the model side. `0`/`-1` are now screened
too — they are a slot that never warms, not a ceiling.

New `test_implausible_ceiling_falls_back_to_the_blanket_cap`, red on the unfixed
code for all of `[0, -1, 99999999, 2**24 + 1, 127]`.

One correction to the finding: `1000000` is *inside* `_CTX_MAX_PLAUSIBLE`
(`2**24` = 16777216), so it is still honored as a plausible ceiling and clamps
to the model's native `262144`. I dropped it from the parametrize rather than
inventing a tighter bound than the model path uses — if the ceiling should be
screened more aggressively than `defaults.context_size`, that is a policy call
worth its own change, not something to slip in here.

---

### [nit] misleading `slot.context_size_no_longer_overrides` hint — agreed, fixed

The hint is now derived-path-aware. On that path the ceiling has already won as
far as it can and the remaining limit is the model's native window, so it reads
"clamped to the model's native window; this slot's larger `[model].context_size`
cannot conjure context the model does not support." The old text stays for the
explicit-model-window path, where it is still correct. Your point about the old
hint inviting an above-native `defaults.context_size` is the reason — that field
is not dense-capped.

---

### [nit] the two "out of scope" items should be issues — agreed, filed

**#1868** — static seeds copied verbatim with no hardware budget (the memory
risk you flagged: every small box's brain slot now warms at 65536), together
with the `hardware/recommend._DENSE_CTX_CAP` tuning question, since the two are
the same decision. Both bullets in the PR body now point at it.

---

### [nit] no on-box verification for a GA blocker — agreed, ran it

On **ct152** (rc.5), branch `container.py` dropped into the service venv,
`hal0-api` restarted:

| | rc.5 baseline | this branch |
|---|---|---|
| `/v1/models` → `brain`, `utility` | `32768` | **`65536`** |
| wrapper vs the real registry row | `32768` | `65536` |
| wrapper vs the same row with `metadata={}` (older-build shape) | **`8192`** | **`65536`** |
| rendered quadlet after a slot restart | `--ctx-size 32768` | `--ctx-size 65536` |
| llama-server `/props` → `n_ctx` | — | **`65536`** |
| `hermes -z "reply with OK"` | timed out | **exit 0, answered `OK`** |

So `--ctx-size 65536` really does reach the unit, llama-server really does
report `n_ctx 65536` back, and `hermes` really does answer. Being honest about
the one weak cell: the rc.5 baseline `hermes -z` run **timed out rather than
emitting the context-window refusal string**, so it is not a clean negative
control — that box's `utility` slot was still up at the new window at the time.
The refusal itself is documented on ct151 and ct150 in #1827.

ct152 was restored to its rc.5 baseline afterwards: original `container.py` back
in place (md5 verified identical to the backup), backups and temp logs removed,
and the `utility` quadlet re-rendered back to `--ctx-size 32768` by a restart
under rc.5 code.

---

### Checks

- `uv run --extra dev pytest tests/providers tests/install tests/slots tests/api`
  → 3285 passed, 1 skipped
- `uv run ruff check src tests` → All checks passed
- `uv run ruff format --check src tests` → 1130 files already formatted

No CHANGELOG hunk, per the wave rule. Not enabling automerge.

Issue #1787, opened 2026-08-09
fix(slots): profile flags never reach llama-server argv for unstamped models — /v1/embeddings and /v1/rerank return 501 while capability slots report ready


thinmintdev:
Found during v1.0.0-rc.4 fresh-install validation (CT151, CPU-only, Ubuntu 26.04 unprivileged LXC) and cross-checked read-only on CT105 (rocm, rc.3). **This is a GA-blocker: the embed/rerank capability path serves 501s for its sole purpose, on every backend.**

## What happens

Enable the embed capability, load the embed slot (`qwen3-embedding-0-6b-q8-0`), slot reports `ready`, health probe passes. Then:

```
POST /v1/embeddings  →  HTTP 501 {"code":501,"message":"This server does not support embeddings. Start it with `--embeddings`"}
POST /v1/rerank      →  HTTP 501 "This server does not support reranking. Start it with `--reranking`"
```

Both direct (`:8083`/`:8086`) and via the `:8080` gateway. The memory reranker path (`[memory.embedding] rerank_model`) is dead as a consequence.

`podman inspect hal0-slot-embed` Cmd: `--host 0.0.0.0 --port 8083 --model .../Qwen3-Embedding-0.6B-Q8_0.gguf --alias qwen3-embedding-0-6b-q8-0 --ctx-size 4096 --jinja -ngl -1` — no `--embedding`, even though `/api/profiles` shows profile `embedding` flags `--embedding -fa on -b 8192 -ub 8192` and `/etc/hal0/slots/embed.toml` has `profile="embedding"`.

## Root cause

The v1.0.0 flags-ownership redesign: `_llama_argv_segments` deliberately discards profile flags (`del profile_flags, slot_parallel, extra_args`, HAL0-SUNSET comment, `src/hal0/providers/container.py` ~line 883). The only surviving path — the #1636 divergence overlay in `_resolve_llama_scalars` (container.py ~1410–1426) — is gated on the model carrying stamped `defaults.profile` provenance, which **auto-scan, pull, and capability-apply never write** (`/api/models` shows `defaults: null` for every auto-scanned model; `capabilities/orchestrator.py` stamps type/device/provider but never model defaults; the `slot_flags_fold.py` migrator is deploy-window-gated and doesn't run for fresh models).

Not backend-conditional — no CPU/GPU branch in the gate. CT105 (rocm, rc.3) shows the same preconditions and its rerank `resolved_command` also lacks `--reranking`.

[...]

PR #1853, opened 2026-08-12
fix(slots): infer the capability profile a slot's type requires at create (#1830)

### LINKED ISSUE #1830: slots: hal0 slot create and the New-slot modal write no profile key, so a capability slot loads ready and 501s its own endpoint
## What happens

On rc.5, every **explicit slot-creation** path writes a slot TOML with no `profile` key — `hal0 slot create` (no `--profile` option exists on `create` or `edit`) and the dashboard "New slot" modal / raw `POST /api/slots`. The #1787 profile-template fix in `src/hal0/providers/container.py:1637` is gated on a truthy `_cfg_profile`, so such a slot bound to an unstamped model (`defaults: null` — auto-scan / `model add` / pull / curated all register that way) resolves argv **with no mode flag**, loads to `state=ready`, and 501s the one endpoint it exists to serve.

Verified for both capability types.

## Repro

```
hal0 slot create zzsk1787 --type reranking --hardware cpu --model rcrerank --port 8097
cat /etc/hal0/slots/zzsk1787.toml            -> name/type/device/provider/port/autoload + [model]; NO `profile` key
curl -s localhost:8080/api/slots/zzsk1787/resolved -> {"name":"zzsk1787","argv":null,"provenance":[],"removed":0}
hal0 slot load zzsk1787                      -> "Loaded zzsk1787 -> state=ready model=rcrerank"
systemctl cat hal0-slot@zzsk1787 | grep -m1 '^Exec='
   -> ... --alias rcrerank --ctx-size 4096 --jinja --metrics      (no --reranking)
curl -s -w 'http=%{http_code}' localhost:8097/v1/rerank \
  -d '{"model":"rcrerank","query":"what is a cat","documents":["a cat is a small feline","steel manufacturing"]}'
   -> http=501 {"error":{"code":501,"message":"This server does not support reranking. Start it with `--reranking`"}}
```

Embedding is the same:
```
hal0 slot create zzskemb --type embedding --hardware cpu --model rcembed --port 8098 ; hal0 slot load zzskemb -> ready
Exec= ... --alias rcembed --ctx-size 4096 --jinja --metrics      (no --embeddings)
curl localhost:8098/v1/embeddings -d '{"model":"rcembed","input":"hello world"}'
   -> http=501 {... "Start it with `--embeddings`"}
```

Dashboard path (`ui/src/dash/slots/CreateSlotModal.jsx:82` body = `{name,type,runtime,model,device,autoload,priority}` — no profile):
```
curl -X POST localhost:8080/api/slots \
  -d '{"name":"zzskui","type":"reranking","runtime":"container","model":"rcrerank","device":"gpu-rocm","autoload":false,"priority":50}' -> 201
cat /etc/hal0/slots/zzskui.toml -> no `profile` key   (device gpu-rocm, i.e. what a GPU box sends)
```

## Decisive control — single variable, via the UI drawer's own endpoint

```
curl -X PUT localhost:8080/api/slots/zzsk1787/config -d '{"profile":"reranking"}'  -> 200
hal0 slot restart zzsk1787
Exec= ... --ctx-size 4096 --reranking -fa auto -b 8192 -ub 8192 --jinja --metrics
same /v1/rerank curl -> http=200 {"results":[{"index":0,"relevance_score":0.1903},{"index":1,"relevance_score":0.0154}]}
```

The `profile` key is the whole cause.

## What should happen

A slot created with `--type reranking` / `--type embedding` either gets a profile inferred (as `orchestrator._profile_for_fit` already does for capability-apply) or is refused at create time. It must not report `ready` while 501ing its own endpoint.

## Not affected, and explicitly not claimed

The installer-seeded `embed`/`rerank` slots (`installer/etc-hal0/slots/rerank.toml` ships `profile = "reranking"`) and capability-apply-created slots launch correctly, so the changelog's "fresh installs serve every capability" claim holds for the seeded path. This is an **incomplete fix, not a regression** — rc.4 broke the seeded path too, so rc.5 is strictly better.

## Nothing warns

No create-time validation (`_fit_check_warning` only covers device/binary), no doctor check, and `GET /api/slots/<name>/resolved` returns `{"argv":null,"provenance":[]}` for a profile-less slot, so even the auditable argv preview shows nothing. Documented-and-broken: `docs/guides/manage-slots.mdx:53` "Create a slot" example is literally `hal0 slot create embed --type embedding ...`.

Source: `src/hal0/providers/container.py:1637`; `src/hal0/cli/slot_commands.py:403-514` (create body has no profile) and `:519-560` (edit has no `--profile`); `src/hal0/api/routes/slots.py:617` `_normalize_create_body` injects no profile default.

## Environment

hal0 1.0.0rc5, ct152 (`10.0.1.152`), fresh install, CPU-only unprivileged LXC. All throwaway slots deleted afterwards.

## Severity — major

The slot reports `state=ready`, which is the silent-wrongness shape that outranks loud brokenness. The create modal's own copy frames profile as an optional runner/image choice, not as the thing that makes a rerank slot rerank.

## Other hardware

Hardware-independent — a config-write gap, not a hardware probe. Confirmed the create path omits `profile` even when the body carries `device: "gpu-rocm"`, and `hal0 slot create --hardware rocm|vulkan` goes through the identical body builder.

Read-only cross-check on ct105 prod (`10.0.1.142`, also 1.0.0rc5): `/etc/hal0/slots/5.toml` is the `embed` slot, `device = "gpu-rocm"`, `type = "embedding"`, **no `profile` key**, bound to `qwen3-embedding-0-6b-q8-0` whose `hal0 model show` reports `defaults | None` — exactly the failing combination (slot offline, not loaded, prod is read-only). **Upgraded boxes are more exposed than fresh ones**: fresh installs get the profile-bearing seeds, while an upgraded box carries whatever profile-less slots its operator created historically, and the seed loop never back-fills an existing TOML.

---
Found by the 1.0.0-rc.5 release-validation run. Full report: `tests/release-validation/reports/v1.0.0-rc.5.md` (branch `h0/rc5-release-testing`). Finding key: `slot-create-no-profile-501`.

thinmintdev:
## What changed

A capability slot is now created with the runtime profile its **type** requires, and the shared `(capability, device) → profile` rule stops pretending the embed/rerank profiles are GPU-only.

- **`SlotManager.create` infers the type-implied profile** (`src/hal0/slots/manager.py`). This is the one chokepoint every creation path funnels through — `hal0 slot create`, the dashboard New-slot modal, a raw `POST /api/slots`, a stack apply, and the capability orchestrator's `_ensure_slot_exists`. An explicit profile still wins, and so does the model's stamped `defaults.profile` (the existing Q1 adoption directly above); `llm` slots infer nothing.
  - **Correction (review round 2):** the chokepoint only covers a caller that supplies a slot `type`, and `_ensure_slot_exists` did not for the non-trio children. A re-created `rerank` slot landed with no `type`, so it inferred nothing, launched llama-server without `--reranking`, reported `state=ready` and 501'd `/v1/rerank` — the reported bug verbatim on the path this PR advertised as covered (and `doctor profiles` missed it for the same reason). `_ensure_slot_exists` now stamps the type its shipped seed carries for `rerank` → `reranking` and `img` → `image`, via a second map so the FLM-trio discriminator keeps exactly its two members.
- **New `hal0.slots.profile_adopt.type_implied_profile`** — translates slot type → capability and delegates to the existing shared rule `capabilities.profile_fit.profile_name_for_fit`, then vetoes the answer twice: with `profile_fits_slot` (catalog + device_class/backend/slot-type) and with the **runner registry** (`_runtime_family_fits_device`). No fourth copy of the mapping.
  - The runner veto exists because the first round regressed TTS: `tts` resolves `qwen3-tts` for *any* GPU device, but `RUNNER_IMAGES["qwen3tts"]` is ROCm-only, and the 1.0 seed profiles are device-agnostic tunes (`device_class = None`, `backend = None`) so `profile_fits_slot` waved it through. `hal0 slot create voice --type tts` with no `--hardware` on a Vulkan-defaulting box went from a working profile-less Kokoro CPU slot to one pinned to a container its device cannot start. Now nothing is inferred there and the `_spec_provider_for` Kokoro fallback stands. The orchestrator's tts/stt engine stamp routes through the same rule, so it gets the veto too.
- **The CLI's historical `--provider llama-server` default is treated as "unspecified"** by the inference. It made `hal0 slot create --type transcription --hardware cpu` infer nothing (the STT rule rejects providers it has no engine for) while the dashboard modal, which sends no provider, inferred Moonshine.
- **`profile_name_for_fit` and `install.profile_derive.derive_profile`: embed/rerank are no longer device-gated.** Both required `device in {gpu-rocm, gpu-vulkan}` — a leftover from the retired per-backend `embed`/`vulkan-embed` seeds. The 1.0 `embedding`/`reranking` seeds are device-agnostic logical tunes (`device_class = None`, `backend = None`), so a CPU box gets them too. NPU still resolves `flm` (different runtime family).
- **`hal0 slot create --profile` and `hal0 slot edit --profile`** — the operator override, and the repair path for a slot already on disk without one.
- **`hal0 doctor profiles` now flags a profile-less embedding/reranking slot** as drift, with the exact repair command — device-aware, reusing `type_implied_profile` so an `npu` embedding slot is told `--profile flm`, not llama-server's `embedding`. It previously skipped every profile-less slot as "base-image, legal". Its failure line no longer claims those slots cannot start (the whole point of #1830 is that they start and 501).
- Docs: `docs/guides/manage-slots.mdx` documents `--profile` on create/edit plus a repair aside.

## Why

Reported symptom (#1830): `hal0 slot create --type reranking` writes a TOML with no `profile` key, the slot loads to `state=ready`, and `/v1/rerank` answers `501 "Start it with --reranking"`.

Root cause is one layer below the create body the issue points at, and one layer wider:

1. **The model's stamped `defaults.profile` was the only create-time profile source.** Auto-scan, `model add`, pull and the curated catalog all register models with `defaults: null`, so the adoption never fired, nothing else supplied a profile, and the #1787 template gate in `providers/container.py` (which needs a truthy `_cfg_profile`) had nothing to apply. Fixing `_normalize_create_body` alone would have missed the capability orchestrator, which calls `SlotManager.create` directly and only ever stamps a profile for the tts/stt engine switch.
2. **The shared derivation rule had a CPU hole.** `profile_name_for_fit("embed", "cpu")` returned `None` and `derive_profile("embed", "cpu")` returned `cpu-chat` — a chat profile that emits no `--embedding`. So on a CPU-only box (which is exactly where the rc.5 repro ran) the guided installer's embed/rerank slots and capability-apply's fit inference were broken by the same defect, independent of the create path. The issue's "capability-apply-created slots launch correctly" holds on GPU only.

For any type but `llm` the profile is a correctness fact, not a tune: it carries llama-server's mode flag or selects the runtime engine.

## How it was verified

- New regression tests, red before the fix:
  - `tests/slots/test_slot_type_profile_inference.py` — the (type, device) → profile matrix, the fit veto, and `SlotManager.create` for embedding/reranking on cpu/rocm/vulkan, plus explicit-profile-wins, stamped-model-wins, `llm` untouched, and the model-less (grey capability slot) case.
  - `tests/api/test_slot_create_profile_inference.py` — the dashboard modal's exact body shape (`{name,type,runtime,model,device,autoload,priority}`, no profile) → the persisted TOML carries `profile`.
  - `tests/cli/test_slot_create_flags.py` — `--profile` forwarded on create and edit, absent when omitted, and the deprecated `slot add` alias sends no typer `OptionInfo` sentinel.
  - `tests/cli/test_doctor_profiles.py` — doctor flags a profile-less embedding/reranking slot, still skips a profile-less `llm` slot, and names the device-correct repair profile.
- Review round 2 regression tests, each red on the branch head before the fix (`5e7e7e26`):
  - `tests/slots/test_slot_type_profile_inference.py::test_type_implied_profile[tts-gpu-vulkan-None]`, `::test_type_implied_profile_vetoes_a_runtime_family_the_device_cannot_run`, `::test_create_leaves_a_vulkan_tts_slot_on_the_kokoro_fallback` — the TTS/Vulkan regression, at the rule and at `SlotManager.create`.
  - `tests/capabilities/test_orchestrator_autocreated_slot_type.py` — `_ensure_slot_exists` for `rerank` (both devices) and `img` writes the seed's `type`, and the rerank slot's TOML therefore carries `profile = "reranking"`; the embed child keeps its trio type.
  - `tests/slots/test_slot_type_profile_inference.py::test_type_implied_profile_ignores_the_cli_legacy_provider_default` and `tests/cli/test_doctor_profiles.py::test_refs_repair_command_is_device_aware`.
  - The old `("image", "img")` matrix row was replaced with the reachable `("image", "gpu-rocm")` one: `img` is a device_class, not a `DeviceLiteral`, so that row was green on an input no slot can have.
- Deliberate, commented updates to the two pinned matrices this changes: `tests/config/test_profile_derivation_parity.py` (`("embed","cpu") → embedding`, `("rerank","cpu") → reranking`) and `tests/install/test_profile_derive.py` (which now also asserts the seeds stay `backend=None`/`device_class=None`, preserving the #807/#834 coherence property the original test protected).
- Full suite green; `ruff check` and `ruff format --check` clean; `scripts/check_sunset.py` clean (scar markers 192 ≤ baseline 192). `mypy src` unchanged from `main`'s pre-existing errors — nothing new in the touched files.
- The full run also caught a real defect in this branch: the deprecated `model assign` / `slot add` aliases call the Typer commands in-process, so the new `profile` parameter arrived as an `OptionInfo` sentinel and was PUT/POSTed. Fixed, with a regression test.

## Deliberately out of scope

- **No launch-time back-fill.** A profile-less capability slot already on disk (an upgraded box, per the issue's ct105 read-only cross-check) is not silently repaired at argv-resolve time — that would inject mode-changing flags into the goldened launch path for slots the operator never re-saved. It gets a visible, auditable repair instead: `hal0 doctor profiles` names it and `hal0 slot edit <name> --profile <name>` fixes it. An updater sweep that back-fills existing TOMLs is a reasonable follow-up.
- **`GET /api/slots/<name>/resolved` returning `{"argv": null}`** for such a slot (issue's "Nothing warns" section) is untouched — separate defect, separate PR.
- No `CHANGELOG.md` hunk (per #1545; the consolidated entry lands at the end of the wave).
- `tts` / `transcription` are inferred at create time (where hal0 ships an engine for the device) but deliberately not flagged by doctor when absent: those providers carry their own profile-less defaults, so an empty profile there is not a demonstrated 501. `image` infers nothing at all — the shared rule's generic GPU branch answers `chat` before its image branch and the fit veto drops it; the CLI help and the guide now say so instead of promising `image → comfyui`. Reordering that shared rule would also move the picker's model-fit evaluation for image models, so it is a separate change.
- **The runner veto compares backend only.** The runner registry keys ComfyUI under the pseudo device class `img` while its slot's device is `gpu-rocm`, so a device_class comparison there would veto correct answers; hardware-class coherence stays `profile_fits_slot`'s job.

Closes #1830

### DIFF
--- a/docs/guides/manage-slots.mdx
+++ b/docs/guides/manage-slots.mdx
@@ -81,6 +81,18 @@ hal0 slot create embed \
 
 6. **`--ctx-size`** — context window in tokens (default `4096`).
 
+7. **`--profile`** — the runtime profile (flag bundle + engine). Omit it
+   and hal0 infers the profile the slot's `--type` implies: `embedding` →
+   `embedding` (`--embedding`), `reranking` → `reranking`
+   (`--reranking`), `tts`/`transcription` → the engine for the chosen
+   device, `image` → `comfyui`. That inference is what makes a capability
+   slot actually serve its endpoint — an `embedding` slot with no
+   `--embedding` flag loads to `ready` and returns HTTP 501 from
+   `/v1/embeddings`. An `llm` slot infers nothing: no mode flag is at
+   stake, so its tune stays the model's (or yours). Change a slot's
+   profile later with `hal0 slot edit <name> --profile <profile>` or the
+   slot drawer.
+
 </Steps>
 
 Creating a slot writes its config and systemd drop-in but does **not**
@@ -180,11 +192,22 @@ setting one) is how you take a slot out of rotation without deleting it.
 ```sh
 hal0 slot edit agent --ctx-size 32768
 hal0 slot edit embed --hardware rocm
+hal0 slot edit rerank --profile reranking
 ```
 
 Pass any subset of `--model`, `--port`, `--ctx-size`, `--provider`,
-`--hardware`. At least one is required. Changes the container baked in
-(port, hardware grid) take effect on the next `restart`.
+`--hardware`, `--profile`. At least one is required. Changes the container
+baked in (port, hardware grid) take effect on the next `restart`.
+
+<Aside type="tip" title="Repairing a profile-less capability slot">
+Capability slots created before v1.0 could land on disk with no `profile`
+key — they load to `ready` but return HTTP 501 from their own endpoint,
+because the profile is what carries llama-server's `--embedding` /
+`--reranking` mode flag. Check with
+`grep profile /etc/hal0/slots/<name>.toml`; repair with
+`hal0 slot edit <name> --profile embedding` (or `reranking`) followed by
+`hal0 slot restart <name>`. Newly created slots infer it automatically.
+</Aside>
 
 ## Dashboard: the inline model-edit pencil
 
--- a/docs/guides/manage-slots.mdx
+++ b/docs/guides/manage-slots.mdx
@@ -84,8 +84,14 @@ hal0 slot create embed \
 7. **`--profile`** — the runtime profile (flag bundle + engine). Omit it
    and hal0 infers the profile the slot's `--type` implies: `embedding` →
    `embedding` (`--embedding`), `reranking` → `reranking`
-   (`--reranking`), `tts`/`transcription` → the engine for the chosen
-   device, `image` → `comfyui`. That inference is what makes a capability
+   (`--reranking`), `tts`/`transcription` → the engine hal0 ships for the
+   chosen device (Kokoro on CPU, Qwen3-TTS on ROCm, Moonshine for CPU
+   speech-to-text). Nothing is inferred where hal0 has no engine for that
+   pairing — a `tts` slot on a Vulkan device, for instance, keeps the CPU
+   Kokoro runtime rather than being pinned to the ROCm-only Qwen3 image.
+   An `image` slot also infers nothing: pass `--profile comfyui`
+   explicitly (the shipped `img` slot already carries it). That inference
+   is what makes a capability
    slot actually serve its endpoint — an `embedding` slot with no
    `--embedding` flag loads to `ready` and returns HTTP 501 from
    `/v1/embeddings`. An `llm` slot infers nothing: no mode flag is at
--- a/src/hal0/capabilities/orchestrator.py
+++ b/src/hal0/capabilities/orchestrator.py
@@ -37,7 +37,7 @@ import logging
 from pathlib import Path
 from typing import Any
 
-from hal0.capabilities.catalog import models_for_capability, tts_profile_for_device
+from hal0.capabilities.catalog import models_for_capability
 from hal0.capabilities.config import (
     CapabilityConfig,
     CapabilitySelection,
@@ -58,6 +58,7 @@ from hal0.registry.store import ModelRegistry
 from hal0.slot_config import SlotConfigStore, SlotSelection
 from hal0.slots.layout import resolve_slot_stem
 from hal0.slots.manager import SlotManager
+from hal0.slots.profile_adopt import type_implied_profile
 
 log = logging.getLogger(__name__)
 
@@ -82,12 +83,27 @@ _SLOT_TO_CHILD: dict[str, tuple[str, str]] = {
 # ── Trio dispatch discriminator: capability child → slot ``type`` ─────────────
 # The FLM trio (one ``flm serve`` process serving chat + embed + asr) is gated
 # in v1._is_npu_trio_request on the slot record's ``type`` field. Only the two
-# trio modalities carry a type; rerank/tts/img are NOT trio members and
-# get no ``type`` key (their auto-created slots are plain llama-server/dedicated
-# providers). Mirrors providers/flm._classify_flm_model emitting "embed"/"stt".
+# trio modalities are trio members — rerank/tts/img carry their own types via
+# ``_NON_TRIO_CHILD_TO_SLOT_TYPE`` below, and must NOT be added here.
+# Mirrors providers/flm._classify_flm_model emitting "embed"/"stt".
 _CHILD_TO_SLOT_TYPE: dict[str, str] = {
     "embed": "embedding",
     "stt": "transcription",
+}
+
+# child → slot ``type`` for the NON-trio children. Deliberately a SECOND map:
+# the trio discriminator above must keep exactly its two members (it also
+# answers "is this an NPU modality?"), but an auto-created slot still needs the
+# ``type`` its shipped seed carries — ``SlotManager.create``'s type-implied
+# profile inference (#1830), ``providers/container._spec_provider_for``'s
+# dispatch and ``hal0 doctor profiles``' profile-less capability check all key
+# on it. Without it a re-created ``rerank`` slot landed profile-less, launched
+# llama-server with no ``--reranking``, reported ``state=ready`` and 501'd
+# ``/v1/rerank``. Values match installer/etc-hal0/slots/{rerank,tts,img}.toml.
+_NON_TRIO_CHILD_TO_SLOT_TYPE: dict[str, str] = {
+    "rerank": "reranking",
+    "tts": "tts",
+    "img": "image",
 }
 
 # The legal capability/child surface — used by HTTP validation.
@@ -741,34 +757,40 @@ class CapabilityOrchestrator:
             # grey/unconfigured slot the dashboard renders as a picker.
             "model": {"default": selection.model or ""},
         }
-        # Stamp the trio dispatch discriminator for embed/stt children so
-        # v1._is_npu_trio_request can gate on it. Non-trio children
-        # (rerank/tts/img) get no ``type`` key.
+        # Stamp the slot ``type`` — the trio dispatch discriminator for the
+        # embed/stt children (so v1._is_npu_trio_request can gate on it), and
+        # the seed-matching type for the non-trio ones. rerank/img used to get
+        # NO ``type`` key here, which left a re-created rerank slot with no
+        # type for SlotManager.create's profile inference (#1830) to key on:
+        # it launched llama-server without ``--reranking``, reported ready and
+        # 501'd /v1/rerank.
         _, child = _SLOT_TO_CHILD.get(slot_name, (None, None))
-        slot_type = _CHILD_TO_SLOT_TYPE.get(child) if child else None
+        slot_type = (
+            _CHILD_TO_SLOT_TYPE.get(child) or _NON_TRIO_CHILD_TO_SLOT_TYPE.get(child)
+            if child
+            else None
+        )
         if slot_type:
             cfg_dict["type"] = slot_type
         # voice.tts engine switch (follow-up to #972): a (re)created tts slot
-        # must carry the engine's runtime profile + type so the spawn routes to
-        # the right provider (cpu→Kokoro, gpu→Qwen3). The builtin tts.toml
-        # normally exists so this path is the deleted-slot fallback.
+        # runs a dedicated container image. The builtin tts.toml normally
+        # exists so this path is the deleted-slot fallback.
         if child == "tts":
-            cfg_dict["type"] = "tts"
             cfg_dict["runtime"] = "container"
-            cfg_dict["profile"] = tts_profile_for_device(selection.device)
-        # voice.stt engine switch (moonshine reinstatement): a (re)created cpu
-        # stt slot carries the Moonshine profile explicitly, mirroring the tts
-        # mechanism above, rather than relying on MoonshineProvider's own
-        # profile-less default (belt-and-suspenders — the slot's on-disk shape
-        # then already matches what a reconciled EXISTING slot would carry via
-        # SlotConfigStore._engine_profile_for). The npu case never reaches
-        # here (`_apply_npu_trio_modality` / `_ensure_slot_exists_npu` handle
-        # it); ``profile_name_for_fit`` returns ``None`` for any device this
-        # branch could see beyond cpu, so the stamp only ever fires there.
-        elif child == "stt":
-            stt_profile = profile_name_for_fit("stt", selection.device, selection.provider or "")
-            if stt_profile is not None:
-                cfg_dict["profile"] = stt_profile
+        # tts/stt engine profile: stamped explicitly so the slot's on-disk
+        # shape already matches what a reconciled EXISTING slot would carry via
+        # SlotConfigStore._engine_profile_for, rather than relying on the
+        # provider's own profile-less default (belt-and-suspenders). It comes
+        # from the SHARED type-implied rule, which vetoes both a profile that
+        # does not fit the slot and a runtime family with no runner for this
+        # device — a Vulkan box must not be pinned to the ROCm-only qwen3tts
+        # image, and an unknown STT engine must not inherit Moonshine's
+        # profile. The npu case never reaches here (`_apply_npu_trio_modality`
+        # / `_ensure_slot_exists_npu` handle it).
+        if child in ("tts", "stt"):
+            engine_profile = type_implied_profile(cfg_dict)
+            if engine_profile is not None:
+                cfg_dict["profile"] = engine_profile
         try:
             await self._slot_manager.create(slot_name, cfg_dict)
         except Exception as exc:
--- a/src/hal0/capabilities/profile_fit.py
+++ b/src/hal0/capabilities/profile_fit.py
@@ -63,15 +63,23 @@ def profile_name_for_fit(capability: str, device: str, provider: str = "") -> st
         return None
     if device == "npu":
         return DEVICE_DEFAULT_PROFILES.get("npu")
-    if capability == "embed" and device in {"gpu-rocm", "gpu-vulkan"}:
-        # Dedicated GPU embed lane (llama-server --embedding), backend-coherent:
-        # gpu-rocm→embed, gpu-vulkan→vulkan-embed. Both non-MTP, so this honours
-        # the resolver's "never force MTP" contract. The base chat profile never
-        # emits --embedding, so embed must not fall through to it.
+    if capability == "embed":
+        # Dedicated embed lane (llama-server --embedding). Non-MTP, so this
+        # honours the resolver's "never force MTP" contract. The base chat
+        # profile never emits --embedding, so embed must not fall through to it.
+        #
+        # NOT device-gated (#1830): the 1.0 seed ``embedding`` profile is a
+        # device-agnostic logical tune (no ``device_class``, no ``backend``) —
+        # the old ``device in {gpu-rocm, gpu-vulkan}`` gate is a leftover from
+        # the retired per-backend ``embed``/``vulkan-embed`` seeds. On a
+        # CPU-only box it made this rule answer "no opinion", so an embed slot
+        # took a chat profile, launched with no ``--embedding``, reported
+        # ``state=ready`` and 501'd every ``/v1/embeddings`` call.
         return "embedding"
-    if capability == "rerank" and device in {"gpu-rocm", "gpu-vulkan"}:
-        # Dedicated GPU rerank lane (llama-server --reranking → /v1/rerank),
-        # backend-coherent: gpu-rocm→rerank, gpu-vulkan→vulkan-rerank.
+    if capability == "rerank":
+        # Dedicated rerank lane (llama-server --reranking → /v1/rerank); MUST
+        # stay a separate instance from embed. Device-agnostic for the same
+        # reason as embed above (#1830).
         return "reranking"
     if device in {"gpu-rocm", "gpu-vulkan"}:
         return DEVICE_DEFAULT_PROFILES.get(device)
--- a/src/hal0/cli/doctor_bundle.py
+++ b/src/hal0/cli/doctor_bundle.py
@@ -333,6 +333,7 @@ def _write_diagnostics_section(out: Path, *, base: str | None = None) -> None:
         profiles = catalog.list()
         valid_names = {p.name for p in profiles}
         slot_profiles: list[tuple[str, str | None]] = []
+        slot_types: dict[str, str] = {}
         # id-aware (P3-runtime-db inc4): report cfg.name (the real display
         # name), not the raw list_slots() stem — see the identical fix +
         # rationale in doctor_commands.doctor_profiles.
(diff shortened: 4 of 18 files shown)


thinmintdev:
## Review — changes requested

Verified against the branch at `5e7e7e2` in a clean worktree, not from the PR description.

**What holds up.** The issue's exact repro path is fixed at the right layer. `hal0 slot create zzsk1787 --type reranking --hardware cpu` → CLI body (no `profile`) → `POST /api/slots` → `_normalize_create_body` → `SlotManager.create` → `type_implied_profile({"type":"reranking","device":"cpu"})` → `"reranking"` persisted. I confirmed the chokepoint choice is correct: `SlotManager.create` is the single funnel, and `_normalize_create_body` alone would not have covered the orchestrator or a stack apply. The second defect (the `device in {gpu-rocm, gpu-vulkan}` gate on the embed/rerank lanes) is a genuine second root cause, not scope creep — it is what made the CPU-only repro box fail through the installer path too.

**Test red-before-green: verified, not taken on faith.** I checked the tests out at PR HEAD, reverted `src/` to the pre-fix base, and re-ran: 1 collection error (`type_implied_profile` missing) + 11 failures across all five layers — the API-route → TOML assertions, both CLI flags, both doctor rows, and the two pinned matrices. Restored and re-ran: 176 passed. Full suite locally: **9996 passed, 15 skipped, 1 xfailed** — matches the claim. No `CHANGELOG.md` hunk. CI green.

Two things block, both outside the issue's blast radius — i.e. exactly the kind of thing the issue never asked you to touch.

---

### Blocking 1 — a `tts` slot on non-ROCm hardware is now routed to a ROCm-only runner

`type_implied_profile` maps `tts` → `tts_profile_for_device`, which answers `qwen3-tts` for **any** GPU device including `gpu-vulkan`. `profile_fits_slot` waves it through (the seed is device-agnostic: `device_class=None`, `backend=None`), so it gets persisted. Measured on the branch:

```
tts  gpu-vulkan -> qwen3-tts          # newly written by this PR
{'type':'tts','device':'gpu-vulkan'}                       -> KokoroProvider     # before
{'type':'tts','device':'gpu-vulkan','profile':'qwen3-tts'} -> Qwen3TTSProvider   # after
```

`RUNNER_IMAGES["qwen3tts"]` is `device_class="gpu"`, `backend="rocm"`, `supported_backends=("rocm",)`, and `Qwen3TTSProvider` unconditionally wires `/dev/kfd` + MIOpen. `providers/container._spec_provider_for` says so in its own comment — *"a profile-less `type=tts` slot still falls through to Kokoro"* — and that fallback is exactly what this PR removes. So:

```
hal0 slot create voice --type tts --model <kokoro model>      # no --hardware
```

on any box whose default hardware resolves to vulkan (including the CLI's documented *"vulkan if no probe"* fallback) used to produce a working Kokoro CPU TTS slot and now produces a slot pinned to a runner its own device cannot run. That is a live regression for a caller #1830 never mentions, and nothing in the new tests covers a vulkan TTS slot.

The fix is already in this file: `runner_fits_slot` sits 40 lines below `type_implied_profile` and answers correctly —

```
runner_fits_slot("qwen3tts", {"device": "gpu-vulkan"}) -> False
runner_fits_slot("qwen3tts", {"device": "gpu-rocm"})   -> True
```

Veto the candidate on the resolved profile's `runtime_family` as well as `profile_fits_slot`, and vulkan TTS goes back to inferring nothing (→ Kokoro dispatch, the pre-PR behaviour). Please add the vulkan case to the `(type, device)` matrix test either way.

### Blocking 2 — the capability orchestrator's `rerank` path still creates a profile-less slot

The PR body and the issue comment both claim the chokepoint covers `CapabilityOrchestrator._ensure_slot_exists`. It does not, for the one child that matters here. `_CHILD_TO_SLOT_TYPE` is `{"embed": "embedding", "stt": "transcription"}` only, and the code says so deliberately:

```python
# Stamp the trio dispatch discriminator for embed/stt children ...
# Non-trio children (rerank/tts/img) get no ``type`` key.
```

So an orchestrator-created `rerank` slot reaches `SlotManager.create` with **no `type` key** → `type_implied_profile` returns `None` → the TOML lands profile-less → `_spec_provider_for` falls through to llama-server → no `--reranking` → `state=ready` + 501 on `/v1/rerank`. The reported bug, reproduced verbatim, on the one path the PR advertises as covered. It is reachable: delete `rerank` (the delete hook flips the selection off), re-enable it in the dashboard, and `_ensure_slot_exists`'s `resolve_slot_stem` probe misses, so it creates. The new doctor check misses it too — no `type` key means `_PROFILELESS_CAPABILITY_REPAIR.get("")` → skip — so it is silent in the same way.

Stamp `type` for the non-trio children in `_ensure_slot_exists` (or pass the capability into the inference explicitly). Either way the PR body's coverage claim needs to match what the code does.

---

### Nits (not blocking)

1. **`doctor profiles`' repair advice is device-blind while create-time inference is not.** `_PROFILELESS_CAPABILITY_REPAIR` keys on type alone, so a profile-less `type=embedding, device=npu` slot is told to run `hal0 slot edit <name> --profile embedding` — llama-server `--embedding -fa auto -b 8192 -ub 8192` onto an FLM runtime. The create path in this same PR correctly answers `flm` for that pair (I checked: `type_implied_profile({"type":"embedding","device":"npu"}) == "flm"`). Two rules in one commit disagree; pass the device in and reuse `type_implied_profile`, or omit the repair when the device isn't `cpu`/`gpu-*`.
2. **Two documented inference promises never fire.** The `--profile` help and `manage-slots.mdx` promise `image → comfyui` and `tts/transcription → the engine for the device`. Measured: `type_implied_profile({"type":"image","device":"gpu-rocm"})` → `None` (the generic GPU branch in `profile_name_for_fit` returns `chat` before its `image` branch, then gets vetoed), and the CLI always sends its legacy default `provider="llama-server"`, which makes the STT guard return `None` — so `hal0 slot create --type transcription --hardware cpu` gets no profile while the dashboard modal (which sends no provider) does. Both are behaviourally harmless (`_spec_provider_for` still falls through to ComfyUI/Moonshine on slot type), but the documented contract is wrong and the CLI/dashboard now disagree. Either fix the inference or trim the claim.
3. **`test_type_implied_profile` pins `("image", "img", "comfyui")`, a case that cannot occur.** `DeviceLiteral` is `gpu-rocm | gpu-vulkan | gpu-cuda | cpu | npu` — `img` is a `device_class`, not a device, and the shipped `installer/etc-hal0/slots/img.toml` is `device = "gpu-rocm"`. That row is green on an unreachable input while the reachable one (nit 2) is uncovered.
4. **Doctor's summary line is now wrong for the new row type**: `"N slot(s) reference a missing profile — fix before those slots can start"`. The whole point of #1830 is that these slots *do* start. Worth a second sentence, since that line is what an operator reads first.

---

### Codex triage

All five P2 comments verified independently; none dismissed.

| # | Claim | Verdict |
|---|---|---|
| `manager.py:2429` | orchestrator omits `type` for rerank → profile-less slot | **Confirmed** → Blocking 2 |
| `profile_adopt.py:100` | vulkan TTS gets the ROCm-only qwen3-tts | **Confirmed**, and understated — the pre-PR fallback (Kokoro) actually worked → Blocking 1 |
| `doctor_commands.py:1714` | NPU embedding repair advice is wrong | **Confirmed** → nit 1 |
| `profile_adopt.py:101` | image never resolves comfyui | **Confirmed**, behaviourally harmless → nit 2/3 |
| `profile_adopt.py:99` | legacy `provider="llama-server"` suppresses STT inference | **Confirmed**, behaviourally harmless → nit 2 |

The bot posted only inline comments and no summary verdict, so treat its silence elsewhere as "did not look", not "approved".

thinmintdev:
## Review round 2 — both blocking findings fixed, with red-first regression tests

Thanks — both were real, and I reproduced each on the branch head (`5e7e7e26`) before touching anything.

### 1. BLOCKING — TTS regression on a Vulkan box (`src/hal0/slots/profile_adopt.py`)

Confirmed exactly as described. On `5e7e7e26`:

```
gpu-vulkan -> qwen3-tts   (family qwen3tts, device_class None, backend None)
gpu-rocm   -> qwen3-tts
runner_fits_slot('qwen3tts', {'device':'gpu-vulkan'}) -> False
runner_fits_slot('qwen3tts', {'device':'gpu-rocm'})   -> True
```

`profile_fits_slot` waved it through precisely because the 1.0 seeds are device-agnostic logical tunes (`device_class = None`, `backend = None`), so nothing compared the slot's backend against the ROCm-only `qwen3tts` runner.

**Fix:** `type_implied_profile` now vetoes twice — `profile_fits_slot` as before, plus a new `_runtime_family_fits_device`, which resolves the candidate's `runtime_family`, looks it up in the runner registry and calls `runner_matches`. A Vulkan/CUDA TTS slot infers nothing again and keeps the profile-less `_spec_provider_for` → Kokoro fallback; ROCm still gets `qwen3-tts`.

Two deliberate details:

* **Backend only, not device_class.** The registry keys ComfyUI under the pseudo device class `img` while its slot's device is `gpu-rocm`, so a device_class comparison would veto correct answers. Hardware-class coherence stays `profile_fits_slot`'s job for any profile that declares one.
* `llama-server` is a multi-runner family and not a runner key, so `get_runner` raises `NotFound` there and the veto abstains — embed/rerank inference is untouched.

The orchestrator's tts/stt engine stamp had the same defect (it called `tts_profile_for_device(selection.device)` directly), so it now routes through the shared `type_implied_profile` and inherits the veto.

### 2. BLOCKING — orchestrator-created `rerank` slot (`src/hal0/capabilities/orchestrator.py`)

Also confirmed, and the PR body's claim was wrong. Red proof on `5e7e7e26` — `_ensure_slot_exists("rerank", …)` wrote a TOML with no `type` and therefore no `profile`.

**Fix:** a second map, `_NON_TRIO_CHILD_TO_SLOT_TYPE = {"rerank": "reranking", "tts": "tts", "img": "image"}`. `_CHILD_TO_SLOT_TYPE` keeps exactly its two members — it is also the "is this an NPU trio modality?" predicate (`orchestrator.py:512`), so adding `rerank` there would have been wrong. The stamped values match the shipped seeds (`installer/etc-hal0/slots/rerank.toml` → `type = "reranking"`, `img.toml` → `type = "image"`). With the type present, `SlotManager.create` infers `reranking`, and `doctor profiles` can see it too. `v1._is_npu_trio_request` is only ever called with an explicit `slot_type` of `embedding`/`transcription`, so the new types cannot reach the trio gate.

**PR body corrected** — the "one chokepoint covers `_ensure_slot_exists`" bullet now carries the correction, plus the rest of this round.

### Non-blocking findings

| Finding | Action |
|---|---|
| `_PROFILELESS_CAPABILITY_REPAIR` is type-only, so an `npu` embedding slot is told `--profile embedding` | **Fixed.** Doctor now reuses `type_implied_profile` (device-keyed), keeping the static map only as a fallback. `npu` → `flm`, `cpu` → `embedding`. Same fix applied to the `doctor bundle` path, which Codex then flagged for the same omission. |
| `doctor profiles`' summary says "fix before those slots can start" | **Fixed.** It now names both shapes — a dangling reference stops the slot starting, a missing one lets it start and 501 — and points at the per-row repair. |
| CLI/docs promise `image → comfyui` and an engine for every tts/transcription device | **Documented as it behaves, not "fixed".** Making `profile_name_for_fit` answer `comfyui` before its generic GPU branch would also move the picker's `evaluate_model_fit` for every image model (`orchestrator._profile_for_fit` feeds it), which is a wider behaviour change than this PR should carry. The `--profile` help and `docs/guides/manage-slots.mdx` now say `image` infers nothing (pass `--profile comfyui`; the shipped `img` seed already carries it) and that no engine is inferred where hal0 ships none for that device. |
| `hal0 slot create` sends `provider=llama-server`, which suppresses STT inference | **Fixed.** The inference treats that historical default as "unspecified", so the CLI and the dashboard modal now agree on `moonshine` for a CPU transcription slot. A named engine hal0 has no runtime for (`whispercpp`) is still rejected. |
| `("image", "img")` matrix row is unreachable | **Fixed.** Replaced with the reachable `("image", "gpu-rocm")` row (and its `None` answer documented). `img` is a `device_class`, not a `DeviceLiteral`. |

Codex raised one more this round I did **not** act on: a model whose `defaults.profile` is a generic chat profile would still win over the type-required mode profile on an embedding slot. It is reachable only by hand-editing a registry entry — every ingest path (auto-scan, `model add`, pull, curated catalog) registers `defaults: null`, which is this issue's root cause — and inverting that precedence would also override a deliberately stamped capability-specialised profile. Left as-is; worth a follow-up issue rather than a precedence flip inside a fix PR.

### Verification

Every new test was run against the unfixed code first. On `5e7e7e26`:

```
FAILED tests/slots/test_slot_type_profile_inference.py::test_type_implied_profile[tts-gpu-vulkan-None]
FAILED tests/slots/test_slot_type_profile_inference.py::test_type_implied_profile_vetoes_a_runtime_family_the_device_cannot_run
FAILED tests/slots/test_slot_type_profile_inference.py::test_type_implied_profile_ignores_the_cli_legacy_provider_default
FAILED tests/slots/test_slot_type_profile_inference.py::test_create_leaves_a_vulkan_tts_slot_on_the_kokoro_fallback
FAILED tests/capabilities/test_orchestrator_autocreated_slot_type.py::test_autocreated_non_trio_slot_carries_its_type[rerank-gpu-vulkan-reranking-reranking]
FAILED tests/capabilities/test_orchestrator_autocreated_slot_type.py::test_autocreated_non_trio_slot_carries_its_type[rerank-cpu-reranking-reranking]
FAILED tests/capabilities/test_orchestrator_autocreated_slot_type.py::test_autocreated_non_trio_slot_carries_its_type[img-gpu-rocm-image-None]
7 failed, 25 passed
```

and for the doctor round (with `src/hal0/cli/doctor_{commands,bundle}.py` reverted):

```
FAILED tests/cli/test_doctor_bundle.py::test_bundle_profile_repair_is_device_aware
FAILED tests/cli/test_doctor_profiles.py::test_refs_never_recommend_a_profile_missing_from_the_catalog
FAILED tests/cli/test_doctor_profiles.py::test_refs_repair_command_is_device_aware  (TypeError: unexpected keyword 'slot_devices')
```

All green after the fix. Full local suite: **10006 passed, 15 skipped, 1 xfailed** (`HAL0_HOME=$(mktemp -d) uv run --extra dev pytest`). `ruff check` and `ruff format --check` clean; `scripts/check_sunset.py` clean (192 ≤ baseline 192). No `CHANGELOG.md` hunk — the consolidated entry lands at the end of the wave.

Not merging or enabling automerge; this is for the wave orchestrator to decide.

PR #1854, opened 2026-08-12
docs/install requirements curl

### LINKED ISSUE #1843: docs(install): Requirements never lists curl, so the documented one-liner dies with 'curl: command not found' on a minimal image
## What

`docs/getting-started/install.mdx` leads with

```sh
curl -fsSL https://hal0.dev/install.sh | sudo bash
```

and its **Requirements** section (line 35) lists Linux + systemd, x86_64, Python ≥ 3.12, podman/docker, and Node — but never `curl`, which is the one thing that command cannot run without.

## Repro

Any minimal image without curl. Found on the Proxmox `local:vztmpl/ubuntu-26.04-standard_26.04-1` LXC template while building the rc.5 validation box:

```
# curl -fsSL https://hal0.dev/install.sh -o /tmp/install.sh
bash: line 1: curl: command not found
```

The failure comes from the user's shell, not from hal0, so there is nothing hal0 can print to explain it. The user's first contact with the product is a bare `command not found` against a command they copied from our docs.

## Impact

Cosmetic-to-minor, but it lands on the very first step of the very first page, and it is invisible to every existing test — `scripts/fresh-test-ct.sh` clones a template that already has curl, and the validation fleet's boxes all had it pre-installed.

## Suggested fix

One line in **Requirements**, e.g.:

> - **curl** and **ca-certificates** — the bootstrap one-liner needs them. Minimal cloud/LXC images often ship without: `apt-get install -y curl ca-certificates`.

Optionally offer the `wget` equivalent alongside the canonical one-liner.

Found during the [v1.0.0-rc.5 validation run](https://github.com/Hal0ai/hal0/blob/main/tests/release-validation/reports/v1.0.0-rc.5.md) (pre-suite, while standing the box up — not part of the 27 verified findings).

thinmintdev:
## Summary

- `docs/getting-started/install.mdx` leads with `curl -fsSL https://hal0.dev/install.sh | sudo bash`, but the Requirements section never listed curl (or ca-certificates) as a prerequisite.
- On a minimal image without curl (e.g. a bare Proxmox LXC template), the documented one-liner dies with a bare `curl: command not found` from the user's own shell, before hal0 ever runs — nothing hal0 prints can explain that failure.
- Added a one-line Requirements bullet calling out `curl` and `ca-certificates`, with the `apt-get install -y curl ca-certificates` remediation.

## Test plan

- [x] Added `tests/docs/test_install_requirements_lists_curl.py`, a regression test that parses `install.mdx`'s Requirements section and asserts it mentions `curl`. Verified it fails against the pre-fix doc and passes after the fix.
- [x] `uv run ruff check` / `ruff format --check` clean on the new test file (repo-wide lint has pre-existing unrelated failures, not touched here).
- [x] Docs-only change, no src/, installer, or CHANGELOG touched (per wave convention, CHANGELOG is handled separately in issue #1545's consolidated PR).

Closes #1843

### DIFF
--- a/docs/getting-started/install.mdx
+++ b/docs/getting-started/install.mdx
@@ -34,6 +34,10 @@ Accepted values: `stable` (default), `preview`, `nightly`.
 
 ## Requirements
 
+- **curl** and **ca-certificates** — the bootstrap one-liner needs them to
+  fetch and verify the release manifest. Minimal cloud/LXC images often
+  ship without either: `apt-get install -y curl ca-certificates` (or the
+  equivalent for your distro).
 - **Linux, x86_64** — hal0 ships amd64-only binaries (the FastFlowLM
   `.deb` and the toolbox container images). ARM and macOS/Windows are not
   supported.
--- a/tests/docs/test_install_requirements_lists_curl.py
+++ b/tests/docs/test_install_requirements_lists_curl.py
@@ -1,0 +1,45 @@
+"""``docs/getting-started/install.mdx`` must list ``curl`` as a requirement.
+
+The install page leads with:
+
+    curl -fsSL https://hal0.dev/install.sh | sudo bash
+
+but its **Requirements** section never mentioned ``curl`` (or
+``ca-certificates``), even though the documented one-liner cannot run without
+it. On a minimal image (e.g. a bare Proxmox LXC template) that command dies
+with a bare ``curl: command not found`` from the user's own shell — before
+hal0 ever runs, so there is nothing hal0 can print to explain it. Found while
+standing up the v1.0.0-rc.5 validation box (issue #1843).
+
+This pins the fix: the Requirements section must call out curl (and
+ca-certificates) explicitly, in the same section as the other prerequisites.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_INSTALL_MDX = _REPO_ROOT / "docs" / "getting-started" / "install.mdx"
+
+
+def _requirements_section(text: str) -> str:
+    """Slice out the ``## Requirements`` section (up to the next ``## `` heading)."""
+    start = text.index("## Requirements")
+    rest = text[start:]
+    next_heading = rest.find("\n## ", 1)
+    return rest if next_heading == -1 else rest[:next_heading]
+
+
+def test_requirements_section_mentions_curl() -> None:
+    assert _INSTALL_MDX.exists(), f"missing {_INSTALL_MDX}"
+    text = _INSTALL_MDX.read_text(encoding="utf-8")
+
+    section = _requirements_section(text)
+
+    assert "curl" in section, (
+        "Requirements section must call out curl — the bootstrap one-liner "
+        "(`curl -fsSL https://hal0.dev/install.sh | sudo bash`) hard-requires "
+        "it and fails with a bare 'command not found' on minimal images "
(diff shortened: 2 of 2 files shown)


thinmintdev:
## Review — approve (post-merge), two follow-ups worth filing

Verified against the issue's repro rather than the PR description.

**1. Does it fix #1843?** Yes, exactly the reported path. #1843 is a docs defect: `docs/getting-started/install.mdx` leads with `curl -fsSL https://hal0.dev/install.sh | sudo bash` while `## Requirements` never lists `curl`. The diff adds the bullet to that same section, with the `apt-get install -y curl ca-certificates` remediation the issue suggested. The failure happens in the user's shell before any hal0 code runs, so a docs line is the only possible fix surface — this is the root cause, not a symptom patch.

**2. Does the new test fail without the fix?** Confirmed by running it, not by reading the description. Extracted `docs/getting-started/install.mdx` at `d8a47d4d~2` (pre-fix) and at the merge, ran the new test against both:

```
post-fix: 1 passed
pre-fix:  1 failed — AssertionError: Requirements section must call out curl
          assert 'curl' in "## Requirements\n\n- **Linux, x86_64** — ..."
```

The section slicer is sound: `text.index("## Requirements")` then `rest.find("\n## ", 1)` stops at `## What the installer does` and does not mis-match `### ` sub-headings. Placing the file in `tests/docs/` without an `__init__.py` is consistent with nine other `tests/` subdirs (`tests/install`, `tests/installer`, `tests/system`, …), so collection is fine.

**3. Regressions / scope.** Zero. Two files, +49/-0, no `src/`, no `installer/`, no CHANGELOG hunk — clean against this wave's convention. Nothing imports the changed code. The added bullet is plain Markdown (no `<` or `{`), so the MDX/Starlight build is unaffected; CI is green across all 8 checks including the 24m python suite.

**4. Codex bot.** It posted only the boilerplate "Reviewed commit 23f3c88d8d" header with no suggestions and no inline comments. Nothing to triage — noting explicitly that this is silence, not a verified approval.

### Follow-ups (non-blocking, worth their own issues)

- **The Requirements list still doesn't match the bootstrap's own preflight.** `installer/bootstrap.sh:138-144` hard-requires `curl`, `tar`, `sha256sum`, `jq`, `python3` — `need jq` is unconditional and there is no fallback (jq does the fail-closed manifest policy pass at line 317 and field extraction at line 372). On the exact image from the repro (a bare `ubuntu-26.04-standard` LXC template), the operator who follows this new bullet and installs curl + ca-certificates gets one step further and then hits `missing dependency: jq — install it and re-run`. That one at least comes from hal0 with a remediation, so it is a different and much milder class than the bare `command not found` #1843 reported — but the same minimal-image path is still not clean end-to-end, and the next validation run will find it. Adding `jq` (and `tar`) to the same bullet would close the class rather than the instance.
- **`docs/getting-started/bare-metal.mdx:22` has a second prerequisites table** that also omits curl/jq and likewise ends with the same one-liner at line 58. Out of scope for #1843, which named only `install.mdx`, but the two lists should not drift.

### Nits

- The test's docstring says it pins "curl (and ca-certificates)" but the assertion only checks `curl`. Either assert both or drop the parenthetical.
- `text.index(...)` raises a bare `ValueError: substring not found` if the heading is ever renamed (e.g. to `## Prerequisites`), which is a confusing failure mode compared to the nice assertion message right below it. A `pytest.fail("install.mdx has no '## Requirements' heading")` guard would read better.

Issue #1796, opened 2026-08-09
polish(cli/api): rc.4 fresh-install validation sweep — 14 items


thinmintdev:
Rollup of minor/cosmetic CLI + API findings from v1.0.0-rc.4 fresh-install validation (CT151, CPU-only). Batched per the existing polish-sweep convention. Each item was reproduced live; ask and I can split any of these out.

**CLI**

1. `hal0 update --check` renders the placeholder stable manifest as a real target: banner `hal0 1.0.0rc4 → 0.0.0 (stable) up to date` with all-zero `digest_sha256` and `released_at 2026-05-15`. Placeholder manifests should render as "no release published on this channel". (Relates to #1530/#1531.)
2. `hal0 system-info` reports impossible CPU topology in a container: `cores / threads 16 / 8` (host physical cores + container cgroup limit mixed) on an 8-vCPU CT (`nproc`=8). Also leaks a debug line (`hardware.probe.nvidia_smi_unavailable ...`) above the table.
3. `hal0 doctor` prints six `!!` warnings then concludes `OK all pre-flight checks passed` (exit 0). Summary contradicts the body. (Relates to #1543.)
4. `hal0 model scan` prints `4 added, [] skipped` — raw empty list in the summary line.
5. `hal0 slot metrics` prints raw float memory (`Mem MB 1223.69921875`); tok/s and KV% show `—` for a live slot even after requests.
6. `hal0 bench results` prints literally nothing when no records exist (no "no records" message).
7. `hal0 slot delete` leaves an empty `/var/lib/hal0/slots/<name>/` dir behind (config TOML, unit, port all cleaned correctly).
8. `hal0 slot capacity` books CPU-resident model memory under `VRAM MB` with `RAM MB 0.0` on a GPU-less box.
9. `hal0 model list` has no default-model marker; after `hal0 model default <id>` the promotion is only visible via `model show`/API.
10. `hal0 model add` misclassifies a reranker gguf as `capabilities: chat` and derives the id from the gguf arch header (`jina-bert-implementation`) instead of the filename — the same weights registered by install-time scan carry the correct rerank capability.
11. `hal0 mcp status` / dashboard MCP tab print a connect URL missing the FastMCP transport suffix — `http://127.0.0.1:8080/mcp/admin` 405s; working endpoint is `/mcp/admin/mcp` (`_connect_url()`, `routes/mcp.py:444`). Docs (`MCP-CLIENTS.md`, `connect-mcp.mdx`) have it right.
[...]

PR #1855, opened 2026-08-12
fix(registry): stop reporting high confidence for a filename-derived gguf capability guess

### LINKED ISSUE #1838: registry: hal0 model add classifies a header-ambiguous gguf by filename and reports confidence high, and accepts a file with no GGUF magic as a chat model
## What happens

Two narrow gaps in `hal0.registry.detect.detect()` (called by `models_service.add_from_path`, i.e. `hal0 model add` / `POST /api/models/add-from-path`).

### (A) Capability detection is filename-dependent when the GGUF header omits `pooling_type`

`detect()` keys capability solely on `<arch>.pooling_type`, then falls back to a filename substring table (`rerank` / `embed` / `bge` / `e5-` / …), then defaults to `chat`. The jina-reranker-v1-tiny-en gguf has **no** `pooling_type` key, so its classification comes entirely from its filename — while `confidence` is still reported `high`. The same header carries three unambiguous non-chat signals that `detect()` ignores: `general.architecture = jina-bert-v2`, `general.tags = ['reranker','cross-encoder',…]`, `jina-bert-v2.attention.causal = False`.

Scope limit: only bites files whose header lacks `pooling_type` **and** whose filename lacks the token. Embedders that carry the key are unaffected (Qwen3-Embedding-0.6B → `pooling=3` → `embed`).

### (B) No format validation

A file with a `.gguf` extension and no valid GGUF magic is accepted and registered as `capabilities: chat`, `backends: [vulkan, rocm, cuda, cpu]`. `detect()` internally records `confidence='low'` and `raw_hints.gguf_header_read='failed'`; **neither is surfaced** in the API response or the CLI output, and the CLI still prints `Next: hal0 model run <id>`.

## Decisive evidence

Controlled A/B, two byte-identical copies (md5 `24f3de7ac2671e1b06a3afc587111aef`):
```
hal0 model add /mnt/ai-models/rc5-scratch/zzscratchrr.gguf                       -> capabilities: chat
hal0 model add /mnt/ai-models/rc5-scratch/jina-reranker-v1-tiny-en.Q8_0.gguf ... -> capabilities: rerank
```
Header proof (installed venv):
```
zzscratchrr.gguf                   pooling_type=None arch=jina-bert-v2 caps=['chat']   conf=high
jina-reranker-v1-tiny-en.Q8_0.gguf pooling_type=None arch=jina-bert-v2 caps=['rerank'] conf=high
Qwen3-Embedding-0.6B-Q8_0.gguf     pooling=3                            caps=['embed']   (header path works when the key exists)
```
Genuinely non-GGUF bytes:
```
head -c 2000000 /dev/urandom > zzskeprand.gguf     (first bytes: 4c8d 2a92 4fc5 8d1e)
hal0 model add ... --id zzskeprand -> "Registered zzskeprand ... capabilities: chat" / "Next: hal0 model run zzskeprand"
detect() internals: ['chat'] low llama {'source': 'filename', 'gguf_header_read': 'failed'}
```

## Refuted parts of the original candidate

1. "rc.4 misclassification, unfixed" — wrong. The rc.4/#1796 case (reranker under its real filename typed `chat`) **is** fixed in rc.5 by `53be7e9d` (tagged v1.0.0-rc.5); it now returns `rerank`.
2. "byte-identical content registered by another path is `['rerank']`" — wrong. `rcrerank` has `metadata.source = add-from-path`, the same code path. The only variable is the filename.
3. Its evidence for (B) is misread: `/var/lib/hal0/models/zzbroken/zzbroken.gguf` begins `47 47 55 46 03` (valid GGUF v3 magic) and parses to `general.name = Qwen3.5-0.8B` — a truncated real chat gguf, so `chat` there is a **correct** header-derived answer. (B) only reproduces with genuinely non-GGUF bytes.

## What is by design (do not "fix")

The chat fallback for an unparseable header is explicit intent — `detect()`'s docstring: "Never raises for an unreadable / missing / non-GGUF file: we fall back to the filename heuristic and lower the confidence." The filename token table is the deliberate rc.5 remedy for #1796. Registering an unloadable file is recoverable with `hal0 model rm`.

## What is not

`hal0 model add --help` says "Reads the file header to derive id, capabilities and backends" and "capabilities auto-detected", with no hedge; `detect()` reports `confidence: high` for a header-read hit even when the capability came from the filename; and `confidence` / `raw_hints.gguf_header_read: failed` are computed and then dropped before the response, so the user sees a confident success line for a guess.

## Recommended fix (small)

- surface `confidence` and `gguf_header_read` in the add-from-path response and CLI line
- warn (not fail) when GGUF magic is missing
- add `general.architecture` / `attention.causal` / `general.tags` as tie-breakers **before** the filename fallback
- add a `--capability` flag to `hal0 model add` mirroring the API's existing `labels` (the CLI exposes only `--id/--name/--license/--overwrite`, which is the real friction)

Workaround verified: `curl -X PUT http://127.0.0.1:8080/api/models/<id> -d '{"capabilities":["rerank"]}'` → 200.

Source: `src/hal0/registry/detect.py:327-420`, `src/hal0/registry/gguf_header.py:205`, `src/hal0/services/models_service.py:1045-1052`, `src/hal0/cli/model_commands.py:391-435`.

## Environment

hal0 1.0.0rc5, ct152 (`10.0.1.152`), fresh CPU-only install. Non-destructive: all four test registrations removed, scratch dir deleted, registry restored.

## Other hardware

Fully hardware-independent — `detect()` is pure Python (mmap of the header window, struct parse, case-insensitive filename substring match). The `backends: [vulkan, rocm, cuda, cpu]` list is a static constant (`_GGUF_BACKENDS`), not a hardware query. GPU boxes, privileged containers and upgraded boxes behave identically. If anything a GPU box makes (B) marginally worse: the bogus `chat` model is more likely to be assigned to a real slot and crash-loop there.

---
Found by the 1.0.0-rc.5 release-validation run. Full report: `tests/release-validation/reports/v1.0.0-rc.5.md` (branch `h0/rc5-release-testing`). Finding key: `model-add-capability-misdetection`.

thinmintdev:
## What changed

Two narrow gaps in `hal0.registry.detect.detect()` (called by `hal0 model add` / `POST /api/models/add-from-path`):

**(A) Root cause fix — capability classification was filename-dependent even though the header carried unambiguous tie-breakers.** When a GGUF header omits `<arch>.pooling_type`, `detect()` fell straight to a filename-token guess and *still* reported `confidence: high` — indistinguishable from a real header-derived read. Two byte-identical files differing only in filename (`zzscratchrr.gguf` vs `jina-reranker-v1-tiny-en.Q8_0.gguf`) landed on different capabilities (`chat` vs `rerank`), both reported as `high` confidence, even though the shared header carries three unambiguous non-chat signals: `general.architecture = jina-bert-v2`, `general.tags = ['reranker', 'cross-encoder', …]`, `<arch>.attention.causal = False`.

  Fix: `gguf_header.py` now also extracts `general.tags` and `<arch>.attention.causal`. `detect()` checks `general.tags` for an unambiguous rerank/embed token as a tie-breaker **before** falling to the filename table, so the header — not the filename — decides capability whenever it can. Confidence is now tied to *how the capability was derived*, not just whether the header parsed:
  - `pooling_type` present, or a header-tags tie-breaker fires → `confidence: high` (a read).
  - Filename token overrides the default → `confidence: medium` (a guess) — this is the actual #1838 bug: it used to report `high` here.
  - No signal anywhere → bare `chat` default → stays `high`, since that's the correct, expected answer for the overwhelming majority of causal-LLM ggufs (which never set `pooling_type` at all) — not a guess that needs hedging.

**(B) Surfacing, not fixing, the by-design fallback.** A `.gguf`-named file with no valid GGUF magic still registers as `chat` (deliberate — see `detect()`'s docstring, recoverable with `model rm`), but `confidence: low` and `gguf_header_read: failed` were computed by `detect()` and then silently dropped before the API response / CLI output. Now `add_from_path` stashes `detection_confidence` / `detection_warning` into the model's `metadata`, and `hal0 model add` prints a warning (or a confidence line) instead of a blanket "Registered" + "Next: hal0 model run" as if the detection had been a clean read.

Also nudged `model add --help` to mention the confidence/warning line and the `PUT /api/models/<id>` capability-fix workaround, since the current copy ("capabilities auto-detected", no hedge) is what the issue flagged as misleading.

## Why

Filed as hal0 issue #1838 from the rc.5 release-validation run (real box repro, `tests/release-validation/reports/v1.0.0-rc.5.md`). The root cause is `detect()` treating a filename-fallback guess as equally trustworthy as a header read, and dropping the confidence signal it already computes before it reaches any caller.

## How verified

- New regression test `test_header_tags_tiebreaker_wins_over_ambiguous_filename` reproduces the exact live A/B (byte-identical file, two filenames) and asserts both now classify identically as `rerank`/`high` via the header tags, not the filename.
- Updated the two existing filename-fallback tests (`test_filename_fallback_for_embed_without_pooling`, `test_filename_fallback_for_rerank_without_pooling`) to assert `confidence == "medium"` and `raw_hints["capability_source"] == "filename"`.
- New API test `test_add_from_path_surfaces_medium_confidence_for_filename_guess` and an added assertion on the existing bad-header test verify `add-from-path`'s response `metadata.detection_confidence` / `detection_warning`.
- New CLI tests `test_model_add_warns_on_failed_gguf_header_read` and `test_model_add_shows_medium_confidence_for_filename_guess` verify the printed warning/confidence line.
- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/` — 9973 passed, 15 skipped, 1 xfailed (pre-existing, unrelated).
- `make lint` and `uv run ruff format --check src tests` — clean.

## Left out of scope

- The issue's optional `--capability` flag for `hal0 model add` (a UX nicety, not the confidence-misreporting bug) — the docstring now points at the existing `PUT /api/models/<id>` workaround instead.
- `general.architecture` alone is not used as a capability tie-breaker (only `general.tags` / `attention.causal`) — mapping specific architecture strings to capabilities would need a maintained lookup table and is a broader, separate change.

Closes #1838

### DIFF
--- a/src/hal0/cli/model_commands.py
+++ b/src/hal0/cli/model_commands.py
@@ -445,9 +445,13 @@ def model_add(
     if warning:
         console.print(f"  [yellow]warning:[/yellow] {warning}")
     elif confidence and confidence != "high":
+        # #1838 (codex review): a medium result can come from a filename
+        # guess OR from a header signal (attention.causal=False) that
+        # contradicts the "chat" default — don't claim it's always the
+        # filename, just that it's not a confident header read.
         console.print(
             f"  [yellow]confidence: {confidence}[/yellow] "
-            "(capabilities/backends guessed from the filename, not the file header)"
+            "(capabilities/backends could not be reliably read from the file header)"
         )
     console.print(
         f"[dim]Next: hal0 model run {mid}   (or: hal0 slot edit <slot> --model {mid})[/dim]"
--- a/src/hal0/cli/model_commands.py
+++ b/src/hal0/cli/model_commands.py
@@ -401,7 +401,11 @@ def model_add(
     """Register an already-downloaded model file — capabilities auto-detected.
 
     Reads the file header to derive id, capabilities and backends; the
-    file stays where it is (no copy). Folds in the explicit-metadata flags
+    file stays where it is (no copy). When the header doesn't carry enough
+    signal (e.g. no ``pooling_type``), capabilities fall back to a filename
+    guess — watch for the confidence/warning line this command prints, and
+    fix a wrong guess with ``PUT /api/models/<id>``
+    (``{"capabilities": [...]}``). Folds in the explicit-metadata flags
     (``--id``, ``--name``, ``--license``) that the old ``model register``
     command exposed, so this command now covers both cases.
     """
@@ -430,6 +434,21 @@ def model_add(
     console.print(f"Registered [bold]{mid}[/bold] → {m.get('path', path)}")
     caps = ", ".join(m.get("capabilities", []) or []) or "—"
     console.print(f"  capabilities: {caps}")
+    # #1838: detect() may have guessed capabilities/backends from the
+    # filename alone (confidence != "high") or failed to read a valid GGUF
+    # header at all — both cases used to print the same confident success
+    # line as a header-derived "high" hit, which reads like a fact when
+    # it's a guess.
+    meta = m.get("metadata") or {}
+    confidence = meta.get("detection_confidence")
+    warning = meta.get("detection_warning")
+    if warning:
+        console.print(f"  [yellow]warning:[/yellow] {warning}")
+    elif confidence and confidence != "high":
+        console.print(
+            f"  [yellow]confidence: {confidence}[/yellow] "
+            "(capabilities/backends guessed from the filename, not the file header)"
+        )
     console.print(
         f"[dim]Next: hal0 model run {mid}   (or: hal0 slot edit <slot> --model {mid})[/dim]"
     )
--- a/src/hal0/registry/detect.py
+++ b/src/hal0/registry/detect.py
@@ -8,12 +8,26 @@ persisting.
 Detection strategy, cheapest first:
 
 1. ``.gguf`` files → :func:`hal0.registry.gguf_header.read_gguf_header`
-   to pull arch + context_length + pooling_type. Strong evidence →
-   ``confidence='high'``. The four GGUF backends are seeded:
-   ``vulkan``, ``rocm``, ``cuda``, ``cpu``. Capability is ``embed`` when
-   ``pooling_type`` is present and non-zero (llama.cpp marks pooling_type
-   = NONE as 0 for chat models, > 0 for embed/rerank pooled outputs),
-   else ``chat``.
+   to pull arch + context_length + pooling_type + tags + attention.causal.
+   The four GGUF backends are seeded: ``vulkan``, ``rocm``, ``cuda``,
+   ``cpu``. Capability is derived, cheapest/strongest signal first:
+
+   a. ``pooling_type`` present → ``rerank`` (RANK=4), ``embed``
+      (MEAN/CLS/LAST, i.e. any other non-zero value), else ``chat``.
+      ``confidence='high'`` — read directly off the header.
+   b. ``pooling_type`` absent but ``general.tags`` carries an
+      unambiguous rerank/embed token (some converters drop the pooling
+      key but keep the tags) → same, ``confidence='high'``.
+   c. Neither present → fall back to the filename token table (also
+      used by the install-time auto-scan). When the filename carries a
+      rerank/embed token, that token alone decides the outcome — a
+      *guess*, not a header read, even though the GGUF header itself
+      parsed fine — ``confidence='medium'`` (#1838: this used to report
+      ``'high'`` for a filename-only guess). When the filename carries
+      no signal either, the bare ``chat`` default applies and stays
+      ``confidence='high'`` — that's the expected, correct answer for
+      the overwhelming majority of causal chat ggufs, which don't set
+      ``pooling_type`` at all.
 2. Non-GGUF: filename heuristic only.  Keywords cover the providers we
    currently ship:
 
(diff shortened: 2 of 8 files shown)


thinmintdev:
## Review — PR #1855 (post-merge)

Verdict: **the fix is correct and lands on the root cause**, with two narrow behavioural regressions I reproduced locally and think are worth a follow-up. CI was green; I re-ran the affected suites myself, both with and without the source hunks.

### 1. Does it fix #1838's exact repro? Yes.

Issue (A) was the byte-identical A/B: `zzscratchrr.gguf` → `chat/high` vs `jina-reranker-v1-tiny-en.Q8_0.gguf` → `rerank/high`, same header. `gguf_header.py` now extracts `general.tags` + `<arch>.attention.causal`, and `detect()` consults the tags **before** the filename table, so both names land on `rerank/high` from the header. That is the reported path, not an adjacent one.

Issue (B) — `confidence` / `raw_hints.gguf_header_read` computed then dropped — is now stashed into `metadata.detection_confidence` / `metadata.detection_warning` and printed by `hal0 model add`. Verified the `--license` path too: the follow-up `PUT /api/models/<id>` returns the full model dump and `merge_update` preserves absent fields, so the warning still prints when `--license` is passed.

### 2. Root cause, not symptom. Agreed.

Tying `confidence` to *how the capability was derived* (`raw_hints.capability_source` ∈ `pooling_type | header_tags | filename | causal_conflict | default`) rather than to "did the header parse" is the right seam — it is the thing that made a filename guess indistinguishable from a header read. Keeping the bare-`chat` default at `high` is the right call and the docstring justifies it.

### 3. Do the new tests fail without the fix? Verified, not taken on faith.

Applied the merged diff to a clean tree, then reverted **only** the four `src/` files and re-ran:

```
18 failed, 50 passed
```

Every new and every modified assertion fails on the old source — including the three `test_gguf_header.py` ordering tests, the CLI trio, and all five `add_from_path` assertions. With the source restored: `68 passed in 7.85s`. No test passes vacuously.

Incidental catch worth calling out: `test_pooling_type_promoted_when_arch_comes_after` fails on old source, i.e. the parser was silently dropping `pooling_type` whenever the key preceded `general.architecture`. That is a pre-existing bug this PR also fixes.

### 4. Regressions — two, both reproduced

**(a) `feature-extraction` is too weak a token to be an "unambiguous" tie-breaker, and it outranks the filename at `high` confidence.**

`feature-extraction` is the generic HF pipeline tag any BERT-family encoder repo can carry, including rerankers uploaded without a proper `pipeline_tag`. Because the tags branch runs before the filename table and, once it fires, the filename fallback is gated off entirely (`if not pooling_present and not is_rerank and not is_embed`), a single weak tag can invert a previously-correct classification and stamp it `high` — which means the new CLI hedge prints nothing at all.

Reproduced (same fixture, before vs after this PR):

```
reranker, general.tags=["feature-extraction","sentence-transformers"],
filename "bge-reranker-v2-m3.gguf"
  before: ['rerank'] high  src=None            <- correct, via the rc.5 #1796 filename remedy
  after:  ['embed']  high  src=header_tags     <- wrong, and silently confident
```

Same shape with a causal chat gguf that happens to carry `sentence-similarity`: `chat/high` → `embed/high`.

The issue text explicitly lists the filename token table as the deliberate rc.5 remedy for #1796 under "what is by design"; this lets a generic header tag override it without a hedge. Suggested follow-up: drop `feature-extraction` (and arguably `sentence-similarity`) from the strong-tie-breaker set, or demote a weak-tag hit to `confidence="medium"` and let a filename token still win. `reranker` / `cross-encoder` / `embedding` are genuinely unambiguous; `feature-extraction` is not.

Note this is the `dc81f537` ambiguity-guard hunk — the head commit, which Codex never re-reviewed.

**(b) `pooling_present = pooling is not None` over-gates: a negative/unspecified `pooling_type` now suppresses every fallback.**

The comment says an explicit `pooling_type` is authoritative, but that only holds for values in the known enum range. `LLAMA_POOLING_TYPE_UNSPECIFIED` is `-1`, and a converter writing it as int32 yields:

```
pooling_type=-1 (int32), filename "jina-reranker-v1.gguf"
  before: ['rerank'] high  src=None
  after:  ['chat']   high  src=pooling_type    <- tags AND filename both gated off
pooling_type=-1 (int32) + general.tags=["reranker"]
  after:  ['chat']   high  src=pooling_type
```

Narrow (gguf-py's `add_pooling_type` writes uint32, so mainstream conversions won't hit it), but the one-line hardening is free: `pooling_present = isinstance(pooling, int) and pooling >= 0`.

Unchanged/fine: `pull.py:1223` only reads `context_length`; `scan_preview` surfaces `confidence` in the response but nothing in `ui/` gates on it, so the new `medium` values are display-only there.

### 5. Scope

Clean. No `CHANGELOG.md` hunk. The `commit_scan_rows` `filename_hint` addition and the `_heuristic_only` hint plumbing are sibling instances of the same defect rather than drive-by refactors, and the `model add --help` copy change is something the issue flagged by name. `--capability` correctly left out of scope with the `PUT` workaround documented.

### 6. Codex bot triage

Eight P2 inline comments across three review passes (`92c03d39`, `dc81f537`, `962ce9ac`). I checked each against the merged source rather than the PR description:

| # | Claim | Verdict |
|---|---|---|
| 1 | Causal header ignored before the chat default | **Valid, addressed** — `causal_conflict` → `medium`. Partial by design (still returns `chat`, since embed-vs-rerank is genuinely unknown); documented. Correct call. |
| 2 | `pooling_type=0` overridable by tags | **Valid, addressed** — `pooling_present` gate + regression test. See (b) for the over-correction. |
| 3 | HF snapshot symlink hides the operator filename | **Valid, addressed** — `filename_hint`, plumbed through both `add_from_path` and `commit_scan_rows`. |
| 4 | Explicit `labels` suppress the invalid-header warning | **Valid, addressed** — warning now survives `labels`; asserted in `test_add_from_path_honours_explicit_id_and_labels`. |
| 5 | Extensionless resolved blob bypasses the failed-GGUF branch | **Valid, addressed** — `claims_gguf` checks the hint's suffix too. |
| 6 | `attention.causal` before `general.architecture` never captured | **Valid, addressed** — speculative capture extended; also fixed the same latent gap for `pooling_type`. |
| 7 | Conflicting tags from both groups | **Valid, addressed** — ambiguity guard falls through to the filename. |
| 8 | CLI blames the filename for every `medium` | **Valid, addressed** — generic wording + a test that asserts "filename" is absent. |

All eight verified genuine and all eight fixed — good triage by the author. Worth noting for the record that Codex's last pass was on `962ce9ac`; the head commit `dc81f537` shipped unreviewed by the bot, and finding (a) above lives in exactly that hunk. Bot silence on a commit is not approval.

### Bottom line

Approve as merged. Suggest a small follow-up issue for (a) — the weak-tag override is the same class of defect #1838 describes (a guess wearing a header read's confidence), just relocated from the filename to `general.tags` — and fold in the one-line (b) hardening while there.

PR #1864, opened 2026-08-12
fix(agents): check chat-anchor routability against the endpoint the probe uses

thinmintdev:
Remediates blocking review findings on the already-merged **#1849**. `Refs #1831`.

#1849 fixed the shipped live-resolve config and stopped there. The independent
review found that the `HAL0_HERMES_LIVE_RESOLVE=0` path the issue explicitly
named is still broken on `main` right now, and — worse — that the regression
register entry meant to catch it passes on the default config while hiding it.

## The defect that is live on main

Under `HAL0_HERMES_LIVE_RESOLVE=0`, `_config_pairs` writes
`model.default = primary["model_id"]` (a chat slot's **raw physical** model id)
and `model.base_url = primary["backend_url"]` (that slot's own llama-server).
The preflight then asks the **hal0 gateway** about that id:

* the gateway's list route suppresses raw chat model ids in favour of the slot's
  alias row (`hal0_chat_slot_model_ids`, which explicitly does not filter on
  loaded state), and
* `get_model`'s `_resolve_virtual_model_entry` returns `None` for anything not
  starting with `hal0/`.

So the by-id route 404s an id that routes perfectly well, `_chat_model_ready`
returns False, and `chat_completions` is skipped forever with the same false
reason as before: `'<physical-id>' not loaded on the gateway`. In that mode
`_smoke_chat_completions` never touches the gateway catalog at all, which also
made #1849's new docstring claim ("matching exactly what `_smoke_chat_completions`
itself dispatches against") false precisely where it mattered.

The root cause the reviewer named is the real one: **the preflight was using a
discovery catalog to answer a routability question.** The by-id route is a better
catalog, not a routability check.

## What changed

**1. Ask the endpoint the probe will actually use.**
`_fetch_model_route_ready(model_id, base_url)` now takes hermes' `model.base_url`
— the origin `_smoke_chat_completions` POSTs `/chat/completions` to — and falls
back from the by-id route to the list route (llama-server has no by-id route but
does list what it has loaded).

**2. A skip must be earned.** The function is three-valued: `True` routes,
`False` the endpoint authoritatively says it does not, `None` inconclusive. Only
a `hal0/*` virtual against the gateway can produce `False` (the gateway resolves
those by id or nowhere). Anything inconclusive — an unknown physical id, a 5xx,
an unreachable endpoint, or a raised exception in the preflight itself — now runs
the probe and reports what really happens. A skip is a claim ("no chat model
exists yet", #1793) and a wrong one is invisible: it reads as a clean install.

The #1793 install-time behaviour is preserved and pinned by a test:
`hal0/agent` + no live slot still skips.

**3. The register entry can now fail.** `regressions.yaml`
`smoke-preflight-skips-chat-probes` probed only the default config, so it would
have reported `fixed` while the named mode stayed broken. It now probes both
resolution modes, names the pinned mode as the failing one to look for, and
requires any recorded skip reason to be cross-checked against `model.base_url`
rather than the gateway catalog. Retiered `judgment` / `promote_ready: false`:
the unit-level half is now in pytest; the rest needs a real install on a box.

## Verification

The reviewer's other complaint about #1849 was that its regression test failed
pre-fix on a **seam signature change**, not on behaviour. This one patches the
real network boundary (`urllib.request.urlopen`) and drives production wiring —
`hp._StepCtx(state=state)` with the default `InstallIO` — so it fails on the
defect itself. Against unfixed `main`:

```
FAILED tests/agents/test_hermes_provision.py::test_pinned_backend_config_does_not_skip_chat_probe
E  AssertionError: chat_completions was skipped on a live pinned backend:
   {'passed': None, 'skipped': True,
    'detail': "skipped: no chat model loaded ('qwen3-30b-a3b-instruct' not loaded on the gateway)"}
   (probed: ['http://127.0.0.1:8080/v1/models/qwen3-30b-a3b-instruct'])
```

Note the probed URL: port 8080, the gateway, while the config points the chat
probe at 8081. That is the bug verbatim.

Also added, closing the reviewer's nits: `TestFetchModelRouteReady` covers the
real default binding (`InstallIO().fetch_model_route_ready is _fetch_model_route_ready`
— nothing asserted this before, so a wiring typo would have shipped silently),
URL shape, slash preservation in virtual names, the list fallback, and each
tri-state outcome.

```
1390 passed, 1 xfailed        # tests/agents + tests/cli
ruff check / ruff format --check   clean
mypy src/hal0/agents/hermes_provision.py — 14 pre-existing errors, none in the changed lines
```

## Judgement on the non-blocking findings

* **memory_roundtrip ungating changes fresh-install output** (nit): correct as
  merged, no change. It goes through `io.mcp_memory_call` with a 5 s timeout and
  has no chat dependency; a memory layer that is not answering yet flips
  `smoke_tests` from a skip to `warn`, which is the honest report. Same hal0-api
  process `admin_tools_list` already requires. Flagged here so the next
  validation run expects it rather than files it.
* **Codex bot posted nothing on #1849** (nit): informational; nothing to fix.
  Recorded so the silence is not mistaken for a clean bot pass.

No CHANGELOG hunk (the consolidated PR owns it). Automerge deliberately not
enabled.

### DIFF
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -3301,11 +3301,13 @@ def _fetch_model_route_ready(model_id: str, base_url: str = "") -> bool | None:
     listed_status, payload = _http_status_json(f"{base}/models")
     if listed_status is None or not (200 <= listed_status < 300):
         return None
-    ids = {
-        str(entry.get("id"))
-        for entry in ((payload or {}).get("data") or [])
-        if isinstance(entry, dict) and entry.get("id")
-    }
+    entries = (payload or {}).get("data")
+    if not isinstance(entries, list):
+        # 2xx with a body we could not parse into a catalog is not evidence of
+        # an EMPTY catalog — it is no evidence at all. An empty ``data`` list
+        # is a real answer; a missing or malformed one is not.
+        return None
+    ids = {str(entry.get("id")) for entry in entries if isinstance(entry, dict) and entry.get("id")}
     if model_id in ids:
         return True
     if model_id.startswith("hal0/"):
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -3283,9 +3283,9 @@ def _fetch_model_route_ready(model_id: str, base_url: str = "") -> bool | None:
        deliberately never advertises them (#1153).
     2. On 404, ``GET {base_url}/models`` — llama-server and friends have no
        by-id route at all but do list the model they have loaded.
-    3. Still nothing: only a ``hal0/*`` name **asked of the hal0 gateway** is
-       conclusively absent — that is the one endpoint that resolves the
-       virtuals by id or not at all. Elsewhere the prefix carries no routing
+    3. Still nothing: only a **canonical virtual name asked of the hal0
+       gateway** is conclusively absent — that pair is exactly what the by-id
+       route resolves. Elsewhere the ``hal0/`` prefix carries no routing
        meaning, and any other id could simply be folded into an alias row, so
        report ``None`` and let the caller run the real probe rather than skip
        on a guess.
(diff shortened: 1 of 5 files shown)


thinmintdev:
Codex P2 (malformed `/models` payload treated as an empty catalog) — valid, fixed in 9ceb1d79. A 2xx list response with no usable `data` now returns `None` (inconclusive) rather than an empty id set, so a `hal0/*` anchor can no longer be skipped on a parse failure. A genuinely empty `data` list stays conclusive. Covered by `test_empty_catalog_is_conclusive_but_malformed_one_is_not`, which fails against the previous commit with `assert False is None`.

thinmintdev:
Codex P2 (virtual-model negatives on a non-gateway backend) — valid, fixed in 8308c8a2. The `hal0/` prefix only carries routing meaning at the hal0 gateway, so the conclusive `False` is now restricted to that endpoint; against a pinned backend an absent id stays inconclusive and the probe runs. Covered by `test_hal0_prefix_is_only_authoritative_on_the_gateway`, which fails against 9ceb1d79.

thinmintdev:
Second Codex round — all three valid, fixed in b4464057.

- **Physical `hal0/`-prefixed ids** — right, the prefix is broader than the virtuals (ADR-0023 §2) and a registry id may legitimately carry it. The conclusive negative now requires membership in the resolver's own `DEFAULT_CHAINS`, read from `hal0.normalize.resolver` rather than re-derived. Test `test_non_canonical_hal0_prefixed_id_is_inconclusive` fails against 8308c8a2.
- **Concurrent mutation of the shared box** — right, and worse than reversible: the other lanes read that config during the window. Mode B no longer re-bootstraps hermes; it drives the installed `_chat_model_ready` against a throwaway hermes home pointed at the live endpoints, so nothing shared is touched.
- **kit_version** — bumped 2 -> 3 with a changelog entry.

PR #1865, opened 2026-08-12
fix(api): evict a deleted slot's model id and stop poisoning the composite catalogue

thinmintdev:
Fix-forward for two blocking review findings on the already-merged **#1850** (`Refs #1837`). Both defects are live on `main` right now.

## 1. The issue's own repro still reproduced (incomplete fix)

`DELETE /api/slots/{name}` calls `SlotManager.delete`, which unloads the slot (`ready -> offline`) and emits **no `ready` event**. The only runtime re-prime, `_refresh_model_cache_on_ready`, gates on `data["to"] == "ready"`, and `/v1/models` (plus the dashboard's synthetic `hal0` tile) read `model_cache["hal0"]` directly rather than calling the fetch. So after deleting a slot, its model id stayed advertised — hard-404ing on dispatch, since neither a slot nor a registry entry backs it — until some *unrelated* slot went ready or `hal0-api` restarted. #1850 converted "never evicted" into "evicted opportunistically on an unrelated edge".

Fixed on the mutation itself: `DELETE /{name}` and `PUT /{name}/config` (which can rebind `model.default`) both call the new best-effort `hal0_refresh_composite_models(app)` — punch the TTL cache, re-derive the bucket.

## 2. Regression introduced by #1850 (the more urgent half)

`_fetch_hal0_composite_models` converted an `iter_configs()` exception into `cfgs = []`, cached that empty list under the 5s TTL, and `_prime_hal0_composite_cache` then wrote it over the bucket unconditionally. The pre-#1850 merge made this harmless. The **partial** case is the likelier one: `iter_configs` swallows a per-slot `SlotConfigError` and continues, so one momentarily unparseable slot TOML silently shortened the catalogue and dropped a healthy slot's model id — permanently, because `/v1/models` reads the bucket rather than the fetch, so there is no self-heal short of another ready event or a restart. (Note the existing `model_cache.hal0_composite_refresh_failed` log line could never fire — the fetch swallowed the error first.)

- `SlotManager.iter_configs_detailed()` now returns `(configs, skipped_names)`; `iter_configs()` delegates and drops the report, so its contract is unchanged.
- The fetch returns `(models, degraded)`. `degraded` = enumeration raised, or a slot was skipped.
- A degraded result is **never** cached under the TTL (so the next call retries immediately) and **never** written over a known-good bucket.
- A *complete* enumeration still REPLACES, so deletion still evicts — that is pinned by its own test.

## Non-blocking findings

- **nit — duplicated FLM multiplex derivation:** addressed. Folded both copies into one `_flm_multiplex_tags(cfg)` helper, and added `TestPrimePathDerivesTheSameTags` so the *live* prime path (not just the now-no-op seeder) is covered for both config shapes, including a re-prime that evicts a deleted slot while keeping the tags.
- **nit — `regressions.yaml` marking #1837 closed:** no change needed. That hunk is not on `main` — #1850's merge (4187e9bb) touched only `src/hal0/api/__init__.py` and `tests/api/test_upstream_dedup.py`. The register entry `v1-models-never-evicts` is unmodified and has no closed/status field.

## Verification

Every new test was run against the unfixed tree first.

Delete path, before the fix:
```
>       assert "qwen3-4b-q4_k_m" not in app.state.upstream_models.get("hal0", [])
E       AssertionError: assert 'qwen3-4b-q4_k_m' not in ['qwen3-4b-q4_k_m']
```

Degraded enumeration, before the fix:
```
E       AssertionError: assert [] == ['m-a', 'm-b']          # transient failure wiped the bucket
E       AssertionError: assert ['m-a'] == ['m-a', 'm-b']     # one skipped slot dropped m-b
```

`iter_configs_detailed` is covered against the **real** `SlotManager` over a real malformed TOML on disk (`tests/slots/test_iter_configs_detailed.py`) so the api-side double's shape is one the production path actually produces.

After: `tests/api` + `tests/slots` green, `ruff check` and `ruff format --check` clean.

No CHANGELOG hunk — a separate consolidated PR owns that.

---

## Review rounds on this PR (Codex, 5 × P2 — all verified, all fixed)

1. **Malformed id-keyed TOML not reported as skipped** (`c22114ef`) — `_iter_configured_slots` dropped a digit-stemmed `143.toml` whose name can't be recovered *before* the parse loop, so the read looked COMPLETE. It now reports pre-enumeration drops.
2. **Degraded read kept the id the caller just deleted** (`c22114ef`) — callers now pass known-dead ids (`evict=`); they're dropped even on the degraded path, and spared if the partial read shows another slot still binding them.
3. **`PATCH /{name}/defaults` not refreshed** (`c22114ef`) — `default` is a `ModelConfig` field, so that route can rebind the model id too. Same refresh.
4. **FLM multiplex tags not evicted with the slot** (`c03ff7bd`) — an FLM slot serves `embed-gemma:300m` / `whisper-v3:turbo` from the same process; `_flm_multiplex_tags(before)` now joins the eviction list.
5. **Shared id belonging to the skipped slot** (`ace73125`) — a partial read can't tell "nobody binds this" from "the binder is the file I couldn't read". The eviction stays (a ghost id hard-404s; an under-advertised id doesn't); the window closes instead. `/v1/models` re-reads the catalogue behind its 5s TTL and refreshes the bucket on a complete read, so a wrongly-dropped id returns with no slot event and no restart — which also removes the structural assumption (every mutation must hit an edge) that made #1837 possible.

Each round has its own failing-first proof in the PR comments. Final: `tests/api` + `tests/slots` → 2547 passed, 1 skipped; `ruff check` + `ruff format --check` clean; all eight CI checks green.

### DIFF
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -181,6 +181,23 @@ async def _fetch_hal0_composite_models(
     now: Callable[[], float] = time.monotonic,
     ttl_seconds: float = _HAL0_MODEL_CACHE_TTL_SECONDS,
 ) -> list[str]:
+    """Aggregated catalogue only — see :func:`_fetch_hal0_composite_models_detailed`.
+
+    Drops the ``degraded`` flag, so callers that REPLACE state from this
+    result should use the detailed form instead (#1837 follow-up).
+    """
+    models, _degraded = await _fetch_hal0_composite_models_detailed(
+        slot_manager, now=now, ttl_seconds=ttl_seconds
+    )
+    return models
+
+
+async def _fetch_hal0_composite_models_detailed(
+    slot_manager: SlotManager,
+    *,
+    now: Callable[[], float] = time.monotonic,
+    ttl_seconds: float = _HAL0_MODEL_CACHE_TTL_SECONDS,
+) -> tuple[list[str], bool]:
     """Aggregate every ready chat-capable slot's model id into one catalogue.
 
     Direct read over the live slot config — no pseudo-upstream is
@@ -204,17 +221,39 @@ async def _fetch_hal0_composite_models(
     The returned list is sorted + deduplicated and cached for
     ``ttl_seconds`` to keep the cold-start fan-out cheap while still
     picking up new slots within a handful of seconds.
+
+    Returns:
+        ``(models, degraded)``. ``degraded=True`` means the enumeration
+        did NOT see every configured slot — ``iter_configs`` raised, or
+        it skipped a slot whose TOML wouldn't parse. A degraded result is
+        a floor, not a catalogue: because callers REPLACE their bucket
+        with it, a degraded result must never be cached under the TTL nor
+        written over a known-good bucket, or one momentarily unparseable
+        slot TOML un-advertises a healthy slot's model until the next
+        refresh (and, since ``/v1/models`` reads the bucket rather than
+        this fetch, there is no self-heal short of another slot-ready
+        event or an ``hal0-api`` restart).
     """
     cached = _HAL0_MODEL_CACHE.get(_HAL0_COMPOSITE_CACHE_KEY)
     monotonic_now = now()
     if cached is not None and cached[0] > monotonic_now:
-        return list(cached[1])
+        return list(cached[1]), False
 
+    skipped: list[str] = []
     try:
-        cfgs = await slot_manager.iter_configs()
-    except Exception as exc:  # pragma: no cover — defensive
+        detailed = getattr(slot_manager, "iter_configs_detailed", None)
+        if callable(detailed):
+            cfgs, skipped = await detailed()
+        else:  # pragma: no cover — test doubles / older managers
+            cfgs = await slot_manager.iter_configs()
+    except Exception as exc:
         log.warning("upstream.hal0_composite_iter_failed", error=str(exc))
-        cfgs = []
+        # Enumeration failed outright — report an empty FLOOR, flagged
+        # degraded, and leave the TTL cache untouched so the next call
+        # retries immediately instead of serving this failure for 5s.
+        return [], True
+    if skipped:
+        log.warning("upstream.hal0_composite_partial_enumeration", skipped=list(skipped))
 
     seen: set[str] = set()
     models: list[str] = []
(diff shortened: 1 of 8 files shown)


thinmintdev:
Codex's three P2s on `6f203adf` were all real — verified each and fixed in `c22114ef`, with a failing-first test per finding.

**P2 — malformed ID-keyed TOMLs weren't surfaced as skipped.** Confirmed: `_iter_configured_slots` drops a digit-stemmed `143.toml` whose name can't be recovered (`slot.id_toml_no_name`, `continue`) *before* `iter_configs_detailed`'s parse loop, so `skipped` came back empty and the read looked COMPLETE. `_iter_configured_slots(dropped=…)` now collects those stems and the detailed enumeration folds them into its skip report — zero extra cost, no re-glob on the hot path.

**P2 — evict known deletions even when enumeration degrades.** Confirmed. `hal0_refresh_composite_models(app, evict=[…])` now carries the ids the caller *knows* it removed (the pre-delete / pre-write `model.default`), and `_prime_hal0_composite_cache` drops them from the preserved bucket on the degraded path. Guarded: an id that the partial read still shows bound to another slot survives, so two slots sharing a model id don't un-advertise each other.

**P2 — refresh after PATCHing model defaults.** Confirmed: `PATCH /{name}/defaults` writes through `update_config(name, {"model": body})` and `default` is a `ModelConfig` field. Same refresh added there.

Red-first proof, all five new cases against the unfixed tree:
```
E  AssertionError: pre-enumeration drop not reported: []          # 143.toml
E  AssertionError: assert 'qwen3-4b-q4_k_m' not in [...]          # PATCH /defaults rebind
E  AssertionError: assert 'qwen3-4b-q4_k_m' not in [...]          # DELETE + malformed sibling
E  TypeError: _prime_hal0_composite_cache() got an unexpected keyword argument 'evict'   (x2)
```
After: `tests/api` + `tests/slots` → 2545 passed, 1 skipped. `ruff check` and `ruff format --check` clean.

thinmintdev:
Codex's fourth P2 (FLM multiplex tags) was also real — fixed in `c03ff7bd`.

An FLM slot serves `embed-gemma:300m` / `whisper-v3:turbo` from the same process as its chat tag, so deleting it removes those ids too; the known-eviction list carried only `_slot_model_id(before)`, so on a degraded read the tags survived in the preserved bucket. `_refresh_composite_catalogue` now derives them with `_flm_multiplex_tags(before)` and evicts them alongside the model id. The still-bound guard in the refresh already spares a tag another live FLM slot still enables.

Red-first proof — with only the multiplex half reverted (delete-path eviction otherwise in place), deleting an FLM slot while an unrelated TOML is malformed:
```
>       assert "embed-gemma:300m" not in after
E       AssertionError: assert 'embed-gemma:300m' not in ['embed-gemma:300m', 'qwen3-4b-q4_k_m', 'whisper-v3:turbo']
```
The test also asserts the unrelated healthy slot's `qwen3-4b-q4_k_m` survives the degraded read, so the eviction stays surgical.

thinmintdev:
Codex's fifth P2 (shared id belonging to the skipped slot) is real too — the partial read genuinely cannot distinguish "nobody binds this id anymore" from "the slot that binds it is the one I couldn't read". Fixed in `ace73125`, but not by dropping the eviction: a ghost id hard-404s on dispatch, which is the reported bug, whereas a briefly under-advertised id is a discovery-surface regression. So the window gets closed instead.

`/v1/models` now re-reads the composite catalogue itself — behind the same 5s TTL cache, so it stays cheap — rather than trusting a bucket maintained purely by edges, and refreshes the bucket when the read comes back complete (a degraded read is not authoritative and falls back to the bucket). Any wrongly-dropped id returns on the first clean read, with no slot-ready event and no restart.

That also removes the structural cause behind this whole issue: the bucket was only ever correct if every mutation hit an edge, which is exactly the assumption #1837 falsified.

Red-first proof (two slots sharing `qwen3-4b-q4_k_m`, the sibling's TOML broken during the delete, then repaired):
```
>       assert "qwen3-4b-q4_k_m" in app.state.upstream_models.get("hal0", [])
E       AssertionError: assert 'qwen3-4b-q4_k_m' in []
```
`tests/api` + `tests/slots`: 2547 passed, 1 skipped. `ruff check` / `ruff format --check` clean.

PR #1866, opened 2026-08-12
fix(cli,mcp): derive slot lifecycle timeouts from the server's own budget

### LINKED ISSUE #1832: cli(slots): every mutating slot verb caps its HTTP read at 10s against a 180s server budget — load/unload/restart print ReadTimeout and exit 1 on success

### LINKED ISSUE #1857: cli/mcp: the 10s-vs-180s lifecycle timeout mismatch survives #1851 — model run, update --restart-slots, MCP admin, and a 220s constant under a 210s+ server floor
Follow-up to #1832 / PR #1851. PR #1851 gave `hal0 slot load/unload/restart/swap` a 220s client timeout, which fixes #1832's literal repro. Three call sites with the identical defect were left behind, and the constant itself does not clear the server's worst case. Two of these were raised by the Codex reviewer on #1851 and merged unaddressed; all four are verified against `main` @ `9c57a83a`.

## 1. `SLOT_LIFECYCLE_TIMEOUT_S = 220.0` is under the server's worst case

- `SlotManager.restart` (`src/hal0/slots/manager.py:1813`) = `unload` then `load`. `unload` is bounded by `_terminate_timeout_s = 30.0` (`manager.py:296`); `load` by `_HEALTH_TIMEOUT_S = 180.0` (`src/hal0/providers/container.py:589`). **210s floor**, before podman/systemd overhead — 220s leaves ~4.7% margin, not the "real headroom" the constant's comment claims.
- With `preload_evict_enabled` (default on), `load` first runs `preload_evict.admit`, which executes `await host.unload(candidate.name)` **sequentially per evicted candidate** inside the load path. Two evicted slots = +60s → **>240s for a plain `load`**; restart/swap add their own unload on top.

Net: on a memory-tight box — exactly where restart/swap are slowest — #1832's ReadTimeout-on-success still fires.

Fix shape: derive the value from the server constants (`_HEALTH_TIMEOUT_S + k * _terminate_timeout_s`) rather than hand-picking, or take #1832's own suggestion and use `timeout=None` on the endpoints that block on the state machine.

## 2. `hal0 model run` still at the 10s default

`src/hal0/cli/model_commands.py:586` posts the same blocking `/api/slots/{target}/load` with no timeout override. Worse than the verbs #1851 fixed:

- Documented path — `docs/guides/pull-and-register-models.mdx:211`: `hal0 model run qwen3-4b --slot agent --timeout 120`.
- The command's own `--timeout` (default 300s) governs only the readiness poll that follows; the 10s POST silently overrides it, and control never reaches the loop.
- The false failure prints `Check compatibility with: hal0 slot show <target>` — misleading diagnostics on a load that in fact succeeded.

`tests/cli/test_model_first_run_commands.py::test_model_run_loads_and_waits_ready` already stubs this call, so this is a one-line fix plus one assertion.

## 3. `hal0 update --restart-slots` still at the 10s default — worst instance in the tree

`src/hal0/cli/update_commands.py:477` posts `/api/updates/restart-slots` at the default. `restart_drifted_slots` (`src/hal0/api/routes/updater.py:1100`) loops `await sm.restart(name)` over **every** drifted slot in a single request: N × 210s server budget against a 10s client cap. The CLI reports failure while the sweep goes on to bounce every slot unattended.

## 4. MCP admin lifecycle tools capped at 30s

`src/hal0/mcp/admin.py:1565` — `_call_rest(timeout_s: float = 30.0)` is the single forwarder for the `slot_load` / `slot_restart` / `model_swap` MCP tools, with no per-tool override. Agents driving slot lifecycle hit the same success-reported-as-failure at 30s against the same 180s+ server budget.

## Root cause

Short generic client defaults (`api_post`'s 10s, `_call_rest`'s 30s) applied to endpoints that block until the slot state machine converges. PR #1851 patched four of seven such call sites with a hand-picked constant; the class is still open.

## Suggested scope

One PR: derive the lifecycle timeout from the server constants, export it from `hal0/cli/_shared.py`, apply it at `model_commands.py:586` and `update_commands.py:477` (the latter scaled by drifted-slot count), and give the MCP admin lifecycle tools a matching `timeout_s`. Consider `httpx.Timeout(read=..., connect=5.0)` so the widened budget doesn't also widen connect (see #1851 review, finding 5).

Found while reviewing PR #1851 for the rc.5 fix wave.

thinmintdev:
Remediates the blocking review findings on the already-merged PR #1851. That PR
fixed #1832 for four `hal0 slot` verbs by hardcoding `SLOT_LIFECYCLE_TIMEOUT_S =
220.0`; an independent review found the number does not clear the server's worst
case and that two blocking call sites were left on the 10s default. Those defects
are live on `main`, so this is a fix-forward.

Refs #1832. Closes #1857.

## What was wrong

**1. 220s does not clear the server budget.** `SlotManager.restart` is
unload-then-load: `_terminate_timeout_s` 30s + `_HEALTH_TIMEOUT_S` 180s = a 210s
floor before any podman/systemd overhead — ~5% margin, not the "real headroom"
the constant's comment claimed. Worse, `preload_evict.admit` (enabled by default)
runs `await host.unload(candidate)` sequentially per evicted candidate *inside*
the load path, so two evictions already push a plain load past 240s. On a
memory-tight box — exactly where load and swap are slowest — the
ReadTimeout-on-success still fires.

**2. `hal0 model run` was never fixed.** `model_commands.py` POSTs the same
blocking `/api/slots/{name}/load` at `api_post`'s 10s default. The command's own
`--timeout` (300s, documented in `docs/guides/pull-and-register-models.mdx`)
governs only the readiness poll that follows, so the 10s POST silently overrode
the advertised contract, control never reached the poll, and the false failure
printed a misleading `Check compatibility with: hal0 slot show <target>`.

**3. `hal0 update --restart-slots` was the worst instance and nobody flagged it.**
`update_commands.py` POSTs `/api/updates/restart-slots` at 10s, while
`restart_drifted_slots` loops `await sm.restart(name)` over every drifted slot in
one request. N × 210s of server budget against a 10s client cap: the CLI reports
failure while the sweep goes on bouncing every slot unattended.

**4. `hal0 capabilities set` is the same defect one level up** (found by the
re-review of this PR, fixed here). `POST /api/capabilities/{slot}/{child}` is not
a config write that returns immediately: `routes/capabilities.py` awaits
`CapabilityOrchestrator.apply`, which awaits `SlotManager.load` / `unload` /
`swap` inline. Both clients were still on a short generic default — the CLI on
`api_post`'s 10s, the MCP bridge on `_call_rest`'s 30s — so
`hal0 capability set voice tts --enabled` reproduced #1832 verbatim.

## What changed

New leaf module `hal0/slot_lifecycle_budget.py` is the single source of truth.
It owns the primitive bounds — `HEALTH_TIMEOUT_S`, `TERMINATE_TIMEOUT_S` — and
`providers/container.py` and `SlotManager` now import them instead of declaring
their own literals, so the server cannot retune out from under its clients.
`slot_lifecycle_timeout_s(loads=, unloads=, slots=)` derives the client budget:
the health poll, one terminate per unload the verb performs, an allowance for the
sequential pre-load evictions, a lock-wait allowance per lock-acquiring phase per
slot (a compound verb does **not** hold one lock — `SlotManager.restart` releases
after `unload` and reacquires inside `load`), and an overhead factor for
podman/systemd and the request hop.

**Every number below is derived, not literal** — they are what
`slot_lifecycle_timeout_s` returns on this branch today, and they move
automatically if the server retunes `_HEALTH_TIMEOUT_S` or `_terminate_timeout_s`:

| Verb shape | Budget as shipped |
|---|---|
| load (`slot load`, `model run`, NPU backend load) | **621s** |
| unload (`slot unload`, `slot delete`, NPU backend unload) | **345s** |
| restart / swap (`slot restart`, `slot swap`, `capabilities set`) | **966s** (16 min) |
| `update --restart-slots`, 3 drifted slots | **2898s** (48 min) |
| `update --restart-slots`, 10 drifted slots | 9660s (2.7 h) |

Read those as ceilings on a *wedged* daemon, not expected durations — a healthy
load returns in seconds and the client never waits. They are deliberately large
because the alternative is what #1832 reported: reporting failure on an operation
that succeeded. But they are also the thing to look at before approving this:
see the "silent wait" note below.

Call sites updated: the four `hal0 slot` verbs plus `slot delete`, `hal0 model
run`, `hal0 update --restart-slots` (scaled by the drift count it already
fetches, and now posting the slot names it counted so the server sweep cannot
outgrow the budget), `hal0 capabilities set`, the hal0-admin MCP lifecycle tools
(`slot_load`/`slot_unload`/`slot_restart`/`slot_delete`/`model_swap`/
`npu_backend_load`/`npu_backend_unload`/`capability_set`/`stack_apply`), and the
bundled hal0-admin MCP client timeout in both Hermes provisioning paths.

`stack_apply` is the one fan-out the bridge cannot size from its arguments:
`StackApplyEngine.converge` loads/swaps/restarts per stack entry and then
unloads every residual running slot, but the tool is handed a slug, not a slot
list. `STACK_APPLY_SLOT_ALLOWANCE = 6` is therefore a documented policy
allowance sitting next to `EVICTION_UNLOAD_ALLOWANCE`, not a derived bound — a
larger stack degrades to today's behaviour instead of regressing.
(`stack_import` was checked and does not converge, so it keeps the short
generic default.)

Two non-blocking review findings addressed while here:

- `_api_request` built `httpx.Client(timeout=<float>)`, which set connect/read/
  write/pool alike — a multi-minute read budget would have let a half-open daemon
  stall the CLI with no output. It now uses `httpx.Timeout(budget, connect=5.0)`.
- `hal0/mcp/admin.py::_call_rest` forwarded the lifecycle MCP tools at a blanket
  30s, i.e. the same success-reported-as-failure for agents. They now carry the
  derived budgets. That was follow-up issue #1857; this closes it.

Docs: `docs/reference/cli.mdx` said `model run --timeout` defaults to `300`; the
code default is now the derived load budget (622). Restated as derived rather
than as a literal that drifts again, and a new "Slot lifecycle timeouts" section
publishes the table above so an operator scripting against these commands knows
what they are signing up for.

## Deliberately not done, and what it costs

The issue's suggested poll-with-progress UX stays out of a fix-forward on a
merged regression. It is now filed as **#1870** with the numbers, because this PR
makes it materially worse rather than merely inheriting it: the silent-wait
ceiling moves from ~220s to 621s per slot op and 2898s for a three-slot sweep,
with no spinner, no state line and no elapsed counter, and
`scripts/release-test.sh:178/197/212` calls `slot load` over ssh with no
per-command timeout (`ConnectTimeout=10` bounds the handshake, not the command)
to bound it from outside.

`EVICTION_UNLOAD_ALLOWANCE = 3` remains the one constant in the budget module
with no server-side source to import — `preload_evict` has no bound on how many
candidates `admit` evicts. That is now documented in the module along with the
slack it has (ten resident slots evicting in series cost 480s against the 621s
load budget). Bounding it server-side is #1869.

## Verification

Regression tests are red without the fix and derive their floor from the *server*
modules (`providers.container._HEALTH_TIMEOUT_S`, `SlotManager._terminate_timeout_s`),
never from the client constant under test — the previous round regressed precisely
because the client number was hand-picked, and a tautological assertion would not
have caught it. Against the pre-fix tree:

```
FAILED tests/cli/test_slot_commands.py::test_slot_load_passes_lifecycle_timeout
FAILED tests/cli/test_slot_commands.py::test_slot_restart_passes_lifecycle_timeout
FAILED tests/cli/test_slot_commands.py::test_slot_swap_passes_lifecycle_timeout
E   AssertionError: timeout=220.0 is under the server's 270.0s worst case (loads=1, unloads=1, slots=1)

FAILED tests/cli/test_update_commands.py::test_restart_slots_scales_timeout_with_drifted_slot_count
FAILED tests/cli/test_model_first_run_commands.py::test_model_run_loads_and_waits_ready
E   AssertionError: blocking load POST passed no explicit timeout kwarg

FAILED tests/mcp/test_admin.py::test_slot_lifecycle_tools_forward_with_the_lifecycle_budget
```

The `capabilities set` fix carries the same proof. With only
`src/hal0/cli/capabilities_commands.py` and `src/hal0/mcp/admin.py` reverted and
the tests untouched:

```
FAILED tests/cli/test_capabilities_commands.py::test_capability_set_passes_the_lifecycle_timeout[argv0]
FAILED tests/cli/test_capabilities_commands.py::test_capability_set_passes_the_lifecycle_timeout[argv1]
FAILED tests/cli/test_capabilities_commands.py::test_capability_set_passes_the_lifecycle_timeout[argv2]
E   AssertionError: capability set passed no explicit timeout kwarg
E   assert 'timeout' in {}

FAILED tests/mcp/test_admin.py::test_slot_lifecycle_tools_forward_with_the_lifecycle_budget
E   AssertionError: capability_set forwarded at 30.0s, under the 270.0s server worst case
E   assert 30.0 >= 270.0
```

`stack_apply` likewise, with only `src/hal0/mcp/admin.py` and
`src/hal0/slot_lifecycle_budget.py` reverted:

```
FAILED tests/mcp/test_admin.py::test_slot_lifecycle_tools_forward_with_the_lifecycle_budget
E   AssertionError: stack_apply forwarded at 30.0s — it converges a whole stack
    sequentially, so it cannot ride a single slot's 270.0s
```

The three CLI cases are parametrized over the three orchestrator branches
(`--enabled` → load, `--disabled` → unload, `--model` → swap) because one request
cannot know which one the server will take and so has to budget for the worst.

All green after the fix. `ruff check` and `ruff format --check` clean across
`src` and `tests`.

No CHANGELOG hunk — the consolidated release PR owns that.

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -156,7 +156,7 @@ selected with `--hardware`, and stored on the slot as `device` (`gpu-rocm`, `gpu
 | `model scan` | — | — |
 | `model add <path>` | `path` | `--id` (default `""`), `--name`, `--license` (default `""`), `--overwrite` |
 | `model store [path]` | optional `path` | `--migrate` |
-| `model run <ref>` | `ref` | `--slot`/`-s` (default `""`), `--timeout` (default `300`) |
+| `model run <ref>` | `ref` | `--slot`/`-s` (default `""`), `--timeout` (default: the derived slot-load budget from `hal0.slot_lifecycle_budget`, `622` at time of writing — see [slot lifecycle timeouts](#slot-lifecycle-timeouts)) |
 | `model import-backup <path>` | `path` (required) | `--force`/`-f`, `--dest` (default `/var/lib/hal0/registry/registry.toml`) |
 
 ## `hal0 memory` — manage hal0 memory (Hindsight engine)
@@ -358,6 +358,29 @@ dashboard/API-only (`/api/profiles`).
 | `board show <task_id>` | `task_id` | `--json` |
 | `board add <title>` | `title` | `--body` (default `""`), `--assignee` (default `""`), `--status` (default `triage`), `--board`, `--priority` (default `0`) |
 | `board move <task_id> <status>` | `task_id`, `status` | — |
+
+## Slot lifecycle timeouts
+
+Commands that drive the slot state machine — `slot load` / `unload` / `restart`
+/ `swap` / `delete`, `model run`, `capabilities set`, `update --restart-slots` —
+block until the server converges the slot. Their client read timeouts are not
+hand-picked; they are derived from the server's own bounds in
+`hal0.slot_lifecycle_budget` (the health poll, one container terminate per
+unload, an allowance for the pre-load evictions and for queueing behind an
+in-flight op on the same slot, plus a podman/systemd overhead factor).
+
+| Verb shape | Budget |
+|---|---|
+| load (`slot load`, `model run`, NPU backend load) | 621s |
+| unload (`slot unload`, `slot delete`, NPU backend unload) | 345s |
+| restart / swap (`slot restart`, `slot swap`, `capabilities set`) | 966s |
+| `update --restart-slots`, per drifted slot | 966s (3 slots = 2898s) |
+
+These are ceilings on a *wedged* daemon, not expected durations — a healthy
+load returns in seconds. But the commands print nothing while they wait, so a
+genuinely stuck slot can hold the terminal for the full budget. If you are
+scripting against them (CI, smoke tests), wrap the call in your own
+`timeout(1)` rather than relying on the CLI to give up first.
 
 ## Deprecated aliases
 
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -55,9 +55,17 @@ from typing import Any
 import structlog
 
 from hal0.agents.role_slots import candidate_from_slot_mapping, resolve_role_slots
+from hal0.slot_lifecycle_budget import slot_lifecycle_timeout_s
 from hal0.system import seam as _seam
 
 log = structlog.get_logger(__name__)
+
+#: Per-server MCP client timeout for ``hal0-admin``. The catalog exposes the
+#: blocking slot lifecycle tools (slot_load / slot_restart / model_swap /
+#: npu_backend_load), so a 60s client cap aborted an agent-driven load long
+#: before the server gave up — the outer half of #1832. Derived from the same
+#: budget the tools themselves forward with, rounded up to whole seconds.
+_HAL0_ADMIN_MCP_TIMEOUT_S = int(slot_lifecycle_timeout_s(loads=1, unloads=1)) + 1
 
 # Canonical last-run-report location. Lives outside $HERMES_HOME — Hermes owns
 # its own tree, and the report must survive a `hermes reset`. This is no longer
@@ -1323,7 +1331,7 @@ def _default_mcp_servers() -> list[dict[str, Any]]:
             "url": "http://127.0.0.1:8080/mcp/admin/mcp",
             "type": "http",
             "private": False,
-            "timeout": 60,
+            "timeout": _HAL0_ADMIN_MCP_TIMEOUT_S,
             "usage_hint": (
                 "query/manage hal0 platform state (slots, services, models, hardware). "
                 "Use when the operator asks about system state or wants to inspect a slot."
@@ -3122,7 +3130,7 @@ def _build_brain_profile_mcp_servers() -> dict[str, Any]:
             "type": "http",
             "url": "http://127.0.0.1:8080/mcp/admin/mcp",
             "headers": admin_headers,
-            "timeout": 60,
+            "timeout": _HAL0_ADMIN_MCP_TIMEOUT_S,
         },
         "hal0-memory": {
             "type": "http",
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -55,17 +55,26 @@ from typing import Any
 import structlog
 
 from hal0.agents.role_slots import candidate_from_slot_mapping, resolve_role_slots
-from hal0.slot_lifecycle_budget import slot_lifecycle_timeout_s
+from hal0.slot_lifecycle_budget import STACK_APPLY_SLOT_ALLOWANCE, slot_lifecycle_timeout_s
 from hal0.system import seam as _seam
 
 log = structlog.get_logger(__name__)
 
 #: Per-server MCP client timeout for ``hal0-admin``. The catalog exposes the
 #: blocking slot lifecycle tools (slot_load / slot_restart / model_swap /
-#: npu_backend_load), so a 60s client cap aborted an agent-driven load long
-#: before the server gave up — the outer half of #1832. Derived from the same
-#: budget the tools themselves forward with, rounded up to whole seconds.
-_HAL0_ADMIN_MCP_TIMEOUT_S = int(slot_lifecycle_timeout_s(loads=1, unloads=1)) + 1
+#: npu_backend_load / capability_set / stack_apply), so a 60s client cap aborted
+#: an agent-driven load long before the server gave up — the outer half of
+#: #1832. It must cover the WORST tool in that catalog, not a representative
+#: one: derived from the restart shape, this outer client still truncated
+#: stack_apply, whose fan-out budget is several times larger.
+#: Computed from the same shape ``hal0.mcp.admin`` budgets its widest tool
+#: (stack_apply) with. It is NOT imported from that table: hal0.mcp.admin
+#: imports this module's package, so reaching into it here is a circular
+#: import that silently breaks MCP mounting. ``test_hal0_admin_mcp_timeout_
+#: covers_the_widest_bridge_tool`` pins the two against drift.
+_HAL0_ADMIN_MCP_TIMEOUT_S = (
+    int(slot_lifecycle_timeout_s(loads=1, unloads=1, slots=STACK_APPLY_SLOT_ALLOWANCE)) + 1
+)
 
 # Canonical last-run-report location. Lives outside $HERMES_HOME — Hermes owns
 # its own tree, and the report must survive a `hermes reset`. This is no longer
--- a/src/hal0/cli/_shared.py
+++ b/src/hal0/cli/_shared.py
@@ -16,6 +16,20 @@ _console = Console(stderr=True)
 
 NOT_IMPLEMENTED = "not implemented yet — see PLAN.md §13"
 
+#: TCP connect bound, kept short and independent of the per-call read budget.
+#: A local hal0-api either accepts immediately or is not there.
+_CONNECT_TIMEOUT_S = 5.0
+
+
+def _client_timeout(timeout: float) -> httpx.Timeout:
+    """Spend ``timeout`` on the read, not on connect/write/pool.
+
+    ``httpx.Client(timeout=<float>)`` sets every phase to that value, so the
+    multi-minute slot lifecycle budgets (#1832) would also let a half-open or
+    wedged daemon hang the CLI for minutes before the first byte.
+    """
+    return httpx.Timeout(timeout, connect=min(_CONNECT_TIMEOUT_S, timeout))
+
 
 def _api_base() -> str:
     """Return the hal0 API base URL, honouring HAL0_API_URL env override."""
@@ -230,8 +244,12 @@ def _api_request(
         headers.update(_auth_headers())
     if headers:
         kwargs["headers"] = headers
+    # ``timeout`` is the READ budget: the slot lifecycle call sites pass
+    # minutes because the server blocks until the slot converges (#1832).
+    # Widening connect to match would let a half-open daemon stall the CLI for
+    # that whole budget with no output, so connect stays short.
     try:
-        with httpx.Client(timeout=timeout) as client:
+        with httpx.Client(timeout=_client_timeout(timeout)) as client:
             resp = client.request(method, url, **kwargs)
     except httpx.HTTPError as exc:
         raise CliApiError(f"{method} {url} failed: {type(exc).__name__}: {exc}") from exc
--- a/src/hal0/cli/capabilities_commands.py
+++ b/src/hal0/cli/capabilities_commands.py
@@ -48,6 +48,17 @@ from hal0.capabilities.orchestrator import _CHILD_TO_CAPABILITY, LEGAL_SLOTS, le
 from hal0.cli._shared import CliApiError, _api_base, _api_unreachable, api_get, api_post, die
 from hal0.config.locking import file_lock
 from hal0.registry.store import ModelRegistry
+from hal0.slot_lifecycle_budget import slot_lifecycle_timeout_s
+
+#: ``POST /api/capabilities/{slot}/{child}`` awaits the slot state machine
+#: inline: ``routes/capabilities.py`` calls ``CapabilityOrchestrator.apply``,
+#: which — depending on which fields changed — awaits ``SlotManager.load``
+#: (off→on), ``SlotManager.unload`` (on→off) or ``SlotManager.swap`` (model or
+#: backend change while on). The client cannot know which branch the server
+#: will take, so it budgets for the worst one: ``swap``, i.e. unload-then-load
+#: (#1832). Without this the command rides ``api_post``'s 10s default and
+#: reports failure on a capability change that in fact succeeded.
+CAPABILITY_SET_TIMEOUT_S = slot_lifecycle_timeout_s(loads=1, unloads=1)
 
 app = typer.Typer(
     name="capabilities",
@@ -161,7 +172,11 @@ def set_capability(
     if _api_unreachable(url):
         raise typer.Exit(1)
     try:
-        result = api_post(f"/api/capabilities/{slot}/{child}", json=body)
+        result = api_post(
+            f"/api/capabilities/{slot}/{child}",
+            json=body,
+            timeout=CAPABILITY_SET_TIMEOUT_S,
+        )
     except CliApiError as exc:
         die(str(exc))
         return
--- a/src/hal0/cli/model_commands.py
+++ b/src/hal0/cli/model_commands.py
@@ -21,6 +21,7 @@ from hal0.cli._shared import (
 )
 from hal0.cli.registry_commands import DEFAULT_REGISTRY_PATH, _do_import_backup
 from hal0.registry.import_toml import import_toml_to_sqlite
+from hal0.slot_lifecycle_budget import slot_lifecycle_timeout_s
 
 app = typer.Typer(help="Manage the local model registry.")
 console = Console()
@@ -582,8 +583,18 @@ def model_run(
         console.print(f"[dim]No --slot given; using slot [bold]{target}[/bold].[/dim]")
 
     # 3. Load with the model assigned.
+    #
+    # /api/slots/{name}/load blocks until the slot converges (#1832), so this
+    # POST needs the derived lifecycle budget: at api_post's generic 10s
+    # default the CLI died with a misleading "check compatibility" hint on a
+    # load that had in fact succeeded, and control never reached the readiness
+    # poll below that --timeout advertises.
     try:
-        api_post(f"/api/slots/{target}/load", json={"model_id": ref})
+        api_post(
+            f"/api/slots/{target}/load",
+            json={"model_id": ref},
+            timeout=slot_lifecycle_timeout_s(loads=1, unloads=0),
+        )
     except CliApiError as exc:
         die(f"{exc}\nCheck compatibility with: hal0 slot show {target}")
         return
--- a/src/hal0/cli/model_commands.py
+++ b/src/hal0/cli/model_commands.py
@@ -535,7 +535,12 @@ def model_run(
         "", "--slot", "-s", help="Slot to run on (default: the first compatible slot)"
     ),
     timeout_s: int = typer.Option(
-        300, "--timeout", help="Seconds to wait for the slot to become ready."
+        # The default clears the server's own worst case for a load; a smaller
+        # explicit value is honoured end to end, including by the blocking POST
+        # below (#1832).
+        int(slot_lifecycle_timeout_s(loads=1, unloads=0)) + 1,
+        "--timeout",
+        help="Seconds to wait for the slot to become ready.",
     ),
 ) -> None:
     """Get a model serving: pull if needed, assign to a slot, load, wait ready.
@@ -588,12 +593,15 @@ def model_run(
     # POST needs the derived lifecycle budget: at api_post's generic 10s
     # default the CLI died with a misleading "check compatibility" hint on a
     # load that had in fact succeeded, and control never reached the readiness
-    # poll below that --timeout advertises.
+    # poll below that --timeout advertises. --timeout is the operator's whole
+    # wait budget, so it caps this POST too: `--timeout 30` must give up at
+    # ~30s, not sit in the request for minutes first.
+    load_budget = min(slot_lifecycle_timeout_s(loads=1, unloads=0), float(max(timeout_s, 1)))
     try:
         api_post(
             f"/api/slots/{target}/load",
             json={"model_id": ref},
-            timeout=slot_lifecycle_timeout_s(loads=1, unloads=0),
+            timeout=load_budget,
         )
     except CliApiError as exc:
         die(f"{exc}\nCheck compatibility with: hal0 slot show {target}")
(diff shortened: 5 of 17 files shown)


thinmintdev:
Codex review triaged — all five findings verified against the code and fixed in 32e17f0a:

- **P1 hal0-admin MCP timeout (60s)** — confirmed at `hermes_provision.py` both in `_default_mcp_servers()` and `_build_brain_profile_mcp_servers()`. Both now use `_HAL0_ADMIN_MCP_TIMEOUT_S`, derived from the same budget the tools forward with. Regression test asserts against the server modules' own bounds and is red at 60s.
- **P1 npu_backend_load / npu_backend_unload** — confirmed: `POST /api/backends/npu/{load,unload}` awaits `SlotManager.load`/`unload` in-request. Both added to the lifecycle timeout map. The test approves the gated call so it actually reaches REST (the first draft of it passed on a stale captured timeout — caught and fixed).
- **P2 honour --timeout on the blocking POST** — right; `--timeout 5` would have sat in the POST for minutes. The POST is now `min(load budget, --timeout)`, and the option default rises to the derived load budget so the default path is still safe.
- **P2 lock-wait allowance** — right, the holder can be a full load including its evictions, not just the health poll. `LOCK_WAIT_ALLOWANCE_S` is now the full load cost.
- **P2 bind the sweep to the counted slots** — right, `restart_drifted_slots` recomputes drift. The CLI now posts `{"slots": [...]}` with the names it counted, so the server sweep cannot outgrow the budget derived from that count.

Budgets after the change: load 621s, unload 345s, restart/swap 655.5s, three-slot restart sweep 1725s.

Local: `tests/cli tests/mcp tests/agents tests/slots` 2507 passed; `ruff check` and `ruff format --check` clean.

thinmintdev:
Second Codex round triaged — three fixed in 05f9f375, two deliberately declined with reasons.

**Fixed**

- **slot_delete needs the unload budget (both surfaces)** — confirmed: `SlotManager.delete` awaits `unload` for a running slot. The CLI `slot delete` was on `api_delete`'s 10s default and the MCP tool on the 30s forward default; both now carry the unload budget. Tests are red without the fix on both surfaces.
- **Lock-wait allowance must scale across the sweep** — right, `restart_drifted_slots` takes each slot's lock separately. The allowance is now charged per slot, so a three-slot sweep is 1966.5s rather than 1242s.
- **Start the readiness deadline before the load POST** — right; `--timeout 100` could approach 200s. The deadline now opens before the POST and both phases draw on it. The regression test drives a fake clock: the load burns 95s, and the command must give up within the operator's 100s rather than starting a fresh window.

**Declined, with reasons**

- **"Account for every possible preload eviction"** — correct that `preload_evict.admit` is bounded only by the resident-slot count, and the module says so. But a *finite* client read timeout cannot cover an unbounded server chain; any number is a policy choice. `EVICTION_UNLOAD_ALLOWANCE = 3` is named, documented, and multiplied by the overhead factor and the per-slot lock allowance on top. Deriving it per call would mean an extra `GET /api/slots` before every lifecycle verb for a bound that is still only an estimate. The real fix is non-blocking lifecycle endpoints, which is out of scope for a fix-forward on a merged regression — see below.
- **"Bound startup subprocesses before using a finite budget"** — verified and agreed: `ContainerProvider._write_and_start_unit` runs `daemon-reload` / `reset-failed` / `systemctl restart` with `timeout=None`, so one server-side phase genuinely has no bound. That is a server-side change to the slot start path, not a client timeout, and papering over it with a still-larger client number would be the same mistake #1851 made. Filed as #1869 with both candidate shapes (bound the subprocesses, or move the lifecycle endpoints to a job handle).

Local: `tests/cli tests/mcp tests/agents tests/slots tests/api` — 4239 passed, 1 skipped, 1 xfailed. `ruff check` and `ruff format --check` clean.

thinmintdev:
Third Codex round — one fixed, one answered in fcbdf127.

- **Charge restart for both lock acquisitions** — confirmed at `manager.py`: `restart` releases the lock after `unload` and `load` reacquires it, so a request can pay two lock waits. The allowance is now charged per lock-acquiring phase per slot; restart/swap go from 655.5s to 966s. Regression test derives the two-lock floor from the server constants and is red on the one-allowance formula.
- **Propagate terminate-timeout overrides** — the drift is real but not fixable in-process: the CLI and the MCP bridge are separate processes, so nothing in them can observe a running `SlotManager` instance's attribute override. Rather than pretend otherwise, `_terminate_timeout_s` now documents that the deployment knob is the module constant in `hal0.slot_lifecycle_budget` — retune there and every client inherits it; override the attribute and you re-open the drift. That keeps one source of truth, which is the whole point of the module.

Budgets now: load 621s, unload 345s, restart/swap 966s, three-slot restart sweep 2898s. These are ceilings on a wedged path, not expected waits.

Local: `tests/cli tests/mcp tests/agents` 1183 passed; `ruff check` and `ruff format --check` clean.

thinmintdev:
Triage of the three Codex findings on `1bedf9f1` / `9012b0d0`:

**P1 `stack_apply` missing from the lifecycle map — accepted and fixed** (`f9b825d3`).
Verified against the code: `POST /api/stacks/{slug}/apply` awaits
`StackApplyEngine.converge`, which sequentially drives `load` / `swap` /
`restart` per stack entry (`stacks/apply.py:472/475/481`) and then unloads every
residual running slot (`:541`), all inside the one request. 30s could not cover
a single unload. It is a fan-out like `/api/updates/restart-slots`, but the
bridge receives a slug rather than a slot list, so the count cannot be derived
without an extra `GET`; `STACK_APPLY_SLOT_ALLOWANCE = 6` is a documented policy
allowance sitting next to `EVICTION_UNLOAD_ALLOWANCE`, and a bigger stack
degrades to today's behaviour rather than regressing. Red-without-fix proof, with
only the production files reverted:

```
FAILED tests/mcp/test_admin.py::test_slot_lifecycle_tools_forward_with_the_lifecycle_budget
E   AssertionError: stack_apply forwarded at 30.0s — it converges a whole stack
    sequentially, so it cannot ride a single slot's 270.0s
```

`stack_import` was checked and does **not** converge — it writes the envelope
only — so it correctly stays on the generic default.

**P2 "account for the global admission-lock wait" — declined, with reasoning.**
The observation is correct: `preload_evict.admit` can hold `_admission_lock`
across another caller's spawn and readiness poll, so a load can queue behind a
per-slot lock *and* the global one. But that chain has no finite bound to charge
— the number of waiters is the number of concurrent lifecycle callers, which is
unbounded — so charging it means picking another arbitrary multiplier and
pushing an already 621s budget higher for a case that only bites under
concurrent lifecycle traffic on one box. This is the same argument as the
earlier "account for every possible preload eviction" finding: no finite client
read timeout can cover an unbounded server-side chain. The real fix is
server-side boundedness (#1869), and the operator-visible cost of these ceilings
is now filed as #1870.

**P2 "bound NPU cleanup before applying an unload timeout" — declined, same
reason, and it is already tracked.** Correct that
`api/routes/backends.py` awaits `proc.wait()` with no timeout after
`SlotManager.unload()`, so a wedged `systemctl reset-failed` outlives any client
budget. That is precisely #1869 ("bound the systemd/podman start subprocesses so
lifecycle waits are finite server-side"), which this PR filed. Widening the
client number does not fix an unbounded server phase — it just moves where the
false failure lands. Not doing it here.

thinmintdev:
## Blocking review finding addressed (9f74fa84)

The re-review flagged that this PR split connect from read in the CLI (`hal0.cli._shared._client_timeout`) but did not carry that split to the MCP bridge whose timeouts it widened from 30s to up to 5796s.

**1. `src/hal0/mcp/admin.py` — connect bound.** `httpx.AsyncClient(base_url=..., timeout=timeout_s)` passed a bare float, so the multi-minute lifecycle budgets applied to TCP connect as well as read. Against a half-open hal0-api, `stack_apply` would hang for up to 96 minutes where it previously failed in 30s. Now `httpx.Timeout(timeout_s, connect=min(_CONNECT_TIMEOUT_S, timeout_s))` with a 5s connect bound, mirroring the CLI.

**2. `_HAL0_ADMIN_MCP_TIMEOUT_S` — outer client.** It was derived from the restart shape (967s) while the catalog it fronts reaches 5796s for `stack_apply`, so the outer client truncated the widest blocking tool it is responsible for. Now derived from the same fan-out shape.

Note on how: it deliberately does **not** import the bridge's `_SLOT_LIFECYCLE_TIMEOUT_S` table. `hal0.mcp.admin` imports `hal0.agents`, so reaching back the other way is a circular import — I tried it first and it broke MCP mounting outright (`cannot import name 'build_server' from partially initialized module 'hal0.mcp.admin'`), while still producing the right number, i.e. it would have passed a value-only test. It now derives from the stdlib-only `slot_lifecycle_budget` leaf, with `test_hal0_admin_mcp_timeout_covers_the_widest_bridge_tool` pinning the two against drift.

### Verification

Both new tests confirmed red without the production change:

```
FAILED tests/mcp/test_admin.py::test_lifecycle_budget_bounds_the_read_not_the_connect
```

`tests/mcp tests/agents tests/cli`: 1719 passed, 1 xfailed. `ruff check` and `ruff format --check` clean on all four touched files.

The `mock_transport` fixture now records `connect_timeout` alongside `timeout`, with `timeout` still meaning the read budget so the existing per-tool assertions read unchanged.

The other items from the re-review (budget table figures, `cli.mdx` 622 default, the silent-wait ceiling note, the `EVICTION_UNLOAD_ALLOWANCE` slack note) were already applied in the PR body and docs before this commit.

thinmintdev:
## Independent review of 9f74fa84: approve, no blocking findings

Reviewed by a separate agent that reverted each production change and confirmed the tests fail:

- Revert the connect split → `AssertionError: connect bound is 966.0s, expected the short 5.0s bound`
- Revert the constant to the old restart shape → `assert 5796.0 <= 967`
- The `mock_transport` fixture change makes **no** existing assertion vacuous: all 14 existing `timeout` assertions are read-budget assertions, and `httpx.Timeout(t, connect=…)` sets `read = t`. The fixture keeps its bare-float `else` branch, which is why the revert fails loudly rather than raising `KeyError`.
- Full checks on a clean tree: `tests/mcp tests/agents tests/cli` 1719 passed / 1 xfailed; `tests/api` 1734 passed; ruff check and format clean.

### The cost this change carries, stated plainly

`_HAL0_ADMIN_MCP_TIMEOUT_S` is a **per-server** timeout for the entire `hal0-admin` catalog, not a per-tool one. Raising it 967s → 5797s so it stops truncating `stack_apply` also means a **wedged read-only tool** (`slot_list`, `hardware_info`) can now stall an agent for 96 minutes instead of 16.

That is the unavoidable consequence of "cover the widest tool" under a per-server timeout. It is the right trade now that connect is bounded at 5s — a wedged *socket* still fails in 5s, so the exposure is narrowed to a server that accepts the connection and then never answers. But the commit message presented this purely as a fix and never named the cost, so recording it here.

The real fix, if this becomes a problem in practice, is a per-tool client timeout on the MCP server side rather than one per-server cap sized to the worst tool.

### Circular import

The commit message's wording ("`hal0.mcp.admin` imports `hal0.agents`") is loose — the cycle is indirect: `admin.py:176` imports `hal0.api._redact` → `api/__init__.py:2578` runs `create_app()` at import time → `:2481` lazily imports `build_server` from the still-partially-initialized `hal0.mcp.admin`. The conclusion stands and the reviewer reproduced the exact error. Worth knowing that `api/__init__.py:2490` **swallows** it into `app.state.mcp_mount_error`, so this failure mode is silent, and the whole suite (3453 tests) passes with the cycle present. The comment at `hermes_provision.py:70-74` is currently the only thing stopping a future change from re-adding that import.

### Follow-ups filed

- **#1873** — `brain/chat.py:1097` and `:828` carry the identical bare-float defect, and `:828`'s 60s *read* is under the 621s `slot_load` floor, i.e. its own #1832-class truncation on a path the original issue never covered.

Two nits not acted on: `_CONNECT_TIMEOUT_S = 5.0` is now duplicated in `cli/_shared.py` and `mcp/admin.py` (a third occurrence would justify hoisting it next to `slot_lifecycle_budget`), and `_call_rest` bypasses `hal0.http_client.async_client` and so pays the per-construction SSL-context build — irrelevant for `http://127.0.0.1`.

Issue #1829, opened 2026-08-11
brain chat: every board tool 401s on a fresh install — the pinned hermes wheel ships no web_dist, so the session token can never be harvested


thinmintdev:
## What happens

On a fresh 1.0.0-rc.5 install, every hal0-api → Hermes kanban forward goes out with **no credential at all** and gets 401. The brain steward's entire board tool surface is dead.

Exact chain, all four links verified on ct152:

1. `installer/agents/hermes/requirements.txt:16` pins `hermes-agent[web] @ git+…@9de9c25f` = version **0.18.2**, whose wheel does **not** ship `hermes_cli/web_dist` (grep of `hermes_agent-0.18.2.dist-info/RECORD` for `web_dist` → 0 hits). hermes-agent **0.19.0** does ship it (21 RECORD hits on ct150).
2. With no `web_dist`, `hermes_cli/web_server.py` takes its no-frontend branch: `GET http://127.0.0.1:9119/` → `404 {"error":"Frontend not built. Run: cd web && npm run build"}`.
3. `HermesKanbanClient._fetch_html_token()` (`src/hal0/board/__init__.py:168-181`) harvests the rotating per-process bearer **only** by scraping `window.__HERMES_SESSION_TOKEN__` out of that HTML. On 404 it returns `None`, and the docstring's own words apply: "the caller then forwards without a token and Hermes's 401 is surfaced honestly." Neither `HERMES_SESSION_TOKEN` (`/etc/hal0/api.env`) nor `HERMES_DASHBOARD_SESSION_TOKEN` (`/etc/hal0/agents/hermes.env`) is provisioned by the installer, so there is no pin to fall back to.
4. `request_json()` 401s on the first attempt and again on the force-refresh retry, then raises `BoardUpstreamError("hermes kanban returned an error")`.

Blast radius: all four read tools (`get_board`, `get_task`, `get_assignees`, `get_orchestration`) route through this client, as do all twelve board mutation tools. Shipped config is `[brain_chat] enabled = true, read_only = true`, so the four reads are the entire board surface the steward has — and all four are dead.

## Second, larger defect found while verifying

[...]

PR #1872, opened 2026-08-12
fix(brain): read the hal0 board store, not the Hermes kanban proxy

thinmintdev:
## What changed

The brain steward's board tools now run against the hal0-owned `BoardStore` —
the same store `GET/POST /api/board/*` serves from — instead of forwarding to
`app.state.hermes_kanban`. This finishes the KB-4 rewire (`e18c25ab`), which
moved the REST routes onto the local store but left `brain/chat.py` on the
proxy.

* new `hal0.board.store.resolve_store(app)` is the ONE resolver both surfaces
  go through, so "the steward and the UI answer from the same board" is
  structural, not a convention two call sites can drift apart on;
* the four reads (`get_board`, `get_task`, `get_assignees`, `get_orchestration`)
  call the store methods the REST GET handlers call;
* the eleven mutations call the store methods the REST mutation handlers call,
  still each wrapped in the same `board.chat.turn` audit row;
* the chat's board-backend precondition is gone: hal0 owns the board, so the
  steward serves with Hermes absent exactly as `/api/board/*` does.

The board store is imported lazily inside `_board_store()`, so the brain module
still has no board import at module scope and no Hermes-shaped import at all.

## Why

Issue #1829. On a fresh 1.0.0-rc.5 install every board tool 401s: the pinned
hermes wheel ships no `web_dist`, so the HTML session-token scrape finds
nothing and the forward goes out with no credential. But the credential is only
half the defect — with a valid token the steward would answer from the Hermes
kanban plugin while `hal0 board list`, `GET /api/board/board`, the board page
and the `/events` WS all answer from the hal0 store. **Fixing only the token
would trade a loud 401 for silent divergence.** This PR is the architectural
half; the credential half (pin bump + provisioned bearer) stays open on #1829,
hence `Refs`, not `Closes`.

## How it was verified

`tests/brain/test_brain_board_store.py` is the regression: every assertion is
cross-surface (write on one surface, read on the other), so a "returns 200"
pass is impossible. Run against the UNFIXED code (call adapted to the old
signature), it fails on exactly the two defects in the report:

```
E       AssertionError: steward answered from a different board: {'tasks': [{'id': 't_hermes', 'title': 'a DIFFERENT board'}]}
E       assert 't_0c78b62d' in ['t_hermes']
```
```
E           RuntimeError: hermes kanban returned an error (401 Unauthorized)
src/hal0/brain/chat.py:1017: in _dispatch_tool
    result = await client.request_json(
```

With the fix: 10/10 pass, and `tests/brain` + `tests/board` are green (392
tests). `ruff check` + `ruff format --check` clean.

## Follow-up left out of scope

* the `hermes-agent` pin bump / `HERMES_DASHBOARD_SESSION_TOKEN` provisioning
  (the credential half of #1829) — the operator is handling that personally;
* the first-boot import in `BoardStore.ensure_initialized` still forwards to
  Hermes unauthenticated; it already degrades to "seed a clean empty board", but
  it is the remaining unauthenticated forward and belongs to that half;
* no CHANGELOG hunk — a separate consolidated PR owns that.

Refs #1829


thinmintdev:
### Codex review triage

All three findings compare the new store path against the OLD Hermes forward. The bar this PR is held to is parity with the `/api/board/*` handler each tool now mirrors — that is the whole point of the rewire (the steward must not behave differently from the surface the operator is looking at). Checked each against `src/hal0/api/routes/board.py`:

1. **P1 "keep dispatcher nudges scoped to the selected board"** — not actionable here. `POST /api/board/dispatch` calls `store.dispatch_nudge(max_dispatch=...)` with no board either, and `BoardStore.dispatch_nudge`'s ready-card query has no board column filter to pass one to. Scoping it is a store change that must move both surfaces together; scoping only the chat would recreate exactly the divergence this PR removes. Commented in place.

2. **P2 "resolve omitted assignee scope to the current board"** — parity, verbatim: the route is `store.list_assignees(board=_board(request))`, and `_board()` is `None` when `?board=` is omitted. The chat now passes the same `None`. If omitted-board-means-current is the intended contract for assignees, it is a store/route fix that both callers inherit.

3. **P1 "run synchronous board operations off the event loop"** — parity again: every `/api/board/*` mutation handler calls the synchronous store directly inside its async handler, `dispatch_nudge` included. The chat doing the same is the existing accepted pattern, not a new blocking path introduced here. If the executor seam needs a thread, it moves for both.

No code behaviour changed in response; a7de47ca records the three parity points in comments so the next reader does not re-derive them.

Verified after: `tests/brain` + `tests/board` 392 passed, full local suite 10133 passed, `ruff check` + `ruff format --check` clean, CI green.

thinmintdev:
## Blocking review finding addressed (df244c66)

The re-review found that `tests/brain/test_brain_board_store.py` — the regression this PR body quotes the output of — was **not on the branch**. Its history:

- added in `c2f20717`
- deleted by `e58f6fd6` ("style: sort `__all__`", -291 test lines)
- re-added by `80ec1fa1`
- deleted again by `a7de47ca` ("docs(brain): note where the board tools deliberately match the REST routes", -291 test lines)

Both deletions were incidental to commits whose messages describe unrelated work, which is why they went unnoticed. `HEAD` ran 382 tests where the body claimed 392 — exactly the 10 that were gone.

Restored from `80ec1fa1`. This is the only proof the steward and the REST surface resolve the *same* store rather than both merely returning 200: it writes via `POST /api/board/tasks` and reads back through the brain's `get_board` tool, writes via chat and reads via `GET /api/board/board`, and asserts `app.state.hermes_kanban.calls == []` against a 401-raising kanban stub.

### Verification

Reverting only `src/hal0/brain/chat.py` to the merge base (`687d27e8`), leaving all tests untouched:

```
FAILED tests/brain/test_brain_board_store.py::test_board_reads_work_with_hermes_absent
FAILED tests/brain/test_brain_board_store.py::test_board_reads_survive_a_401ing_hermes
FAILED tests/brain/test_brain_board_store.py::test_chat_create_task_lands_on_the_rest_board
FAILED tests/brain/test_brain_board_store.py::test_chat_move_task_moves_the_rest_card
FAILED tests/brain/test_brain_comment_and_block_land_on_the_rest_card
FAILED tests/brain/test_brain_board_store.py::test_chat_update_orchestration_lands_on_the_rest_surface
FAILED tests/brain/test_brain_board_store.py::test_brain_and_routes_resolve_one_store_object
10 failed in 0.26s
```

With the fix in place: `tests/brain tests/board` = **392 passed**, matching the figure the body advertised.

Scope is unchanged and still excludes the credential half — no `hermes-agent` pin edit, no token generation. `Refs #1829`, so the issue stays open for the operator's half.

Issue #1867, opened 2026-08-12
hermes: an upgraded box keeps an old slot ceiling below Hermes' floor, and nothing preflights the anchor window


thinmintdev:
Split out of #1827 (PR #1852). #1852 fixes the **fresh-install** half of that
blocker: a slot's on-disk `[model].context_size` ceiling is now honored as the
cap for the derived window, so the shipped 65536 seeds actually reach
llama-server and `/v1/models`. It does **not** fix the ct150 half, and #1827
must stay open until this one lands.

## The failure #1852 does not touch

From the ct150 comment on #1827 (rc.3 → rc.5 in-place upgrade, CPU-only box):

```
hal0 model show hal0-brain-sft-q8-rocmfpx   -> defaults.context_size = 96000
/etc/hal0/slots/agent.toml                  -> context_size = 4096
hal0/agent  -> resolves to the agent slot   -> 4096  -> below Hermes' 64,000 floor -> refuse
```

The model already declares 96000, comfortably above the floor. The agent slot's
own ceiling is 4096, so the effective window is 4096 and Hermes hard-errors.
Making the dense cap ceiling-aware (#1852) changes nothing here — the ceiling is
the thing that is too low, and a ceiling can only clamp down.

## Two defects behind it

**1. In-place upgrades never reconcile a slot seed the operator never chose.**
The shipped rc.5 seeds all say `context_size = 65536`:

```
/usr/lib/hal0/current/installer/etc-hal0/slots/{agent,brain,utility}.toml -> 65536
```

but the live ct150 box still has `agent.toml -> 4096`, mtime 2026-08-09,
untouched by the upgrade. An in-place upgrade never revisits an existing slot
TOML, so a box seeded by an older release keeps its old ceiling forever. That is
correct for an operator-tuned value and wrong for a value the operator never
chose. This needs an explicit decision — provenance tracking on seeded values, a
one-shot reconcile for untouched seeds, or a `hal0 doctor --fix` repair — not an
implicit "whatever is on disk wins".

**2. Nothing checks the anchor window before handing it to Hermes.**
The refusal surfaces ~1.6 s later as an upstream agent error with no pointer at
the slot that caused it. A preflight on the resolved window behind `hal0/agent`
— refuse (or repair) with an actionable message naming the slot, its ceiling,
and the floor it is under — would turn a mystery upstream error into a
one-line fix instruction. `hal0 doctor` is the natural home; the smoke/preflight
path is the other candidate.

[...]

PR #1877, opened 2026-08-13
fix(hermes): preflight the anchor context window against Hermes' floor

### LINKED ISSUE #1827: hermes: fresh install cannot chat — the installer-pulled brain model resolves to a 32,768 ctx, below Hermes' hard 64,000 floor

thinmintdev:
## What

Preflights the context window behind Hermes' anchor model and fails loudly, with
both numbers and the repair command, when it is under Hermes' hard floor.

`Refs #1867` — this is the **first half** only. The upgrade-time reconciliation of
a slot seed the operator never chose (defect 1 in the issue) stays open.

## Why

Hermes raises below `agent/model_metadata.py::MINIMUM_CONTEXT_LENGTH` (64,000)
rather than degrading. A slot's effective window is
`min(model's declared window, slot's [model].context_size ceiling)`, so either
half can put a box under the floor. #1852 fixed the model side for fresh
installs; it cannot help the ct150 shape, where the model declares 96000 and the
agent slot's own ceiling is still the 4096 an older release seeded — a ceiling
only ever clamps down.

Nothing checked this before handing the anchor to Hermes, so the operator saw an
opaque upstream agent error ~1.6 s into every turn that named neither the slot
nor the ceiling behind it, and reasonably concluded rc.6 had not fixed #1827.

## What changed

- **New `hal0.agents.anchor_window`** — a pure, fully-injected resolver: anchor
  model id → effective window (from `GET /v1/models/{id}`, which resolves the
  routing fallback chain the same way chat dispatch does) → the slot the gateway
  actually served it from → that slot's on-disk ceiling → verdict
  (`ok` / `below_floor` / `unknown`) + an operator-facing message.
- **New `anchor_context_window` smoke probe** in `hermes_provision`, next to
  #1831's routability preflight. Below the floor it fails with:

  ```
  anchor 'hal0/agent' resolves to slot 'agent' with an effective context window
  of 4,096 tokens, below the 64,000 tokens Hermes requires — Hermes raises
  rather than degrading, so EVERY turn will fail until this is raised. The limit
  is this slot's own configured ceiling: [model].context_size = 4,096 in
  /etc/hal0/slots/agent.toml. Fix: hal0 slot edit agent --ctx-size 65536 &&
  hal0 slot restart agent
  ```

  When the ceiling is *not* the binding constraint the message blames the model
  instead and deliberately does **not** hand over the `--ctx-size` command,
  which would be useless advice there.
- **`unknown` is never a pass.** A window that cannot be read (nothing loaded,
  endpoint unreachable, a backend advertising no `context_length`) reports
  `skipped` — from the probe itself, not only from the install-time gate — and
  the message says in words that it is not a pass. The probe also joins
  `_MODEL_DEPENDENT_PROBES` so an install-time run skips before it dials.
- **A `hal0 doctor all` row (`hermes_anchor_window`).** `installer/install.sh`
  calls `install_hermes` only when the hermes venv is ABSENT ("covers the
  upgrade path where no provision ran"), so on an in-place upgrade — the one
  shape a stale `[model].context_size` survives in — no smoke test runs at all.
  Below the floor is a `fail` with the repair command; a window that cannot be
  read is a `warn`; an unprovisioned Hermes passes.
- **The floor is read FROM Hermes** — `read_hermes_minimum_context()` asks the
  provisioned Hermes venv's interpreter for its own `MINIMUM_CONTEXT_LENGTH`
  (Hermes is not importable from hal0's venv, so this is the only honest read)
  and degrades to one pinned constant with the reason recorded, which the message
  discloses. The two pre-existing test-side copies of `64_000` (#1827/#1852) now
  import that constant, and a drift test compares it against the real Hermes on
  any host with a provisioned venv.

## Explicitly out of scope

Automatically rewriting the operator's slot ceiling during an update. That is a
config mutation nobody has approved and is the other half of #1867.
`resolve_anchor_window` makes that repair easier when it is decided: it already
names the slot, its ceiling, and (via `recommended_ceiling`/`fix_command`) the
exact value the repair would write.

## Round 2 — the two blocking findings

**1. It never ran on the in-place-upgrade path.** `_phase_smoke_tests` is only
reachable from `install_hermes()`, which the installer skips when the hermes
venv already exists. Fixed by the `hal0 doctor all` row above — the tool an
operator reaches for when the agent misbehaves.

**2. It green-passed a box that genuinely cannot chat.** Reproduced on live
ct152 (rc.5): the agent slot is offline, so `hal0/agent` resolves through the
fallback chain to the **brain** slot's 32,768-token window (under the 64,000
floor), while `GET /v1/models` lists only `brain` and `utility` — no `agent`
row at all. The alias lookup found nothing, called the window `unknown`, and
returned `passed=True`.

Both halves are fixed: the window now comes from the by-id route, and the
serving slot is identified from the gateway's own construction (the by-id row
is the resolved slot's catalog row with `id` rewritten) rather than from the
anchor's spelling — so the fix command names `brain`, not the innocent
`agent.toml` sitting right next to it on that box. Where the slot is ambiguous
or unprovable there is no slot, no ceiling and no command. And `unknown` now
reports SKIPPED.

Codex's P1 is covered too: the anchor is asked of `model.base_url` (the
endpoint Hermes dispatches to) rather than a gateway catalog that deliberately
suppresses a pinned physical id, and a raw physical model id no longer derives
a slot TOML path that was never on the box.

## Verification

**Round 2.** The ct152 regression drives the real `_phase_smoke_tests` over
payloads captured verbatim from 10.0.1.152, so the expectation comes from the
live gateway, not from anything hal0's resolver computes. Against the code as
it stood before this round:

```
>       assert result["passed"] is not True, result  # the whole point of #1867
E       AssertionError: {'detail': "anchor 'hal0/agent' (slot 'agent') advertises
E         no context window right now — cannot check it against Hermes'
E         64,000-token floor", 'passed': True}
E       assert True is not True
FAILED tests/agents/test_hermes_provision.py::TestAnchorContextWindowPreflight::test_ct152_the_anchor_falls_back_to_another_slot_and_must_not_pass
FAILED tests/agents/test_hermes_provision.py::TestAnchorContextWindowPreflight::test_an_unreadable_window_skips_rather_than_passing
FAILED tests/agents/test_anchor_window.py::TestCt152FalseNegative::test_the_by_id_route_resolves_the_chain_and_names_the_serving_slot
FAILED tests/agents/test_anchor_window.py::TestCt152FalseNegative::test_a_virtual_is_never_named_from_its_spelling
FAILED tests/agents/test_anchor_window.py::TestCt152FalseNegative::test_two_identical_rows_are_ambiguous_not_a_coin_flip
5 failed, 1 passed
```

**Round 1.** Regression test run against **unfixed** code (`hermes_provision.py` reverted to
`github/main`, `anchor_window.py` removed):

```
    def test_a_slot_ceiling_under_the_floor_fails_loudly_with_the_fix(...):
        state = self._ct150_box(tmp_path, monkeypatch, slot_ceiling=4096, advertised=4096)
        out = hp._phase_smoke_tests(hp._StepCtx(state=state))
>       result = out.details["results"]["anchor_context_window"]
E       KeyError: 'anchor_context_window'
FAILED ...::test_a_slot_ceiling_under_the_floor_fails_loudly_with_the_fix
FAILED ...::test_a_ceiling_above_the_floor_passes
FAILED ...::test_no_routable_anchor_skips_instead_of_inventing_a_verdict
3 failed in 0.39s
```

With the fix: `tests/agents tests/install tests/providers` → 1476 passed,
1 skipped (the Hermes-venv drift check, absent on a dev host), 1 xfailed.
`ruff check` and `ruff format --check` clean; `mypy` clean on the new module.

The `TestAnchorContextWindowPreflight` tests drive the **real** `_phase_smoke_tests`
over a fake network and a real slot TOML on disk — no hand-stamped result shapes.

### DIFF
--- a/src/hal0/agents/anchor_window.py
+++ b/src/hal0/agents/anchor_window.py
@@ -1,0 +1,305 @@
+"""Preflight the context window behind Hermes' anchor model (#1867).
+
+Hermes refuses to run at all when the model it is pointed at advertises a
+context window below ``agent/model_metadata.py::MINIMUM_CONTEXT_LENGTH``
+(64,000 at the pinned commit) — it raises rather than degrading. hal0's side of
+that contract is the EFFECTIVE window of the slot behind the anchor
+(``hal0/agent`` by default), which is::
+
+    effective = min(model's native/declared window, slot's [model].context_size)
+
+so either half can put the box under the floor:
+
+* the MODEL side — a registry row with no ``defaults.context_size`` resolving
+  to a small derived window (#1827's fresh-install half, fixed by #1852), and
+* the SLOT side — a configured ceiling that is itself below the floor. An
+  in-place upgrade never revisits an existing ``/etc/hal0/slots/<slot>.toml``,
+  so a box seeded by an older release keeps e.g. ``context_size = 4096``
+  forever even though its model happily declares 96000. #1852's "the ceiling
+  wins over the blanket dense cap" change cannot help there: a ceiling only
+  ever clamps DOWN.
+
+Nothing used to check this before handing the anchor to Hermes, so the second
+shape surfaced ~1.6s into the first turn as an opaque upstream agent error that
+named neither the slot nor the ceiling behind it. This module is the check: it
+resolves the anchor's effective window, compares it against Hermes' own floor
+(read FROM the installed Hermes when possible, see
+:func:`read_hermes_minimum_context`), and renders a message an operator can act
+on without reading any source — which slot, that slot's configured ceiling, the
+required floor, and the exact command that repairs it.
+
+Deliberately read-only. Rewriting an operator's slot ceiling during an update is
+a config mutation nobody has approved yet and is tracked as the other half of
+#1867; :func:`resolve_anchor_window` is the resolver that repair would be built
+on when it is decided (it already names the slot, its ceiling, and the target
+value the repair would need to write).
+"""
+
+from __future__ import annotations
+
+import subprocess  # nosec B404 — asks the Hermes venv python for its own constant
+import tomllib
+from collections.abc import Callable, Mapping
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import structlog
+
+log = structlog.get_logger(__name__)
+
+#: hal0's ONLY copy of Hermes' hard context floor — a fallback for when the
+#: installed Hermes cannot be asked (see :func:`read_hermes_minimum_context`).
+#: Mirrors ``agent/model_metadata.py::MINIMUM_CONTEXT_LENGTH`` in the bundled
+#: Hermes (``[tool.hal0.upstream-hermes]`` pin in pyproject.toml). Hermes is not
+#: importable from hal0's own venv, so this cannot be a plain re-export; the
+#: drift guard is ``tests/agents/test_anchor_window.py``, which asserts every
+#: hal0-side copy of the number resolves to this constant and compares against
+#: the real Hermes constant whenever a Hermes install is reachable.
+HERMES_MINIMUM_CONTEXT_LENGTH = 64_000
+
+#: Where the slot TOMLs an operator (or the installer) writes ceilings into live.
+SLOTS_DIR = Path("/etc/hal0/slots")
+
+#: Prefix hal0's virtual model ids carry (``hal0/agent`` → slot ``agent``).
+_VIRTUAL_PREFIX = "hal0/"
+
+_FLOOR_PROBE = "from agent.model_metadata import MINIMUM_CONTEXT_LENGTH as m; print(int(m))"
+
+
+def read_hermes_minimum_context(
+    venv_python: str | Path | None,
+    *,
+    run: Callable[..., Any] = subprocess.run,
+) -> tuple[int, str]:
+    """Return ``(floor, source)`` — Hermes' own floor when it can be asked.
+
+    hal0 runs from its own venv and never imports Hermes, so the number is read
+    the only way it honestly can be: by asking the Hermes venv's interpreter for
+    ``agent.model_metadata.MINIMUM_CONTEXT_LENGTH``. Any failure (no venv yet, an
+    older Hermes without the constant, a Hermes that cannot import) degrades to
+    :data:`HERMES_MINIMUM_CONTEXT_LENGTH` with the reason recorded in ``source``,
+    because a preflight that raises is worse than one that uses the pinned value.
+
+    ``source`` is ``"hermes"`` when the live constant was read, else
+    ``"fallback:<why>"`` — so a message built off the fallback can say so.
+    """
+    if not venv_python:
+        return (HERMES_MINIMUM_CONTEXT_LENGTH, "fallback:no-venv")
+    try:
+        proc = run(  # nosec B603 — fixed argv, no shell, hal0-owned venv path
+            [str(venv_python), "-c", _FLOOR_PROBE],
+            check=False,
+            capture_output=True,
+            text=True,
+            timeout=15,
+        )
+    except (OSError, subprocess.SubprocessError) as exc:
+        return (HERMES_MINIMUM_CONTEXT_LENGTH, f"fallback:{type(exc).__name__}")
+    if getattr(proc, "returncode", 1) != 0:
+        return (HERMES_MINIMUM_CONTEXT_LENGTH, "fallback:probe-failed")
+    try:
+        value = int(str(getattr(proc, "stdout", "") or "").strip())
+    except (TypeError, ValueError):
+        return (HERMES_MINIMUM_CONTEXT_LENGTH, "fallback:unparsable")
+    if value <= 0:
+        return (HERMES_MINIMUM_CONTEXT_LENGTH, "fallback:non-positive")
+    return (value, "hermes")
+
+
+def anchor_slot_name(model_name: str) -> str:
+    """Slot alias behind an anchor model id (``hal0/agent`` → ``agent``).
+
+    A bare id (a slot pinned directly, e.g. under
+    ``HAL0_HERMES_LIVE_RESOLVE=0``) is already the alias.
+    """
+    name = (model_name or "").strip()
+    if name.startswith(_VIRTUAL_PREFIX):
+        return name[len(_VIRTUAL_PREFIX) :]
+    return name
+
+
+def read_slot_ceiling(slot: str, *, slots_dir: Path = SLOTS_DIR) -> int | None:
+    """The slot's on-disk ``[model].context_size`` ceiling, or ``None``.
+
+    ``None`` covers every "no usable ceiling" case alike — no TOML, no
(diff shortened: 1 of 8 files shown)


thinmintdev:
## Round-3 regression addressed (520ab801)

The re-review found a new, nondeterministic regression introduced by this branch: `_row_identity` (`src/hal0/agents/anchor_window.py`) compared every field except `id` — **including `created`**.

The gateway stamps `created = int(time.time())` per request (`api/routes/v1.py:840,893`), and `_fetch_anchor_models` issues the by-id read and the `/v1/models` read as **two separate requests**. Any pair straddling a second boundary compared unequal, so `serving_slot` returned `None` for a `hal0/*` anchor and the `below_floor` message silently lost both the slot name and the repair command.

That is precisely the deliverable of #1867 — "told exactly what to change and by which command" — degrading at random. Measured on an **idle** ct152: 1 of 40 paired requests mismatched (~2.5%), and the window scales with by-id latency against what `doctor_all.py` documents as the slowest read-only route hal0 serves.

### Fix

Identity now ignores volatile fields (`id`, `created`) via `_VOLATILE_ROW_FIELDS`, with a comment recording why `created` cannot participate.

### Verification

Reverted only the production change (`if k not in _VOLATILE_ROW_FIELDS` → `if k != "id"`), tests untouched:

```
FAILED tests/agents/test_anchor_window.py::TestCt152FalseNegative::test_a_created_skew_still_names_the_serving_slot
E  AssertionError: a one-second created skew lost the serving slot
1 failed, 27 deselected
```

With the fix: `tests/agents/test_anchor_window.py` 27 passed; `tests/agents` 704 passed, 1 skipped, 1 xfailed. `ruff check` and `ruff format --check` clean.

The test pins the real ct152 payloads and asserts `hal0 slot edit brain --ctx-size 65536` still reaches the operator when the by-id row's `created` is one second ahead of the catalog row's.

Issue #1740, opened 2026-08-09
SECURITY P1: hal0-systemctl write-unit/write-quadlet take an unvalidated root unit body → root RCE from hal0 account


thinmintdev:
Found by the pre-GA verification review (adversarially confirmed). #1718 hardened the two drop-in verbs, but write-unit/write-quadlet in the same case statement (installer/wrappers/hal0-systemctl:224-246) install /dev/stdin verbatim as a root-owned unit/quadlet with ZERO validation. Sudoers grant is pinned to the binary, not argv, so the hal0 account can write a [Service] ExecStartPre=/bin/sh -c 'id>/root/pwned' quadlet, daemon-reload, start → unconditional root RCE. Every other seam hardening sits behind this open door, and the wrapper/sudoers comments claim the opposite.

Fix: allow-list the rendered unit/quadlet shape (providers/container.py:_render_quadlet_text emits a fixed key set) rejecting unexpected [Service] directives; OR explicitly document hal0-account compromise as root-equivalent and stop advertising the seam as a boundary. PRIVILEGED SURFACE — operator review required.

PR #1881, opened 2026-08-15
chore(security): document the bundled agent trust boundary for 1.0 (ADR-0002 Option C)

thinmintdev:
Lands ADR-0002 **Option C** items 1–3 (see #1880 for the full analysis and the
evidence — this PR does not re-derive it).

## Why

`hal0-api.service` and `hal0-agent@<instance>.service` both ship `User=hal0`.
Same UID, so the bundled agent — provisioned with `terminal.backend = local`,
i.e. it has a shell — can read the API process's `/proc/<pid>/environ` and with
it every credential the services hold. `kernel.yama.ptrace_scope` does not
prevent this: it gates `PTRACE_MODE_ATTACH`, not same-UID reads. The `hal0`
account is additionally root-equivalent by design — slots run under rootful
podman, so whatever can author a `[Container]` spec can `Volume=`-mount host
paths, which `installer/wrappers/hal0-systemctl`'s own HONEST BOUNDARY block has
stated since #1740. It has simply never been said anywhere an operator reads.

This PR does not change that posture. It documents it, surfaces it in
`hal0 doctor`, and takes the one free tightening that was available.

## What changed

1. **`SECURITY.md` — "Bundled agent trust boundary".** States the posture
   plainly, then a control-by-control table of what each compensating control
   *buys* and what it *does not*: unprivileged `User=hal0` (not root, but the
   same UID as the API), `ProtectSystem=strict`/`PrivateTmp`/`ReadWritePaths=`
   (integrity, not confidentiality against a same-UID read), `0600 root:root`
   on the per-agent env file, `0600` on `api.env` (useless against its owner).
   One honest qualifier added on top of the ADR: `hal0-agent@.service` sets
   `NoNewPrivileges=yes`, so a process inside the agent unit cannot invoke the
   setuid sudo wrappers *directly* — a speed bump, not a boundary, because
   `hal0-api.service` sets no such restriction and the admin key read out of
   `/proc` drives the same privileged paths through the API. Status is stated
   outright: accepted, documented risk for 1.0, agent-UID split tracked for
   1.1, link to ADR-0002, plus what to do if you cannot accept it.
2. **`hal0 doctor` row `agent-uid` / "Agent UID split"**
   (`src/hal0/cli/doctor_all.py`). Warns — never fails, this is the shipped
   default — when a `hal0-agent@*` unit resolves to the same `User=` as
   `hal0-api.service`, says in one line what that implies, and points at the
   SECURITY.md section and the ADR. Both probes (`systemctl show -p User`,
   `systemctl list-units 'hal0-agent@*.service'`) are injectable seams,
   mirroring `check_hal0_target`, so classification is tested without systemd.
   No agent unit installed → pass; `systemctl` unable to answer → warn
   "unknown", not a silent pass.
3. **`upstreams.toml` → `0640`** (`src/hal0/install/perms.py`). Metadata only,
   but free.

## Consumer check before the mode change (item 3)

Asked and answered before touching the `PermRow`:

- `grep` for `upstreams.toml` / `UPSTREAMS` across `src`, `installer`,
  `packaging`, `tests`: the readers are `src/hal0/config/loader.py`
  (`load_upstreams_config`), `src/hal0/api/__init__.py` startup priming,
  `src/hal0/api/routes/providers.py` CRUD, and `hal0 config show/edit`.
- Those run in `hal0-api.service` (`User=hal0`), `hal0-agent@*.service`
  (`User=hal0`) or the CLI (root, or `hal0`). The file is `hal0:hal0`, so
  `0640` is readable by all of them.
- No container mounts `/etc/hal0` — searched every `Volume=`/bind-mount site in
  `src`, `installer`, `packaging`; OpenWebUI and the slot Quadlets do not touch
  it.
- Writers: `installer/install.sh` (`[[ ! -f ]]`-guarded create, default umask)
  and `write_toml_atomic`; the perms engine converges the mode either way.
- **The one behaviour change worth naming:** an operator running the `hal0` CLI
  as their *own* unprivileged account, not in group `hal0`, could previously
  `hal0 config show upstreams`. After this they need `sudo` or group
  membership. ADR-0002 Q5 flagged exactly this and found no in-tree dependency;
  I found none either, but it is out-of-repo-script-shaped, so it is called out
  here.
- Stale comment in `installer/systemd/hal0-agent@.service` (which quoted
  "upstreams.toml 0644") updated to match.

## Regression tests — each fails without its fix

`tests/install/test_perms.py::test_upstreams_toml_is_not_world_readable`, with
the `perms.py` hunk reverted:

```
>       assert row.mode & 0o007 == 0, f"upstreams.toml is world-readable: {row.mode:o}"
E       AssertionError: upstreams.toml is world-readable: 644
E       assert (420 & 7) == 0
1 failed, 33 deselected
```

`tests/docs/test_security_md_documents_agent_boundary.py`, with the
`SECURITY.md` hunk reverted (5/5 fail):

```
E       ValueError: substring not found   # "## Bundled agent trust boundary"
FAILED test_security_md_has_a_bundled_agent_section
FAILED test_section_names_the_shared_uid_and_the_proc_read
FAILED test_section_names_compensating_controls_and_their_limits
FAILED test_section_declares_the_status_and_links_the_adr
FAILED test_section_points_at_the_doctor_row
5 failed
```

`tests/cli/test_doctor_all.py`, with the `doctor_all.py` hunk reverted (8 new
row tests + the pre-existing composition test):

```
tests/cli/test_doctor_all.py:803: AttributeError
FAILED test_agent_uid_shared_with_api_warns
FAILED test_agent_uid_split_passes
FAILED test_agent_uid_no_agent_unit_passes
FAILED test_agent_uid_unknown_api_user_warns
FAILED test_agent_uid_flags_only_the_sharing_instance
FAILED test_agent_uid_probes_degrade_without_systemctl
FAILED test_unit_user_reports_root_for_an_empty_user_directive
FAILED test_agent_units_parses_list_units_output
FAILED test_build_all_checks_composes_verify_plus_extras
9 failed, 61 passed
```

The doc tests assert the *content* of `SECURITY.md` directly rather than
reimplementing a rule; the perms test asserts the shipped ownership table, not a
copy of it.

## Verification

- `uv run ruff check src tests` → All checks passed
- `uv run ruff format --check src tests` → 1142 files already formatted
- `pytest tests/docs tests/cli/test_doctor_all.py tests/install/test_perms.py tests/cli/test_doctor_perms.py` → 123 passed
- No box was written to. No secret value was read, printed, or committed.

## Out of scope

- ADR-0002 item 4 (the agent's missing-`HAL0_MCP_TOKEN` degraded state) — not
  requested here, still tracked in the ADR.
- The ADR file itself lands with #1880; the `SECURITY.md` link resolves once
  that merges.
- No CHANGELOG hunk — #1860 owns the rc.6 section. Suggested line:
  `Added: SECURITY.md now documents the bundled agent's trust boundary — the agent shares hal0-api's UID and can therefore reach every credential the box holds — plus a hal0 doctor "Agent UID split" row reporting the posture, and upstreams.toml tightened to 0640 (ADR-0002).`

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -66,8 +66,17 @@ the agent can `cat`. The axis that matters is which UID the agent runs as.
 
 This is the shipped default for hal0 1.0 and an accepted risk, disclosed here
 rather than papered over. Giving the agent its own system user with no sudoers
-grants — the change that actually closes the chain — is tracked for 1.1. The
-analysis, the evidence, and the rejected alternatives are in
+grants is tracked for 1.1, but a new `User=` alone does not close the chain:
+`/etc/hal0` is `hal0:hal0 2775` (setgid, see
+`hal0.install.perms.ownership_table()`), so a split agent left in the shared
+`hal0` *group* could still create or replace `hal0.toml`/`api.env` there —
+and an unreadable or malformed `hal0.toml` makes the API's own auth gate
+(`hal0.api.auth._config_require_auth()`) treat `require_auth` as unset and
+fall back to its OFF default, reopening the exact privileged-API path this
+document warns about. Closing the chain for 1.1 means the UID split *and*
+dropping the agent out of the `hal0` group / config-tree write access — not
+the UID split alone. The analysis, the evidence, and the rejected alternatives
+are in
 ADR-0002 — agent credential isolation, under review in
 [PR #1880](https://github.com/Hal0ai/hal0/pull/1880) and landing at
 `docs/adr/0002-agent-credential-isolation.md`. (Linked as a PR, not as a
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -51,7 +51,7 @@ injection from content it reads — it can reach every credential the box holds.
 | `User=hal0` (unprivileged, not root) | The agent is not UID 0; kernel-level and other-user data stay out of reach | Any separation from `hal0-api`, which is the same UID |
 | systemd sandboxing on the units (`ProtectSystem=strict`, `PrivateTmp`, restricted `ReadWritePaths=`) | Integrity: the agent cannot rewrite `/usr`, `/boot`, or most of `/etc` | Confidentiality of anything the `hal0` user may read, including `/proc` of same-UID processes |
 | `NoNewPrivileges=yes` on `hal0-agent@` | The agent's own processes cannot invoke the setuid `sudo` wrappers | Anything about `hal0-api`, which sets no such restriction and is reachable with the admin key the agent can read |
-| `0600 root:root` on `/etc/hal0/agents/<instance>.env` | The agent cannot read *another* agent's MCP token | Protection of `api.env`-sourced values, which are already in the API's environment |
+| `0600 root:root` on `/etc/hal0/agents/<instance>.env` | The agent cannot *open another instance's file*, and cannot create, replace or unlink one | Confidentiality of that instance's `HAL0_MCP_TOKEN`: with two agent instances running as the same user, either can read the other's `/proc/<pid>/environ`, so the file mode does not isolate the token itself |
 | Owner-only `0600` on `/etc/hal0/api.env` | Other local accounts cannot read it | Anything against the agent, which is the owner |
 
 Moving secrets into a `root:root 0600` file loaded by pid1 does **not** close
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -24,12 +24,16 @@ runs as. Two consequences follow, and neither is hypothetical:
   `PTRACE_MODE_READ_FSCREDS`, which a same-UID process passes. Yama's
   `kernel.yama.ptrace_scope` does not prevent it — that knob gates
   `PTRACE_MODE_ATTACH`, not same-UID reads. So every secret in the API's
-  environment (`HAL0_ADMIN_KEY`, provider API keys, `HF_TOKEN`, the agent's own
-  session token) is reachable by the agent regardless of the mode or ownership
-  of the file those values were loaded from. This has been verified on a live
-  box, not merely reasoned about.
+  environment (`HAL0_ADMIN_KEY`, provider API keys, `HF_TOKEN`) is reachable by
+  the agent regardless of the mode or ownership of the file those values were
+  loaded from. The chat-proxy session secret is not an API env var — it is a
+  file at `/var/lib/hal0/agents/secret.bin` (`0600`, owned by `hal0`) — but the
+  same same-UID logic applies to it too: the agent is the file's owner, so it
+  can simply read the file directly, no `/proc` trick required. This has been
+  verified on a live box, not merely reasoned about.
 - **The `hal0` account is root-equivalent by design.** The sudo grants hal0
-  installs (`hal0-systemctl`, `hal0-agentenv`, `hal0-benchctl`, `hal0-update`)
+  installs (`hal0-systemctl`, `hal0-agentenv`, `hal0-benchctl`, `hal0-podman-ro`,
+  `hal0-update`)
   are issued to the *user* `hal0`, which the agent also is. Slots run under
   rootful podman, so anything that can author a slot's `[Container]` spec can
   `Volume=`-mount arbitrary host paths — `installer/wrappers/hal0-systemctl`'s
@@ -49,7 +53,7 @@ injection from content it reads — it can reach every credential the box holds.
 | Control | Buys | Does not buy |
 | --- | --- | --- |
 | `User=hal0` (unprivileged, not root) | The agent is not UID 0; kernel-level and other-user data stay out of reach | Any separation from `hal0-api`, which is the same UID |
-| systemd sandboxing on the units (`ProtectSystem=strict`, `PrivateTmp`, restricted `ReadWritePaths=`) | Integrity: the agent cannot rewrite `/usr`, `/boot`, or most of `/etc` | Confidentiality of anything the `hal0` user may read, including `/proc` of same-UID processes |
+| systemd sandboxing on `hal0-agent@` (`ProtectSystem=strict`, `PrivateTmp`, restricted `ReadWritePaths=`) | Integrity: the agent cannot rewrite `/usr`, `/boot`, or most of `/etc` | Confidentiality of anything the `hal0` user may read, including `/proc` of same-UID processes; `hal0-api.service` sets none of these directives |
 | `NoNewPrivileges=yes` on `hal0-agent@` | The agent's own processes cannot invoke the setuid `sudo` wrappers | Anything about `hal0-api`, which sets no such restriction and is reachable with the admin key the agent can read |
 | `0600 root:root` on `/etc/hal0/agents/<instance>.env` | The agent cannot *open another instance's file*, and cannot create, replace or unlink one | Confidentiality of that instance's `HAL0_MCP_TOKEN`: with two agent instances running as the same user, either can read the other's `/proc/<pid>/environ`, so the file mode does not isolate the token itself |
 | Owner-only `0600` on `/etc/hal0/api.env` | Other local accounts cannot read it | Anything against the agent, which is the owner |
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,6 +11,72 @@ back-patched.
 | `main` (latest)| :white_check_mark: |
 | tagged alphas  | latest only        |
 
+## Bundled agent trust boundary
+
+**Treat the bundled agent as root-equivalent on the box it runs on.**
+
+hal0 ships a bundled agent (Hermes) provisioned with `terminal.backend = local`,
+so it can execute arbitrary shell commands on the host. It runs under
+`hal0-agent@<instance>.service` as `User=hal0` — the same user `hal0-api.service`
+runs as. Two consequences follow, and neither is hypothetical:
+
+- **Same UID means same credentials.** Reading `/proc/<pid>/environ` requires
+  `PTRACE_MODE_READ_FSCREDS`, which a same-UID process passes. Yama's
+  `kernel.yama.ptrace_scope` does not prevent it — that knob gates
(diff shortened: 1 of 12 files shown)


thinmintdev:
### Review rounds applied (Codex, 3 rounds — 9 findings, all fixed)

Round 1 (`a002a259`):
- `_agent_units()` conflated "no agent" with "could not ask" → returns `None` for unknown, which warns.
- An unreadable agent `User=` fell through to PASS → unresolved units now warn.
- `User=` strings compared instead of resolved UIDs → numeric/alias spellings now compare as the same UID.
- SECURITY.md claimed the per-agent env file's `0600` isolates a peer's `HAL0_MCP_TOKEN` — false by this section's own same-UID `/proc` argument; qualified.
- `hal0 config show`'s PermissionError hint recommended `sudo chmod 0644`, i.e. hand-reverting the `0640` this PR lands; now points at sudo.

Round 2 (`ad867a94`, `6d465ed6`):
- `save_upstreams_config()` dropped the mode on every provider CRUD (`mkstemp` 0600 + `os.replace`) → `write_toml_atomic` grows an opt-in `mode=`, canonical value now in `paths.UPSTREAMS_TOML_MODE`, shared with the perms table.
- Discovery missed a stopped *and disabled* instance → unions the install-state artifacts.
- The permission hint offered `hal0` group membership even for `hal0.toml`, which is `0600` by design → mode-aware remedy.
- CodeQL flagged the path-based `chmod` as a path-injection sink → applied via `os.fchmod` on the descriptor already owned. CodeQL now clean.

Round 3 (`2d9637f2`):
- `installer/install.sh` ships `hal0-agent@hermes.service.d/` on every install "whether or not bootstrap has been run", so counting drop-in dirs would warn about an agent an `HAL0_SKIP_HERMES=1` operator deliberately declined → discovery uses enable symlinks + `/etc/hal0/agents/<i>.{toml,env}` only.
- The relative ADR link 404s in-tree (the ADR is still a proposal in #1880) → SECURITY.md points at the PR and names the path it lands at.

Each fix carries a test that fails without it; verified by restoring the previous file and re-running. Latest local run: `ruff check` clean, `ruff format --check` clean (1142 files), `pytest tests/cli tests/install tests/docs tests/config tests/api/test_providers_routes.py` → 1813 passed, 9 skipped. All 8 CI checks green on `2d9637f2`.

Not merging and not enabling automerge — the operator owns that call.

thinmintdev:
Addressed a follow-up round of findings from an independent Opus review (commit a5f4fe5d):

**SECURITY.md corrections**
- The compensating-control table's "systemd sandboxing on the units" row was unscoped. Only `hal0-agent@.service` sets `ProtectSystem=strict` / `PrivateTmp` / restricted `ReadWritePaths=`; the installer-generated `hal0-api.service` (installer/install.sh) sets none of them. Scoped the row to `hal0-agent@`, matching how the adjacent `NoNewPrivileges` row already scopes itself.
- The secret enumeration in the trust-boundary section wrongly listed "the agent's own session token" as living in the API's environment. It's actually two different things, neither of which is an API env var: `HAL0_MCP_TOKEN` is loaded by the AGENT unit from `/etc/hal0/agents/<i>.env`, and the chat-proxy session secret is a file at `/var/lib/hal0/agents/secret.bin`. Reworded to drop the wrong claim and note that the same same-UID logic lets the agent read that file directly (it owns it) — the overall conclusion is unchanged.
- The sudo-grant list named four of `packaging/sudoers/`'s five grants to user `hal0`, missing `hal0-podman-ro`. Added it.

**Seed-mode fix (config_commands.py)**
`hal0 config edit upstreams`/`hal0` seeding a *missing* config file went through a bare `Path.write_text()`, landing at 0644 (umask-mediated) instead of the canonical mode — 0640 for upstreams.toml (`UPSTREAMS_TOML_MODE`, ADR-0002), 0600 for hal0.toml (perms.py) — that `write_toml_atomic`'s fchmod-before-rename already enforces on every real rewrite. The seed path just never got that treatment. Added a `_seed_config_file()` helper mirroring that fchmod-on-owned-descriptor pattern (the seeded text has a comment header, so it isn't a TOML-encodable dict and can't go through `write_toml_atomic` directly), and applied it to hal0.toml, upstreams.toml, and providers.toml (0600, matching what `save_providers_config`'s writer already produces).

TDD: `test_config_edit_upstreams_seeds_at_canonical_mode` and `test_config_edit_hal0_seeds_at_canonical_mode` confirmed red (0o644) against the unfixed code, green (0o640 / 0o600 respectively) after.

Honest note: my earlier "9 findings, all fixed" comment on this PR missed this seed-mode gap entirely — it's now fixed here, along with the two SECURITY.md corrections above.

Verification:
- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/docs/test_security_md_documents_agent_boundary.py tests/cli/test_doctor_all.py tests/install/test_perms.py tests/config/test_loader.py tests/cli/test_config_show_edit_selector.py -q` → 184 passed
- `uv run ruff check` / `uv run ruff format --check` on touched files → clean

thinmintdev:
Addendum (commit 54fa4fa9): Codex reviewed a5f4fe5d and raised two more valid P2s, both fixed.

**1. Ownership regression in `_seed_config_file` (config_commands.py)**
The mode fix alone was incomplete: fchmod'ing a freshly-seeded `hal0.toml`/`upstreams.toml`/`providers.toml` to 0600/0640 without also fixing ownership meant `sudo hal0 config edit` (or any invocation by root, or a third unprivileged user) left the seed owned by the invoker — unreadable by `hal0-api.service` (`User=hal0`). `hal0.api.auth._config_require_auth()` swallows that `PermissionError` into a bare "unset" and falls back to `require_auth` OFF, so this could silently reopen the privileged API path. The previous bare `write_text()` at 0644 never hit this since 0644 stays world-readable regardless of owner — the mode-only fix was the regression.

Fixed with `_fchown_to_service_owner()`, mirroring `hal0.install.perms.ownership_table()`'s default `service_user`/`service_group` ("hal0"): no-op if already running as `hal0`, succeeds if running as root, and is caught + swallowed (leaving prior invoking-user ownership) for any other unprivileged user, since POSIX only lets root chown to an arbitrary uid.

TDD: `test_config_edit_seed_chowns_to_service_owner_when_privileged` faked a privileged invocation via monkeypatched `pwd`/`grp`/`os.fchown` and asserted the chown target — confirmed red (0 chown attempts) against the unfixed code, green after. `test_config_edit_seed_survives_chown_permission_denied` covers the unprivileged-third-user path staying non-fatal.

**2. SECURITY.md 1.1 remediation overstatement**
The forward-looking text claimed a new agent `User=` with no sudoers grants "actually closes the chain." Verified in-tree that's incomplete: `ownership_table()`'s default flip leaves `/etc/hal0` at `hal0:hal0 2775` (setgid), so a split agent left in the shared `hal0` group could still create/replace `hal0.toml` or `api.env` directly — hitting the exact `_config_require_auth()` fallback above and reopening the privileged API path. Reworded to state the 1.1 remediation needs the UID split *and* dropping the agent out of the `hal0` group / config-tree write access, not the UID split alone.

**Verification**
- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/cli/test_config_show_edit_selector.py tests/docs/test_security_md_documents_agent_boundary.py tests/cli/test_doctor_all.py -q` → 97 passed (11 tests in the edit-selector file, including the 2 new ones, red confirmed before the fchown wiring was added)
- `uv run ruff check` / `uv run ruff format --check` on touched files → clean

thinmintdev:
Addendum (commit ca3295ff): Codex P2 on 54fa4fa9, independently nitted by the Opus reviewer — fixed.

`_fchown_to_service_owner()` was silently suppressing a failed chown (PermissionError). An unprivileged operator in the `hal0` group running `hal0 config edit` on the missing-file path got a seed that landed owned by themselves at 0600/0640 — unreadable by `hal0-api.service` (`User=hal0`), and for `hal0.toml` that read failure quietly falls back to `require_auth` OFF via `_config_require_auth()`.

Now: when a system `hal0` account exists but the chown fails (widened to any `OSError`, not just `PermissionError`, per the reviewer's EROFS nit), the seed is aborted — temp file closed and unlinked, nothing published — and the command exits nonzero with a message pointing the operator at `sudo`. The no-system-account case (dev/CI sandboxes) still seeds fine, unchanged.

`test_config_edit_seed_survives_chown_permission_denied` → `test_config_edit_seed_rejects_seed_when_chown_denied` (nonzero exit, no file left, "sudo" in output) plus a new `test_config_edit_seed_rejects_seed_on_other_oserror`. Both confirmed red against 54fa4fa9's suppress-and-continue behavior, green after.

Also added an autouse fixture forcing "no system hal0 account" as the default for the whole test file — this repo's own dev boxes genuinely have a `hal0` system user installed, so the seed tests were incidentally depending on host state the moment ownership failures stopped being silently swallowed.

Verification: `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/cli/test_config_show_edit_selector.py tests/cli/test_doctor_all.py -q` → 93 passed. `ruff check` / `ruff format --check` clean.

thinmintdev:
Addendum (66e543fc): fixed a reviewer major on ca3295ff — the ownership requirement now keys off whether `hal0.config.paths.etc()` is the real `/etc/hal0` (i.e. no `HAL0_HOME`), not off whether a system `hal0` account happens to exist, so `HAL0_HOME=$(mktemp -d) hal0 config edit` no longer dies with the sudo hint on boxes (like this repo's own dev/CI boxes) that have a real hal0 account installed; the autouse test fixture that had been masking this is replaced with an opt-in helper used only by the privileged/chown-denied tests, plus a new regression test that touches neither pwd/grp nor `etc()` (confirmed red on ca3295ff, green after). `tests/cli/test_config_show_edit_selector.py` + `tests/cli/test_doctor_all.py` → 94 passed; ruff check/format clean.

PR #1882, opened 2026-08-15
feat(hermes)!: make the terminal tool an explicit opt-in, default off

thinmintdev:
## This is a deliberate default change

**Hermes no longer gets a terminal tool unless the operator asks for it.** On a
fresh 1.0 install the agent can read the box through the hal0 MCP tools but
cannot run shell commands.

`src/hal0/agents/hermes_provision.py` wrote `("terminal.backend", "local")`
unconditionally. `local` means the agent executes shell commands as the `hal0`
user, which is root-equivalent on this box (rootful podman + the
`hal0-systemctl` / `hal0-agentenv` sudo wrappers). Hermes also ingests untrusted
content — web pages, stored memories, MCP tool output — so shipping that on by
default handed a prompt-injectable LLM a root-equivalent shell, with no egress
restriction, without ever asking. Refs #1863.

There is no sandboxed middle ground on this appliance. The pinned Hermes does
offer `docker | singularity | modal | daytona | ssh`, but the box ships podman
and no docker, singularity is absent, modal/daytona are paid cloud services, and
a containerised shell cannot reach the host or `127.0.0.1:8080` — which would
break the bundled host-management skills by design. So the choice is on/off.

## What "off" actually is

Verified READ-ONLY against the pinned `hermes-agent` 0.18.2 installed at
`/var/lib/hal0/venvs/hermes` on a test box, not guessed:

- Dropping `terminal.backend` disables **nothing**. `hermes_cli/config.py`
  `_DEFAULTS` still carries `terminal.backend: local` and bridges it to
  `TERMINAL_ENV`.
- What removes the tool from the model-facing schema is
  `agent.disabled_toolsets`. `model_tools._compute_tool_definitions` applies it
  as a strict subtraction *after* toolset resolution, with an explicit comment
  that this is so a composite bundle (`toolsets: [hermes-cli]`, what hal0 runs
  with) cannot re-add a disabled toolset's tools.
- `terminal` resolves to `terminal` + `process`. `code_execution` resolves to
  `execute_code`, which runs arbitrary Python as the same user — disabling one
  without the other would leave the shell reachable under another name, so both
  go.

- `delegation` (`delegate_task`) goes too. A subagent is built from the
  toolsets named in the *call*: the pinned build passes only `enabled_toolsets`
  to the child and never forwards the parent's `disabled_toolsets`, and its own
  `DELEGATE_BLOCKED_TOOLS` blocks `execute_code` but not `terminal`/`process`.
  Leaving it reachable would let a prompt-injected agent call
  `delegate_task(toolsets=["terminal"])` and get the declined shell one hop
  away. Same reasoning that bundles `code_execution` with `terminal`.

So: off writes `agent.disabled_toolsets: [terminal, code_execution, delegation]`
and no `terminal.*` key at all. On writes exactly today's config —
`terminal.backend: local`, `terminal.cwd: $HERMES_HOME/scratch` — plus
`disabled_toolsets: []`, so opting in later positively clears a previous
subtraction (convergent in both directions).

## How the operator opts in

| Path | Behaviour |
|---|---|
| Interactive `install.sh` | Asks once, defaults No. The prompt states the consequence in plain language: the agent will be able to run commands as a root-equivalent user on this box. |
| `curl … \| bash`, headless, `HAL0_NONINTERACTIVE=1` | Never asked, always OFF. A silent install cannot enable it. |
| `HAL0_HERMES_TERMINAL=1` | Unattended opt-in. `0` is an explicit opt-out. |
| `hal0 agent install hermes --terminal-tool` / `--no-terminal-tool` | Tri-state flag; omitting it means "not answered here". |

The answer is forwarded across the §7.4 privilege drop as an explicit
`env NAME=VALUE` assignment rather than by inheritance, because `_run_as_hal0`'s
`sudo` fallback resets the environment and would otherwise silently swallow the
opt-in on boxes without `runuser`/`setpriv`.

## Existing boxes keep working

**Every box hal0 has ever provisioned with a terminal stays enabled.** hal0's
own overlay is a `terminal.backend` *plus* a `terminal.cwd` of its own, and
`git log -S'terminal.cwd' -- src/hal0/agents/hermes_provision.py` shows exactly
two shapes across the release history:

| Shape | Written by | Releases |
|---|---|---|
| `cwd: /etc/hal0` | the original `config.yaml.j2`, then commit `ce4f9ded` | `v0.7.3-nightly.20260621` … `v0.9.8` |
| `cwd: $HERMES_HOME/scratch` | commit `cda957c3` | `v0.9.8-nightly.20260720084155` … today |

The `local` branch of the detection accepts **any** `cwd` that is not hermes'
own default, so both shapes and any directory the operator chose themselves are
recognised. Only hermes' untouched default — backend `local` with no `cwd`, or
`cwd: "."`, which `hermes config migrate` materialises on a box hal0 never
finished provisioning — is not consent; those start off, like a fresh install.
The decision is still taken from the **pre-`hermes config migrate`** content for
the same reason. The installer also does not re-ask on an upgrade (the hermes
venv already exists), so a distracted Enter cannot disable a working agent.

An operator who wants it off on such a box runs
`sudo hal0 agent install hermes --no-terminal-tool`.

## What you lose with it off

Documented in `docs/guides/run-agents.mdx` and as a note at the top of each
affected `SKILL.md`: `hal0-service-management`, `hal0-bench`, `hal0-tune`,
`hal0-quantize` drive the host through shell commands and stop working.

`delegate_task` goes with them (see above). Every bundled skill that calls it is
one of those shell workflows, no persona or prelude asks for it, the kanban
board spawns its workers through hal0's own dispatcher rather than
`delegate_task`, and hal0's cross-slot `route_to_chat` delegation is a separate
path that is untouched. What is genuinely given up is parallel fan-out for
subagent work that needs no shell; opting the terminal in returns it.

Chat, memory, personas, the board and every hal0 MCP tool are unaffected — those
go through the API, not a shell.

## Verification

`tests/agents/test_hermes_terminal_optin.py` drives the real
`_phase_config_write` against a temp `$HERMES_HOME` and asserts on the resulting
`config.yaml` — the file hermes actually reads — rather than re-deriving the rule
under test. Covers: fresh install off; non-interactive off; explicit opt-out;
opt-in reproduces today's config; opt-in survives a later re-provision; off is
byte-convergent across re-runs; an existing `local` setting survives an update
(with unrelated hermes-owned keys preserved); an existing box can still be turned
off explicitly.

Added on review (all written test-first, each confirmed red before its fix):

| Case | Red before | Green after |
|---|---|---|
| Legacy `cwd: /etc/hal0` survives an update | `assert 'default-off' == 'existing-config'` | ✔ |
| An operator-chosen `cwd` survives an update | same | ✔ |
| Hermes' own default shape (`local`, no `cwd` / `cwd: "."`) is **not** consent | already green — pins the boundary | ✔ |
| Terminal off also subtracts `delegation` | `assert 'delegation' in ['terminal', 'code_execution']` | ✔ |
| Terminal on keeps `delegation`; opting in later clears the subtraction | already green | ✔ |
| The `delegation:` config block is still written when the terminal is off | already green | ✔ |
| The readback predicate stays at the two execution toolsets | already green — guards boxes provisioned by an earlier build of this branch | ✔ |

`tests/installer/test_hermes_terminal_prompt.py` is new: it slices the gating
block out of `install.sh` by its comment banner and runs it under real bash with
install.sh's own `_interactive`, replacing `_tty_read` with a stub that records
having been called. So the assertion is "the operator was never asked", not "the
answer happened to be no". Covers headless/piped installs, `--dev`,
`--no-start`, `HAL0_SKIP_HERMES=1`, an environment answer surviving and being
re-exported, and the prompt defaulting to No. Mutation-checked: replacing the
`_interactive` gate with `true` turns the headless case red.

The original behavioural case fails on the pre-fix code for a behavioural
reason, not a missing symbol:

```
    assert (doc.get("terminal") or {}).get("backend") != "local", (
        "a fresh install must not hand the agent a root-equivalent shell"
    )
E   AssertionError: a fresh install must not hand the agent a root-equivalent shell
E   assert 'local' != 'local'
```

Full suite on this branch (`HAL0_HOME=$(mktemp -d) uv run --extra dev pytest
tests -q`, 15m32s, at the pushed head): **`10260 passed, 16 skipped, 1
xfailed`**. The `1477 passed` figure quoted here earlier was a subset, not the
full run. `ruff check` and `ruff format --check` clean.

`test_terminal_backend_default_is_sandboxed` stays xfail — it asserts on the
**upstream** default, which is still `local`, so it remains the drift watch on
the pin. Its reason now records that the hal0 side of the deviation is answered.

## Changelog line (no CHANGELOG hunk here — #1860 owns the rc.6 section)

Under **Changed** (breaking default):

> **Hermes' terminal tool is now off by default.** A fresh install no longer
> gives the agent the ability to run shell commands as the root-equivalent
> `hal0` user; the installer asks, and non-interactive installs never enable it.
> Opt in with `hal0 agent install hermes --terminal-tool` or
> `HAL0_HERMES_TERMINAL=1`. Boxes hal0 already provisioned with a terminal keep
> it on across the update. With it off, the `hal0-service-management`,
> `hal0-bench`, `hal0-tune` and `hal0-quantize` skills stop working, and the
> agent can no longer spawn subagents (`delegate_task`) — a subagent's tools are
> chosen at call time, so leaving it reachable would hand back the same shell.

## Out of scope

No release publish, no tag, no CHANGELOG hunk. Not deployed to any box — all
inspection of 10.0.1.152 was read-only.


thinmintdev:
Codex review addressed in 96bc4221 — all four findings were real:

- **P1 preserve unrelated disabled toolsets** — fixed. hal0 now adds/removes only `terminal` + `code_execution` and preserves any other entry in place. Test: `test_unrelated_disabled_toolsets_are_preserved` (a config with `[browser, cronjob]` keeps both across both an off run and a later opt-in).
- **P1 partial migration read as consent** — fixed, and thank you: this was the sharpest of the four. The decision is now recorded at `$HERMES_HOME/.hal0-terminal-tool` *before* `hermes config migrate` runs and outranks config sniffing, so an interrupted run leaves the posture on disk instead of letting the next run re-derive it from hermes' own defaults. The legacy-config path additionally now requires evidence that hal0's overlay actually landed (`memory.provider: hal0-memory` or the hal0 skills dirs), so a bare migrate result is never read as an opt-in. Tests: `test_bare_migrate_output_is_not_read_as_an_opt_in`, `test_recorded_state_outranks_a_reappearing_backend`.
- **P2 prompt on runs that never provision Hermes** — fixed. The prompt is gated on `DEV_MODE == 0` and `HAL0_SKIP_HERMES != 1`, matching the later provisioning block. Persisting the answer for a deferred install is deliberately not done: the deferred path is its own command with its own `--terminal-tool` flag, and stashing consent in a file for a provision that may never happen is worse than asking again.
- **P2 report the effective post-override posture** — fixed. `terminal_tool` is now read back off the final merged config (after `overrides.yaml`), with the decision reported separately as `terminal_tool_decision`. Test: `test_reported_posture_follows_an_operator_override` — an override that clears the list reports `decision: off` / `terminal_tool: on`.

thinmintdev:
Second codex pass addressed in 4e5e5e4d — again all four were real:

- **P1 preserve an existing sandboxed backend** — fixed. A backend the operator moved to `docker`/`ssh` is a *narrower* boundary than hal0's default, and re-asserting `local` on top would have widened it back to root-equivalent host execution. hal0 now writes no `terminal.*` key when the existing backend is anything other than `local`. Test: `test_an_operator_moved_backend_is_not_widened_back_to_local`.
- **P2 require both execution toolsets before reporting off** — fixed. `terminal_enabled_in_config` now reports off only when both `terminal` and `code_execution` are subtracted. Test: `test_effective_off_requires_both_execution_toolsets`.
- **P1 fail closed when the marker cannot be written** — fixed, two ways. `write_terminal_state` now returns whether it landed and the phase forces the tool OFF (reason `state-unwritable`) rather than enabling something it cannot record — chosen over aborting the install, since fail-closed here costs the operator a re-run of `--terminal-tool` and never breaks a box. Separately, the legacy-config path no longer keys on a generic hal0 marker like `memory.provider`: it requires `terminal.cwd == $HERMES_HOME/scratch`, which only this provisioner ever writes (hermes' own default is `"."`), so a partially provisioned fresh box cannot present as a legacy opt-in even with no marker at all. Tests: `test_unwritable_state_file_fails_closed`, `test_bare_migrate_output_is_not_read_as_an_opt_in`.
- **P1 reload Hermes after changing terminal access** — fixed. An answered flag now `systemctl try-restart`s both consumers (`hal0-agent@hermes.service`, `hermes-gateway.service`), and says so loudly if the restart fails. `try-restart` rather than `restart` so this never starts a gateway on a box that runs without one. Tests: `test_posture_change_restarts_the_running_hermes`, `test_posture_restart_only_fires_when_the_flag_was_answered`.

The earlier install.sh gate also missed `--no-start`, which skips Hermes provisioning along with `--dev`; added.

thinmintdev:
Third codex pass (2980ea77). Two of the six were new and both are fixed; three were re-postings of findings already fixed in 96bc4221 / 4e5e5e4d (the bot re-anchors them to the changed lines each pass), and one was a docs omission:

- **P2 non-local backend with a custom cwd** — fixed, good catch. My previous fix preserved a `docker` backend only while it still carried hal0's scratch cwd; an operator with their own cwd fell through to default-off and would have had their backend disabled. Since hermes' own default is `local`, any other value cannot be a materialised default — it is now recognised as an operator decision regardless of cwd. The cwd evidence still gates the `local` branch, which is exactly where it is needed to separate a legacy opt-in from a half-migrated fresh box. Test: `test_non_local_backend_is_recognised_whatever_its_cwd`.
- **P1 restart after an environment-driven posture change** — fixed. `HAL0_HERMES_TERMINAL=0` on a running box rewrote the config but skipped the restart, because only the flag triggered it. The env interface is an explicit answer too and now triggers it. Test extended in `test_posture_restart_only_fires_when_the_flag_was_answered`.
- **P2 hal0-bench-autopilot** — correct, it drives `hal0 bench …` through a shell block. Added to the terminal-dependent table in the guide and given the same skill-level guard note as the other four.

Already fixed, re-flagged: partial migration as consent (marker written pre-migrate + fail-closed + the cwd evidence above), the install.sh prompt gate (`--dev` / `--no-start` / `HAL0_SKIP_HERMES`), and the restart-after-flag path.

thinmintdev:
Fourth codex pass (3a5143a1). Both new findings fixed:

- **P1 reload for every env-driven provisioning path** — fixed. The restart was in `_install_hermes`, so `HAL0_HERMES_TERMINAL=0 hal0 agent reprovision|bootstrap|upgrade hermes` rewrote the config and left the live agent on the old posture. It now lives in `_provision_hermes`, which all of those funnel through, and is skipped on a dry run (nothing written) or a failed provision (nothing to apply). While there: the in-process branch now restores the environment it sets rather than mutating this process' env permanently. Test: `test_posture_restart_covers_every_provisioning_entry_point` (no answer / flag / env / dry-run / failed-rc).
- **P2 undecodable posture marker** — fixed. `read_terminal_state` caught only `OSError`, so non-UTF-8 bytes raised `UnicodeDecodeError` and aborted provisioning, including the explicit `HAL0_HERMES_TERMINAL=0` run that would have repaired the marker. Both are now treated as "no usable record".

Everything else in that pass was a re-posting of findings already fixed in earlier commits on this branch.

thinmintdev:
Fifth codex pass (6b8290f9). All three fixed:

- **P1 keep the consent marker outside the agent-writable home** — fixed, and this was the best finding of the review. $HERMES_HOME is owned by the same `hal0` user the agent runs as and is a `ReadWritePath` of the agent unit, so the marker was agent-writable and a prompt-injected agent could have laundered its own `1` into operator consent at the next re-provision. It now lives at `/etc/hal0/agents/hermes-terminal-tool` — that directory is `root:root 0755` on a live box (checked read-only), so the `hal0` user cannot create files in it. The CLI's root prelude resolves and records the decision before the §7.4 privilege drop (this is also where a pre-#1863 box's existing setting gets captured for the first time), and the unprivileged provisioner only reads it. Stated plainly in the code comment and the docs: this is **not** a containment boundary against an already-compromised agent — one with `write_file` can edit its own `config.yaml` and get the tools back at the next restart — what it closes is the laundering path into provisioning-time consent. Tests: `test_state_marker_lives_outside_the_agent_writable_home`, `test_persist_terminal_decision_records_a_legacy_box`, `test_persist_terminal_decision_records_a_fresh_box_as_off`.
- **P1 restart after posture writes made by a dry run** — correct, I had assumed `dry_run` meant "writes nothing"; in this pipeline it only sets `write_report=False` and the config is still written. The restart is no longer gated on it. Test updated accordingly.
- **P2 surface the posture in normal status output** — fixed. `hal0 agent status hermes` now prints a dedicated line (effective posture, decision, reason) instead of leaving it inside the 60-char-truncated Detail blob.

Full local run green (`tests/agents tests/cli tests/installer`: 1969 passed), lint and format clean.

thinmintdev:
Sixth codex pass (501b696e). Both fixed — both were mine, introduced by the marker move:

- **P1 pass the CLI posture to the root prelude** — fixed. The prelude is the only place that can write the root-only marker, but it resolved from this process' environment and never saw `--terminal-tool`. With the flag set and `HAL0_HERMES_TERMINAL` unset it recorded the old value, and the privilege-dropped child could not correct it: an opt-in failed closed as `state-unwritable` and an opt-out was undone at the next unflagged reprovision. The resolved answer (flag first, then inherited env) is now passed into the prelude. Test: `test_root_prelude_receives_the_flag_answer`.
- **P2 readable mode on the marker** — fixed. `_atomic_write` inherits the process umask, so a root umask of `077` left it `0600` and the `hal0` child could not read it — same fail-closed-despite-consent outcome. Explicit `0644`; it is a posture, not a secret. Test: `test_marker_is_world_readable_under_a_restrictive_umask` (runs under `umask 077`).

Local: `tests/agents tests/cli` 1493 passed, lint/format/sunset clean.

thinmintdev:
Seventh codex pass (d70019a3). Both fixed:

- **P1 route non-root posture reloads through the privileged seam** — fixed, and correct: the reload runs from every provisioning entry point, including the ones the unprivileged `hal0` user drives, where a bare `systemctl` escalates through polkit (prompts with nobody there, or fails) while the command still reported success. Root keeps plain `systemctl`; anything else goes through the seam — `restart-agent hermes` via the existing `agent_unit_argv` router, and a new `hermes-gateway` key added to the wrapper's **closed** svc-* name→unit map (reviewable one-line widening, same shape as `openwebui`/`hindsight`). Tests: `test_posture_change_restarts_the_running_hermes_as_root`, `test_posture_restart_routes_through_the_seam_when_unprivileged`, plus the wrapper-level `svc-restart hermes-gateway` case in `tests/system/test_seam_agent_units.py`.
- **P1 reject opt-outs that cannot replace an enabled marker** — fixed. The write result was discarded, so an unprivileged opt-out disabled the tools and returned success while the still-`1` marker would re-enable them at the next unflagged provision. That case is now applied (off now beats on now — refusing would leave the tools live) **and** reported as a `warn` phase with a reason naming the marker and telling the operator to re-run as root, plus a `terminal_state_stale` detail; `hal0 agent status` surfaces warn phases prominently. The opposite direction — an opt-in that cannot be recorded — still fails closed outright. Test: `test_optout_that_cannot_be_recorded_is_applied_but_warns`.

Local: `tests/agents tests/cli tests/system tests/installer` 2146 passed; lint, format, sunset and `bash -n` on the wrapper clean.

thinmintdev:
Review findings addressed in three commits on top of `d70019a3`.

**Finding 1 (blocker) — legacy-box detection gap.** `git log -S'terminal.cwd' -- src/hal0/agents/hermes_provision.py` turns up exactly two `cwd` shapes hal0 has ever written: `/etc/hal0`, from the original `config.yaml.j2` through commit `ce4f9ded` and shipped in every tag from `v0.7.3-nightly.20260621` to `v0.9.8`; and `$HERMES_HOME/scratch`, from commit `cda957c3` in `v0.9.8-nightly.20260720084155` onward. Only the second was recognised, so a legacy box with a working `local` terminal would have been silently disabled on update. The `local` branch now treats any `cwd` that is not hermes' own default as evidence of a hal0-provisioned or operator-chosen setup, with the `/etc/hal0` literal named explicitly so the release history stays readable in the source. The `$HERMES_HOME`-relative shape still matches dynamically, which also covers a box whose `HERMES_HOME` has since moved. Hermes' untouched default — backend `local` with no `cwd`, or `cwd: "."` — is still not consent, so a half-finished fresh provision stays default-off.

**Finding 2 (blocker) — the docs claim.** With finding 1 fixed there is no uncovered hal0-provisioned population left, so the claim holds. It is still reworded, because the old phrasing ("a box provisioned before this change already has `terminal.backend: local`") described evidence the code deliberately does not accept on its own. The guide now names the two shapes, says an operator's own `cwd` counts too, and says which single shape does not and why.

**Finding 3 (major) — delegation escape. Decision: option (a), disable the `delegation` toolset while the terminal is off.** The impact investigation found nothing load-bearing that works terminal-less and depends on it. Every `delegate_task` reference in the bundled skills is inside `hal0-service-management` and is explicitly a shell workflow (`references/mcp-tool-quirks.md` lines 22-28, `SKILL.md` lines 22 and 186), so all of it is already dead with the terminal off — that skill is one the guide already lists as broken. No provisioning prompt, prelude template or persona mentions it: no hits in `src/hal0/agents/hermes_templates/*.j2` or in `personas.py`. The kanban board, which the guide promises still works with the terminal off, spawns its workers through its own dispatcher and upstream's `kanban_tools.py` explicitly rejects `delegate_task` children, so that promise is unaffected. hal0's own cross-slot delegation (`route_to_chat`) is a separate, terminal-free path and is untouched. The one real cost is parallel fan-out for pure-MCP subagent work, which nothing in the repo ships today; it is now documented under "What you lose with it off", and opting the terminal in returns delegation with it.

Two implementation details worth flagging for the re-review. The subtraction is a new `TERMINAL_OFF_TOOLSETS` write set; the readback predicate `terminal_enabled_in_config` deliberately stays at the two execution toolsets, because it is an `issubset` test and growing it would make a box already provisioned off by an earlier build of this branch start reporting "on". And only the toolset goes — the `delegation.{model,provider,base_url}` block that routes subagents to the `agent` role slot is still written, so the posture stays a pure toolset decision and re-enabling stays a one-key change. Both are pinned by tests.

Since option (a) closes the hole rather than documenting it, no residual is stated in SECURITY.md and the installer prompt is unchanged.

**Finding 4 (minor) — installer test.** A harness was workable. `tests/installer/test_hermes_terminal_prompt.py` slices the gating block out of `install.sh` by its comment banner and runs it under real bash with install.sh's own `_interactive`, following the pattern in `test_preflight_python_floor.py`. `_tty_read` is replaced by a stub that records having been called, so the assertion is "the operator was never asked" rather than "the answer happened to be no". It covers the requested case (a non-interactive install never prompts and leaves `HAL0_HERMES_TERMINAL` unset) plus `--dev`, `--no-start`, `HAL0_SKIP_HERMES=1`, an environment answer surviving and being re-exported, and the prompt defaulting to No. Mutation-checked: replacing the `_interactive` gate with `true` turns the headless case red.

**Finding 5 (nit) — verification counts.** Corrected in the PR body. The `1477 passed` figure was a subset. The genuine full run at the pushed head is `10260 passed, 16 skipped, 1 xfailed` in 15m32s; `ruff check` and `ruff format --check` are clean.

Each behavioural fix was written test-first. The legacy-`cwd` and operator-`cwd` cases were red with `assert 'default-off' == 'existing-config'` before the detection change and green after; the delegation case was red with `assert 'delegation' in ['terminal', 'code_execution']` before the subtraction change and green after.

thinmintdev:
Second review round. Three more commits, `a6528e6c` / `0b57d96c` / `6699d5ff`, all on top of `d7986827`.

Worth stating plainly: two of these are defects in `d70019a3`, a commit that landed on this branch from another author while my first round was in flight and that I rebased onto without reviewing. Reviewing it post-hoc was the right catch and I should have done it at the rebase rather than treating a clean rebase as a clean diff.

**Blocker — an unwritable marker could disable a working terminal.** `d70019a3`'s fail-closed branch keyed on `terminal_enabled` alone, so every enable that could not be recorded was downgraded to `state-unwritable`, `existing-config` included. The comment three lines above still promised the opposite ("A box that is already working is left alone — refusing to record must not disable something already in use"), and the gap is reachable rather than theoretical: the marker lives in root-only `/etc/hal0/agents/`, and `hal0 agent reprovision hermes` run by a non-root operator takes the in-process branch with no root prelude, so the write cannot land. A legacy box with a working `terminal.backend: local` would have lost its shell to a command that asked for nothing of the sort.

Fail-closed is now restricted to a fresh `opt-in`, which is the case it was written for — an unrecorded "on" that nothing on disk attests to is what a later run could re-derive permissively. An unrecordable `existing-config` enable gets the treatment the unrecordable opt-out already had: applied, marked `+state-stale`, surfaced as a WARN. Its evidence is in `config.yaml`, so the next run re-derives the same answer rather than a permissive one. The operator-facing stale line now names the direction of the disagreement too; the two cases point opposite ways and the text described only the opt-out.

Red before, green after: `assert 'existing-config+state-stale' == 'state-unwritable'`, with the config's `disabled_toolsets` non-empty on the old code and `[]` on the new.

**Minor — the posture restart could start a stopped Hermes. Fixed properly rather than documented.** The seam gained the narrower verb instead. `try-restart-agent` and `svc-try-restart` are both strictly weaker than the restart verbs already in the wrapper — `try-restart` is a no-op on an inactive unit and cannot start anything — and the sudoers grant is pinned to the helper binary rather than an argv list, so widening the wrapper's own case statement was the entire privileged-surface change. That felt clearly preferable to documenting a footgun that fires on `hal0 agent install hermes --no-terminal-tool`.

While there I fixed the gateway's mirror image of the same bug: root used `try-restart` but the unprivileged path used `svc-restart`, so the invariant held for root only. Both consumers now use try-restart on both privilege paths. Red before, green after: five cases across `test_seam_agent_units.py` and `test_agent_install_hermes.py`, including a new one asserting the wrapper emits `try-restart hal0-agent@hermes.service` and never a bare restart.

**Minor — the readback predicate. Decision: require all three toolsets.** The old rationale (protecting boxes provisioned by an earlier build of this branch) protects nobody, since the branch is unmerged — a fair hit. The reachable shape is an operator `overrides.yaml` handing one toolset back, and there the old predicate reported a clean "off" on a box where `delegate_task` is live and a subagent's toolsets are chosen at call time: a shell one hop away, reported as no shell.

I picked "require all three" over a distinct partially-disabled state because it is not a new rule, just the existing one extended. The function already refused to call `terminal` disabled + `code_execution` live an "off" posture, on exactly the grounds that each is independently sufficient to reach command execution. `delegation` is the third such route. Reporting "on" for a partially-subtracted config is also the least-surprise answer — it is what the agent can actually do — and it needs no third value in the status field or in anything that consumes it. Red before, green after: `terminal_enabled_in_config` on the two-toolset shape, plus an end-to-end case driving an `overrides.yaml` that restores only `delegation`.

Full suite after all three: **`10268 passed, 16 skipped, 1 xfailed`** in 15m59s. `ruff check` and `ruff format --check` clean; `bash -n` clean on the wrapper.

PR #1907, opened 2026-08-17
fix(ui): derive footer services group from /api/services/health

### LINKED ISSUE #1899: ui: dashboard footer services LED group is a hardcoded 3-member set that omits ComfyUI
**Severity:** minor · **Box:** ct151-cpu-fresh; bites hardest on prod where ComfyUI is a real running service.

### Summary

The dashboard footer's `services` LED group is a hardcoded three-member array (`hal0`, `hermes`, `openwebui`) in `ui/src/dash/chrome.jsx`; `comfyui` is not a member, so on a box where ComfyUI is a real running service an outage leaves the always-visible footer at an unqualified green "services 3 / 3 ready". Not a miscount (the three rendered pips are truthful and labelled), not a regression (identical in every bundle back to `0.7.3-beta.2`), and no repo artifact supports the fixed set as intentional. The sibling runtimes group proves the ribbon can render degradation.

### Repro

```
curl -s $API/api/services/health | jq 'keys'
# Playwright: [data-testid="foot-health-services"] innerText + pip titles; diff the chip
# id set against the /api/services/health id set.
```

### Expected

The footer services group enumerates (or explicitly summarises) every id in `/api/services/health` — a down `comfyui` must move the count or trip the warn styling.

### Test to promote

A `ui/tests/e2e` fixture spec diffing chip ids against a services fixture.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `ui-footer-services-omits-comfyui`.

thinmintdev:
## Summary

The dashboard footer's `services` LED group (`ui/src/dash/chrome.jsx`) was a hardcoded three-member array (`hal0`, `hermes`, `openwebui`). `comfyui` was never a member, so on a box where ComfyUI is a real running service an outage left the always-visible footer reporting an unqualified green "services 3 / 3 ready" — the three rendered pips were individually truthful, but the group as a whole silently dropped a real backing service.

## Root cause

The reported symptom (comfyui missing) is one layer above the actual defect: the footer never read the `id` set out of the `/api/services/health` payload at all — it hand-rolled three specific service objects (`hermes`, `openwebui`) alongside `hal0`. The backend (`src/hal0/api/routes/services_health.py`) already reports `comfyui`, `hermes`, and `openwebui` every call. Adding `comfyui` as a fourth hardcoded entry would have fixed today's symptom but left the same defect for the next service the backend adds. Fixed by mapping `footerIndicators` directly over `serviceHealth.services` (prepending only the `hal0` self-check, since hal0 IS the API answering the request and isn't in that payload) — the footer now always mirrors whatever the backend actually reports, with no fixed id list to fall out of sync.

## What changed

- `ui/src/dash/chrome.jsx`: `footerIndicators` (the `services` LED group) now derives from `serviceHealth.services` instead of a hardcoded `[hermes, openwebui]` pair alongside `hal0`.
- `ui/tests/e2e/specs/footer-services-omits-comfyui-v3.spec.ts`: new regression spec that mocks `GET /api/services/health` with the real 3-service payload shape and asserts the footer renders 4 pips (hal0 + comfyui + hermes + openwebui) with a correct ready count, and that a down `comfyui` both keeps its own pip and trips the warn styling / drops the ready count.

## Verification

- New spec confirmed RED on the pre-fix hardcoded array (3 pips, comfyui absent) and GREEN after the fix (4 pips, `4/4` all-up / `3/4` + warn on comfyui-down) via `git stash` bisection of the fix commit.
- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run test:unit` (vitest) — 127/127 passed
- `npm run build` — succeeds
- `npx playwright test tests/e2e/specs/footer-*` — 18 passed, 1 pre-existing skip (unrelated bundle-hygiene test gated on a prod build)

## Out of scope

- The sibling `runtimes` group (already dynamic, used as the working reference for this fix) — untouched.
- No changes to `/api/services/health` itself; the backend already reported comfyui correctly, the defect was entirely on the frontend consumption side.
- CHANGELOG.md intentionally left untouched per the fix-wave convention (#1545) — consolidated at the end of the wave.

Closes #1899

### DIFF
--- a/tests/release-validation/regressions.yaml
+++ b/tests/release-validation/regressions.yaml
@@ -1203,7 +1203,7 @@ regressions:
     severity: minor
     tier: mechanical
     lane: ui
-    promote_ready: true
+    promoted_to: ui/tests/e2e/specs/footer-services-omits-comfyui-v3.spec.ts
     summary: >-
       The dashboard footer's `services` LED group is a hardcoded three-member array (hal0,
       hermes, openwebui) in ui/src/dash/chrome.jsx; comfyui is not a member, so on a box where
--- a/ui/src/dash/chrome.jsx
+++ b/ui/src/dash/chrome.jsx
@@ -816,7 +816,14 @@ function Footer({ updateAvailable, expanded = false, onToggle }) {
   const runtimeTitle = degraded
     ? `degraded — ${failing.length ? failing.join("; ") : "see /api/health/system"}`
     : `${L.ready}/${L.total} slot container${L.total === 1 ? "" : "s"} ready`;
-  const serviceById = Object.fromEntries((serviceHealth.services || []).map((s) => [s.id, s]));
+  // The `services` group is driven by whatever /api/services/health actually
+  // reports (#1899) — NOT a hardcoded id list. The three-member array here
+  // used to omit `comfyui` (and would silently omit any future service too),
+  // leaving the always-visible footer reporting an unqualified "3 / 3 ready"
+  // while a real backing service sat down. `hal0` itself is prepended since
+  // it isn't in the services-health payload (it IS the API answering the
+  // request) but every backend-reported id — including comfyui — renders a
+  // pip and counts toward the ready total.
   const footerIndicators = [
     {
       id: 'hal0',
@@ -824,18 +831,12 @@ function Footer({ updateAvailable, expanded = false, onToggle }) {
       tone: degraded ? 'warn' : health.isError ? 'warn' : 'up',
       title: degraded ? `hal0 degraded — ${failing.join("; ") || "see /api/health/system"}` : 'hal0 api ok',
     },
-    {
-      id: 'hermes',
-      label: 'hermes',
-      tone: serviceHealth.pending ? 'warn' : serviceById.hermes?.up ? 'up' : 'err',
-      title: serviceById.hermes?.detail || (serviceHealth.pending ? 'service health pending' : 'hermes down'),
-    },
-    {
-      id: 'openwebui',
-      label: 'openwebui',
-      tone: serviceHealth.pending ? 'warn' : serviceById.openwebui?.up ? 'up' : 'err',
-      title: serviceById.openwebui?.detail || (serviceHealth.pending ? 'service health pending' : 'openwebui down'),
-    },
+    ...(serviceHealth.services || []).map((s) => ({
+      id: s.id,
+      label: s.name || s.id,
+      tone: serviceHealth.pending ? 'warn' : s.up ? 'up' : 'err',
+      title: s.detail || (serviceHealth.pending ? 'service health pending' : `${s.name || s.id} down`),
+    })),
   ];
   // runtimes group — one LED pip per CONFIGURED slot (a model bound is what
   // makes a slot real, #1369), plus the ready count.
--- a/ui/src/dash/chrome.jsx
+++ b/ui/src/dash/chrome.jsx
@@ -824,6 +824,16 @@ function Footer({ updateAvailable, expanded = false, onToggle }) {
   // it isn't in the services-health payload (it IS the API answering the
   // request) but every backend-reported id — including comfyui — renders a
   // pip and counts toward the ready total.
+  //
+  // When /api/services/health itself has not answered — first paint, 404
+  // (older API), non-2xx, or network error all surface as `pending` with an
+  // empty `services` list (useServicesHealth fails soft) — the group must NOT
+  // collapse to a green "1 / 1 ready" hal0 self-check: the state of the real
+  // backing services is unknown, which is exactly the unqualified-green defect
+  // #1899 describes. Render a single warn-toned "services unknown" pip so the
+  // count stays honest and the warn styling trips until real data arrives.
+  const servicesUnknown =
+    (serviceHealth.pending || serviceHealth.error) && !(serviceHealth.services || []).length;
   const footerIndicators = [
     {
       id: 'hal0',
@@ -831,12 +841,19 @@ function Footer({ updateAvailable, expanded = false, onToggle }) {
       tone: degraded ? 'warn' : health.isError ? 'warn' : 'up',
       title: degraded ? `hal0 degraded — ${failing.join("; ") || "see /api/health/system"}` : 'hal0 api ok',
     },
-    ...(serviceHealth.services || []).map((s) => ({
-      id: s.id,
-      label: s.name || s.id,
-      tone: serviceHealth.pending ? 'warn' : s.up ? 'up' : 'err',
-      title: s.detail || (serviceHealth.pending ? 'service health pending' : `${s.name || s.id} down`),
-    })),
+    ...(servicesUnknown
+      ? [{
+          id: 'services-unknown',
+          label: 'services',
+          tone: 'warn',
+          title: 'service health unknown — /api/services/health not answering',
+        }]
+      : (serviceHealth.services || []).map((s) => ({
+          id: s.id,
+          label: s.name || s.id,
+          tone: s.up ? 'up' : 'err',
+          title: s.detail || `${s.name || s.id} down`,
+        }))),
   ];
   // runtimes group — one LED pip per CONFIGURED slot (a model bound is what
   // makes a slot real, #1369), plus the ready count.
@@ -993,6 +1010,7 @@ function Footer({ updateAvailable, expanded = false, onToggle }) {
             {footerIndicators.map((svc) => (
               <span
                 key={svc.id}
+                data-service-id={svc.id}
                 className={"pip " + _svcPip(svc.tone)}
                 title={`${svc.label}: ${svc.title}`}
                 aria-label={`${svc.label}: ${svc.tone}`}
--- a/ui/tests/e2e/specs/footer-services-omits-comfyui-v3.spec.ts
+++ b/ui/tests/e2e/specs/footer-services-omits-comfyui-v3.spec.ts
@@ -14,6 +14,15 @@
  * pip + in the ready count) every id GET /api/services/health reports —
  * not a fixed set baked into the component. A down `comfyui` must both
  * appear as its own pip and move the ready count / trip warn styling.
+ * The pips carry `data-service-id`, so the assertion diffs the chip id set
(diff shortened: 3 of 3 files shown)


thinmintdev:
## Review — changes requested

The fix is aimed at the right layer: the defect really was "the footer hand-rolls an id list" and not "comfyui is missing from the payload", so deriving `footerIndicators` from `serviceHealth.services` is the root-cause fix, not a symptom patch. The regression spec is genuine — I reverted `ui/src/dash/chrome.jsx` to the parent commit while keeping the new spec and it goes red on the pip count:

```
Locator:  locator('[data-testid="foot-health-services"]').locator('.pip')
Expected: 4
Received: 3
```

and green with the fix restored (2 passed). Scope is clean: two files, no CHANGELOG hunk, no drive-by refactor. `footer-chips`, `footer-health-tooltip`, `dashboard-redesign`, `ui-sweep-a`, `system-card` still pass at head (32 passed, 6 skipped), and no other spec or unit test asserts on the services chip, so the `hermes` → `Hermes` label change breaks nothing today.

One blocking problem, though.

### Blocking — the "unqualified green" defect moves to the pending/unavailable path

Removing the two hardcoded entries also removed the only thing that made the group render degradation when `/api/services/health` itself is unavailable. `useServicesHealth` returns `services: []` whenever the payload is null — 404, non-2xx, network error, or first paint (`useServicesHealth.ts:80-81`) — so `footerIndicators` collapses to just the `hal0` self-check.

Measured on both sides with a probe spec that fulfils `**/api/services/health` with a 500:

| | pips | chip text | `.v` class |
|---|---|---|---|
| before this PR | 3 | `SERVICES 1 / 3 ready` | `v warn` |
| this PR | 1 | `SERVICES 1 / 1 ready` | `v` |

That is the same failure #1899 describes — an always-visible footer reporting unqualified green while the state of three real backing services is unknown — just reached through a different door. `/api/health/system` answering fine while `/api/services/health` 500s or 404s (older API, proxy hiccup, route not yet built — the hook's whole reason for failing soft) is exactly the case that now reads all-clear. The next validation run will re-find this under the same register key.

Corroborating evidence that the path wasn't considered: the `serviceHealth.pending ? 'warn' : …` and `'service health pending'` fallbacks inside the new `.map()` are unreachable. `pending` is true only when `services` is `[]`, so the map body never executes while pending.

Suggested shape (not prescriptive): when `serviceHealth.pending` (or `error`) is true and `services` is empty, keep the group honest — a single `warn`-toned "services unknown" pip, or force the `.v warn` styling and drop the bare `n / n` — rather than silently shrinking the denominator to 1. Please add a spec case for the 500/404 payload alongside the two you already have; that is the case that will otherwise regress next.

### Nits (non-blocking)

1. **Register entry not promoted.** `tests/release-validation/regressions.yaml:1200` carries `ui-footer-services-omits-comfyui` with `promote_ready: true`, and the file header says the promoting PR should "set `promoted_to:` and drop it in the same PR". This PR writes exactly the promotion the entry asks for but leaves the entry untouched. If the wave is consolidating register edits at the end (like CHANGELOG), say so in the PR body; otherwise set `promoted_to: ui/tests/e2e/specs/footer-services-omits-comfyui-v3.spec.ts` here.

2. **The spec pins display copy, not the id set.** `[aria-label^="ComfyUI:"]` asserts on the backend's `name` field. The contract in the issue is the *id* set (`.services[].id`), and the pips carry no id attribute. Renaming `"ComfyUI"` → `"Comfy UI"` in `services_health.py` would break the regression spec without the defect returning. Consider emitting `data-service-id={svc.id}` on the pip and asserting the id set against the fixture — that also matches the issue's "diff the chip id set against the /api/services/health id set" repro verbatim.

### Codex bot

The Codex review on `1954c9ed` posted its boilerplate header with zero suggestions — no findings to triage. Noting explicitly that this is not an approval signal.

thinmintdev:
## Revision — all review findings addressed (`03d62115`)

Rebased onto current main (past #1909/#1913/#1915/#1919).

### Blocking — unqualified green moved to the pending/unavailable path: **fixed**

`chrome.jsx` now detects the unavailable state explicitly: when `serviceHealth.pending` (or `error`) is true and `services` is empty — first paint, 404, non-2xx, network error, all the hook's fail-soft doors — the group renders a single warn-toned **`services unknown`** placeholder pip (`data-service-id="services-unknown"`, title `service health unknown — /api/services/health not answering`) instead of collapsing to the hal0 self-check. Measured on the 500 path: **2 pips, `1 / 2 ready`, `.v warn`** — never an unqualified green `1 / 1`. The corroborating dead code you flagged is gone too: the per-entry `serviceHealth.pending` fallbacks inside the `.map()` (unreachable, since pending implies an empty list) are removed; the map now handles only real entries.

New spec case added as requested: `services health endpoint 500s: group shows a warn "unknown" pip, not unqualified green` fulfils `**/api/services/health` with a 500 and asserts pips=2, `1 / 2`, `.v warn`, and the warn-toned placeholder pip.

### Nit 1 — register entry not promoted: **fixed**

`tests/release-validation/regressions.yaml` `ui-footer-services-omits-comfyui`: `promote_ready: true` replaced with `promoted_to: ui/tests/e2e/specs/footer-services-omits-comfyui-v3.spec.ts` in this PR, per the file header and your suggestion. (Left the entry in place rather than deleting it — first `promoted_to` in the file, so no precedent to follow; happy to drop the entry instead if that's the intended convention.)

### Nit 2 — spec pins display copy, not the id set: **fixed**

Pips now emit `data-service-id={svc.id}`, and the all-up spec diffs the rendered chip id set against the fixture's `.services[].id` set plus `hal0` — the issue's "diff the chip id set against the /api/services/health id set" repro verbatim. The comfyui-down assertion targets `.pip[data-service-id="comfyui"]` instead of `[aria-label^="ComfyUI:"]`, so a backend display-name rename can no longer break (or mask) the regression spec.

### Codex bot

The `1954c9ed` Codex review's single P1 inline comment is the same finding as your blocker (group-level pending state lost) — resolved by the same fix; no other Codex findings to triage.

### Verification

- `npx playwright test footer-services-omits-comfyui-v3.spec.ts` — 3 passed
- neighbouring specs (`footer-chips`, `footer-health-tooltip`, `footer-journal-pane`, `footer-throughput-chip`, `footer-update-chip`, `dashboard-redesign`, `ui-sweep-a`, `system-card`) — 45 passed, 6 skipped
- `npm run lint`, `npm run typecheck`, `make lint`, `ruff format --check`, `scripts/check_sunset.py` — all clean

thinmintdev:
## Re-review — approved (`03d62115`)

Re-verified from scratch in a clean worktree at `03d62115` (`git reset`, `git diff HEAD` empty, three files in the diff: `ui/src/dash/chrome.jsx`, the new spec, one line of `regressions.yaml`).

### Blocking finding — resolved at the root

`servicesUnknown = (serviceHealth.pending || serviceHealth.error) && !services.length` gates a single warn-toned `services-unknown` placeholder into `footerIndicators`, and `servicesReady` (`tone === 'up'`) excludes it, so the denominator no longer shrinks to 1 and the `.v warn` class trips. The unreachable per-entry `serviceHealth.pending` fallbacks inside the `.map()` are gone.

Measured three ways, not taken on the PR body's word:

| tree | 4-pip case | comfyui-down case | 500 case |
|---|---|---|---|
| `add97eb3` (pre-fix parent) | 3 pips ✗ | 3 pips ✗ | 3 pips ✗ |
| `30e12b93` (the version I rejected) | pass | pass | **1 pip ✗** — the unqualified green |
| `03d62115` (head) | pass | pass | 2 pips, `1 / 2`, `.v warn` ✓ |

That middle row is the important one: the new third spec case is genuinely red on the exact tree I filed the blocker against and green here, so the fix is pinned rather than asserted. Full run at head: `3 passed`.

### No regression of the original #1899 fix

The issue's contract — "diff the chip id set against the `/api/services/health` id set" — is now asserted literally: pips carry `data-service-id={svc.id}` and the all-up case compares the rendered id set to `fixture.services.map(s => s.id)` plus `hal0`. The comfyui-down case targets `.pip[data-service-id="comfyui"]`, so a `name` rename in `services_health.py` can no longer break or mask the spec (nit 2, fixed as asked). Reverting `chrome.jsx` to the parent still takes all three cases red, so the derive-from-payload behaviour is still what the spec measures.

### No new defects

- `npm run lint`, `npm run typecheck` — clean; `test:unit` 127/127.
- Neighbouring e2e (`footer-*`, `dashboard-redesign`, `ui-sweep-a`, `system-card`) — 48 passed, 6 skipped (pre-existing prod-build skips).
- `data-service-id` is additive markup; grep shows no other spec, unit test, or component reads the services pips, so the `hermes` → `Hermes` label change from `s.name || s.id` still breaks nothing.
- `regressions.yaml` still parses; 44 entries, no structural damage.
- CI green on `03d62115` on all 8 check-runs (`python (3.12)`, `ui`, `sunset`, `γ-suite (chromium)` 10m45s, CodeQL ×4) — no cancelled runs masquerading as failures.
- Residual behaviour worth naming, not a defect: `error` with a *non-empty* stale list keeps rendering the stale entries rather than the unknown pip. `useServicesHealth` never throws, so `error` is unreachable today; the guard is defensive.

### Codex triage

The P1 on `1954c9ed` was the same defect as my blocker — resolved by the same hunk, and I confirmed it independently above. Two P2s on `03d62115` remain, both about kit bookkeeping rather than the fix:

**Non-blocking, but please land it.** Codex is right on the documented convention and I steered you wrong in nit 1 — `regressions.yaml:5` says "set `promoted_to:` and drop it in the same PR", and `README.md:116-118` says the entry "should be deleted from the kit in the PR that writes the test". So the intended end state is deletion, not a `promoted_to` key: `.claude/workflows/rc-validate.js:316` selects entries by `tier`/`lane` and never reads `promoted_to`, so as written this still spends a live-box probe every release on a case CI now covers. Second P2 is also correct: `README.md:122` requires bumping `kit.toml`'s `kit_version` (4 → 5) with a changelog line in the same PR as any `regressions.yaml` change, otherwise two materially different registers both report as kit v4.

Neither has runtime or user impact, so I am not holding the fix for them — but they should land here or be explicitly tracked as end-of-wave kit curation.

Approving.

Issue #1834, opened 2026-08-11
memory: hindsight extraction is wired to a shared slot with no concurrency cap, no max_tokens and a requeueing 300s x4 ladder — the retain queue never drains on a CPU-only install


thinmintdev:
> Covers two validation findings with one root cause: `memory-retain-never-lands` and `utility-slot-contention`.

## What happens

hal0 owns `installer/systemd/hindsight-api.service` and wires Hindsight fact extraction at a **shared inference slot** (`HINDSIGHT_API_LLM_MODEL=hal0/utility`) with **no worker-concurrency cap, no `max_tokens` cap, and a 300 s × 4 retry ladder that requeues on failure**. On any box where one extraction cannot finish inside `HINDSIGHT_API_LLM_TIMEOUT=300`, the retain queue enters a non-draining loop that holds the shared slot permanently busy.

On the CPU-only fresh install (ct152, `HAL0_ALLOW_CPU_ONLY=1`, `utility` serving a 0.8 B model):

- Every `POST /api/memory/add` returns HTTP 200 in 0.046 s, and **zero facts are ever stored**: bank `shared` `fact_count=0`, `GET /api/memory/list` → `{"items":[]}`.
- The extraction prompt is 3253 tokens and the slot's own llama-server reports `llamacpp:prompt_tokens_seconds 14.6459` — ~223 s of prefill before a single token — so every attempt hits the 300 s dispatcher read timeout.
- Four concurrent extractions + one deferred pile onto that one llama-server, each with `params.n_predict: 64000`, and each timeout re-queues — so the retry ladder **amplifies the congestion it is retrying**.
- A 14-token `Say OK` to `hal0/utility` did not return in 180 s while this was running.

## Decisive evidence (ct152, 2026-08-11 02:40–02:55 UTC, hal0 1.0.0rc5)

```
$ systemctl show hindsight-api.service -p Environment | tr ' ' '\n' | grep LLM
HINDSIGHT_API_LLM_MODEL=hal0/utility   HINDSIGHT_API_LLM_TIMEOUT=300   HINDSIGHT_API_LLM_BASE_URL=http://127.0.0.1:8080/v1

$ curl -s http://127.0.0.1:8090/metrics   (sampled 02:48:48 / 02:49:50 / 02:50:55)
llamacpp:requests_processing 4      at every sample
llamacpp:requests_deferred  5 -> 3 -> 1
[...]

Issue #1613, opened 2026-08-07
memory: provider silently sticks on the pgvector fallback after a boot-time hindsight race


thinmintdev:
Observed live on CT105 during the #1612 deploy: `systemctl restart hindsight-api hal0-api` together made hal0-api's boot probe hit the engine mid-restart (embedded postgres takes ~60-90s), so provider selection degraded to `PgVectorProvider` (#1301 behavior) and **stayed there** — every `memory_reflect` returned the ABC stub `{"status": "unsupported"}` even after the engine came up healthy. Only a second hal0-api restart recovered it.

Suggestions:
1. Runtime re-probe: on a degraded provider, periodically (or on next memory call after N seconds) re-probe hindsight health and swap back to `HindsightProvider` when it answers — the reverse transition of the existing degrade path.
2. Unit ordering mitigation: `After=hindsight-api.service` (+ `Wants=`) on hal0-api narrows the boot-time window but doesn't cover live restarts.

The new `hal0 doctor all` rows (`mcp_mounts`, `hindsight_llm_auth`) don't catch this class — the mount answers fine, the key is fine, only the selected provider is wrong. A doctor row asserting the ACTIVE provider matches `[memory].engine` config would close the gap.

Refs #1543, #1301, #1612

thinmintdev:
Reproduced on a **pristine fresh install** (CT151, ubuntu-26.04 rollback, main @ 55edfa6d): the installer starts hindsight-api and hal0-api in the same pass, the engine's first boot (embedded postgres init + model load) outlasts hal0-api's 2s probe, and the very first boot of a fresh box lands with `Hindsight / banks — disabled (no memory engine)` shown as PASS while `hindsight-api` is active. A single `systemctl restart hal0-api` afterwards flips it to `reachable — 2 bank(s)`. So this isn't just a restart-race edge case — it's the default fresh-install end state, which is also exactly the #1543 report.

PR #1909, opened 2026-08-17
fix(registry): drop unloadable bge-reranker-base curated row

thinmintdev:
## What changed

Removed the `bge-reranker-base-q4_k_m` row (`cstr/bge-reranker-base-GGUF`) from the curated model catalogue in `src/hal0/registry/curated.py`, and its false "classifier-head fix" note. Added a comment in its place recording why, plus two regression tests in `tests/registry/test_curated.py`.

## Why

That row was offered **first and smallest** in the rerank catalog with `fit_status: allowed`, so hal0 pulled it unvalidated. cstr's GGUF mirror carries a non-mainline fork's bert KV schema with no `bert.context_length`, so the runner fails to load it: `"error loading model hyperparameters: key not found in model: bert.context_length"`. The slot then crash-loops invisibly. The row's note — "cstr's GGUF mirror includes the classifier-head fix needed for llama-server --reranking" — was also false.

This is not a rerank-capability defect: the seeded default `bge-reranker-v2-m3-q4_k_m` already loads and returns correctly-ordered scores on the identical runner image, so dropping the row (rather than repointing it at another mirror) leaves no gap in rerank coverage.

Root cause is the curated-catalog data itself (an unvalidated third-party GGUF re-quant claimed to fix an upstream gap it didn't actually fix) — not the pull layer, the runner, or the flag plumbing (`#1787`), which are explicitly out of scope here per the issue.

## How verified

- Added `test_bge_reranker_base_row_removed` (fails before the fix — the row and its false note were present; passes after) and `test_every_curated_rerank_row_has_hf_coordinates` as a forward guard against a bad rerank row silently becoming the offered default again.
- `HAL0_HOME=$(mktemp -d) uv run pytest tests/registry/ tests/install/test_curated_seed_wins.py -q` → 410 passed.
- `uv run ruff check src/hal0/registry/curated.py tests/registry/test_curated.py` → clean.
- `uv run ruff format --check src/hal0/registry/curated.py tests/registry/test_curated.py` → already formatted.
- Confirmed `/api/capabilities` fit_status sourcing (`src/hal0/capabilities/catalog.py`) reads directly from `CURATED`/`CURATED_BY_ID`, so removing the catalog row removes it from the offered rerank picks with no route change needed (read-only touch, per plan).

## Left out of scope

- The runner, flag plumbing, and `#1787` — explicitly excluded by the issue.
[...]

Issue #448, opened 2026-06-04
Hermes namespace_register: verify memory_delete count, don't trust HTTP 200


thinmintdev:
## Context

Follow-up from PR #446 (Peer-memory stale-card flood). The root cause was a silently-failing `memory_delete` that returned HTTP 200 with `{"deleted": 0}`. `_phase_namespace_register` in `src/hal0/agents/hermes_provision.py` treated the 200 as success and proceeded to add a fresh card, leaking one identity card per bootstrap/`--repair` run (19 accumulated).

## Problem

The dedup logic checks `deleted["ok"]` (HTTP success) but never inspects the returned `deleted` **count**:

```python
deleted = _mcp_memory_call("tools/call", {"name": "memory_delete", "arguments": {"ids": existing_ids}}, ...)
if not deleted["ok"]:
    warnings.append(...)
# proceeds to add regardless — never checks deleted["result"]["deleted"]
```

So a delete that finds priors but removes none looks identical to a successful prune.

## Proposed fix

After delete, assert the count matches the request and warn/abort the rewrite on mismatch:

```python
removed = (deleted.get("result") or {}).get("deleted", 0)
if existing_ids and removed != len(existing_ids):
    warnings.append(f"memory_delete: requested {len(existing_ids)}, removed {removed} — not rewriting to avoid duplicate accumulation")
    # skip the add, or fail the phase
```

This is defense-in-depth: even with the wrapper bug fixed (#446), a future delete-scope regression would be caught at the call site instead of silently flooding the Peer view.

## Acceptance
- [ ] `_phase_namespace_register` inspects the delete count
- [ ] On count mismatch it warns and does not add a duplicate
- [ ] Regression test covers the "delete reported fewer than requested" path

thinmintdev:
> *This was generated by AI during triage.*

## Agent Brief

**Triage:** `bug` · `ready-for-agent` · confidence high · repro: confirmed-by-trace

**Claims verified vs `main` @ `f3b3c44`:**
- __phase_namespace_register checks deleted['ok'] but never the deleted count_ → **confirmed** — src/hal0/agents/hermes_provision.py:1650-1651 only `if not deleted["ok"]: warnings.append(...)`, then add at 1653 unconditionally
- _memory_delete returns a count under result.deleted_ → **confirmed** — cognee_wrapper.py:999 `return {"deleted": deleted}`; shim wraps as deleted["result"]["deleted"] (_mcp_memory_call:1601)
[...]

PR #1912, opened 2026-08-17
fix(memory): replay boot-lane writes lost to the degraded provider

### LINKED ISSUE #1897: hal0-api: agents bank writes during hindsight's degraded-provider boot window are silently lost, never replayed after provider_healed
**Severity:** minor · **Box:** ct151-cpu-fresh, cross-checked on ct150-update.

### Summary

hal0-api's install-time boot runs agents-dataset seeding while hindsight is still cold-starting, so the identity-card writes land in the in-memory pgvector fallback ("VOLATILE and will be LOST"), and the `#1613` runtime self-heal swaps the provider without replaying them — the durable `agents` bank does not exist until some later `hal0-api` restart re-runs `brain_lane` against a healthy engine.

Generalised defect: writes issued by hal0-api's own boot phases during the degraded-provider window are silently lost and never re-issued after `provider_healed`.

Even after the bank appears, `hal0 agent peers` can stay empty because the card retains are stuck in extraction (that half is tracked as `hindsight-extraction-shares-utility-slot` / #1834).

**Spin-off bug found while cross-checking, included here rather than filed separately:** `hal0 agent peers` on ct150 crashes with `AttributeError` at `agent_commands.py:1372` when a card's `hal0_state` metadata comes back as a JSON string instead of a dict.

### Repro

```
# On a fresh install, before any manual restart:
curl -s $API/api/memory/banks | jq '.[] | {bank_id, created_at}'   # agents absent = defect
journalctl -u hal0-api | grep -E 'boot_degraded|degraded_write|provider_healed'
# then: systemctl restart hal0-api; the bank appears (created lazily by the re-run write)
```

### Expected

The `agents` bank exists with `created_at` ~= install time on a fresh box, OR the heal path re-runs the (idempotent) `brain_lane` card publishes from the `provider_healed` branch. A degraded-window boot write lost with no journal warning, or the bank still missing after a restart with a durable provider, is the hard fail.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `agents-bank-boot-writes-lost`.

thinmintdev:
Closes #1897

## What was wrong

On a fresh install the installer starts `hal0-api` and the hindsight daemon in the same pass, so hal0-api's terminal `brain_lane` boot phase always runs inside the memory degrade window: `create_app` has already handed every consumer the volatile in-memory `PgVectorProvider`, and the agent identity cards + boot self-report are accepted, warned about ("VOLATILE and will be LOST"), and dropped. The #1613 self-heal later swaps the durable engine in, but it cannot recover writes that engine never saw — so the durable `agents` bank did not exist until some unrelated `hal0-api` restart happened to re-run the lane against a healthy provider.

## Root cause

Deeper than "brain_lane runs too early": it is the missing half of the #1613 self-heal. `SelfHealingMemoryProvider.try_heal()` rebinds its delegate and tells nobody, so no caller that wrote through the volatile fallback can learn about the swap and re-issue. Any boot phase that writes during the degrade window loses its data, silently — the brain lane is simply the one that always does.

## What changed

- `src/hal0/memory/__init__.py` — `SelfHealingMemoryProvider` grows a replay contract: `add_heal_hook(hook)` (returns `False` when the provider has already healed, so the caller knows to replay itself) and `run_heal_hooks()`, a best-effort drain that runs each hook once. A hook that raises is logged and skipped.
- `src/hal0/api/__init__.py` — the re-probe loop moves out of a closure in `_boot_background_tasks` into a module-level `_memory_reprobe_loop(provider, interval_s)` that drains the heal hooks right after a successful swap (and is directly testable).
- `src/hal0/api/__init__.py` — `_boot_brain_lane(..., replay=False)`: when the provider is degraded on entry, it arms a one-shot replay of itself on the heal path. The three publishes are already idempotent (each searches + deletes its prior card before rewriting), so replaying is safe whether or not the first pass reached a durable engine. A heal that lands mid-phase — after the drain — replays inline instead. With a deliberate `[memory].engine = "pgvector"` (no self-heal shell) it logs `brain_lane.degraded_writes_volatile` instead, so the loss is no longer silent in the journal.
- `src/hal0/cli/agent_commands.py` — spin-off crash reported in the same issue: `hal0 agent peers` died with `AttributeError: 'str' object has no attribute 'get'` when a card's `hal0_state` (or `endpoint`) came back from the engine as a JSON string rather than a dict, taking the whole listing down. Nested metadata is now coerced tolerantly, and a non-list `roles` no longer renders one character per column cell.

## How it was verified

- `tests/api/test_boot_brain_lane_replay.py` (new): the degraded lane replays after heal; the replay does not re-arm; a healthy boot arms nothing; a mid-lane heal replays inline; the re-probe loop drains hooks after the swap; a raising hook does not break the drain. Verified red before the fix (with the arming disabled, `test_degraded_boot_lane_replays_after_provider_heals` and `test_lane_replays_inline_when_provider_healed_mid_lane` both fail), green after.
- `tests/cli/test_agent_peers_metadata.py` (new): both cases fail on `main` with the reported `AttributeError`, pass after.
- `uv run --extra dev pytest tests/api tests/cli tests/agents tests/memory tests/install` → **3761 passed, 2 skipped, 1 xfailed**.
- `uv run ruff check src tests` → clean. `uv run ruff format --check src tests` → 1146 files already formatted.

## Deliberately out of scope

- `#1834` (hindsight extraction sharing the utility slot) — the other half of why `hal0 agent peers` can read empty even once the bank exists. Untouched here.
- No `CHANGELOG.md` hunk, per the wave's consolidated-changelog rule (#1545).
- `tests/release-validation/regressions.yaml` left as-is; the register has no `fixed_in` convention and is owned by the validation kit.
- Replay is armed only for the brain lane. The hook mechanism is general, but no other boot phase writes to memory today.

### DIFF
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -1760,7 +1760,7 @@ async def _boot_metrics_state(app: FastAPI, ctx: BootState) -> None:
 
 
 async def _memory_reprobe_loop(
-    provider: Any, interval_s: float, *, max_replay_attempts: int = 10
+    provider: Any, interval_s: float, *, max_replay_attempts: int = 20
 ) -> None:
     """Poll a boot-degraded memory provider until the durable engine answers.
 
@@ -1774,10 +1774,12 @@ async def _memory_reprobe_loop(
     replay. Draining here, on the loop, is what makes those writes durable
     without waiting for the next hal0-api restart.
 
-    A drain that reports failure (a healthy engine can still fail an
-    individual retain) keeps its hooks armed, so the loop keeps ticking and
-    retries — up to ``max_replay_attempts``, after which it gives up loudly
-    rather than retrying forever.
+    A drain that reports failure keeps its hooks armed, so the loop keeps
+    ticking and retries — up to ``max_replay_attempts``, after which it gives
+    up loudly rather than retrying forever. "Failure" covers both a replay
+    whose writes did not land (a healthy engine can still fail an individual
+    retain) and a replay armed by a lane that has not finished yet, so the
+    budget is generous relative to the interval.
     """
     attempts = 0
     while True:
@@ -2045,13 +2047,47 @@ def _boot_mcp_memory_call(
         return {"ok": False, "error": str(exc)}
 
 
+class _BrainLaneReplay:
+    """Heal hook that re-issues the brain-lane publishes lost while degraded (#1897).
+
+    Armed on the self-healing provider BEFORE the lane's publishes run, so a
+    heal that lands mid-lane cannot drain an empty replay and leave the lane
+    with nothing to register onto. Until :meth:`finish` hands over the result,
+    the hook reports "not done" (``False``) so the drain keeps it armed.
+
+    Each run narrows ``pending`` to the publishes that still did not land, so
+    a partially successful retry never repeats one that already succeeded —
+    ``_phase_self_report`` has no dedupe, and re-running it would leave one
+    extra private self-report per tick.
+    """
+
+    def __init__(self, app: FastAPI, ctx: BootState) -> None:
+        self._app = app
+        self._ctx = ctx
+        self.pending: tuple[str, ...] = ()
+        self.ready = False
+
+    def finish(self, owed: tuple[str, ...]) -> None:
+        """Hand over the publishes the degraded window lost (may be empty)."""
+        self.pending = owed
+        self.ready = True
+
+    async def __call__(self) -> bool:
+        if not self.ready:
+            return False
+        if not self.pending:
+            return True
+        self.pending = await _boot_brain_lane(self._app, self._ctx, only=self.pending, replay=True)
+        return not self.pending
+
+
 async def _boot_brain_lane(
     app: FastAPI,
     ctx: BootState,
     *,
     only: tuple[str, ...] | None = None,
     replay: bool = False,
-) -> bool:
+) -> tuple[str, ...]:
     """Phase — RELOCATE(brain-lane): identity-card + self-report memory publishes.
 
     Runs LAST (after ``background_tasks``, the final named phase before this
@@ -2093,14 +2129,28 @@ async def _boot_brain_lane(
     the lost ones are replayed: a replay of a write that already landed
     durably races the engine's asynchronous retain and can duplicate the card.
 
-    Returns True when every publish it ran reported success; the heal-hook
-    drain keeps the hook armed on False so a transient post-heal failure is
-    retried rather than losing the data until the next restart.
+    Returns the names of the publishes it ran that did NOT report success —
+    empty means everything landed. :class:`_BrainLaneReplay` narrows its retry
+    set to exactly that, so a retry never repeats a publish that already
+    succeeded (``_phase_self_report`` has no dedupe of its own; repeating it
+    would leave a second private self-report per tick).
     """
     import functools
 
     provider = getattr(app.state, "memory_provider", None)
     degraded_on_entry = bool(getattr(provider, "degraded", False))
+
+    # #1897: arm the replay BEFORE the publishes run. The hook reports
+    # "not done" until this lane hands it its result, so a heal that lands
+    # mid-lane cannot drain-and-drop an empty replay and leave the lane with
+    # nothing to register onto afterwards.
+    replay_state: _BrainLaneReplay | None = None
+    if not replay and degraded_on_entry:
+        _add_heal_hook = getattr(provider, "add_heal_hook", None)
+        if _add_heal_hook is not None:
+            replay_state = _BrainLaneReplay(app, ctx)
+            if not _add_heal_hook(replay_state):  # pragma: no cover — heal beat the arm
+                replay_state = None
 
     from hal0.agents.hermes_provision import (
         BootstrapState,
@@ -2143,8 +2193,8 @@ async def _boot_brain_lane(
     def _volatile_writes() -> int:
         return int(getattr(fallback, "volatile_writes", 0) or 0)
 
-    lost: list[str] = []
-    ok = True
+    owed: list[str] = []
+    failed: list[str] = []
     for name, phase, done_key, prior in steps:
         if only is not None and name not in only:
             continue
@@ -2157,54 +2207,66 @@ async def _boot_brain_lane(
         try:
             result = await asyncio.to_thread(phase, step_ctx)
             if not result.details.get(done_key):
-                ok = False
+                failed.append(name)
                 log.info(
                     f"brain_lane.{name}_degraded",
                     warnings=result.details.get("warnings") or result.details.get("warning"),
                 )
         except Exception as exc:  # pragma: no cover — defensive, must never block boot
-            ok = False
+            failed.append(name)
             log.warning(f"brain_lane.{name}_failed", error=str(exc))
-        if _volatile_writes() > before:
-            lost.append(name)
+        # Owed a replay when this step's write was swallowed by the volatile
+        # fallback, or when it did not land at all — in both cases no durable
+        # card exists, so re-running cannot duplicate one.
+        if _volatile_writes() > before or name in failed:
+            owed.append(name)
 
     if replay:
-        # Report to the heal-hook drain whether this replay actually landed;
-        # a False keeps the hook armed for the next re-probe tick.
-        return ok
-    if not degraded_on_entry or not lost:
-        return ok
-    add_heal_hook = getattr(provider, "add_heal_hook", None)
-    if add_heal_hook is None:
+        # The drain narrows the retry set to what came back here.
+        return tuple(failed)
+    if replay_state is not None:
+        # Hand the armed hook its work; empty means the drain drops it and
+        # the re-probe loop can exit.
+        replay_state.finish(tuple(owed))
+        if owed:
+            log.warning(
+                "brain_lane.degraded_replay_armed",
+                steps=owed,
+                detail="brain-lane publishes landed in the volatile pgvector fallback — "
+                "queued for replay once the durable memory engine heals",
+            )
+        return tuple(failed)
+    if not owed:
+        return tuple(failed)
+    if getattr(provider, "add_heal_hook", None) is None:
         # Deliberate [memory].engine = "pgvector", or memory disabled: no
         # durable engine is ever coming, so there is nothing to replay onto.
         log.warning(
             "brain_lane.degraded_writes_volatile",
-            steps=lost,
+            steps=owed,
             detail="brain-lane publishes accepted by a volatile provider with no self-heal "
             "shell — the agents dataset will not survive a restart",
         )
-        return ok
-    hook = functools.partial(_boot_brain_lane, app, ctx, only=tuple(lost), replay=True)
-    if add_heal_hook(hook):
-        log.warning(
-            "brain_lane.degraded_replay_armed",
-            steps=lost,
-            detail="brain-lane publishes landed in the volatile pgvector fallback — "
-            "queued for replay once the durable memory engine heals",
+        return tuple(failed)
+    # Provider healed between the degraded-on-entry check and the arming, so
+    # the shell refused the hook and nothing else will run the replay: do it
+    # inline. A replay that still does not land is out of retries here — say
+    # so loudly rather than pretending the cards are durable.
+    log.warning(
+        "brain_lane.degraded_replay_inline",
+        steps=owed,
+        detail="memory provider healed before the replay could be armed — replaying "
+        "the lost brain-lane publishes against the durable engine now",
+    )
+    still_owed = await _boot_brain_lane(app, ctx, only=tuple(owed), replay=True)
+    if still_owed:
+        log.error(
+            "brain_lane.replay_incomplete",
+            steps=still_owed,
+            detail="inline replay of the degraded-window publishes did not land — "
+            "restart hal0-api to re-run them against the durable engine",
         )
-    else:
-        # The provider healed while this phase was mid-flight, after its heal
-        # hooks were drained: nothing would ever run our replay. Do it inline,
-        # for the lost steps only.
-        log.warning(
-            "brain_lane.degraded_replay_inline",
-            steps=lost,
-            detail="memory provider healed mid-phase — replaying the lost brain-lane "
-            "publishes against the durable engine now",
-        )
-        await _boot_brain_lane(app, ctx, only=tuple(lost), replay=True)
-    return ok
+    return tuple(failed)
 
 
 @asynccontextmanager
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -1759,6 +1759,28 @@ async def _boot_metrics_state(app: FastAPI, ctx: BootState) -> None:
     app.state.metrics_seam = ctx.metrics_seam
 
 
+async def _memory_reprobe_loop(provider: Any, interval_s: float) -> None:
+    """Poll a boot-degraded memory provider until the durable engine answers.
+
+    Armed by ``_boot_background_tasks`` only when ``create_app`` wrapped a
+    degraded boot result in ``SelfHealingMemoryProvider`` (#1613). ``try_heal``
+    is synchronous, probe-bounded I/O, so it runs off-loop via ``to_thread``.
+
+    After the swap, drain the provider's heal hooks (#1897) — the writes hal0-api
+    made through the volatile fallback during the degrade window are gone from
+    the engine that just came up, and the phases that issued them registered a
+    replay. Draining here, on the loop, is what makes those writes durable
+    without waiting for the next hal0-api restart.
+    """
+    while True:
+        await asyncio.sleep(interval_s)
+        if await asyncio.to_thread(provider.try_heal):
+            run_hooks = getattr(provider, "run_heal_hooks", None)
+            if run_hooks is not None:
+                await run_hooks()
+            return
+
+
 async def _boot_background_tasks(app: FastAPI, ctx: BootState) -> None:
     """Phase — MCP session managers + refresh / GPU-arbiter loops + omni & NPU routers.
 
@@ -1826,14 +1848,9 @@ async def _boot_background_tasks(app: FastAPI, ctx: BootState) -> None:
     _mem_provider = getattr(app.state, "memory_provider", None)
     if hasattr(_mem_provider, "try_heal"):
         _reprobe_interval = float(os.environ.get("HAL0_MEMORY_REPROBE_INTERVAL_S", "30") or "30")
-
-        async def _memory_reprobe_loop(provider: Any = _mem_provider) -> None:
-            while True:
-                await asyncio.sleep(_reprobe_interval)
-                if await asyncio.to_thread(provider.try_heal):
-                    return
-
-        ctx.memory_reprobe_task = asyncio.create_task(_memory_reprobe_loop())
+        ctx.memory_reprobe_task = asyncio.create_task(
+            _memory_reprobe_loop(_mem_provider, _reprobe_interval)
+        )
         log.info("hal0.memory.reprobe_loop_started", interval_s=_reprobe_interval)
 
     async def _stop_memory_reprobe_task() -> None:
@@ -2010,7 +2027,7 @@ def _boot_mcp_memory_call(
         return {"ok": False, "error": str(exc)}
 
 
-async def _boot_brain_lane(app: FastAPI, ctx: BootState) -> None:
+async def _boot_brain_lane(app: FastAPI, ctx: BootState, *, replay: bool = False) -> None:
     """Phase — RELOCATE(brain-lane): identity-card + self-report memory publishes.
 
     Runs LAST (after ``background_tasks``, the final named phase before this
@@ -2033,8 +2050,27 @@ async def _boot_brain_lane(app: FastAPI, ctx: BootState) -> None:
     self-test of chat/memory/etc. Flagged here and in the relocation
     handoff notes; a future lane could add a lightweight post-boot
     self-check if a closer smoke_tests equivalent is wanted.
+
+    #1897 — degraded-window replay. On a fresh install this phase always runs
+    inside the memory degrade window: the installer starts hal0-api and the
+    hindsight daemon in the same pass, so the engine is still cold-starting
+    (embedded postgres init + model load) and ``create_app`` handed every
+    consumer the volatile in-memory ``PgVectorProvider``. All three publishes
+    below are then accepted, warned about ("VOLATILE and will be LOST"), and
+    dropped — the durable ``agents`` bank is never created. The #1613 self-heal
+    swaps the engine in later but cannot recover writes the engine never saw,
+    so the bank stayed missing until an unrelated hal0-api restart happened to
+    re-run this phase against a healthy provider. When the provider is degraded
+    on entry we therefore arm a one-shot replay of this same phase on the heal
+    path; the publishes are idempotent (each searches + deletes its prior card
+    before rewriting), so replaying is safe whether or not the first pass
+    reached a durable engine. ``replay=True`` marks the re-run so it does not
+    arm another.
     """
     import functools
+
+    provider = getattr(app.state, "memory_provider", None)
+    degraded_on_entry = bool(getattr(provider, "degraded", False))
 
     from hal0.agents.hermes_provision import (
         BootstrapState,
@@ -2086,6 +2122,35 @@ async def _boot_brain_lane(app: FastAPI, ctx: BootState) -> None:
             )
     except Exception as exc:  # pragma: no cover — defensive, must never block boot
         log.warning("brain_lane.self_report_failed", error=str(exc))
+
+    # #1897: the writes above went to a volatile fallback — owe a replay.
+    if replay or not degraded_on_entry:
+        return
+    add_heal_hook = getattr(provider, "add_heal_hook", None)
+    if add_heal_hook is None:
+        # Deliberate [memory].engine = "pgvector", or memory disabled: no
+        # durable engine is ever coming, so there is nothing to replay onto.
+        log.warning(
+            "brain_lane.degraded_writes_volatile",
+            detail="brain-lane publishes accepted by a volatile provider with no self-heal "
+            "shell — the agents dataset will not survive a restart",
+        )
+        return
+    if add_heal_hook(functools.partial(_boot_brain_lane, app, ctx, replay=True)):
+        log.warning(
+            "brain_lane.degraded_replay_armed",
+            detail="brain-lane publishes landed in the volatile pgvector fallback — "
+            "queued for replay once the durable memory engine heals",
+        )
+    else:
+        # The provider healed while this phase was mid-flight, after its heal
+        # hooks were drained: nothing would ever run our replay. Do it inline.
+        log.warning(
+            "brain_lane.degraded_replay_inline",
+            detail="memory provider healed mid-phase — replaying the brain-lane "
+            "publishes against the durable engine now",
+        )
+        await _boot_brain_lane(app, ctx, replay=True)
 
 
 @asynccontextmanager
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -2188,12 +2188,29 @@ async def _boot_brain_lane(
     # durably is what would duplicate cards: the dedupe search that makes a
     # replay safe races the engine's asynchronous retain of the write it is
     # meant to supersede.
+    #
+    # Known, pre-existing, very narrow window (Codex review on #1912): if the
+    # swap lands between a phase's dedupe search (ran against the empty
+    # fallback) and its add (durable), the counter does not move, the step is
+    # not replayed, and the durable dedupe never ran — a degraded RESTART on a
+    # box whose agents bank already holds a prior card can duplicate it. The
+    # window is one in-process search round-trip against a 30s reprobe tick,
+    # and it predates this replay (the #1613 loop already ran concurrently
+    # with the lane).
     fallback = getattr(provider, "target", provider)
 
     def _volatile_writes() -> int:
         return int(getattr(fallback, "volatile_writes", 0) or 0)
 
-    owed: list[str] = []
+    # ``volatile`` and ``failed`` are DIFFERENT findings and drive different
+    # messaging: "the fallback swallowed this write" (durable data loss unless
+    # replayed) vs "this publish just failed" (a normal warn-as-OK outcome the
+    # per-step log above already covers — including every boot with
+    # [memory].enabled = false, where all three fail). Only their union feeds
+    # the replay set; the degraded-window warnings key off ``volatile`` alone,
+    # so a healthy-provider publish failure can never fire a WARNING claiming
+    # volatile data loss (rc-validate greps the journal for exactly that key).
+    volatile: list[str] = []
     failed: list[str] = []
     for name, phase, done_key, prior in steps:
         if only is not None and name not in only:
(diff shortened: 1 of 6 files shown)


thinmintdev:
Codex review triage — all three findings addressed in 5110f1c5.

**P1 (keep retrying until replay writes succeed) — fixed.** `_boot_brain_lane` now returns whether its publishes landed; `SelfHealingMemoryProvider.run_heal_hooks()` keeps a hook that returned `False` (or raised) armed and reports the drain as failed, and `_memory_reprobe_loop` keeps ticking instead of exiting. Bounded by `max_replay_attempts` (10) so a permanently failing replay gives up loudly (`hal0.memory.heal_replay_abandoned`) rather than spinning forever. Covered by `test_failed_replay_stays_armed_for_the_next_tick` and `test_reprobe_loop_retries_a_failed_drain_then_gives_up`.

**P2 (avoid replaying durable writes during a mid-lane heal) — fixed, and it improved the root-cause signal.** `PgVectorProvider` now counts the writes it swallows (`volatile_writes`), and the lane reads that counter around each step. So the replay set is "the publishes whose writes the fallback actually ate", not "everything, because we were degraded on entry". A heal that lands mid-lane replays only the lost steps; the ones that reached the engine are left alone, which is what avoids the duplicate cards you describe (`_phase_self_report` does not dedupe at all, and the identity-card dedupe search races the engine's async retain). Covered by `test_only_the_steps_that_hit_the_fallback_are_replayed`, which asserts `self_report` runs exactly once when the heal lands before its write.

**P2 (parse the nested metadata format Hindsight emits) — fixed, good catch.** Confirmed at `src/hal0/memory/hindsight_provider.py:956`: metadata goes out as `{k: str(v) for ...}`, so nested values come back as Python reprs, not JSON — the original fallback would have turned every production card's Endpoint/Registered column into `—`. `hal0 agent peers` now tries `json.loads` then `ast.literal_eval` (literals only, never `eval`), and `roles` gets the same treatment so a flattened list no longer renders one character per cell. New test `test_peers_renders_hindsight_repr_metadata` uses `str({...})` — the producer format — alongside the JSON case.

Verified: `uv run --extra dev pytest tests/api tests/cli tests/agents tests/memory tests/install tests/mcp` → 4095 passed, 2 skipped, 1 xfailed. `ruff check` + `ruff format --check` clean.

thinmintdev:
Second codex pass triaged — two fixed in 325e5374, one acknowledged as out of scope.

**P2 (keep failed inline replays armed) — fixed, and the race it lives in is closed.** The replay hook is now armed on the provider *before* the publishes run, and reports "not done" until the lane hands it its result — so a heal draining mid-lane can no longer find an empty replay, drop it, and leave the lane with nothing to register onto. That removes the inline path for the common case entirely. The remaining inline fallback (heal beat the arming outright, so `add_heal_hook` refused) now checks its own return and logs `brain_lane.replay_incomplete` at ERROR with the steps and the "restart hal0-api" remedy, rather than discarding a failed replay silently. New tests: `test_replay_is_armed_before_the_publishes_run`, `test_replay_hook_reports_not_done_until_the_lane_hands_over`.

**P2 (retry only the steps that remain failed) — fixed.** `_boot_brain_lane` returns the names of the publishes that did not land, and `_BrainLaneReplay` narrows its pending set to exactly those on each attempt. `test_failed_replay_stays_armed_for_the_next_tick` now asserts the two steps that landed are not repeated while the failing one is retried.

**P2 (wait for asynchronous replay writes to finish) — acknowledged, out of scope.** Correct that a successful `memory_add` on hindsight means the retain was *accepted*, not that extraction finished. That is a pre-existing property of every brain-lane publish — the install-time path and the pre-#1897 boot path report success on the same signal — so the replay is no worse than the write it re-issues, and this PR does not regress it. Closing it needs `sync` writes or operation-id polling plumbed through the `memory_add` MCP surface and the boot dispatch adapter, which is a change of its own; the extraction half of this failure mode is already tracked as #1834. Flagging it here rather than expanding this PR's blast radius mid-fix-wave.

Verified: `uv run --extra dev pytest tests/api tests/cli tests/agents tests/memory tests/install` → 3767 passed, 2 skipped, 1 xfailed. `ruff check` + `ruff format --check` clean.

thinmintdev:
CI note — two red checks on 325e5374, both traced, neither in this diff:

- **Analyze (python)** failed at `Uploading code scanning results` with `##[error]No server is currently available to service your request` — a GitHub-side 503 during the SARIF upload (the same incident that was 503-ing `api.github.com/graphql` and `rerun-failed-jobs` for ~40 minutes). CodeQL itself completed: "CodeQL scanned 1162 out of 1162 Python files". The run is a default-setup (`dynamic`) run and cannot be retried, so 0f0ca5f0 is an empty commit purely to re-trigger it.
- **python (3.12)** failed on `tests/slots/test_fail_watcher_warming.py::test_warming_slot_recovers_when_stale` — a pre-existing race in that test, not a regression here (this branch touches `memory/`, the boot lane and the peers CLI; nothing under `slots/`). The test breaks out of its polling loop as soon as the `unload`/`load` spy calls are recorded, then asserts `_current_state("chat") == SlotState.READY` with no wait for the reload's state transition, so a slow runner observes `STARTING`. It passed on the two earlier CI runs of this same branch, passes 8/8 locally, and passed on the re-run (25m23s). Left alone deliberately — fixing it belongs on its own branch, not mixed into this one.

thinmintdev:
Status on 0f0ca5f0 (the empty CodeQL re-trigger commit):

- `python (3.12)` — **pass**
- `γ-suite (chromium)`, `ui`, `sunset`, `Analyze (javascript-typescript)` — **pass**
- `Analyze (python)` and `Analyze (actions)` — **fail, GitHub-side**. The re-triggered run hit the same incident earlier in `Initialize CodeQL`, before any analysis ran: `##[error]Encountered an error while trying to determine feature enablement: HttpError: No server is currently available to service your request`. `api.github.com` has been intermittently 503-ing for hours (it is also refusing `gh pr checks`, `rerun-failed-jobs`, and `gh run rerun`). These are default-setup (`dynamic`) runs, so they cannot be retried from the API — they need either a push or a re-run once GitHub is healthy.

Nothing further to do on the code from my side. Handing over for review; whoever lands it will need to re-trigger the two CodeQL jobs first.

thinmintdev:
# Review (agent, delegated) — CHANGES REQUESTED

## Review — verified against #1897's repro, not the PR description

Verdict: **changes requested**, but only for one contained defect in the *new* code. The core fix is right and the tests are load-bearing.

### Does it fix the reported defect, at the root cause?

Yes, and at the root cause rather than the symptom. #1897 asked for exactly one of two outcomes: the `agents` bank exists at install time, **or** the heal path re-runs the idempotent `brain_lane` publishes from `provider_healed`. This implements the second, and it does it as the missing half of the #1613 contract (`add_heal_hook` / `run_heal_hooks` on `SelfHealingMemoryProvider`) rather than as a special case bolted onto the lane. Any future boot phase that writes inside the degrade window can register onto the same seam. The per-step `volatile_writes` counter is the right discriminator: "replay the writes the fallback actually swallowed", not "replay everything because we were degraded on entry".

Traced the whole write path to confirm the counter is the real signal, not an assumption: `_boot_mcp_memory_call` → `_boot_memory_dispatch` re-resolves `app.state.memory_provider` per call → `make_dispatcher` → `SelfHealingMemoryProvider.__getattr__` → the live delegate's `add`. So a write made while the fallback is live does bump `PgVectorProvider._volatile_writes`, and once the shell rebinds, the captured `fallback` object stops moving — later steps are correctly not replayed. Also confirmed the publishes are genuinely re-runnable: `_phase_namespace_register` / `_phase_brain_profile_seed` search-then-delete before rewriting, and neither of them (nor `_phase_self_report`) can return `registered/published: False` *after* a completed `memory_add`, so a retry of a failed step can never duplicate a card.

### Do the new tests fail without the fix?

Verified by mutation, not by reading the PR body. On `0f0ca5f0` with the fix selectively disabled:

- disable the hook arming (`if False and not replay and degraded_on_entry:`) → `test_degraded_boot_lane_replays_after_provider_heals`, `test_replay_is_armed_before_the_publishes_run`, `test_failed_replay_stays_armed_for_the_next_tick` **fail**
- replace the per-step counter check with `if degraded_on_entry or name in failed:` → `test_only_the_steps_that_hit_the_fallback_are_replayed` **fails**
- stop re-arming a failed hook in `run_heal_hooks` → `test_failed_replay_stays_armed_for_the_next_tick`, `test_heal_hook_failure_does_not_break_the_drain` **fail**
- `tests/cli/test_agent_peers_metadata.py` — all 3 fail on `main` with the reported `AttributeError`

Suites: `tests/api tests/cli tests/memory` → 2764 passed, 1 pre-existing environmental failure (`test_health_degraded.py::test_health_system_ok_when_no_errored_slots`, fails identically on `main` here because no hindsight daemon is listening on `127.0.0.1:9177`). `ruff check` + `ruff format --check` clean. The two red CodeQL jobs are GitHub-side — read the logs: `Analyze (python)` completed the scan ("1162 out of 1162 Python files") and died in `Uploading code scanning results`, `Analyze (actions)` died in `Initialize CodeQL`, both on `No server is currently available to service your request`. The CI note on this PR is accurate.

---

## Blocking

### `brain_lane.degraded_writes_volatile` now fires on boots where nothing was volatile

`owed` conflates two different things: "the fallback swallowed this write" (`_volatile_writes() > before`) and "this publish just failed" (`name in failed`). The replay logic wants the union; the **messaging and the inline-replay branch** key off it too, and that is wrong for the second case.

Reproduced against this branch (`0f0ca5f0`) with a throwaway probe:

- **Healthy durable provider** (no shell, `HindsightProvider`), one publish returns `published: False` — e.g. a transient `memory_add` error, which is a normal warn-as-OK outcome:
  `EVENTS: ['brain_lane.self_report_degraded', 'brain_lane.degraded_writes_volatile']`
- **`[memory].enabled = false`** (`app.state.memory_provider is None`, so all three phases fail with "memory provider not configured") — this fires on **every** boot of such a box:
  `EVENTS: [...3x _degraded, 'brain_lane.degraded_writes_volatile']`

The message is *"brain-lane publishes accepted by a volatile provider with no self-heal shell — the agents dataset will not survive a restart"*. In both cases no volatile provider exists and nothing was accepted. This matters more than a normal log nit for two reasons: (1) it is a WARNING claiming durable data loss where there is none, and (2) #1897's own repro greps the journal for exactly this class of key (`grep -E 'boot_degraded|degraded_write|provider_healed'`), so the next `rc-validate` pass will read it as evidence of the defect this PR closes. The PR's stated goal is "the loss is no longer silent in the journal"; a false positive on that key costs more than the silence did.

Same conflation, second symptom: on a shell that has **already healed** (post-heal restart of the lane, `degraded_on_entry` False), a merely-failed publish takes the `brain_lane.degraded_replay_inline` branch and logs *"memory provider healed before the replay could be armed"* before re-running it. The extra retry is harmless — the retry-safety argument above holds — but the label is fiction.

Fix is small: track the volatile-swallowed set and the failed set separately, and gate `degraded_writes_volatile` / `degraded_replay_inline` on the former (and on `degraded_on_entry`), while still feeding the union to the replay. A test for the `[memory].enabled = false` boot would pin it.

---

## Non-blocking

### The third Codex pass was never triaged

Passes 1 and 2 got proper triage comments — good ones. Pass 3 on `325e5374` ("Reconcile identity phases split across the provider swap") has no response. Verified the claim: it is **real but pre-existing and very narrow**. If the swap lands between a phase's `memory_search` (empty fallback → `existing_ids` empty → no delete) and its `memory_add` (durable), `volatile_writes` never moves, the step is correctly not replayed, but the durable dedupe never ran — so on a box whose `agents` bank already holds a prior card you get a duplicate, which is the failure #448 exists to prevent. The window is one in-process search round-trip against a 30 s reprobe tick, and it predates this PR (the #1613 reprobe loop already ran concurrently with the lane). Not a reason to hold this PR — but it should be answered on the thread and either commented in the code or filed as a follow-up, not left silent. Codex silence is not approval, and neither is ours.

### `ast.literal_eval` can still take the whole `peers` listing down

`_unflatten` catches `ValueError, SyntaxError`. `literal_eval` is also documented to raise `MemoryError` / `RecursionError` on sufficiently nested input, and can raise `TypeError` on some malformed values. Engine-produced metadata makes this unlikely, but the entire point of this hunk is that one bad card must cost only its own columns, not the listing. Broaden the except.

### The inline-replay path has no test

`if not _add_heal_hook(replay_state):` carries `# pragma: no cover`, and nothing exercises `add_heal_hook` refusing → inline `_boot_brain_lane(..., replay=True)` → `brain_lane.replay_incomplete`. That path is reachable (my mutation run drove it and the ERROR log fired correctly), and it is precisely what Codex's pass-2 P2 was about, so it deserves a pin. The first commit's PR body claimed a mid-lane inline test; the final test file has none.

### The retry budget can be spent before the lane finishes

`_memory_reprobe_loop` counts a tick as a failed attempt even when the drain only failed because `_BrainLaneReplay.ready` is still False. At the default 30 s interval the budget is ~10 minutes and the lane's worst case is ~90 s, so this is safe as shipped. But `HAL0_MEMORY_REPROBE_INTERVAL_S` is operator-settable: at 1 s, 20 ticks elapse in 20 s, the loop logs `heal_replay_abandoned` and exits while the lane is still publishing, leaving the hook armed with nothing left to drain it — #1897 returns, now with an error log instead of silence. Cheap hardening: don't consume an attempt while every pending hook is merely not-ready.

---

## Scope

Clean. Four source files, two test files, all on the reported path. No `CHANGELOG.md` hunk, no drive-by refactor — the `_memory_reprobe_loop` closure-to-module-level move is load-bearing for the drain and for testability. The `hal0 agent peers` hunk is a separate concern but #1897 explicitly folds it in ("included here rather than filed separately"), and the deliberate exclusions (#1834, `regressions.yaml`) are the right calls.

thinmintdev:
## Review — changes requested (one blocking, rest are nits)

Verified against `pr/1912` head `0f0ca5f0` in a clean worktree, not from the PR description.

### 1. Does it fix #1897's exact path? Yes.

The issue's `expect` is "the `agents` bank exists at install time, **or** the heal path re-runs the
idempotent `brain_lane` publishes from the `provider_healed` branch". This does the second, and the
wiring is real: `create_app` wraps only a degraded boot in `SelfHealingMemoryProvider`
(`api/__init__.py:2750`), `_boot_background_tasks` arms `_memory_reprobe_loop` on it, and the loop
now drains `run_heal_hooks()` immediately after a successful `try_heal()`. The brain-lane publishes
reach the fallback through `_boot_memory_dispatch` → `make_dispatcher(app.state.memory_provider)` →
the shell's `__getattr__` → `PgVectorProvider.add`, which is exactly where the new
`volatile_writes` counter increments — so the per-step accounting reads the object it claims to.

### 2. Root cause, not symptom. Agreed.

The obvious symptom patch here was "move `brain_lane` later" or "re-run the lane once after heal".
Putting the replay contract on `SelfHealingMemoryProvider` instead is the right level: `try_heal()`
rebinding its delegate while telling nobody is the general defect, and the brain lane is only its
first victim. Nothing about the mechanism is brain-lane-specific.

### 3. Do the tests fail without the fix? Yes — verified by mutation, not by trusting the PR body.

* Disabling only the arming (`if False and not replay and degraded_on_entry:`):
  `test_degraded_boot_lane_replays_after_provider_heals`,
  `test_replay_is_armed_before_the_publishes_run`, and
  `test_failed_replay_stays_armed_for_the_next_tick` all fail. 3 failed, 8 passed.
* Reverting the four metadata lines in `agent_peers()` to their `main` form: all three CLI tests
  fail, and the failure is literally
  `Result AttributeError("'str' object has no attribute 'get'")` — the reported crash.

`tests/memory tests/cli` → 999 passed. `tests/api -k "boot or lifespan or memory or brain"` → 157
passed. `ruff check` / `ruff format --check` clean. The two red CodeQL jobs are a GitHub-side SARIF
upload 503 (`##[error]No server is currently available to service your request` after
`Exported results to SARIF`), not a finding.

### 4. Regressions

---

## 🔴 Blocking — `brain_lane.degraded_writes_volatile` now fires on healthy providers

`owed` conflates two different things: "this step's write was swallowed by the volatile fallback"
and "this step didn't land at all".

```python
if _volatile_writes() > before or name in failed:
    owed.append(name)
```

The terminal branch then treats a non-empty `owed` as proof the writes went to a volatile store:

```python
if getattr(provider, "add_heal_hook", None) is None:
    log.warning("brain_lane.degraded_writes_volatile", steps=owed,
        detail="brain-lane publishes accepted by a volatile provider with no self-heal "
               "shell — the agents dataset will not survive a restart")
```

A healthy `HindsightProvider` has no `add_heal_hook` and no `volatile_writes`, so **any** transient
publish failure on a perfectly healthy box lands in that branch. Reproduced (healthy provider,
`degraded = False`, one publish returning `published: False`):

```
[info    ] brain_lane.self_report_degraded warnings=boom
[warning ] brain_lane.degraded_writes_volatile steps=['self_report']
          detail='brain-lane publishes accepted by a volatile provider with no self-heal shell —
                  the agents dataset will not survive a restart'
```

That claim is false — nothing volatile was involved and nothing was accepted. The same branch also
fires with `[memory].enabled = false`, where `provider is None`, all three phases fail with
"memory provider not configured", and boot warns that a dataset that was never written "will not
survive a restart".

This matters more than usual for this wave: `rc-validate`'s memory lane and the register entry for
`agents-bank-boot-writes-lost` both attribute by grepping the boot journal for
`boot_degraded|degraded_write|provider_healed`. A false `degraded_write*` line on a healthy box is
a fresh false positive planted in the exact evidence channel this issue was found through.

Fix is small — keep the two sets apart:

```python
volatile: list[str] = []     # counter moved → really lost
...
if _volatile_writes() > before:
    volatile.append(name)
if name in failed and name not in volatile:
    ...
```

and gate the `degraded_writes_volatile` log on `volatile`, not on `owed`. Arming the replay for
`failed` steps too is fine and should stay — it is only the log that overclaims.

---

### Nits (non-blocking)

**A successful replay is invisible in the journal.** On the happy path the operator sees
`boot_degraded` → `brain_lane.degraded_replay_armed` → `provider_healed` → *silence*.
`_memory_reprobe_loop` returns without a word on the success branch, and
`_boot_brain_lane(replay=True)` logs nothing when it lands. The failure paths are loud
(`heal_replay_retry`, `heal_replay_abandoned`, `replay_incomplete`) but the fix working produces no
evidence that it worked. One `log.info("hal0.memory.heal_replay_landed", steps=...)` before the
`return` would let the next `rc-validate` run prove the remedy instead of inferring it from the
bank's `created_at`.

**`max_replay_attempts` is 20 in code, 10 in the prose.** The PR body and the `5110f1c` commit
message both say "Bounded by `max_replay_attempts` (10)"; the signature default is `20`. At the
30s default interval that is 10 minutes, not 5. Fix one or the other.

**Two new mypy errors.** `_run_boot_phase` is typed `Callable[[], Awaitable[None]]` and
`_boot_brain_lane` now returns `tuple[str, ...]`:

```
src/hal0/api/__init__.py:2308: error: Argument 3 to "_run_boot_phase" has incompatible type
  "Callable[[], Coroutine[Any, Any, tuple[str, ...]]]"; expected "Callable[[], Awaitable[None]]"
src/hal0/api/__init__.py:2308: error: Incompatible return value type
  (got "Coroutine[Any, Any, tuple[str, ...]]", expected "Awaitable[None]")
```

`mypy` is not in CI and this file already carries 32 pre-existing errors, so `make typecheck` was
red before — but this adds to the pile for free. Widening the phase parameter to
`Callable[[], Awaitable[Any]]` clears both.

**The validation kit's acceptance criterion no longer matches the shipped remedy.**
`tests/release-validation/lanes/stateful/memory.md` accepts "(b) the durable bank exists ... and its
`created_at` is **<=** the timestamp of the first `provider_healed` journal line". A replay-based
remedy necessarily creates the bank *after* `provider_healed`. The governing sentence that follows
("only fail if the bank is STILL absent ... after a `provider_healed` line has appeared") still
grades this correctly, so it is not a blocker — but leaving `<=` in place will mis-grade someone.
Reasonable as a follow-up rather than in this PR.

**The inline-replay race is loud but terminal.** If the heal lands between the `degraded_on_entry`
read and `add_heal_hook`, the inline replay gets exactly one attempt; a transient failure there logs
`brain_lane.replay_incomplete` at error and the data waits for a restart. `add_heal_hook` returns
`False` once healed, so there is genuinely nothing to re-arm onto. Accepting this with an ERROR line
is the right call for a window that narrow — noting it so it is a decision and not an oversight.

---

### Codex bot triage

Seven comments, all verified against the code before acting:

| Finding | Verdict |
|---|---|
| P1 keep retrying until replay writes succeed | **Valid, fixed** in `5110f1c` — `run_heal_hooks()` re-arms a `False`/raising hook, loop keeps ticking, bounded by `max_replay_attempts`. Confirmed by `test_failed_replay_stays_armed_for_the_next_tick`. |
| P2 parse the metadata format the producer emits | **Valid, fixed** — `HindsightProvider.add` does store `{k: str(v)}`; JSON→`literal_eval` fallback added with a repr-form test. `literal_eval` is safe here: deep nesting raises `SyntaxError` (caught) on 3.12/3.13, and the int-str limit turns huge literals into `ValueError` (caught). |
| P2 avoid replaying durable writes during a mid-lane heal | **Valid, fixed** via the `volatile_writes` counter; `test_only_the_steps_that_hit_the_fallback_are_replayed` pins it. |
| P2 keep failed inline replays armed | **Valid, partially addressed** — now logs `replay_incomplete` at error rather than discarding silently, but does not retry. See the nit above; accepted deliberately. |
| P2 retry only the steps that remain failed | **Valid, fixed** in `325e537` — `_boot_brain_lane(replay=True)` returns `tuple(failed)` and `_BrainLaneReplay` narrows `pending` to it. |
| P2 wait for asynchronous replay writes | **Valid but pre-existing** — a successful `memory_add` means the retain was accepted, not extracted, for *every* brain-lane publish on `main` too. Correctly deferred; closing it needs sync writes or op-id polling on the `memory_add` MCP surface. |
| P2 reconcile identity phases split across the provider swap | **Valid but pre-existing and unchanged here.** If a heal lands between a phase's dedupe search and its `memory_add`, the search saw the empty fallback and the add goes durable — a duplicate card on a box that already had one. That race exists identically on `main` (the re-probe loop predates this PR); the new counter correctly declines to replay that step, so this PR neither introduces nor worsens it. Worth its own issue for delegate-generation detection, not a blocker on this one. |

The bot's silence is not approval, but in this case it did post — and five of seven findings were
real and were fixed. Good triage discipline on the author's side.

### Scope

Clean. No `CHANGELOG.md` hunk (correct for this wave, #1545). The `agent_commands.py` change is the
spin-off crash reported in the body of #1897, not a drive-by. The `_boot_brain_lane` rewrite from
three try-blocks into a data-driven loop is a real refactor, but it is what makes the per-step
volatile accounting expressible — justified, not gratuitous.

Fix the `degraded_writes_volatile` gating and this is good to merge.

thinmintdev:
Changes-requested review addressed in 7e29c5cb (branch also rebased onto current main — #1909/#1915/#1913/#1919 — clean, no conflicts).

**Blocking: `brain_lane.degraded_writes_volatile` fired on boots where nothing was volatile — fixed exactly as prescribed.** The lane now tracks the volatile-swallowed set (`_volatile_writes() > before`) and the failed set separately. Only their union still feeds the replay (`replay_state.finish` and the inline path are unchanged in what they retry), but the degraded-window messaging keys off `volatile` alone:

- `[memory].enabled = false` boot (provider absent, all three publishes fail) → no `degraded_writes_volatile`, no inline replay. Pinned by `test_memory_disabled_boot_emits_no_volatile_warning`, whose docstring records the rc-validate journal-grep hazard you called out.
- Healthy durable provider, one transient `published: False` → no volatile warning, no `degraded_replay_inline` fiction, the step runs exactly once. Pinned by `test_healthy_provider_publish_failure_is_not_called_volatile`.
- Deliberate `[memory].engine = "pgvector"` (volatile writes, no self-heal shell) still fires the warning — that path is real data loss and `test_deliberate_pgvector_engine_still_warns_volatile` keeps it from being gated away.

The `degraded_replay_inline` branch is now only reachable when something actually hit the fallback (`volatile` non-empty implies the arm-time race, since a healthy-on-entry provider never exposes a moving `volatile_writes` counter), which closes the second symptom — the already-healed-shell boot can no longer log "healed before the replay could be armed" over a plain publish failure.

**Non-blocking: third Codex pass ("Reconcile identity phases split across the provider swap") — now triaged.** Your verification matches mine: real, pre-existing, and very narrow. The window is one in-process dedupe-search round-trip racing the 30 s reprobe tick, it only bites a degraded *restart* on a box whose `agents` bank already holds a prior card, and it predates this PR (the #1613 loop already ran concurrently with the lane — this PR did not add the concurrency). Recorded as a code comment at the `fallback` capture in `_boot_brain_lane` so the next person tracing the counter sees it; closing it needs delegate-generation detection around each search/delete/add sequence, which is #448-adjacent follow-up work, not a bolt-on here.

Verified on the rebased branch: `ruff check` + `ruff format --check` clean; `tests/api tests/memory` → 1983 passed, 1 skipped; `tests/cli` → 783 passed (2 failures are basetemp-path artifacts of running with `--basetemp` inside the repo worktree — `AF_UNIX path too long` and `detect_editable_root` finding the enclosing checkout — both pass with a short basetemp and touch nothing in this diff).

thinmintdev:
Fourth Codex pass triaged — the one finding is real and fixed in f997fdbf.

**P2 (don't spend replay attempts before the lane is ready) — fixed.** Correct: with a short `HAL0_MEMORY_REPROBE_INTERVAL_S`, every tick where the armed `_BrainLaneReplay` merely reported not-ready consumed one of the capped attempts, so the loop could hit `heal_replay_abandoned` and exit before the lane ever called `finish()` — leaving the hook armed with nothing left to drain it. `run_heal_hooks` now keeps a hook whose `ready` attribute is `False` armed **without running it** and exposes `last_drain_attempted`; `_memory_reprobe_loop` counts an attempt only when a replay was actually tried. Pinned by `test_not_ready_replay_does_not_spend_the_attempt_budget`: with a 0 s interval and `max_replay_attempts=3`, the loop ticks far past the cap while the hook is not ready, stays alive, and drains the replay once the lane hands over.

Verified: `tests/api` → 1770 passed, 1 skipped; `tests/api/test_boot_brain_lane_replay.py` + `tests/memory` → 229 passed. `ruff check` + `ruff format --check` clean.

CI note on the earlier head (7e29c5cb): γ-suite needed re-runs for infrastructure, not code — runs 1–2 died in `Install system deps` with apt stalled on the Ubuntu mirror (cancelled at the 12 min cap before a single test ran), run 3 reached 601/653 specs with zero failures before the same 12 min cap (the #1926 slow-runner profile), run 4 passed clean in 9m23s. All other checks (python 3.12, ui, sunset, CodeQL ×3) passed first try.

thinmintdev:
# Re-review (agent, delegated) — APPROVE

Verified on head `f997fdbf` in the branch worktree (`git reset` first; `git diff HEAD` empty — no phantom index). Everything below was re-run, not read off the revision summary.

## 1. Blocking finding from the changes-requested review — resolved at the root

**`brain_lane.degraded_writes_volatile` fired on boots where nothing was volatile.** Fixed in `7e29c5cb` exactly as prescribed: `volatile` (counter delta) and `failed` are now separate lists; `owed = [name for name, *_ in steps if name in volatile or name in failed]` still feeds the replay, and a new `if not volatile: return tuple(failed)` gate sits between the armed-replay branch and both degraded-window log branches. So `degraded_writes_volatile` **and** `degraded_replay_inline` are now unreachable unless something actually hit the fallback — which is the second symptom I raised (the already-healed-shell boot logging "healed before the replay could be armed" over a plain publish failure) closed as well.

Confirmed by mutation, not by reading the diff. Restoring the old gating (`if not owed:` + `steps=owed`):

```
FAILED test_memory_disabled_boot_emits_no_volatile_warning
FAILED test_healthy_provider_publish_failure_is_not_called_volatile
2 failed, 13 passed
```

with the failure message being the exact false positive I reported:
`assert 'brain_lane.degraded_writes_volatile' not in ['brain_lane.self_report_degraded', 'brain_lane.degraded_writes_volatile']`.

`test_deliberate_pgvector_engine_still_warns_volatile` keeps the legitimate loss warning from being gated away, and `test_volatile_write_counter_moves_on_the_real_fallback` pins the fake counter against the real `PgVectorProvider.add` — the fakes are not tautological.

**Non-blocking: third Codex pass triaged.** Answered on the thread and recorded as a code comment at the `fallback` capture in `_boot_brain_lane` (lines 2197–2204). The recorded characterisation matches my own verification: real, pre-existing, one in-process dedupe-search round-trip against a 30s reprobe tick, not introduced or worsened here.

## 2. No regression of the original #1897 fix

The issue's `expect` is "the `agents` bank exists at install time, **or** the heal path re-runs the idempotent `brain_lane` publishes from `provider_healed`". That second path is untouched by the revision: the `not volatile` gate sits **after** the `replay_state is not None` branch, so a real degraded boot still arms the hook before the publishes, calls `finish(owed)`, logs `degraded_replay_armed`, and gets drained by `_memory_reprobe_loop` after `try_heal()`.

Revert-and-confirm-red on the core fix, on the revised head:

* disable only the arming (`if False and not replay and degraded_on_entry:`) → `test_degraded_boot_lane_replays_after_provider_heals`, `test_replay_is_armed_before_the_publishes_run`, `test_failed_replay_stays_armed_for_the_next_tick` **fail** (3 failed, 12 passed)
* `src/hal0/cli/agent_commands.py` restored to `main` → all 3 of `tests/cli/test_agent_peers_metadata.py` **fail**, with the reported `AttributeError`

Green on the branch: `tests/api/test_boot_brain_lane_replay.py tests/cli/test_agent_peers_metadata.py tests/memory` → **232 passed**.

## 3. The new Codex P2 fix (`f997fdbf`) is real and correctly scoped

`run_heal_hooks` now keeps a hook with `ready is False` armed *without running it* and records `last_drain_attempted`; `_memory_reprobe_loop` only spends an attempt when a replay was actually tried. The `continue` lands before `attempts += 1` and after `await asyncio.sleep(interval_s)` at the top of the loop, so there is no busy-spin. Mutating the guard to `if False:` fails `test_not_ready_replay_does_not_spend_the_attempt_budget` with the exact symptom described (`heal_replay_retry` ×2 → `heal_replay_abandoned` while the lane was still running). Mixed-hook case (one ready-and-failed + one not-ready) still counts the attempt — correct.

## 4. No new defects; CI green

`ruff check src tests` clean, `ruff format --check src tests` → 1147 files already formatted. All 8 checks on `f997fdbf` are `success`, including `python (3.12)` and `γ-suite (chromium)` (no cancelled-run ambiguity — verified via `commits/f997fdbf/check-runs`).

One observation, explicitly **not** a finding: if `_boot_brain_lane` were to raise between arming the hook and `finish()`, the hook would stay armed and not-ready forever and the reprobe loop would now tick indefinitely instead of giving up loudly after 20 attempts. `_run_boot_phase` re-raises, so that path fails the lifespan boot rather than leaving a live process with a dangling task — and the tick is a no-op `try_heal` short-circuit. Noting it so the trade is on record.

Carry-over nits from the previous review, all still open and all still the author's call — none of them blocks:

* `_unflatten` still catches only `(ValueError, SyntaxError)` (`literal_eval` deep-nesting behaviour makes this fine in practice on 3.12/3.13)
* no `heal_replay_landed`-style success log — a working replay is still invisible in the journal
* `max_replay_attempts` is 20 in code, "10" in the `5110f1c` commit message
* the two mypy errors from `_run_boot_phase`'s `Awaitable[None]` signature (mypy is not in CI)
* `tests/release-validation/lanes/stateful/memory.md`'s `created_at <= provider_healed` criterion now mis-describes a replay-based remedy — follow-up, not this PR
* the inline-replay path (`# pragma: no cover`) still has no test

Blocking count: **0**. Ship it.

PR #1913, opened 2026-08-17
fix(bench): print no-records notice for empty history results

### LINKED ISSUE #1904: cli: hal0 bench history prints zero bytes at rc=0 on an empty result, indistinguishable from a hang

### Summary

`hal0 bench history` prints ZERO bytes at rc=0 whenever its query matches no rows — empty store OR a non-matching `--model`/`--cell` filter on a populated store — while the sibling `hal0 bench results` prints "no records" (that branch was added by `#1807` for exactly this "indistinguishable from a hang" reason; `cmd_history` was missed). `--json` correctly prints `[]` on both.

### Repro

```
hal0 bench history --model nosuchmodel >/tmp/h.out 2>/tmp/h.err; echo rc=$?; wc -c /tmp/h.*
```

### Expected

An empty result prints "no records" (mirroring `cmd_results`) at rc=0.

### Test to promote

`tests/bench/test_cli_history.py` mirroring `test_cli_results`'s two empty-state cases — this entry should not survive to the next release.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `bench-history-blank-empty-state`.

thinmintdev:
## What
`hal0 bench history` now prints `no records` at rc=0 when its query
matches no rows (empty store, or a non-matching `--model`/`--cell`
filter on a populated store), mirroring the sibling `cmd_results`
branch added by #1807. `--json` already correctly printed `[]` and is
unchanged.

## Why
Before this fix, an empty non-`--json` result printed zero bytes at
rc=0 — indistinguishable from a hung command. `cmd_results` got this
fix in #1807 for the same reason; `cmd_history` was missed.

## Verified
- New regression test `tests/bench/test_cli_history.py` (mirrors
  `test_cli_results.py`'s two empty-state cases): confirmed it fails
  against the pre-fix code (`assert 'no records' in ''`) and passes
  after the fix.
- `uv run pytest tests/bench/` — 387 passed.
- `make lint` — clean.
- `uv run ruff format --check` on changed files — clean.

## Out of scope
Nothing else touched; this is a one-line mirror of the existing
`cmd_results` empty-state branch plus its regression test.

Closes #1904

### DIFF
--- a/src/hal0/bench/cli.py
+++ b/src/hal0/bench/cli.py
@@ -548,6 +548,10 @@ def cmd_history(args: argparse.Namespace) -> int:
     rows = store.history(cell_key=args.cell, model=args.model, limit=args.limit)
     if args.json:
         print(json.dumps(rows, indent=2))
+    elif not rows:
+        # Same "empty table" vs. "nothing has ever run" ambiguity as
+        # cmd_results (#1796) — say so instead of printing nothing (#1904).
+        print("no records")
     else:
         for r in rows:
             print(f"  {r['ts']}  {r['model_id']:40s} {r['decode_ts_med']}")
--- a/tests/bench/test_cli_history.py
+++ b/tests/bench/test_cli_history.py
@@ -1,0 +1,27 @@
+"""`hal0 bench history` — empty-store output (#1904).
+
+Before this fix an empty query result (empty store, or a non-matching
+``--model``/``--cell`` filter on a populated store) printed nothing at all
+for the non-``--json`` path, indistinguishable from a hung/broken command.
+``cmd_results`` got this fix in #1807; ``cmd_history`` was missed.
+"""
+
+from __future__ import annotations
+
+from hal0.bench.cli import main
+
+
+def test_history_empty_prints_no_records(tmp_path, monkeypatch, capsys):
+    monkeypatch.setenv("HAL0_BENCH_STATE", str(tmp_path))
+    rc = main(["history"])
+    out = capsys.readouterr().out
+    assert rc == 0
+    assert "no records" in out
+
+
+def test_history_empty_json_stays_empty_list(tmp_path, monkeypatch, capsys):
+    monkeypatch.setenv("HAL0_BENCH_STATE", str(tmp_path))
+    rc = main(["history", "--json"])
+    out = capsys.readouterr().out
+    assert rc == 0
+    assert out.strip() == "[]"


thinmintdev:
## Review — approve

Verified independently against `main` (6e45d717), not from the PR description.

**1. Fixes the reported defect — yes, the exact path.** Issue #1904's repro is
`hal0 bench history --model nosuchmodel`. `hal0 bench` is an argv passthrough to
`hal0.bench.cli.main` (`src/hal0/cli/main.py:106-109`), and `Store.history()`
(`src/hal0/bench/store.py:311-334`) applies `--cell`/`--model` inside the SQL and returns
`[]` for a non-matching filter exactly as it does for an empty store. The new `elif not rows`
branch keys off the returned list, so both arms of the issue ("empty store OR a non-matching
filter on a populated store") land on the same `print("no records")`. `--json` is untouched
and still prints `[]`.

**2. Root cause vs symptom — root cause, for this verb.** The cause was a missing empty-state
arm in `cmd_history`, and the arm is what the diff adds. It is a literal mirror of the
`cmd_results` branch at `cli.py:533-536`, which is the shape the register entry
(`bench-history-blank-empty-state`) and `lanes/readonly/cli.md:40-45` ask for. I swept the
other `cmd_*` in `bench/cli.py` for the same class: `cmd_status` always prints its
`state root:` / `records:` header lines, and the remaining verbs print an unconditional
result line, so `cmd_history` was the last blank-at-rc-0 read verb in this module. Nothing
broader is left in place to be re-found.

**3. Does the test fail without the fix — yes, confirmed by running it.** I applied only the
`tests/` half of the diff onto unpatched `main`:

```
FAILED tests/bench/test_cli_history.py::test_history_empty_prints_no_records
E       AssertionError: assert 'no records' in ''
1 failed, 1 passed
```

The second case (`--json` → `[]`) passes pre-fix, which is correct — it is a pin against the
fix regressing the JSON path, not a red-first case. With the source hunk applied,
`uv run pytest tests/bench/` → **387 passed**. `ruff format --check` and `ruff check` on both
changed files: clean.

**4. Regressions — none found.** `cmd_history` has exactly one caller
(`p.set_defaults(func=cmd_history)`, `cli.py:914`), no test asserted the old zero-byte
behaviour, and nothing in `src/`, `installer/`, or the docs parses `bench history` stdout
(`installer/agent-skills/hal0-bench/SKILL.md` and `hal0-tune/SKILL.md` only mention the verb
in prose). The only behaviour change for any caller is the added line on the empty non-JSON
path, which is the requested change.

**5. Scope — clean.** Two files, +31/-0. No CHANGELOG.md hunk (forbidden in this wave), no
drive-by refactor, no unrelated edits. Not touching `tests/release-validation/regressions.yaml`
matches precedent: #1807 fixed the sibling `cmd_results` case and left the register to the
report-time curation pass.

**6. Codex bot triage.** No Codex review comments and no reviews of any kind are present on
this PR. Its silence is not an approval — I did not treat the absence as a signal and reviewed
the diff from scratch. Nothing to fix, file, or dismiss.

### Nit (non-blocking)

`tests/bench/test_cli_history.py` covers only the empty-store arm; the issue's literal repro
(non-matching `--model` on a *populated* store) is not exercised. It is the same `not rows`
branch, and the issue itself asked for a mirror of `test_cli_results`'s two cases, so this
matches the requested scope — but a third case seeding one record and querying
`--model nosuchmodel` would pin the filter arm the validation run actually hit. Fine to land
as-is.

CI is green across all 8 checks. LGTM.

PR #1915, opened 2026-08-17
fix(providers): catch httpx.TransportError in container health probe

### LINKED ISSUE #1898: ContainerProvider.health() doesn't catch httpx.ReadError — a mid-read RST aborts slot load at ~7s with an empty error message

### Summary

`ContainerProvider.health()` catches only `(httpx.ConnectError, httpx.TimeoutException)`. A runner that binds its port and then dies mid-read raises `httpx.ReadError` (`str()` is `""`), unguarded up through `wait_ready`/`_await_ready`/`SlotManager.load`: the load aborts at ~7s instead of running the full 180s window, the ERROR stamp gets an EMPTY `metadata.message` (the explicit `still_report_if` clause of the known `crash-loop-warming-180s-window` entry), and the route answers a generic `system.internal` 500 (the symptom is already covered by `#1424` — record that half as known).

Every other `httpx` catch in `src/hal0` uses `HTTPError`/`RequestError`; the sibling call in `slots/watchdog.py` is already guarded correctly. The same unguarded probe sits on the dispatch hot path (`SlotWatchdog.readiness_check` via `Dispatcher._check_container_slot_ready`) — code-read only, would 500 an inference request instead of the retryable `slot.loading` 503. No test in the suite simulates anything but `ConnectError`.

### Repro

```
# Deterministic: bind a socket, accept one connection, SO_LINGER{1,0} abortive close, then
# await container_provider().health(port) -> raises httpx.ReadError str=''.
# In the wild: intermittent during a crash-looping load (observed 1 in 5 failed loads on rc.6);
# decisive journal shape: `hal0.unhandled  error=` (empty) + ReadError + POST .../load 500.
```

### Expected

`health()` catches `httpx.TransportError`; a mid-read RST is treated as "unhealthy", the load runs its full window, `metadata.message` is never empty on error, and the dispatch gate keeps its `slot.loading` 503 contract.

### Test to promote

A pytest `ReadError` case beside the existing `ConnectError` simulations.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `container-health-readerror-unhandled`.

thinmintdev:
## Summary

`ContainerProvider.health()` (`src/hal0/providers/container.py`) caught only
`(httpx.ConnectError, httpx.TimeoutException)`. A runner that binds its port
and then dies mid-read raises `httpx.ReadError` (`str()` is `""`), which sat
outside that narrow catch and escaped `health()` as an unhandled exception.

That propagated through three layers, matching the exact symptoms in #1898:

1. `wait_ready()`'s poll loop calls `self.health(port)` unguarded — the
   unhandled `ReadError` aborts the loop on the first strike instead of
   continuing to poll for the full `_HEALTH_TIMEOUT_S` (~180s) window, so the
   load dies at ~7s.
2. `SlotManager.load()`'s outer `except Exception as exc: ... message=str(exc)`
   then stamps the slot `ERROR` with an **empty** `metadata.message`, since
   `str(ReadError(""))` is `""` — the `still_report_if` clause of the known
   `crash-loop-warming-180s-window` entry.
3. The same unguarded probe sits on the dispatch hot path
   (`SlotWatchdog.readiness_check` → `Dispatcher._check_container_slot_ready`).
   `readiness_check`'s second probe (`container_provider().health(port, cfg)`)
   is not wrapped in its own try/except (unlike the sibling `is_healthy`
   method just above it, which already catches broadly) — an unhandled
   `ReadError` there would bubble out of `_check_container_slot_ready` as a
   raw 500 instead of the documented retryable `slot.loading` 503.

## Root cause

The root cause is entirely in `ContainerProvider.health()`'s exception
handling — every downstream site (`load()`'s outer except, `readiness_check`'s
sibling probe, `is_healthy`) is already written to defensively catch broadly.
They only misbehave because `health()` itself leaks a transport error instead
of converting it to `{"ok": False, ...}`. Fixing the single narrow `except`
clause resolves all three symptoms without touching `manager.py`, `router.py`,
or `watchdog.py` — no broadening of blast radius needed beyond the one method.

## Fix

- Catch `httpx.TransportError` instead of the two named subclasses — the
  base class covering `ConnectError`, `TimeoutException`, `ReadError`,
  `WriteError`, and every other failure mode where the connection dies
  without producing an HTTP response.
- Fall back to the exception's class name (`type(exc).__name__`) when
  `str(exc)` is empty, so the returned `status` (and therefore any
  downstream `metadata.message`) is never blank.

## Test

Added `test_health_read_error_stays_unhealthy` beside the existing
`test_health_connect_refused_stays_unhealthy` in
`tests/providers/test_container_npu.py` — mocks the client's `GET` to raise
`httpx.ReadError("")` and asserts `health()` returns `{"ok": False, ...}`
instead of letting the exception propagate. Confirmed it fails on `main`
(pre-fix) and passes after the one-line `except` change.

## Verified

- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/providers/test_container_npu.py tests/providers/test_container_health_gating.py tests/providers/test_container.py tests/slots/ tests/dispatcher/ -q` → 1207 passed
- `uv run ruff check src/hal0/providers/container.py tests/providers/test_container_npu.py` → all checks passed
- `uv run ruff format --check src/hal0/providers/container.py tests/providers/test_container_npu.py` → already formatted

(A full-repo `pytest -q` was also kicked off but the box was under heavy
contention from several other parallel fix-wave worktrees running full
suites simultaneously at the same time; the scoped run above covers every
module touched by this change plus their direct dependents.)

## Out of scope

- Did not touch `CHANGELOG.md` (per fix-wave process — consolidated at the
  end of the wave).
- Did not add defensive try/except wrapping in `watchdog.py`'s
  `readiness_check` or `router.py`'s `_check_container_slot_ready` — those
  are the *symptom* sites and are already written correctly to trust that
  `health()` never raises; the fix belongs in `health()` itself.

Closes #1898

### DIFF
--- a/src/hal0/providers/container.py
+++ b/src/hal0/providers/container.py
@@ -2203,8 +2203,19 @@ class ContainerProvider(Provider):
                     "ok": ok,
                     "status": "healthy" if ok else f"http_{models_resp.status_code}",
                 }
-        except (httpx.ConnectError, httpx.TimeoutException) as exc:
-            return {"ok": False, "status": str(exc)}
+        except httpx.TransportError as exc:
+            # TransportError is the base class for ConnectError,
+            # TimeoutException, ReadError, WriteError, etc. — every failure
+            # mode where the connection dies without producing an HTTP
+            # response (#1898: a runner that binds its port and then dies
+            # mid-read raises ReadError, whose str() is empty; that escaped
+            # the old narrow (ConnectError, TimeoutException) catch as an
+            # unhandled exception, abandoning wait_ready()'s poll loop after
+            # a single strike instead of reporting "unhealthy" and retrying
+            # for the full window). status falls back to the exception's
+            # class name so metadata.message is never empty even when
+            # str(exc) is "".
+            return {"ok": False, "status": str(exc) or type(exc).__name__}
 
     async def wait_ready(self, port: int) -> None:
         """Poll /health until 200 or HEALTH_TIMEOUT_S exceeded.
--- a/tests/providers/test_container_npu.py
+++ b/tests/providers/test_container_npu.py
@@ -302,6 +302,32 @@ async def test_health_connect_refused_stays_unhealthy() -> None:
 
 
 @pytest.mark.anyio
+async def test_health_read_error_stays_unhealthy() -> None:
+    """A mid-read RST (httpx.ReadError, str()=="") must not escape health() as
+    an unhandled exception (#1898). A runner that binds its port and then
+    dies mid-read raises ReadError, which sat outside the narrow
+    (ConnectError, TimeoutException) catch — that unhandled exception
+    aborted the caller's poll loop instead of reporting "unhealthy" for the
+    next retry."""
+    import httpx
+
+    provider = ContainerProvider()
+
+    async def fake_get(url: str, **kwargs: Any) -> MagicMock:
+        raise httpx.ReadError("")
+
+    mock_client = AsyncMock()
+    mock_client.__aenter__ = AsyncMock(return_value=mock_client)
+    mock_client.__aexit__ = AsyncMock(return_value=False)
+    mock_client.get = fake_get
+
+    with patch("hal0.providers.container.httpx.AsyncClient", return_value=mock_client):
+        result = await provider.health(8088)
+
+    assert result["ok"] is False
+
+
+@pytest.mark.anyio
 async def test_health_200_on_health_still_healthy() -> None:
     """Existing /health→200 path still works (GPU slots use it)."""
     provider = ContainerProvider()


thinmintdev:
## Review: fix(providers): catch httpx.TransportError in container health probe

**Verdict: approve.** The fix is at the root cause, on the exact path #1898 describes, and the new test genuinely fails without it. Notes below are non-blocking.

### Verified against the issue

- **Exact path, not an adjacent one.** #1898 names `ContainerProvider.health()`'s narrow `except (httpx.ConnectError, httpx.TimeoutException)` and asks for `httpx.TransportError`. That is precisely the hunk. Checked the httpx hierarchy in-tree: `ReadError`, `WriteError`, `CloseError`, `ConnectError`, `TimeoutException`, `RemoteProtocolError`, `ProxyError`, `PoolTimeout` are all `TransportError` subclasses, so the mid-read-RST case and its likely cousin (`RemoteProtocolError`, "server disconnected without sending a response") are both covered.
- **Root cause, not symptom.** `health()`'s documented contract is `Returns {"ok": bool, "status": str}`; `wait_ready()`, `SlotManager.load()`'s outer except, `SlotWatchdog.is_healthy` and `readiness_check` are all written to trust that. Patching those sites would have been the symptom fix. Agreed with the PR's "out of scope" reasoning.
- **Delegation branch checked.** `health()` returns early to `FLMProvider.health` / `ComfyUIProvider.health` before the `try`. Both already end in `except httpx.HTTPError` + `except Exception`, so no ReadError leaks through that branch either. `_spec_provider_for` is already wrapped. `MoonshineProvider.health` still has the old narrow tuple but has a catch-all `except Exception` after it and is documented dead code on this path — no residual leak.

### Test verified independently (not taken on trust)

Reproduced the new test against **unpatched** `main`:

```
src/hal0/providers/container.py:2186: in health
    resp = await client.get(health_url)
E       httpx.ReadError
1 failed in 0.13s
```

It fails for the right reason — the exception escapes `health()` rather than the assert tripping. Applied the one-line `except` change locally and it passes. Also confirms the `patch("hal0.providers.container.httpx.AsyncClient", ...)` indirection actually reaches `async_client()`'s construction.

Regression sweep with the patch applied locally: `tests/providers tests/slots tests/dispatcher` → **1540 passed**. PR CI is green (python 3.12, γ-suite, ui, CodeQL).

### Scope

Clean. Two files, no `CHANGELOG.md` hunk (correct for this wave), no drive-by refactor.

### Codex bot

`pulls/1915/reviews` and `issues/1915/comments` are both empty — the bot posted nothing at all on this PR. Recording that explicitly: that is silence, not an approval, so there was nothing to triage.

### Non-blocking notes

1. **The test only covers half the fix.** It asserts `result["ok"] is False` but never asserts the `status`, which is the `str(exc) or type(exc).__name__` fallback — and a non-empty message is the *named* symptom in #1898 (`still_report_if`: empty `metadata.message`). A future refactor could drop the fallback and this test would stay green. Suggest adding one line:
   ```python
   assert result["status"] == "ReadError"
   ```
   The regression-register key `container-health-readerror-unhandled` will be re-probed by rc-validate, so it is worth locking down here.

2. **Semantic change for a caller the issue does not mention.** `SlotWatchdog.is_healthy` (`slots/watchdog.py:417`) previously saw a `ReadError` as an *exception* and returned `True` — its docstring says "a probe exception returns `True` so the fail-watch never demotes a slot it cannot actually judge". Post-fix the same RST returns `ok=False`, i.e. it now counts as a fail-watch strike toward demotion. I think this is correct and consistent (`ConnectError` already behaved exactly this way, and `_HEALTH_FAIL_STRIKES` absorbs transients), but it is a real behaviour change on the fail-watcher and should be called out in the wave notes rather than discovered later.

3. **`TransportError` is narrower than the repo convention.** The issue observed that every other httpx catch in `src/hal0` uses `HTTPError`/`RequestError`. `httpx.DecodingError` and `TooManyRedirects` are `RequestError` but *not* `TransportError`, and `DecodingError` is raised at read time inside the `try` — the `except ValueError` around `resp.json()` does not cover it. Practically unreachable against a local llama-server; noted only for completeness.

4. **`readiness_check` stays the one unguarded probe on the dispatch hot path** (`watchdog.py:466`). The PR's rationale for leaving it alone is sound, but `_maybe_load_config` and `is_active` above it are unguarded too, so a raw 500 there is still reachable from a non-`health()` source. Worth a separate follow-up issue, not a blocker for #1898.

PR #1916, opened 2026-08-17
docs(mcp): fix bare mount URLs and enumerate every MCP tool

### LINKED ISSUE #1902: docs: connect-mcp.mdx gives bare mount URLs that 405 on loopback; mcp-tools.mdx claims to list every tool but omits 21 gated destructive ones

### Summary

Two shipped-docs defects:

1. `docs/guides/connect-mcp.mdx` "### Connect" gives `http://localhost:8080/mcp/admin` and `/mcp/memory` as the URLs to point a client at — both 405 on loopback (no redirect; the working transport URL is `<mount>/mcp`, as `/etc/hal0/MCP-CLIENTS.md` itself says). This is the literal `still_report_if` of the known `mcp-lan-421-invalid-host` entry firing.
2. `docs/reference/mcp-tools.mdx` claims to enumerate "every MCP tool" but its tables name 91 of the 180 admin tools (**21 GATED/destructive tools undocumented**) and 5 of the 26 memory tools; the source frozensets were already complete when the page was written.

### Repro

```
sed -n 69,74p docs/guides/connect-mcp.mdx     # bare mount URLs = defect
# tools/list over /mcp/admin/mcp -> count; regex backticked names out of the doc's table
# rows; comm the two name sets.
```

### Expected

The Connect section's URLs end in `/mcp` and `initialize` returns 200 against them verbatim; the tool reference either enumerates every served tool (assert by name-set diff, not count) or stops claiming to.

### Test to promote

A doc-lint pytest diffing the mdx tables against the `AUTONOMOUS_READ`/`WRITE` + `GATED` frozensets, and grepping `connect-mcp.mdx` for bare mount URLs.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `mcp-docs-connect-url-and-tool-tables`.

thinmintdev:
## What changed

1. **`docs/guides/connect-mcp.mdx`** — the "### Connect" section told a client to point at the bare mount URL (`http://localhost:8080/mcp/admin`, `.../mcp/memory`). Those 405 with no redirect: FastMCP's Streamable-HTTP sub-app only serves `POST` on `/mcp/admin/mcp` / `/mcp/memory/mcp`, the mount root itself is not the transport endpoint (`hal0.api.routes.mcp.mcp_url` and #1796 already pinned this exact rule for the REST introspection surface — the curl example a few lines further down the same doc already used the correct `/mcp/admin/mcp` path; only the prose above it was wrong).
2. **`docs/reference/mcp-tools.mdx`** — claimed to enumerate "every MCP tool" but its tables named a stale subset: 89 of 180 admin tools (`AUTONOMOUS_READ_TOOLS`/`AUTONOMOUS_WRITE_TOOLS`/`GATED_TOOLS` in `hal0.mcp.admin`) and 21 of 26 memory tools (`@server.tool` registrations in `hal0.mcp.memory`) were missing — mostly tools added in later waves that never got backfilled into the doc. Regenerated both tables directly from source: the admin tables now list all 180 tools with their REST route (via `TOOL_NAME_ALIASES`), and the memory section is restructured into the same three gating tiers as the admin section (read / write / gated), since `hal0.mcp.admin` is the single owner of that classification even for memory-only tools.

## Why

Found by `rc-validate` v1.0.0-rc.6 on `ct151-cpu-fresh` (issue #1902): an agent following the doc's Connect section got a bare 405 with the URLs as written, and a name-set diff of the tool tables against the source frozensets showed the "every tool" claim was false.

## How verified

- New regression test `tests/docs/test_mcp_docs_match_source.py`:
  - `test_connect_section_has_no_bare_mount_urls` / `test_connect_section_urls_end_in_mcp` — reproduce the bare-URL defect (failed before the fix, pass after).
  - `test_admin_tool_tables_name_every_classified_tool` — diffs the mdx admin tables against `AUTONOMOUS_READ_TOOLS | AUTONOMOUS_WRITE_TOOLS | GATED_TOOLS` by name set. Failed with the exact 89 missing tool names before the fix; passes after.
  - `test_memory_tool_table_names_every_registered_tool` — same diff against the memory module's `@server.tool` registrations. Failed with the exact 21 missing tool names before the fix; passes after.
- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/docs/ -q` → 10 passed.
- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest -q -x` (full suite) → 1170 passed, 1 unrelated pre-existing flake (`test_health_degraded.py::test_health_system_ok_when_no_errored_slots`, fails only under full-suite CPU/global-state contention, passes standalone; not touched by this change), 1 skipped, 1 xfailed.
- `uv run --extra dev ruff check tests/docs/test_mcp_docs_match_source.py docs` → all checks passed.
- `uv run --extra dev ruff format --check tests/docs/test_mcp_docs_match_source.py` → already formatted (after running `ruff format` once to apply the project's line-wrap convention).

## Left out of scope

- Per-tool argument schemas for the 89 newly-added admin rows were intentionally left out of the table (only the REST route is listed) — reproducing every tool's full arg list by hand across 180 rows is itself a drift-prone manual doc, and the table now points readers at `TOOL_PARAM_HINTS` in source for that detail, consistent with the intro paragraph's existing "treat the frozensets/route map in source as ground truth if this table drifts" guidance.
- Did not touch `CHANGELOG.md` per the fix-wave convention (tracked in #1545 — consolidated at the end of the wave).

Closes #1902

### DIFF
--- a/docs/guides/connect-mcp.mdx
+++ b/docs/guides/connect-mcp.mdx
@@ -67,11 +67,12 @@ or behind your own reverse proxy — do not expose `/mcp/*` to the internet.
 
 ### Connect
 
-Point your MCP client at the mount URL on the hal0 host:
+Point your MCP client at the Streamable-HTTP endpoint on the hal0 host — the
+mount root itself 405s (no redirect); the working path always ends in `/mcp`:
 
 ```
-http://localhost:8080/mcp/admin
-http://localhost:8080/mcp/memory
+http://localhost:8080/mcp/admin/mcp
+http://localhost:8080/mcp/memory/mcp
 ```
 
 ### Authenticate
--- a/docs/reference/mcp-tools.mdx
+++ b/docs/reference/mcp-tools.mdx
@@ -33,6 +33,11 @@ Tools fall into three gating tiers. Treat the frozensets in source
 (`AUTONOMOUS_READ_TOOLS`, `AUTONOMOUS_WRITE_TOOLS`, `GATED_TOOLS`) as ground truth if
 this table drifts.
 
+Full argument schemas are not reproduced per tool here — the REST route each
+tool passes through documents its own body/query shape; hand-authored
+overrides for the handful that don't map 1:1 onto their route live in
+`TOOL_PARAM_HINTS` in source.
+
 ### Autonomous — read (no approval)
 
 | Tool | REST route |
@@ -45,7 +50,6 @@ this table drifts.
 | `slot_by_id` | `GET /api/slots/by-id/{slot_id}` |
 | `slot_resolved` | `GET /api/slots/{name}/resolved` |
 | `slot_state` | `GET /api/slots/{name}/state` |
-| `port_list` | `GET /api/ports` |
 | `model_list` | `GET /api/models` |
 | `model_show` | `GET /api/models/{model_id}` |
 | `model_scan_preview` | `POST /api/models/scan/preview` |
@@ -53,16 +57,15 @@ this table drifts.
 | `model_update_check` | `GET /api/models/updates/check` |
 | `model_pulls_list` | `GET /api/models/pulls` |
 | `model_pull_status` | `GET /api/models/{model_id}/pull/status` |
-| `model_inspect` | `POST /api/models/inspect` — args: `hf_repo` (`org/name`) or `hf_url` |
+| `model_inspect` | `POST /api/models/inspect` |
 | `model_store` | `GET /api/settings/models/store` |
 | `hardware_probe` | `GET /api/stats/hardware` |
 | `system_info` | `GET /api/system-info` |
+| `port_list` | `GET /api/ports` |
 | `capability_list` | `GET /api/capabilities` |
 | `provider_list` | `GET /api/providers` |
 | `version_info` | `GET /api/status` |
 | `upstream_list` | `GET /api/upstreams` |
-| `upstream_get` | `GET /api/upstreams/{name}` |
-| `upstream_test` | `POST /api/upstreams/{name}/test` |
 | `stack_list` | `GET /api/stacks` |
 | `stack_status` | `GET /api/stacks/{slug}` |
 | `profile_list` | `GET /api/profiles` |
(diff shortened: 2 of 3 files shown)


thinmintdev:
## Review verdict: **changes requested**

## Review — verified against source, not against the PR description

Reproduced everything below in a clean worktree at `0081078e`.

### What checks out

- **The Connect fix is the right path.** `src/hal0/api/routes/mcp.py:451-456` says exactly this: the trailing `/mcp` is the FastMCP Streamable-HTTP transport path, the bare `/mcp/<mount>` root 405s (#1796). `connect-mcp.mdx` is now internally consistent — lines 74, 75, 99, 178, 185 all end in `/mcp`, and no other doc still tells a client to point at a bare mount root. The rest of the repo agrees with the same convention (`hermes_provision.py:2560` strips one trailing `/mcp` precisely because the probe helper appends it).
- **The tables really do match source now.** I diffed each tier's table rows against the frozensets by name set, per tier, not just by presence on the page:
  `read 102 doc / 102 src`, `write 24 / 24`, `gated 54 / 54` — no missing, no extras. The 26 `@server.tool` memory registrations are all present, and every memory row's tier matches `AUTONOMOUS_READ_TOOLS`/`AUTONOMOUS_WRITE_TOOLS`/`GATED_TOOLS` (`memory_mental_model_delete` and `memory_directive_delete` are indeed in `GATED_TOOLS`). The `Deliberately excluded` seven are all genuinely unclassified. Spot-checked the new memory `Required` columns against `_require`/`_optional_id` in `hal0/mcp/memory.py` — correct except for `memory_curate` (below).
- **The tests are red before the fix.** Ran the new file against `main`'s `docs/` tree: all 4 fail — 89 admin names, 21 memory names, plus both Connect assertions. Green on the PR tree (10 passed in `tests/docs/`). Ruff check + format clean. CI green. No `CHANGELOG.md` hunk, no source changes, no drive-by refactors — scope is clean.

### Blocking — the regression ratchet is weaker than the issue asked for

`test_admin_tool_tables_name_every_classified_tool` / `test_memory_tool_table_names_every_registered_tool` run `_backticked_names(_tools_text())` over the **whole page** and only assert `src - doc == {}`. That is a "name appears somewhere on the page" check, not the table diff #1902 asked to promote ("a doc-lint pytest diffing the mdx **tables** against the frozensets"). Demonstrated, not theorised — I deleted the entire `hal0-memory` tool section (all 26 rows of the thing this PR just wrote) and reran:

```
$ pytest tests/docs/test_mcp_docs_match_source.py -q
....                                                    [100%]
4 passed in 0.02s
```

It passes because the names survive in the admin delegate rows. Three drift modes this leaves open, all of them the exact failure class #1902 reported:

1. A whole documented section can be deleted or gutted silently.
2. **Tier drift is invisible** — moving `service_action` from Gated to Autonomous-read still passes. The issue's headline complaint was specifically "21 GATED/destructive tools undocumented", so tier is the load-bearing bit.
3. Stale rows are invisible (the diff is one-directional): rename a tool in source and the doc keeps a phantom row for the old name forever while the test only nags about the new one.

Please parse the three `### Autonomous — read` / `— write` / `Gated` table bodies (and the three memory sub-tables) and assert **set equality** per tier against the corresponding frozenset. It is ~20 lines — this is what I used to verify the tables above:

```python
def _rows(section: str) -> set[str]:
    out: set[str] = set()
    for line in section.splitlines():
        if line.startswith("|") and not line.startswith("|---"):
            out |= set(re.findall(r"`([a-z][a-z_0-9]*)`", line.split("|")[1]))
    return out
# then: assert _rows(read_section) == set(AUTONOMOUS_READ_TOOLS)  # etc.
```

(Independently flagged by Codex, P2 on `tests/docs/test_mcp_docs_match_source.py:103` — that comment is correct, both of its examples reproduce.)

### Should fix — two claims this page now makes that source contradicts

Both are Codex P2s; I verified both against source rather than taking the bot's word.

1. **`docs/reference/mcp-tools.mdx:27-28`** — "success (autonomous): the raw REST JSON response — `slot_list`, `provider_list`, and `upstream_list` wrap their bare-list REST response". `_LIST_TOOL_WRAP_KEY` (`hal0/mcp/admin.py:217-225`) also contains `runner_image_pulls → "pulls"` and `backend_list → "backends"`. Pre-existing staleness, but this PR is the one that newly documents both tools, so a client reading this page will now parse a top-level array that never arrives. Two names to add.
2. **`memory_curate` row** — `Required: id`. `_memory_curate` (`hal0/mcp/memory.py:682-685`) raises `MemorySchemaError("memory_curate requires at least one field to change")` unless one of `text/context/occurred_start/occurred_end/fact_type/entities/state` is also present, so the documented minimum call returns `mcp.memory_schema`. `memory_mental_model_update` / `memory_directive_update` do **not** have this rule, so it is genuinely `memory_curate`-only.

### Nits (no action required)

- Dropping the `Args` column from the write/gated tables deletes real content (`model_swap`, `model_assign`, `model_scan`, `model_set_default`, `slot_create`, `slot_rename`, `model_pull`, `upstream_create`, `upstream_update`, `model_inspect`). I checked each one: every removed hint still exists in `TOOL_PARAM_HINTS` and is served through `tools/list`, and the new intro paragraph points there — so the tradeoff is defensible, just noting it is a net loss in the human-readable page beyond what #1902 asked for.
- "…`hal0.mcp.admin`, which is the single owner of that classification even for tools that only ever run through the standalone `/mcp/memory` mount" overstates it slightly: `_needs_approval` checks `memory.py`'s own `_LOCAL_GATED_TOOLS` **first** (`hal0/mcp/memory.py:1161`). The classification in the table is still right; only the "single owner" sentence is. (Source-side: that frozenset's comment claiming admin "doesn't yet classify" these is itself now stale — admin's `GATED_TOOLS` does classify both. Follow-up, not this PR.)
- Out of scope, but worth a thought for the register: the doc-side fix cannot stop this class recurring in third-party configs. Making the mount root 307 to `<mount>/mcp` would kill the 405 at the source. #1902 asked for the doc fix, so not a change request here.

Requesting changes on the ratchet strength plus the two contradicted claims. The doc content itself is accurate and the reported defect is genuinely fixed.

thinmintdev:
Revision pushed as `9ec12cfb` (rebased onto current main first). Per finding:

**Blocking — ratchet strength.** Replaced both whole-page one-directional checks with per-tier table parsing and set equality. `test_admin_tool_tables_match_frozensets_per_tier` slices the three `### Autonomous — read / — write / Gated` table bodies and asserts first-cell name sets equal `AUTONOMOUS_READ_TOOLS` / `AUTONOMOUS_WRITE_TOOLS` / `GATED_TOOLS` exactly (both directions). `test_memory_tool_tables_match_registrations_per_tier` does the same for the three `hal0-memory` sub-tables against the `@server.tool` registrations intersected with each frozenset, plus an unclassified-leftover assertion. Reproduced all three drift modes from the review before committing:
- gutted the entire memory read table → `FAILED ... missing ['memory_bank_stats', ...]`
- moved `service_action` Gated → Autonomous-read → `FAILED ... stale/misplaced rows ['service_action']`
- renamed `slot_metrics` in the doc only → `FAILED ... missing ['slot_metrics'], stale/misplaced rows ['slot_metrics_old']`

First-cell-only extraction is deliberate: delegate rows list several names in one cell (wanted), while Returns/route cells cross-reference other tools (would poison equality). This also resolves the Codex P2 on `test_mcp_docs_match_source.py:103`.

**Should fix 1 — wrap-key list.** Return-shape contract now names `runner_image_pulls` and `backend_list` alongside the original three, and points at `_LIST_TOOL_WRAP_KEY` as the authoritative list. Verified against source: the dict maps exactly those five names.

**Should fix 2 — `memory_curate`.** Required column now reads `id`, plus at least one field to change (`text`/`context`/`occurred_start`/`occurred_end`/`fact_type`/`entities`/`state`), and notes that `id` alone returns `mcp.memory_schema` — matching `_memory_curate`'s guard.

**Nit — "single owner" sentence.** Softened even though flagged no-action: it now says the standalone mount additionally hard-gates its two delete tools through its own local list, checked first (`_LOCAL_GATED_TOOLS`). The source-side stale comment in `memory.py` is left for the follow-up you noted, keeping this PR docs+tests only.

Verified: `tests/docs/` 10 passed, `make lint` clean, `ruff format --check src tests` clean.

thinmintdev:
## Review verdict: **approve**

Re-reviewed the revision (`9ec12cfb`) in a clean worktree from the head SHA — I re-ran every claim rather than reading the summary.

### Blocking finding from the prior round — resolved at the root

The whole-page one-directional `_backticked_names` checks are gone; both catalog tests now slice the per-tier table bodies and assert **set equality** in both directions (`_section` + `_table_row_names` + `_assert_tier_matches`, `tests/docs/test_mcp_docs_match_source.py:72-108`). I reproduced all three drift modes I raised, myself, against the new test:

1. **Gutted section** — deleted the entire `hal0-memory` **Autonomous — read** table body: `FAILED test_memory_tool_tables_match_registrations_per_tier ... missing ['memory_bank_stats', 'memory_directive_get', ... 'memory_tags_list']`. (Under the old test this passed 4/4 — that was the demonstration that motivated the change.)
2. **Tier drift** — moved the `service_action` row from Gated into Autonomous — read: `FAILED test_admin_tool_tables_match_frozensets_per_tier ... stale/misplaced rows ['service_action']`.
3. **Stale row** — renamed `slot_metrics` → `slot_metrics_old` in the doc only: `FAILED ... missing/stale`.

Also re-ran the full file against `main`'s docs tree (`git checkout add97eb3 -- docs/`): **4 failed** — both Connect assertions plus both catalog tests. Green on the branch: `tests/docs/` **10 passed**. First-cell-only extraction is the right call and the docstring explains why (delegate rows put several wanted names in one cell; Returns/route cells cross-reference other tools and would poison equality).

### Should-fix items — both verified against source

1. **Wrap-key list.** `_LIST_TOOL_WRAP_KEY` (`hal0/mcp/admin.py:217-225`) maps exactly `slot_list`, `provider_list`, `upstream_list`, `runner_image_pulls`, `backend_list` — the return-shape contract now names all five and points at the symbol as authoritative. Correct and complete.
2. **`memory_curate`.** `_memory_curate` raises `MemorySchemaError("memory_curate requires at least one field to change")` when `text/context/occurred_start/occurred_end/fact_type/entities/state` are all `None` (`hal0/mcp/memory.py:682-685`). The Required cell now states exactly that set and that `id` alone returns `mcp.memory_schema`. Matches source; the neighbouring `*_update` rows correctly do **not** carry the rule.
3. **"single owner" nit** (was no-action) — softened accurately: `_needs_approval` does check `_LOCAL_GATED_TOOLS` first (`memory.py:1161`), and the two tools in it are exactly the two deletes.

### No regression to the #1902 fix

The Connect section still reads `http://localhost:8080/mcp/admin/mcp` / `/mcp/memory/mcp` with the "mount root 405s, no redirect" prose, and both Connect tests still go red against `main`'s doc. The catalog side is now strictly stronger than the round-1 state: per-tier equality holds for `read 102`, `write 24`, `gated 54` and all 26 memory registrations, with the unclassified-leftover assertion catching any future `@server.tool` that no frozenset covers. Diff is still docs + one test file; no source changes, no `CHANGELOG.md` hunk.

CI: green on `9ec12cfb` — `python (3.12)`, `γ-suite (chromium)`, `ui`, `sunset`, CodeQL all `success` (verified via `check-runs`, not just the glyph). Locally: `make lint` clean, `ruff format --check src tests` clean (1146 files).

### Non-blocking — three round-2 Codex P2s, triaged

- **`test_mcp_docs_match_source.py:126` (both mount URLs).** Reproduces: I stripped every `/mcp/memory` line from the Connect section and the suite still passed 4/4, because `test_connect_section_urls_end_in_mcp` only requires ≥1 matching URL. Real gap, but it guards a *disappearance* of documented content rather than the wrong-URL defect #1902 filed, and it is cheap: assert the captured mount set equals `{"admin", "memory"}`. Worth a follow-up commit if you touch this file again; not worth another round-trip on a docs PR.
- **`test_mcp_docs_match_source.py:148` (validate route cells against `TOOL_NAME_ALIASES`).** True by construction — only first cells are parsed, so a mistyped route in the second column stays green. This is a new ratchet beyond what #1902 asked to promote, and it is genuinely additive value given the page calls the alias map ground truth. Follow-up, not a blocker.
- **`mcp-tools.mdx:242` (Hindsight-only availability).** I checked this one against source because it is a docs-accuracy claim on the rows this PR adds: `PgVectorProvider` (`hal0/memory/pgvector_provider.py:35`) implements only `add/search/list_items/delete/graph_status/set_*`; `curate`, `list_mental_models`, `list_tags`, `bank_stats`, directives and operations exist only on the Hindsight provider, so under an explicit pgvector config those handlers raise into `mcp.memory_failed` (`memory.py:1246`). Codex is right that the table gives an engine caveat only for `memory_reflect`. Same severity class as the two should-fixes above and a one-line note under the memory heading would close it — filing as a follow-up rather than blocking, since #1902's ask (enumerate every served tool, per tier) is fully met.

Prior blocking finding is genuinely closed, both should-fixes are correct against source, nothing regressed, CI green. Approving.

Issue #1888, opened 2026-08-15
runner ade07ba: the Vulkan backend emits invalid tokens for every model — documented-recommended lane produces garbage while every health surface reads green


thinmintdev:
## GA blocker — found by rc-validate v1.0.0-rc.6

**Severity:** GA-blocker · **Box:** ct151-cpu-fresh (10.0.1.151), fresh install, no `/dev/kfd`; A/B confirmed on ct150-update (10.0.1.150), privileged, has `/dev/kfd`.

### Summary

`ghcr.io/hal0ai/hal0-rocmfpx:ade07ba` (the release-pinned runner image) is a HIP+Vulkan single build. llama.cpp picks the ROCm backend when `/dev/kfd` is visible inside the container and silently falls back to Vulkan when `ggml_rocm_init` fails. `providers/_gpu.resolve_gpu_device_paths` only passes `/dev/kfd` through to the container if it exists on the host. On any box without `/dev/kfd` (container/LXC deployments without amdkfd passthrough — exactly what a fresh unprivileged install produces), **every slot runs the Vulkan lane, and the Vulkan lane emits invalid tokens for every model it serves**, at full nominal speed (60–120 tok/s), while HTTP 200, container health, `hal0 doctor` ("OK gpu"), and the SSE `done` frame all read green.

### Mechanism, proven decisively

An identical-argv A/B on the *privileged, prod-class* ct150 box, varying only `-dev`:

- `-dev ROCm0` → `"Paris, a city that has been the center of French culture"` (correct)
- `-dev Vulkan0` → 12 tokens of garbage, no coherent text

Not model-specific — an unrelated Qwen3.5-0.8B Q4_K_XL model is equally garbage on Vulkan. Not contention — measured 42 tok/s prompt / 123 tok/s eval in a standalone container. Not privilege-related — reproduces identically on both the unprivileged fresh box and the privileged update box when `/dev/kfd` is absent from the container's view.

### Three release-owned consequences (not just an environment quirk)

1. **The `device` field lies.** hal0 seeds and reports these slots as `device = "gpu-vulkan"` / `backend: vulkan`, but on every box where `/dev/kfd` exists, they actually execute on **ROCm** — meaning the Vulkan lane hal0 advertises has never actually been exercised by any validation run to date.
2. **The docs recommend the broken lane.** `docs/hal0-install-migration-guide.html`: *"RADV/Vulkan recommended on Strix Halo … v1.0 standardizes shared APU models on Vulkan."*
[...]

PR #1918, opened 2026-08-17
feat(dispatcher): bound relayed upstream streams with a stall guard

### LINKED ISSUE #1893: hermes/provider plugin: no stall guard for a never-terminating upstream stream — completely dead terminal with no diagnostic

### Summary

Neither hermes nor the hal0 provider plugin has a stall guard or partial-rendering fallback for a never-terminating upstream stream. Fed an unterminated repetition (`finish_reason` forever `null`, the shape produced by the Vulkan-backend garbage-token defect — see #1888), `hermes -z ... --cli` prints zero bytes, never self-terminates, and leaves no session file: a completely dead terminal with no diagnostic for as long as the operator waits.

This is residue split out from the rc.6 `hermes-cannot-chat` finding: the root cause (garbage tokens from the Vulkan backend) is tracked as the GA-blocker in #1888. This issue is the **independent gap** that remains even once the model is fixed — hal0 has no defense against a pathological/never-terminating stream from any upstream.

### Repro

```
# Stream a request to the anchor's serving slot directly and confirm the pathological shape
# (finish_reason null past max_tokens' worth of tokens), then run a hermes turn against it:
sudo -u hal0 HOME=/var/lib/hal0 timeout 300 \
  /usr/lib/hal0/venvs/hermes/bin/hermes -z "Reply with exactly: HERMES-RC-OK" --cli
# (via /usr/local/bin/hermes for a supported invocation; the raw path is fine here because
# the cwd is controlled.)
```

### Expected

A never-terminating stream produces SOMETHING operator-visible within a bounded time: partial output, a stall warning, or a client-side cutoff naming the cause — never an indefinitely silent turn.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `hermes-stream-no-stall-guard`.

thinmintdev:
## What

A never-terminating upstream stream had no bound anywhere in hal0. Adds a **stall guard** to the gateway's streaming relay, plus the consumer-side half in the hal0 Hermes provider plugin.

## Why — the root cause is one layer below the report

#1893 was filed against hermes / the provider plugin: `hermes -z ... --cli` printed zero bytes, never self-terminated, and left no session file. The provider plugin is not where the defect lives.

`Dispatcher._forward_streaming` relayed `resp.aiter_raw()` to exhaustion, and **httpx applies no read timeout once a stream is open** (the existing `direct_read_timeout_s` is explicitly documented as non-streaming-only). So an upstream that never terminates — `finish_reason` forever `null`, the shape the Vulkan garbage-token defect (#1888) produces — was piped forever to *every* hal0 client: Hermes `--cli`, the dashboard, any OpenAI-compatible caller. None of them could produce a diagnostic because none of them ever saw an end of stream. Fixing this only in the hermes plugin would have left the gap open for every other client, and would not even have fixed hermes (Hermes' own OpenAI client consumes the stream, not hal0's helpers).

## What changed

**`src/hal0/dispatcher/router.py`** — the relay is now bounded by two independent budgets:

| Key | Default | Meaning |
|---|---|---|
| `[dispatcher].stream_total_timeout_s` | `900.0` | Max wall clock for one relayed stream |
| `[dispatcher].stream_idle_timeout_s` | `300.0` | Max gap between two upstream chunks |

Both are deliberately generous — prompt processing on a cold CPU-only slot is genuinely silent for minutes. The guard converts "hangs forever" into "ends with a named reason"; it does not police slow-but-working inference. `0` disables a bound; a negative value is coerced to `0` so a bad config can never cut every stream off instantly.

On a trip: the upstream stream is closed, `dispatch.stream_stall_guard_tripped` is logged (reason, elapsed, upstream, both budgets), and — **on `text/event-stream` only** — one terminal `chat.completion.chunk` is appended before `data: [DONE]`:

- `delta.content` carries a human-readable notice, so the operator sees the cutoff instead of a silent terminal;
- `finish_reason` is the standard `"length"`, so strict clients terminate cleanly;
- `x_hal0_stall` carries the precise reason, elapsed time, and budgets.

Non-SSE streaming bodies (audio, image bytes) are cut off with no injected frame — rewriting them would corrupt them. Chunks that do arrive still pass through byte-for-byte, so partial output is preserved.

**`provider_hal0/profile.py`** (+ byte-identical installer seed) — the consumer-side half: `iter_sse_data` / `parse_sse_chunks` take a `max_events` ceiling (default 200k) so a direct slot connection can't spin them forever, and `assemble_stream` now reports `stalled` — True when the stream ended with no `finish_reason` at all, or when the gateway stamped `x_hal0_stall` — plus the gateway's structured `stall` dict when present. A cut-off turn no longer looks identical to a clean one.

**Config plumbing** — `DispatcherConfig` schema fields, `_settings_apply` registry entries (`service-restart[hal0-api]`, matching `direct_read_timeout_s`), and the `_boot_dispatcher` wiring.

**Docs** — `docs/reference/config-schema.mdx` and a new "Stall guard" section in `docs/reference/api/streaming.mdx` (which previously documented "there is no separate configurable read timeout once bytes start flowing" as a caution — now it has one).

## How it was verified

New `tests/dispatcher/test_stream_stall_guard.py` drives a real never-terminating `httpx.AsyncByteStream` through `Dispatcher.forward`:

- chatty endless stream → cut at the total bound, terminal frame names `reason: "total"`, partial tokens still delivered;
- opened-then-silent stream → cut at the idle bound, `reason: "idle"`;
- non-SSE endless stream → bounded, no `[DONE]` / no `x_hal0_stall` injected;
- well-behaved stream → relayed byte-for-byte, untouched;
- `0` → that bound is disabled (all 20 chunks relayed, only the idle bound ends it).

Every one of these hangs or fails on `main` (the first fails at construction — the knobs don't exist; the relay itself has no exit).

Plus 4 provider-plugin tests for the `max_events` ceiling and the `stalled` / `stall` surfacing.

```
uv run --extra dev pytest tests/ -q -p no:randomly
  -> 10314 passed, 16 skipped, 1 xfailed in 1057s
uv run --extra dev ruff check src tests           -> All checks passed!
uv run --extra dev ruff format --check src tests  -> 1145 files already formatted
```

No test needs a live hermes or a live gateway (CI has neither).

## Deliberately out of scope

- **#1888** — the garbage-token defect that produces this stream shape. This is the independent gap that remains once the model is fixed, exactly as the issue frames it.
- **No `StreamStalled` type in `hermes/core/errors.py`.** The blast-radius plan suggested one, but `core/client.py` is a non-streaming transport and the provider plugin runs inside the Hermes venv where hal0 core is not importable — the error would have had no raiser and no consumer. The stall is surfaced through the SSE contract instead, which is the seam that actually crosses the process boundary.
- **`hermes/driver.py` partial-render surfacing.** The driver installs and health-checks Hermes; it is not in the token path. Hermes' own CLI renders the stream, and it renders our terminal chunk like any other assistant content.
- **CHANGELOG.md** — per the fix-wave convention (#1545), one consolidated entry lands at the end of the wave.

Closes #1893

### DIFF
--- a/docs/reference/api/streaming.mdx
+++ b/docs/reference/api/streaming.mdx
@@ -16,11 +16,12 @@ hal0's streaming forward path in the dispatcher (`hal0.dispatcher.router`).
   specifically so a connect failure surfaces as a clean `502 UpstreamUnavailable`
   rather than failing partway through an already-started response generator.
 - The client response is a Starlette `StreamingResponse` whose body iterator does a
-  **raw byte passthrough** — `async for chunk in resp.aiter_raw(): yield chunk`. hal0
+  **raw byte passthrough** — every upstream chunk is yielded unmodified. hal0
   does **not** parse or re-emit SSE events itself; the OpenAI-standard
   `data: {...}\n\n` ... `data: [DONE]\n\n` framing and the `text/event-stream`
   content-type come from whatever upstream is serving the slot (llama-server, FLM, or
-  a remote provider), copied through verbatim.
+  a remote provider), copied through verbatim. The one exception is the stall guard
+  below, which may **append** a terminal chunk after cutting a stream off.
 - Status code and `content-type` on the client response are copied verbatim from the
   upstream response.
 
@@ -34,6 +35,45 @@ apply — only the underlying HTTP client's `connect` (5.0s), `write` (10.0s), a
 `pool` (5.0s) timeouts constrain the streaming path's connection setup. There is no
 separate configurable read timeout once bytes start flowing.
 </Aside>
+
+## Stall guard
+
+Because no read timeout applies to an open stream, a pathological upstream that
+never terminates (`finish_reason` forever `null`) would be relayed forever, leaving
+the client — Hermes `--cli`, the dashboard, any OpenAI-compatible caller — waiting
+on a socket that never closes, with no output and no diagnostic. Two bounds prevent
+that:
+
+| Key | Default | Meaning |
+|---|---|---|
+| `[dispatcher].stream_total_timeout_s` | `900.0` | Max wall clock for one relayed stream |
+| `[dispatcher].stream_idle_timeout_s` | `300.0` | Max gap between two upstream chunks |
+
+Either bound tripping closes the upstream stream and logs
+`dispatch.stream_stall_guard_tripped` (with `reason`, `elapsed_s`, and the upstream
+name). On a `text/event-stream` response hal0 then appends a final
+`chat.completion.chunk` before `data: [DONE]`:
+
+```json
+{
+  "id": "hal0-stall-guard",
+  "object": "chat.completion.chunk",
+  "choices": [
+    {
+      "index": 0,
+      "delta": { "role": "assistant", "content": "\n\n[hal0] stall guard: ..." },
+      "finish_reason": "length"
+    }
+  ],
+  "x_hal0_stall": { "reason": "total", "elapsed_s": 900.0, "upstream": "llm" }
+}
+```
+
+The notice is rendered as ordinary assistant content so the operator sees the cutoff,
+`finish_reason` is the standard `"length"` so strict clients terminate cleanly, and
+the precise cause lives in the `x_hal0_stall` extension. Non-SSE streaming bodies
+(audio, images) are cut off without any injected frame. Set either key to `0` to
+disable that bound.
 
 ## Slot state during a stream
 
--- a/docs/reference/api/streaming.mdx
+++ b/docs/reference/api/streaming.mdx
@@ -32,9 +32,12 @@ hal0's streaming forward path in the dispatcher (`hal0.dispatcher.router`).
 [Config schema](/docs/reference/config-schema/#dispatcher)) sets the HTTP client's
 `read` timeout, which httpx applies to **every** body read — streaming included.
 While the stall guard is active, streaming requests therefore disable it
-per-request so the guard owns the read bound and a stall always produces the
+per-request so the guard owns the *body* bound and a stall always produces the
 diagnostic frame below (a bare `httpx.ReadTimeout` would tear the stream with no
-terminal frame). Only with **both** stall-guard bounds set to `0` is the client's
+terminal frame). `direct_read_timeout_s` still bounds the wait for response
+headers on those requests — the guard's own clocks only start once headers
+arrive, so a header wait needs its own bound. Only with **both** stall-guard
+bounds set to `0` is the client's
 read timeout kept on the streaming path, as a per-read transport backstop; if it
 trips, the guard still converts it into a diagnostic frame with reason `read`.
 The `connect` (5.0s), `write` (10.0s), and `pool` (5.0s) timeouts constrain the
--- a/docs/reference/api/streaming.mdx
+++ b/docs/reference/api/streaming.mdx
@@ -29,20 +29,27 @@ hal0's streaming forward path in the dispatcher (`hal0.dispatcher.router`).
 
 <Aside type="caution">
 `[dispatcher].direct_read_timeout_s` (default `300.0`, range `30.0-600.0` — see
-[Config schema](/docs/reference/config-schema/#dispatcher)) governs the
-**non-streaming** forward path only. Once a stream is open, this timeout does not
-apply — only the underlying HTTP client's `connect` (5.0s), `write` (10.0s), and
-`pool` (5.0s) timeouts constrain the streaming path's connection setup. There is no
-separate configurable read timeout once bytes start flowing.
+[Config schema](/docs/reference/config-schema/#dispatcher)) sets the HTTP client's
+`read` timeout, which httpx applies to **every** body read — streaming included.
+While the stall guard is active, streaming requests therefore disable it
+per-request so the guard owns the read bound and a stall always produces the
+diagnostic frame below (a bare `httpx.ReadTimeout` would tear the stream with no
+terminal frame). Only with **both** stall-guard bounds set to `0` is the client's
+read timeout kept on the streaming path, as a per-read transport backstop; if it
+trips, the guard still converts it into a diagnostic frame with reason `read`.
+The `connect` (5.0s), `write` (10.0s), and `pool` (5.0s) timeouts constrain the
+streaming path's connection setup as usual.
 </Aside>
 
 ## Stall guard
 
-Because no read timeout applies to an open stream, a pathological upstream that
-never terminates (`finish_reason` forever `null`) would be relayed forever, leaving
-the client — Hermes `--cli`, the dashboard, any OpenAI-compatible caller — waiting
-on a socket that never closes, with no output and no diagnostic. Two bounds prevent
-that:
+httpx's per-read timeout is no defense against a *chatty* pathological upstream
+that never terminates (`finish_reason` forever `null`): every chunk resets the
+read timer, so the stream would be relayed forever, leaving the client — Hermes
+`--cli`, the dashboard, any OpenAI-compatible caller — waiting on a socket that
+never closes, with no output and no diagnostic. And against a *silent* upstream a
+bare transport timeout tears the stream without any terminal frame. Two
+guard-owned bounds prevent both:
 
 | Key | Default | Meaning |
 |---|---|---|
@@ -51,8 +58,14 @@ that:
 
 Either bound tripping closes the upstream stream and logs
 `dispatch.stream_stall_guard_tripped` (with `reason`, `elapsed_s`, and the upstream
-name). On a `text/event-stream` response hal0 then appends a final
-`chat.completion.chunk` before `data: [DONE]`:
+name). `reason` is `total`, `idle`, or `read` (transport read timeout — only
+possible with the guard fully disabled). On a `text/event-stream` response hal0
+then appends a final chunk before `data: [DONE]`, preceded by a blank line so it
+parses as its own SSE event even when the upstream stalled midway through a
+`data:` line. For `/chat/completions` (and any other chat-shaped SSE stream) it is
+a `chat.completion.chunk`; for the legacy `/v1/completions` it is a
+`text_completion` chunk carrying the notice in `choices[].text` instead of
+`choices[].delta`:
 
 ```json
 {
@@ -75,6 +88,14 @@ the precise cause lives in the `x_hal0_stall` extension. Non-SSE streaming bodie
 (audio, images) are cut off without any injected frame. Set either key to `0` to
 disable that bound.
 
+<Aside type="note">
+`stream_total_timeout_s` is wall clock, not a health signal: a genuinely healthy
+long generation — e.g. a big turn at a few tok/s on a CPU-only box — that outlives
+the budget is also cut, with the notice injected as assistant content. On slow
+hardware that routinely runs long turns, raise the total bound (up to `86400`) or
+set it to `0` and rely on the idle bound, which only fires on true silence.
+</Aside>
+
 ## Slot state during a stream
 
 A slot dispatching a streaming request is held in the `SERVING` state until the client
@@ -92,6 +113,11 @@ sent to the client:
 - **Time-to-first-token (TTFT)**: recorded at the first chunk containing a `"delta":`
   marker, deliberately skipping llama-server's initial role-only chunk — stored in a
   per-slot TTFT event deque.
+
+The synthetic stall-guard frame is recognized by its `x_hal0_stall` marker and
+excluded from both: it never counts as a generated token or a TTFT sample, and the
+request-metrics row for a guard-tripped stream is recorded as a failure
+(`ok = 0`, `error_code = stream_stall_<reason>`) rather than a clean success.
 
 Non-streaming responses get an analogous hook that pulls `usage.completion_tokens`
 (and, for FLM/NPU responses, the hal0-specific `usage.decoding_speed_tps` /
--- a/docs/reference/config-schema.mdx
+++ b/docs/reference/config-schema.mdx
@@ -76,11 +76,16 @@ Root model `Hal0Config`, `extra="allow"`.
 | `prefetch_parallel_cap` | int | `4` | `>= 1` |
 
 `stream_total_timeout_s` / `stream_idle_timeout_s` are the streaming **stall
-guard**: no read timeout applies once a stream is open, so a never-terminating
-upstream would otherwise be relayed forever. When either bound trips, hal0
-closes the upstream stream and — on `text/event-stream` responses — appends one
-terminal chunk carrying a human-readable notice, `finish_reason: "length"`, and
-an `x_hal0_stall` object naming the reason, followed by `data: [DONE]`.
+guard**: a chatty never-terminating upstream resets the transport read timer
+with every chunk, so it would otherwise be relayed forever. While the guard is
+active, streaming requests disable the HTTP client's read timeout per-request so
+the guard owns the bound (`direct_read_timeout_s` then applies to non-streaming
+requests only); with both bounds set to `0` the client read timeout stays on as
+a transport backstop. When a bound trips, hal0 closes the upstream stream and —
+on `text/event-stream` responses — appends one terminal chunk carrying a
+human-readable notice, `finish_reason: "length"`, and an `x_hal0_stall` object
+naming the reason, followed by `data: [DONE]`. Note the total bound is wall
+clock: a healthy long generation on slow hardware that outlives it is also cut.
 
 ### `[telemetry]`
 
--- a/docs/reference/config-schema.mdx
+++ b/docs/reference/config-schema.mdx
@@ -71,7 +71,16 @@ Root model `Hal0Config`, `extra="allow"`.
 |---|---|---|---|
 | `prefetch_timeout_s` | float | `8.0` | `> 0.0` |
 | `direct_read_timeout_s` | float | `300.0` | `30.0-600.0` |
+| `stream_total_timeout_s` | float | `900.0` | `0.0-86400.0` (`0` disables) |
+| `stream_idle_timeout_s` | float | `300.0` | `0.0-86400.0` (`0` disables) |
 | `prefetch_parallel_cap` | int | `4` | `>= 1` |
+
+`stream_total_timeout_s` / `stream_idle_timeout_s` are the streaming **stall
+guard**: no read timeout applies once a stream is open, so a never-terminating
+upstream would otherwise be relayed forever. When either bound trips, hal0
+closes the upstream stream and — on `text/event-stream` responses — appends one
+terminal chunk carrying a human-readable notice, `finish_reason: "length"`, and
+an `x_hal0_stall` object naming the reason, followed by `data: [DONE]`.
 
 ### `[telemetry]`
 
--- a/docs/reference/config-schema.mdx
+++ b/docs/reference/config-schema.mdx
@@ -79,8 +79,9 @@ Root model `Hal0Config`, `extra="allow"`.
 guard**: a chatty never-terminating upstream resets the transport read timer
 with every chunk, so it would otherwise be relayed forever. While the guard is
 active, streaming requests disable the HTTP client's read timeout per-request so
-the guard owns the bound (`direct_read_timeout_s` then applies to non-streaming
-requests only); with both bounds set to `0` the client read timeout stays on as
+the guard owns the body bound; `direct_read_timeout_s` still bounds the wait for
+response headers on those requests, since the guard's own clocks only start once
+headers arrive. With both bounds set to `0` the client read timeout stays on as
 a transport backstop. When a bound trips, hal0 closes the upstream stream and —
 on `text/event-stream` responses — appends one terminal chunk carrying a
 human-readable notice, `finish_reason: "length"`, and an `x_hal0_stall` object
--- a/installer/agents/hermes/plugins/hal0-provider/profile.py
+++ b/installer/agents/hermes/plugins/hal0-provider/profile.py
@@ -167,15 +167,37 @@ def is_alias(model_id: str) -> bool:
 
 _DONE_SENTINEL = "[DONE]"
 
+# hal0's stall-guard extension field (see hal0.dispatcher.router). The gateway
+# stamps it on the terminal chunk when it cuts off a never-terminating upstream
+# so the consumer can name the cutoff instead of reporting a clean stop.
+_STALL_FIELD = "x_hal0_stall"
 
-def iter_sse_data(lines: Iterable[bytes | str]) -> Iterator[str]:
+# Consumer-side ceiling on SSE events per turn (#1893). A pathological upstream
+# that the gateway's guard somehow does not bound (a direct slot connection, a
+# gateway on an older build) still cannot spin this consumer forever.
+DEFAULT_MAX_SSE_EVENTS = 200_000
+
+
+def iter_sse_data(
+    lines: Iterable[bytes | str],
+    *,
+    max_events: int | None = DEFAULT_MAX_SSE_EVENTS,
+) -> Iterator[str]:
     """Yield the ``data:`` payloads of an SSE stream.
 
     Tolerates gaps: blank keep-alive lines and ``:``-comment heartbeats are
     skipped, non-``data:`` fields are ignored, and the terminal ``[DONE]``
     sentinel stops iteration.
+
+    ``max_events`` bounds how many payloads are consumed before iteration
+    stops regardless of what the upstream is doing — a never-terminating
+    stream must never become an unbounded loop here (#1893). Pass ``None``
+    to disable the ceiling.
     """
+    emitted = 0
     for raw in lines:
+        if max_events is not None and emitted >= max_events:
+            return
         line = raw.decode("utf-8", "replace") if isinstance(raw, bytes) else raw
         line = line.rstrip("\r\n")
         if not line or line.startswith(":"):
@@ -185,12 +207,17 @@ def iter_sse_data(lines: Iterable[bytes | str]) -> Iterator[str]:
         payload = line[len("data:") :].strip()
         if payload == _DONE_SENTINEL:
             return
+        emitted += 1
         yield payload
 
 
-def parse_sse_chunks(lines: Iterable[bytes | str]) -> Iterator[dict[str, Any]]:
+def parse_sse_chunks(
+    lines: Iterable[bytes | str],
+    *,
+    max_events: int | None = DEFAULT_MAX_SSE_EVENTS,
+) -> Iterator[dict[str, Any]]:
     """Parse SSE ``data:`` payloads into chunk dicts, dropping malformed ones."""
-    for payload in iter_sse_data(lines):
+    for payload in iter_sse_data(lines, max_events=max_events):
         try:
             obj = json.loads(payload)
         except (ValueError, TypeError):
@@ -206,14 +233,26 @@ def assemble_stream(chunks: Iterable[dict[str, Any]]) -> dict[str, Any]:
     across chunks, preserves ``reasoning_content`` verbatim, and reassembles
     indexed ``tool_calls`` with gap tolerance (missing / non-contiguous /
     repeated indices are folded by index, then emitted in index order).
+
+    Also reports whether the turn actually *finished* (#1893). ``stalled`` is
+    True when the stream ended without any ``finish_reason`` — the shape a
+    never-terminating upstream produces once something bounds it — or when
+    hal0's gateway stamped its ``x_hal0_stall`` stall-guard extension on the
+    terminal chunk. ``stall`` carries the gateway's structured reason when
+    there is one. Callers must render a cutoff notice rather than presenting
+    a stalled turn as a clean, complete answer.
     """
     content_parts: list[str] = []
     reasoning_parts: list[str] = []
     tool_calls: dict[int, dict[str, Any]] = {}
     role: str | None = None
     finish_reason: str | None = None
+    stall: dict[str, Any] | None = None
 
     for chunk in chunks:
+        raw_stall = chunk.get(_STALL_FIELD)
+        if isinstance(raw_stall, dict):
+            stall = dict(raw_stall)
         for choice in chunk.get("choices") or []:
             if not isinstance(choice, dict):
                 continue
@@ -249,7 +288,14 @@ def assemble_stream(chunks: Iterable[dict[str, Any]]) -> dict[str, Any]:
         message["reasoning_content"] = "".join(reasoning_parts)
     if tool_calls:
         message["tool_calls"] = [tool_calls[i] for i in sorted(tool_calls)]
-    return {"message": message, "finish_reason": finish_reason}
+    result: dict[str, Any] = {
+        "message": message,
+        "finish_reason": finish_reason,
+        "stalled": stall is not None or finish_reason is None,
+    }
+    if stall is not None:
+        result["stall"] = stall
+    return result
 
 
 # ── The profile ────────────────────────────────────────────────────────────
--- a/src/hal0/agents/hermes/plugins/provider_hal0/profile.py
+++ b/src/hal0/agents/hermes/plugins/provider_hal0/profile.py
@@ -167,15 +167,37 @@ def is_alias(model_id: str) -> bool:
 
 _DONE_SENTINEL = "[DONE]"
 
+# hal0's stall-guard extension field (see hal0.dispatcher.router). The gateway
+# stamps it on the terminal chunk when it cuts off a never-terminating upstream
+# so the consumer can name the cutoff instead of reporting a clean stop.
+_STALL_FIELD = "x_hal0_stall"
 
-def iter_sse_data(lines: Iterable[bytes | str]) -> Iterator[str]:
+# Consumer-side ceiling on SSE events per turn (#1893). A pathological upstream
+# that the gateway's guard somehow does not bound (a direct slot connection, a
+# gateway on an older build) still cannot spin this consumer forever.
+DEFAULT_MAX_SSE_EVENTS = 200_000
+
+
+def iter_sse_data(
+    lines: Iterable[bytes | str],
+    *,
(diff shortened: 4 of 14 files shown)


thinmintdev:
## Review — changes requested

The diagnosis is right and the layer choice is right. #1893 was filed against hermes, but the plugin points at `http://127.0.0.1:8080/v1` (`profile.py:112`), so the gateway relay is genuinely the layer where the defect lives, and bounding it there fixes the reported repro path *and* every other client. Scope is tight, the installer seed is byte-identical (verified `sha256`), and there is no CHANGELOG hunk. New dispatcher tests do fail on `main` (the constructor kwargs don't exist).

Two blocking defects, both of which destroy the diagnostic in exactly the situation the guard exists for. Both were raised by Codex; I reproduced both rather than taking the bot's word for it.

---

### Blocking 1 — `httpx.ReadTimeout` escapes the guard; the premise of the PR is false

The comment at `router.py:135`/`:138` and the docs both assert that "httpx applies NO read timeout once a stream is open". That is not true. httpx applies the client's `read` timeout to **every** body read, including during `aiter_raw()`, and it raises `httpx.ReadTimeout` — which is **not** a subclass of the builtin `TimeoutError` the guard catches (`httpx.ReadTimeout → TimeoutException → TransportError → RequestError → HTTPError → Exception`).

Live repro against a real socket (not `MockTransport`), server sends one chunk then goes silent:

```python
d = Dispatcher(direct_read_timeout_s=0.3, stream_total_timeout_s=60.0,
               stream_idle_timeout_s=0.0)   # docs: "0 disables that bound"
resp = await d.forward(call)
async for c in resp.body_iterator: ...
```
```
BODY: b'data: {"id":"1","choices":[{"delta":{"content":"hi"},"finish_reason":null}]}\n\n'
ESCAPED EXCEPTION: httpx.ReadTimeout
AssertionError: read timeout escaped the stall guard
```

Consequences:

* `stream_idle_timeout_s = 0` does **not** disable the idle cutoff, contrary to the schema description, `config-schema.mdx`, and `streaming.mdx`. It silently hands the cutoff to httpx, which tears the stream with no terminal frame and no `[DONE]` — the exact "dead socket, no diagnostic" outcome #1893 is about.
* Same for any `stream_idle_timeout_s > direct_read_timeout_s`.
* **With the shipped defaults it is a coin flip.** `stream_idle_timeout_s` defaults to `300.0` and `direct_read_timeout_s` defaults to `300.0`; the `asyncio.wait_for` budget is `min(300, 900-elapsed) = 300`, and httpx's read timer starts at the same instant. Which one wins is nondeterministic, so the guarded path for the silent-upstream case is not the one the tests pin. The idle test can't catch this because `MockTransport` ignores timeouts entirely.

Fix: catch `httpx.ReadTimeout` alongside `TimeoutError` and classify it as an idle stall, and/or send streaming requests with `read=None` so the guard owns the bound outright. Whichever you pick, `router.py:135`, `router.py:138` and the `<Aside type="caution">` in `docs/reference/api/streaming.mdx` (still says "There is no separate configurable read timeout once bytes start flowing") need correcting — the PR adds a Stall guard section right underneath a caution that is factually wrong.

### Blocking 2 — the terminal frame is not started on an SSE event boundary

`aiter_raw()` yields transport chunks, not complete SSE events. An upstream can stall after hal0 has relayed only part of a `data:` line, and `_stall_sse_frames()` starts immediately with `data: ` and no leading separator, so it is concatenated onto the partial line.

Failing test against this branch — upstream emits one clean event, then half of the next, then goes silent (`idle=0.2`):

```
UNPARSEABLE EVENTS: ['data: {"id":"1","choices":[{"delta":{"content":"wordata: {"id":"hal0-stall-guard","object":"chat.completion.chunk",...}']
AssertionError: stall frame concatenated onto a partial upstream event
```

The whole diagnostic — the operator-visible notice, `finish_reason`, and `x_hal0_stall` — is swallowed into one malformed event. A strict client raises a JSON parse error instead of showing the cause; a permissive one drops it. `data: [DONE]` does survive, so it is not a hang, but "operator sees a JSON error" is not what the issue asks for ("a stall warning ... naming the cause").

One-line fix: prefix the injected bytes with `\n\n`. A stray blank line before a well-formed event is a no-op for every SSE parser, so it is safe unconditionally. Please add a test that splits an event across chunk boundaries — the current suite only ever stalls on clean boundaries, which is why this passed.

---

### Non-blocking

**3 — the provider-plugin half has no production callers.** `iter_sse_data`, `parse_sse_chunks` and `assemble_stream` are referenced nowhere outside `tests/`:

```
grep -rn "assemble_stream\|iter_sse_data\|parse_sse_chunks" --include=*.py .
  → the two profile.py definitions + tests only
```

The PR body already concedes Hermes' own OpenAI client consumes the stream, so `max_events` and `stalled` do not change what `hermes -z ... --cli` prints — only the gateway change does. Keeping them as a defensive seam is fine; describing them as "the consumer-side half" of the fix overstates it, and the issue's operator-visible outcome rests entirely on Blocking 1 and 2 being right.

**4 — legacy `/v1/completions` gets a chat-shaped chunk.** `_dispatch_and_forward` also serves `/v1/completions` (see the comment at `api/routes/v1.py:1109`), where a stream chunk needs `choices[].text`, not `choices[].delta`. Either select the payload shape from `call.target_url` or limit the injection to `/chat/completions`. Low blast radius, cheap to close.

**5 — the synthetic chunk pollutes success metrics.** It contains `"delta":`, so `_instrument_streaming_throughput` counts it as a generated token and — for an upstream that never sent anything — records a **bogus ~300s TTFT sample** into the slot's TTFT deque that feeds the dashboard. `RequestSeam.wrap_streaming` sees no exception, so the row lands with `ok=True` and the stall chunk as `last_payload`, i.e. `finish_reason: "length"`. Every guard-triggered request is persisted as a clean success. Given the PR's own thesis is "a cut-off turn no longer looks identical to a clean one", the gateway's own telemetry should honour it — have both wrappers recognise `x_hal0_stall`.

**6 — a wall-clock total bound also polices slow-but-working inference.** The claim that the guard "does not police slow-but-working inference" holds for the idle bound but not for `stream_total_timeout_s`. On a CPU-only box — `ct151-cpu-fresh`, literally the box in this issue — a long generation at a few tok/s can pass 900s while working correctly, and gets truncated with `finish_reason: "length"` plus a stall notice injected as assistant *content*, which then enters conversation history for any client that persists it. The distinguishing property of the reported defect is unbounded output, not elapsed time; an output-event/byte ceiling would target it exactly and wouldn't penalise slow hardware. Worth considering before GA, or at least worth a doc note that the total bound can cut a healthy long turn.

**7 —** `iter_sse_data`'s 200k ceiling truncates with no marker of its own; the caller only learns via `stalled` (from the absent `finish_reason`). Acceptable, just noting it.

---

### Codex triage

| Codex finding | Verdict |
|---|---|
| Handle HTTPX read timeouts inside the stall guard | **Confirmed** — reproduced against a real socket. Blocking 1. |
| Start terminal frames on a fresh SSE event boundary | **Confirmed** — reproduced with a split-event upstream. Blocking 2. |
| Emit endpoint-compatible terminal SSE chunks | **Confirmed** but low impact — non-blocking 4. |
| Exclude synthetic stall chunks from success metrics | **Confirmed** by reading `seam.py` / `v1.py`; the false TTFT is the sharper half — non-blocking 5. |

Nothing to dismiss this round.

thinmintdev:
## Second-pass review — changes requested (independent verification)

I re-reviewed this from scratch rather than re-reading the earlier comment, and reproduced every
claim against this branch (`d35f2ab0`) instead of trusting the PR body or the bot.

**What holds up.** The layer choice is right: the provider plugin's `DEFAULT_BASE_URL` is
`http://127.0.0.1:8080/v1` (`provider_hal0/profile.py:112`), so `hermes -z ... --cli` genuinely
goes through `Dispatcher._forward_streaming` — bounding the relay there fixes the issue's exact
repro path and every other client with it. Installer seed is byte-identical
(`sha256 8f0fc6ee…` on both copies). No CHANGELOG hunk. No drive-by refactors. The new tests do
fail on `main` (`Dispatcher.__init__` on `main` has no `stream_*` kwargs; `iter_sse_data` has no
`max_events`; `assemble_stream` returns no `stalled`). On this branch:
`37 passed in 1.25s` for the two touched test files.

Two blocking defects survive verification. Both destroy the diagnostic in precisely the situation
the guard exists for, so the issue's "Expected" is not met in those paths.

---

### Blocking 1 — `httpx.ReadTimeout` escapes the guard; the PR's stated premise is false

`router.py:138` and `docs/reference/api/streaming.mdx` both assert that httpx applies **no** read
timeout once a stream is open. That is not true. The dispatcher-owned client sets
`read=self._direct_read_timeout_s` (`router.py:472`), and httpx applies that per body read,
including inside `aiter_raw()`. It raises `httpx.ReadTimeout`, which is **not** a subclass of the
builtin `TimeoutError` the guard catches:

```
httpx.ReadTimeout → TimeoutException → TransportError → RequestError → HTTPError → Exception
issubclass(httpx.ReadTimeout, TimeoutError) -> False      # httpx 0.28.1
```

Live repro against a real socket (not `MockTransport`): server sends one SSE event, then goes
silent forever; `direct_read_timeout_s=0.4`, `stream_total_timeout_s=60.0`,
`stream_idle_timeout_s=0.0` — the value the schema, `config-schema.mdx` and `streaming.mdx` all
document as "disables the bound":

```
ESCAPED EXCEPTION: httpx.ReadTimeout
BODY: b'data: {"id":"1","choices":[{"delta":{"content":"hi"},"finish_reason":null}]}\n\n'
has DONE: False | has x_hal0_stall: False
```

Same result with `direct_read_timeout_s=0.4, stream_idle_timeout_s=2.0` (idle bound merely set
*above* the read timeout) — `ESCAPED ReadTimeout`, no stall frame. The exception propagates out
of `_guarded_stream` after headers are already on the wire, so the post-loop log, the terminal
chunk and `data: [DONE]` are all skipped: a torn stream with no diagnostic, which is the outcome
#1893 is filed about.

**Correction to my earlier comment, for the record:** I claimed the shipped defaults are a coin
flip. They are not. `asyncio.wait_for` registers its deadline before the inner read begins, so at
equal values the guard wins deterministically — 6/6 runs at `read=0.5 / idle=0.5` tripped the
guard, and once elapsed time eats into `remaining_total` the guard's budget only gets smaller.
So the **default** config (`idle=300`, `direct_read=300`) is fine. The broken configurations are
the two the docs actively invite:

* `stream_idle_timeout_s = 0` — documented as an escape hatch, actually hands the cutoff to httpx
  with no diagnostic;
* any `stream_idle_timeout_s > direct_read_timeout_s` — and the schema allows `0.0-86400.0` for
  the idle bound while `direct_read_timeout_s` is capped at `600.0`, so most of the permitted
  range is dead config.

Fix: catch `httpx.ReadTimeout` alongside `TimeoutError` and classify it as an idle stall, and/or
send streaming requests with `read=None` so the guard owns the bound outright. Either way,
`router.py:135`, `router.py:138`, the `<Aside type="caution">` in `streaming.mdx` (still says
"There is no separate configurable read timeout once bytes start flowing", directly above the new
Stall guard section) and the `0 disables` wording in `schema.py` / `config-schema.mdx` need
correcting.

**Test gap that let this through:** every dispatcher test uses `httpx.MockTransport`, which
ignores client timeouts entirely, so the suite structurally cannot observe the httpx interaction.
`test_guard_can_be_disabled_with_zero` only exercises `total=0`; `idle=0` — the broken one — is
untested. A real-socket test is ~20 lines (`asyncio.start_server`) and is what caught this.

### Blocking 2 — the terminal frame is not started on an SSE event boundary

`aiter_raw()` yields transport chunks, not complete SSE events, so an upstream can stall after
hal0 has relayed only part of a `data:` line. `_stall_sse_frames()` begins immediately with
`data: ` and no leading separator, so it is concatenated onto the partial line. Reproduced on
this branch (upstream emits one clean event, then half of the next, then goes silent):

```
RAW: data: {"id":"1","choices":[{"delta":{"content":"word"}}]}

data: {"id":"2","chodata: {"id":"hal0-stall-guard","object":"chat.completion.chunk",...}

UNPARSEABLE: ['{"id":"2","chodata: {"id":"hal0-stall-guard","object":"chat.completion.chunk",...']
```

The entire diagnostic — the operator-visible notice, `finish_reason`, and `x_hal0_stall` — is
swallowed into one malformed event. A strict client raises a JSON parse error; a permissive one
drops it silently. `data: [DONE]` does survive, so it is not a hang, but "operator gets a JSON
error" is not "a stall warning naming the cause".

One-line fix: prefix the injected bytes with `\n\n`. A stray blank line before a well-formed event
is a no-op for every SSE parser, so it is safe unconditionally. Please add a test that splits an
event across a chunk boundary — the current suite only ever stalls on clean boundaries.

---

### Non-blocking

**3 — the provider-plugin half has no production caller.** `iter_sse_data`, `parse_sse_chunks`
and `assemble_stream` are referenced nowhere outside `tests/`. The PR body already concedes Hermes'
own OpenAI client consumes the stream, so `max_events` and `stalled` do not change what
`hermes -z ... --cli` prints — only the gateway change does. Fine as a defensive seam; calling it
"the consumer-side half" of the fix overstates it, and the issue's operator-visible outcome rests
entirely on Blocking 1 and 2.

**4 — legacy `/v1/completions` gets a chat-shaped chunk.** `completions()` at `v1.py:1358` routes
through the same `_dispatch_and_forward`, but a legacy completions stream chunk needs
`choices[].text`, not `choices[].delta`. Select the payload shape from `call.target_url`, or limit
the injection to `/chat/completions`. Low blast radius, cheap to close. (Codex finding —
confirmed.)

**5 — the synthetic chunk is recorded as clean success + a bogus TTFT.** It contains `"delta":`,
so `_instrument_streaming_throughput` (`v1.py:101`) counts it as a generated token and, for an
upstream that never sent content, records `now - dispatch_started` — i.e. a ~300s sample — into
the slot TTFT deque that feeds the dashboard. `RequestSeam.wrap_streaming` (`metrics/seam.py:130`)
sees no exception, so the row lands `ok=True` with `last_payload.finish_reason == "length"`. The
PR's own thesis is "a cut-off turn no longer looks identical to a clean one"; the gateway's own
telemetry should honour it. Have both wrappers recognise `x_hal0_stall`. (Codex finding —
confirmed by reading both call sites.)

**6 — the wall-clock total bound also policies slow-but-working inference.** The claim that the
guard "does not police slow-but-working inference" holds for the idle bound, not for
`stream_total_timeout_s`. On `ct151-cpu-fresh` — literally the box in this issue — a long
generation at a few tok/s can pass 900s while working correctly, and gets truncated with
`finish_reason: "length"` plus a stall notice injected as assistant *content*, which then enters
conversation history for any client that persists it. The distinguishing property of the reported
defect is unbounded output, not elapsed time. An output-event/byte ceiling would target it exactly
without penalising slow hardware. At minimum, document that the total bound can cut a healthy
long turn.

**7 —** `iter_sse_data`'s 200k ceiling truncates with no marker of its own; the caller only learns
via `stalled` (from the absent `finish_reason`). Acceptable, noted.

---

### Codex triage

| Codex finding | Verdict |
|---|---|
| Handle HTTPX read timeouts inside the stall guard | **Confirmed** — reproduced against a real socket. Blocking 1. |
| Start terminal frames on a fresh SSE event boundary | **Confirmed** — reproduced with a split-event upstream. Blocking 2. |
| Emit endpoint-compatible terminal SSE chunks | **Confirmed**, low impact — non-blocking 4. |
| Exclude synthetic stall chunks from success metrics | **Confirmed** at `v1.py:101` and `metrics/seam.py:130`; the false TTFT is the sharper half — non-blocking 5. |

Nothing dismissed.

thinmintdev:
## Revision — all review findings addressed (`2db77dc4`)

Rebased onto current `main` (`add97eb3`) and pushed one fix commit. Per finding:

### Blocking 1 — `httpx.ReadTimeout` escapes the guard — **fixed**

Two-layer fix, per the review's suggestion (both halves):

1. **The guard owns the read bound.** `_forward_streaming` now builds the streaming request with a per-request `timeout` that sets `read=None` whenever the guard is active (either bound > 0). httpx's read timer is out of the picture, so `stream_idle_timeout_s = 0` genuinely disables the idle cutoff and `idle > direct_read_timeout_s` is no longer dead config.
2. **`httpx.ReadTimeout` is caught anyway.** With the guard fully disabled (both bounds `0`) the client's read timeout is deliberately kept as the transport backstop — and `_guarded_stream` now converts its `ReadTimeout` into a terminal frame with the new reason `"read"` instead of letting it escape mid-stream. Same protection applies to injected `http_client`s that carry their own read bound.

The false premise is corrected everywhere the review listed it: the `router.py` constants comment, the `_get_http_client` comment, `schema.py` descriptions (`direct_read_timeout_s`, `stream_total_timeout_s`), the `<Aside type="caution">` in `streaming.mdx`, and the stall-guard paragraph in `config-schema.mdx`.

Test gap closed with **real-socket tests** (`asyncio.start_server`; MockTransport ignores timeouts):
- `test_idle_bound_above_read_timeout_still_produces_the_diagnostic` — `read=0.2 / idle=0.8`, the review's escape repro; asserts the diagnostic frame and that nothing escapes.
- `test_idle_zero_disables_the_idle_cutoff_and_total_still_bounds` — `idle=0`, `read=0.2`, `total=0.7`; asserts the stream outlives the read timeout (`elapsed_s >= 0.7`) and ends with a `total` frame.
- `test_transport_read_timeout_becomes_a_read_stall_frame` — guard fully disabled, upstream raises `ReadTimeout`; asserts a `"read"` frame + `[DONE]`, not an escape.

### Blocking 2 — terminal frame not on an SSE event boundary — **fixed**

`_stall_sse_frames` now emits a leading `\n\n`, terminating any partial `data:` line so the diagnostic always parses as its own event (a stray blank line before a complete event is a no-op for SSE parsers). Test `test_stall_frame_after_a_mid_event_stall_parses_standalone` stalls the upstream halfway through a `data:` line and asserts the stall frame is a standalone, parseable event.

### Non-blocking 4 — `/v1/completions` got a chat-shaped chunk — **fixed**

The frame shape is now selected from `call.target_url`: `/completions` (non-chat) gets a `text_completion` chunk with `choices[].text` + `logprobs: null`; everything else stays `chat.completion.chunk`. Test `test_legacy_completions_stall_frame_is_text_shaped`.

### Non-blocking 5 — synthetic chunk polluted success metrics — **fixed**

Both wrappers now recognize the `x_hal0_stall` marker (the frame is always yielded as its own chunk, so a per-chunk check is sound):
- `_instrument_streaming_throughput` counts neither a token event nor a TTFT sample for it — the bogus ~300s TTFT for a silent upstream is gone (tests in `test_throughput_history.py`).
- `RequestSeam.wrap_streaming` keeps it out of token/TTFT/`last_payload` accounting and records the row as `ok=0`, `error_code=stream_stall_<reason>` — a guard-tripped request no longer persists as a clean success with `finish_reason: "length"` (tests in `test_seam.py`).

### Non-blocking 6 — total bound polices slow-but-working inference — **documented**

Added a note in `streaming.mdx` and the schema description: `stream_total_timeout_s` is wall clock and can cut a healthy long generation on slow hardware; guidance is to raise it or set it to `0` and rely on the idle bound, which only fires on true silence. An output-event/byte ceiling is a fair GA-era idea but a different bound with its own config surface — left out of scope here rather than grown onto this PR.

### Non-blocking 3 / 7 — acknowledged, no code change

Agreed the provider-plugin helpers are a defensive seam, not the operative half — the PR body already frames the gateway change as what fixes the repro path. Keeping them as-is. The 200k-event ceiling surfacing only via `stalled` is likewise left as noted.

Verification: `make lint`, `ruff format --check src tests`, `scripts/check_sunset.py` all green; `tests/dispatcher/ tests/metrics/test_seam.py tests/api/test_throughput_history.py tests/agents/hermes/plugins/ tests/config tests/api/test_chat_proxy.py` — 1169 passed locally.

thinmintdev:
## Re-review — changes requested (one new blocking defect, introduced by the revision)

Both prior blocking findings are genuinely fixed at the root, and I verified every claim rather
than reading the revision summary. But the Blocking-1 fix over-reached: disabling httpx's `read`
timeout for the whole streaming request also removed the only bound on the **response-header**
wait, and the guard cannot cover that phase because it starts after `send()` returns. On the
shipped defaults, a wedged-but-listening upstream now hangs `forward()` forever — the exact
failure mode #1893 is filed about, moved one step earlier in the request.

---

### What I verified as fixed

**Blocking 1 — `httpx.ReadTimeout` escapes the guard — fixed at the root.** `_forward_streaming`
now builds the streaming request with `httpx.Timeout(..., read=None, ...)` whenever either bound
is set, so the guard owns the body-read bound outright; and `_guarded_stream` catches
`httpx.ReadTimeout` as reason `"read"` for the guard-fully-disabled case. The real-socket tests
are the right shape (`asyncio.start_server`, not `MockTransport`). I ran them against the
pre-revision commit `a1cebe12` with only the test files copied over:

```
FAILED test_idle_bound_above_read_timeout_still_produces_the_diagnostic
FAILED test_idle_zero_disables_the_idle_cutoff_and_total_still_bounds
FAILED test_transport_read_timeout_becomes_a_read_stall_frame
```
…and green on `2db77dc4`. The `idle=0` escape hatch and `idle > direct_read_timeout_s` are now
genuinely honoured. The false premise is corrected in `router.py:131-146`, `router.py:472`,
`schema.py`, `streaming.mdx` and `config-schema.mdx` — I re-read each; none still claims "no read
timeout applies once a stream is open".

**Blocking 2 — SSE event boundary — fixed at the root.** `_stall_sse_frames` emits a leading
`\n\n`; `test_stall_frame_after_a_mid_event_stall_parses_standalone` splits an event across
transport chunks and asserts the diagnostic parses standalone. Red pre-revision, green on head.

**Non-blocking 4 and 5 — both fixed**, both with tests that are red pre-revision:
`test_legacy_completions_stall_frame_is_text_shaped`, `test_stall_guard_frame_recorded_as_failure_not_success`,
`test_stall_frame_alone_records_no_ttft`, `test_stall_guard_frame_excluded_from_throughput_and_ttft`,
`test_stall_guard_frame_alone_records_nothing`. 9 new tests fail on `a1cebe12`, all 10 stall-guard
tests fail on `main` (`add97eb3`). On the branch: `202 passed` for
`tests/dispatcher/test_stream_stall_guard.py tests/metrics/test_seam.py tests/api/test_throughput_history.py tests/agents/hermes/plugins/`.
**Non-blocking 6** documented in both `streaming.mdx` and the schema description. The original
#1893 path is not regressed — the chatty-endless-stream test still cuts at the total bound with a
named reason and partial output preserved.

CI on `2db77dc4`: all 8 check-runs `success` (no cancelled runs masquerading as failures).

---

### Blocking — the header-read wait is now unbounded (regression from the revision)

`_forward_streaming` opens the stream eagerly, and `await client.send(req, stream=True)` blocks
until the upstream sends response **headers**. With `read=None` there is nothing bounding that
wait: `connect` (5.0s) only covers the TCP handshake, and `_guarded_stream` — where both the
total and idle clocks start — is not constructed until `send()` has already returned. An upstream
that accepts the connection and then wedges (never emits status line + headers) now hangs
`forward()` forever, holding a pool connection and the slot in `SERVING`.

Live repro against a real socket (accepts, reads the request, never writes), guard active with
`total=2.0 / idle=1.0` — i.e. the guard should have cut it at 2s:

```
# on 2db77dc4 (this branch)
HANG: forward() did not return within 8s (elapsed 8.01)

# on a1cebe12 (pre-revision), same probe, direct_read_timeout_s=1.0
dispatch.forward_stream_open_failed  error_type=ReadTimeout  upstream=primary
raised UpstreamUnavailable after 1.04s
```

Before the revision this path terminated with a clean `503 UpstreamUnavailable` bounded by
`direct_read_timeout_s`. After it, it is unbounded — and with the shipped defaults the guard is
always active, so every streaming request takes this path. This is not an exotic configuration:
it is the default one, and "client waits forever on a socket, no output, no diagnostic" is
verbatim what #1893 reports. Codex flagged the same thing at P1 on this commit
("Bound response-header reads before disabling the timeout"); I reproduced it before agreeing.

Fix: keep the header phase bounded while still letting the guard own body reads — e.g. wrap the
eager `await client.send(req, stream=True)` in `asyncio.wait_for(..., timeout=self._direct_read_timeout_s)`
(or a dedicated header budget) and map the timeout onto the existing `UpstreamUnavailable` path,
so only `aiter_raw()` runs with `read=None`. Please add a real-socket test with a server that
accepts and never sends headers — the current suite's `_silent_sse_upstream` always sends headers
first, which is why this passed.

---

### Non-blocking

**8 — the per-request timeout hardcodes the non-read bounds.**
`httpx.Timeout(connect=5.0, read=None, write=10.0, pool=5.0)` duplicates the literals from
`_get_http_client` rather than deriving from the client in hand, so an injected `http_client` with
different connect/write/pool budgets silently gets the dispatcher's. The revision comment says
"same protection applies to injected `http_client`s that carry their own read bound", which is
only true when the guard is fully disabled. Cheap to close by copying `client.timeout` and
overriding `read`.

**9 — `_stall_reason` can never return `"read"` from a guarded read**, which is correct, but
`_STALL_MESSAGES["read"]` says "at the transport" without saying the guard was disabled. Minor
wording; the docs cover it.

**3 / 6 / 7 — acknowledged as-is.** Agreed the provider-plugin helpers are a defensive seam with
no production caller (Codex re-raised this on `profile.py:184`); the PR body no longer rests the
repro fix on them, which is the honest framing. The total-bound wall-clock caveat is now
documented in both places. No change needed.

### Codex triage (head commit `2db77dc4`)

| Codex finding | Verdict |
|---|---|
| **P1** Bound response-header reads before disabling the timeout | **Confirmed** — reproduced against a real socket. Blocking. |
| P2 Handle HTTPX read timeouts inside the stall guard (`router.py:1270`) | **Stale** — re-run on head: with the guard active `read=None` so no `ReadTimeout` is raised, and `_guarded_stream` catches it anyway for the disabled case. Both real-socket tests pass. Dismissed. |
| P2 Wire the event ceiling into Hermes' real streaming path (`profile.py:184`) | **Confirmed but out of scope** — same as non-blocking 3; the gateway change is what fixes the repro. No action. |
| P2 endpoint-compatible chunks / P2 SSE boundary / P2 metrics (older commit) | **Fixed** in this revision, verified above. |

thinmintdev:
Fixed the header-wedge blocking defect from the last review round (https://github.com/Hal0ai/hal0/pull/1918#issuecomment-5322682202).

**Root cause:** `_forward_streaming` set `httpx.Timeout(read=None, ...)` for the whole request while the guard is active, so the guard's own clocks (`_guarded_stream`) would own the body bound. But that same disabled read timeout also covered `client.send(req, stream=True)`, which blocks until response *headers* arrive — and the guard's clocks only start after `send()` returns. An upstream that accepts the TCP connection and then wedges before writing a status line hung `forward()` forever, holding a pool connection and the slot in `SERVING`.

**Fix:** bound the header wait explicitly with `direct_read_timeout_s` via `asyncio.wait_for` when the guard is active. `asyncio.TimeoutError` is a builtin `TimeoutError`, itself an `OSError` subclass, so it falls straight into the existing `UpstreamUnavailable` handling alongside other transport errors — matching the pre-guard behavior of a clean bounded 503 instead of a hang. The body path is untouched: `read=None` is still in effect for streamed chunks, and `_guarded_stream`'s own `asyncio.wait_for`/idle/total clocks still own that bound.

**TDD:** added a red-first real-socket regression test (`test_header_wedge_with_guard_active_terminates_bounded`, mirrors the reviewer's exact repro shape: `asyncio.start_server` that reads the request and never writes a response). Confirmed it hung past a 5s `asyncio.wait_for` outer bound before the fix, and passes (raises `UpstreamUnavailable`, no hang) after.

Verified all previously-fixed behaviors from this review round still hold (full `tests/dispatcher/test_stream_stall_guard.py` suite, 11 tests, green): `httpx.ReadTimeout` → reason `"read"` when the guard is fully disabled, the SSE leading-boundary fix, completion-shaped frames on `/v1/completions`, and the idle/total stall diagnostics.

Head SHA: `98e8c8d6a46a26377f75199facce66f6d48d0cc2` (rebased onto `main` @ `b26eec0a`, including #1910's `guard_capability_mismatch` in the same file — resolved clean, no conflicts).

Checks:
- `make lint` — pass
- `uv run ruff format --check src tests` — pass
- `tests/dispatcher tests/metrics tests/api/test_v1_dispatch.py tests/agents/hermes/plugins` — 440 passed
- CI on the PR — all 8 checks green (`python (3.12)` full suite, `γ-suite (chromium)` e2e, CodeQL, lint/format analyzers, ui, sunset)

Not merging — leaving that to the reviewer/operator.

thinmintdev:
## Re-review #3 — head `98e8c8d6` — **approve** (1 non-blocking docs nit)

I did not write this code. Independent verification below; every claim ran on my own harness plus the PR's suite.

### 1. Live repro — header wedge (`asyncio.start_server` that reads the request and never writes)

Reviewer-written probe (not the PR's test), real sockets, `forward()` end-to-end, with an outer watchdog to catch a hang:

| Mode (`total`/`idle`/`direct_read`) | Outcome | Elapsed |
|---|---|---|
| A guard active `60/60/1.0` | `UpstreamUnavailable` | **1.00 s** |
| B guard **disabled** `0/0/1.0` | `UpstreamUnavailable` | **1.00 s** |
| C total only `60/0/1.0` | `UpstreamUnavailable` | **1.00 s** |
| D idle only `0/60/1.0` | `UpstreamUnavailable` | **1.00 s** |

All four terminate at the bound — no hang. **No new hole in the guard-disabled mode**: there the client's default `read=direct_read_timeout_s` still covers `send()` and surfaces as `httpx.ReadTimeout` → `UpstreamUnavailable` (case B, `error_type=ReadTimeout`). The `guard_active` flag correctly covers the "only one bound set" cases (C, D), which are exactly the configs where `read=None` is installed.

**Slow-but-legit upstream tradeoff.** `direct_read_timeout_s` defaults to `300.0` (schema; range 30–600) and prod wiring passes the configured value through `_boot_dispatcher`, so the effective header bound is 300 s — the same value that bounded `send()` on `main` before this PR (the client-level `read`). So this is a **restoration of pre-PR behavior, not a new cut**: an upstream whose headers take >300 s got a bounded 503 on `main` too. It also lines up with `stream_idle_timeout_s=300.0`, so the first-chunk and pre-header silences share one budget. Deliberate — but see the docs nit.

### 2. `asyncio.TimeoutError` → `TimeoutError` → `OSError`, on the real interpreter

`requires-python = ">=3.12"`; ran under **CPython 3.12.13** (and again on 3.13.14):

```
asyncio.TimeoutError is TimeoutError: True
issubclass(TimeoutError, OSError):    True
```

It genuinely lands in `except (httpx.HTTPError, OSError)`. Error shape observed:

```
type        UpstreamUnavailable
str         "upstream 'llm' unreachable: TimeoutError"
code        dispatch.upstream_unavailable
details     {'upstream': 'llm', 'target': 'http://…/v1/chat/completions', 'error': ''}
log         dispatch.forward_stream_open_failed error_type=TimeoutError
```

`str(TimeoutError())` is `''`, so `details["error"]` is empty — but the **message is not** (`type(exc).__name__` is interpolated into it), and `error_type` is on the warning log. That is the same shape the guard-disabled path already produces for `ReadTimeout` (whose `str()` is also `''`), so the #1898 empty-message failure mode does not recur here.

### 3. Cancellation / connection leak

`wait_for` cancels `send()`; I checked what that leaves behind. Wedge server in a **separate process**, so every socket fd counted belongs to the httpx side. 10 consecutive timed-out requests:

```
client-process socket fds before: 2
iter 0..9: socket_fds=2 (+0)  pool_conns=0
after aclose:  socket_fds=2 (base 2)
```

Zero pool connections retained, zero socket growth. No leaked connection, which was part of the original defect's harm.

### 4. Revert-and-confirm-red

Dropped the `wait_for` wrapper, left everything else intact:

```
FAILED tests/dispatcher/test_stream_stall_guard.py::test_header_wedge_with_guard_active_terminates_bounded
  raise TimeoutError from exc_val   ← the test's own 5 s watchdog; forward() never returned
1 failed, 10 deselected
```

Restored → `11 passed in 3.32s`. The test genuinely pins the fix.

### 5. Previously-verified behavior still holds

```
tests/dispatcher tests/metrics tests/api/test_throughput_history.py \
tests/agents/hermes/plugins/test_hal0_provider_plugin.py tests/config
→ 1065 passed, 9 skipped
```

Spot-checked by name: `test_transport_read_timeout_becomes_a_read_stall_frame` (reason `read` with the guard disabled), `test_stall_frame_after_a_mid_event_stall_parses_standalone` (SSE leading boundary), `test_stall_guard_frame_recorded_as_failure_not_success` / `test_stall_frame_alone_records_no_ttft` (seam metrics), plus the full 11-test stall-guard suite. `ruff format --check` and `ruff check` clean on both touched files.

### 6. Rebase / #1910 collision

`git merge-base 98e8c8d6 main = b26eec0a` = current `main` head, so the branch is properly rebased on top of #1910 and #1920. `guard_capability_mismatch` is present and unmodified (`router.py:537`, shifted from `:494` purely by the `Dispatcher.__init__` additions); no diff hunk touches that region. It lives in `dispatch()`, this PR touches `forward()` / `_forward_streaming` — different call paths, no semantic collision.

---

### Non-blocking nit — the docs now contradict the code

`docs/reference/config-schema.mdx` (new in this PR) says:

> While the guard is active, streaming requests disable the HTTP client's read timeout per-request so the guard owns the bound (`direct_read_timeout_s` **then applies to non-streaming requests only**)

That is no longer true as of `98e8c8d6`: `direct_read_timeout_s` also bounds the **response-header wait** on streaming requests while the guard is active. Same stale claim in `docs/reference/api/streaming.mdx:34-35` ("streaming requests therefore disable it per-request so the guard owns the read bound") and in the field description at `src/hal0/config/schema.py:2178`.

Worth one sentence in each — an operator debugging a bounded 503 on a very slow prefill will read exactly this line. Suggested: *"…the guard owns the body bound; the knob still bounds the wait for response headers, which the guard's clocks cannot cover because they only start once headers arrive."* Docs-only, no behavior change — I'd fix it in this PR rather than block on it.

**Verdict: approve.** Blocking findings: **0**. (`--approve` is refused on my own account, hence a comment.)

PR #1919, opened 2026-08-17
fix(cli): catch mkstemp PermissionError in config edit seed with sudo hint

### LINKED ISSUE #1885: config edit: seeding under /etc/hal0 as a non-hal0-group user dies with a raw PermissionError instead of the sudo hint
## Summary

PR #1881 added `_seed_config_file()` / `_fchown_to_service_owner()` so that seeding a missing config file on a real install either lands service-owned or fails with a clean message telling the operator to re-run under sudo. That message only covers one of the two failure points.

For an ordinary non-root user who is **not** in the `hal0` group, `/etc/hal0` (mode `2775 hal0:hal0`) is not writable, so `tempfile.mkstemp` raises `PermissionError` before `_fchown_to_service_owner` is ever reached. `config_edit` catches only `ConfigSeedOwnershipError`, so this surfaces as an unhandled traceback rather than the sudo hint.

Flagged by Codex review on PR #1884 (which documents the shipped behavior); the changelog wording is being narrowed to match reality, and this issue tracks making the code match the intended UX.

## Fix sketch

In `src/hal0/cli/config_commands.py`, catch `PermissionError`/`OSError` from the mkstemp/seed path in `config_edit` (scoped to the real `/etc/hal0` root, same gate as the ownership check) and `die()` with the same re-run-with-sudo message. Regression test: invoke the seed path with a config root whose directory denies writes and assert nonzero exit, no traceback, "sudo" in the output.

## Severity

Polish — the failure is loud either way; it is the presentation that is wrong. Not an rc.6 blocker.

thinmintdev:
## What changed

`config_edit`'s seed path caught only `ConfigSeedOwnershipError` from
`_seed_config_file()`. That error is raised by `_fchown_to_service_owner()`
when the `os.fchown()` call fails — but there is an earlier failure point in
the same function: `tempfile.mkstemp(dir=path.parent)`. A non-root user who
is not in the `hal0` group cannot write into `/etc/hal0` (mode `2775
hal0:hal0`), so `mkstemp` raises a bare `PermissionError` before
`_fchown_to_service_owner` is ever reached. That escaped as an unhandled
traceback instead of the sudo hint the ownership path already gives.

Added an `except OSError` clause around the `_seed_config_file()` call in
`config_edit`, alongside the existing `ConfigSeedOwnershipError` clause. The
sudo-hinted message is only shown when the target is the real `/etc/hal0`
(same gate `_fchown_to_service_owner` uses) — a `HAL0_HOME` sandbox failing
to seed is an unrelated problem (e.g. genuinely out of space), and the
"this box has a hal0 service account" framing would be misleading there.

## Why

Flagged by Codex review on #1884 as a gap in #1881's ownership-hint fix:
the seed path has two distinct failure modes (can't create the temp file at
all vs. can create it but can't chown it), and only one of them got the
clean sudo message.

## How verified

- Added `test_config_edit_seed_hints_sudo_when_mkstemp_denied`, which
  monkeypatches `tempfile.mkstemp` to raise `PermissionError` against the
  real-`/etc/hal0` gate. Confirmed it fails (raw `PermissionError` traceback,
  empty CLI output) on the pre-fix code and passes after the fix.
- `uv run --extra dev pytest tests/cli -q` — 783 passed.
- `uv run --extra dev pytest -q` (full suite) — passed.
- `make lint` — clean.
- `uv run ruff format --check` on changed files — already formatted.

## Out of scope

`config_show`'s analogous `PermissionError` handling on `path.read_text()`
was already correct (that's the pre-existing hint this issue's title
references as the reference UX) — untouched.

Closes #1885

### DIFF
--- a/src/hal0/cli/config_commands.py
+++ b/src/hal0/cli/config_commands.py
@@ -292,6 +292,21 @@ def config_edit(
                 "a seed owned by you at this file's canonical mode. Re-run "
                 "with [bold]sudo[/bold] so the seed lands hal0-owned."
             )
+        except OSError as exc:
+            # tempfile.mkstemp() itself can fail before _fchown_to_service_owner
+            # is ever reached: /etc/hal0 is mode 2775 hal0:hal0, so an
+            # unprivileged caller who isn't in the hal0 group can't create the
+            # temp file there at all (PermissionError), and a read-only mount
+            # would raise EROFS the same way. Only worth the sudo hint against
+            # the real /etc/hal0 — same gate _fchown_to_service_owner uses — a
+            # HAL0_HOME sandbox failing to seed is a different, unrelated
+            # problem (e.g. a genuinely full or missing filesystem).
+            if _config_paths.etc() == Path("/etc/hal0"):
+                die(
+                    f"cannot seed {path}: {exc}. Re-run with [bold]sudo[/bold] "
+                    "so the seed lands hal0-owned."
+                )
+            die(f"cannot seed {path}: {exc}")
     try:
         subprocess.run([editor, str(path)], check=True)
     except (FileNotFoundError, subprocess.CalledProcessError) as exc:
--- a/tests/cli/test_config_show_edit_selector.py
+++ b/tests/cli/test_config_show_edit_selector.py
@@ -227,6 +227,29 @@ def test_config_edit_seed_rejects_seed_on_other_oserror(monkeypatch, tmp_path: P
     assert "sudo" in result.output
 
 
+def test_config_edit_seed_hints_sudo_when_mkstemp_denied(monkeypatch, tmp_path: Path) -> None:
+    """/etc/hal0 itself (mode 2775 hal0:hal0) is not writable by an ordinary
+    user outside the ``hal0`` group, so ``tempfile.mkstemp`` raises
+    ``PermissionError`` before ``_fchown_to_service_owner`` is ever reached —
+    a failure point the ownership-focused ``ConfigSeedOwnershipError`` catch
+    doesn't cover. Must abort with the same sudo hint, not an unhandled
+    traceback (#1885).
+    """
+    cfg_dir = _set_home(monkeypatch, tmp_path)
+    _simulate_real_etc_hal0(monkeypatch)
+    monkeypatch.setenv("EDITOR", "true")
+
+    def _denied_mkstemp(*_a, **_k):
+        raise PermissionError(13, "Permission denied")
+
+    monkeypatch.setattr(config_commands.tempfile, "mkstemp", _denied_mkstemp)
+
+    result = runner.invoke(config_commands.app, ["edit", "hal0"])
+    assert result.exit_code != 0, "a permission-denied seed dir must abort, not traceback"
+    assert not (cfg_dir / "hal0.toml").exists()
+    assert "sudo" in result.output, f"error must point the operator at sudo: {result.output}"
+
+
 def test_config_edit_seed_ignores_host_hal0_account_in_sandbox(monkeypatch, tmp_path: Path) -> None:
     """A HAL0_HOME sandbox seed must succeed regardless of whether the real
     host machine happens to have a system ``hal0`` account.


thinmintdev:
## Review — approve (nits only)

Verified independently against a checkout of `1134430c`, not from the PR description.

### 1. Does it fix the defect #1885 reports? Yes, the exact path.

The issue's repro is: `/etc/hal0` exists at `2775 hal0:hal0`, caller is unprivileged and outside the `hal0` group, so `tempfile.mkstemp(dir=path.parent)` denies before `_fchown_to_service_owner` runs.

I checked the one thing that could have made this an adjacent-path fix — `config_edit` calls `path.parent.mkdir(parents=True, exist_ok=True)` at `config_commands.py:271`, *outside* the guarded block, so if that raised first the new `except OSError` would never be reached. It doesn't, for this repro: `pathlib.Path.mkdir` swallows *any* `OSError` when `exist_ok=True` and the path is already a directory (`except OSError: if not exist_ok or not self.is_dir(): raise`). Confirmed empirically — existing dir under a mode-`0555` parent returns cleanly, and `mkstemp` in the same dir then raises `PermissionError`. So control does reach the new handler on the reported path.

### 2. Root cause vs. symptom: right layer.

`_seed_config_file` is deliberately the "atomic seed or raise" primitive and has exactly one caller (`config_edit`, line 287) — translating errno into operator-facing prose belongs at the CLI boundary, not inside the seeder. Catching `OSError` rather than only `PermissionError` also covers the `os.fchmod`/`write`/`os.replace` legs of the same function, which had the same bare-traceback exposure. No cause left in place for a validation run to re-find on an installed box.

### 3. Does the test fail without the fix? Yes — verified, not assumed.

Reverted `src/hal0/cli/config_commands.py` to `1134430c~1`, kept the test:

```
>       assert "sudo" in result.output, f"error must point the operator at sudo: {result.output}"
E       AssertionError: error must point the operator at sudo:
E       assert 'sudo' in ''
E        +  where '' = <Result PermissionError(13, 'Permission denied')>.output
1 failed, 13 passed
```

That is the issue's symptom verbatim. With the fix restored: `tests/cli` — **783 passed**. All 8 CI checks green.

I also checked the rendered message, because `str(exc)` is interpolated into a Rich markup string and errno text is bracketed. `[Errno 13]` renders literally rather than being eaten as a style tag or raising `MissingStyle`:

```
Error: cannot seed /.../hal0.toml: [Errno 13] Permission denied. Re-run with sudo so the seed lands hal0-owned.
```

### 4. Regressions / 5. Scope

`ConfigSeedOwnershipError` is a `RuntimeError`, so the new `except OSError` cannot shadow the existing clause. `_seed_config_file` has no other callers, and its `finally` still unlinks the temp file, so no half-published config. The only behaviour change for an unmentioned caller is the `HAL0_HOME` sandbox seed, which now dies with a clean `cannot seed <path>: <errno>` instead of a traceback — strictly better. Two files, +38/-0, no `CHANGELOG.md` hunk, no drive-by refactors.

### 6. Codex review bot

Nothing posted on this PR — zero reviews, zero issue comments, zero line comments. Recording that as bot silence, not as an approval. Nothing to triage.

---

### Nits (non-blocking)

**a. `config_commands.py:271` still tracebacks when `/etc/hal0` itself is missing.** The `mkdir` swallow above only applies when the directory already exists. On a box where the installer never created `/etc/hal0` (pipx / manual install), `hal0 config edit` as a non-root user raises `PermissionError: '/etc/hal0'` from line 271, before the try — the same raw traceback #1885 is about, one step earlier. Verified: `Path('/unwritable/nope').mkdir(parents=True, exist_ok=True)` → `PermissionError`. `_seed_config_file` already does the identical `path.parent.mkdir(parents=True, exist_ok=True)` at line 179, so line 271 is redundant; deleting it puts that failure under the new handler for free. Not an installed-fleet path, so not a blocker.

**b. The sudo hint fires for errnos sudo cannot fix.** Under the real `/etc/hal0`, *any* `OSError` gets "Re-run with **sudo** so the seed lands hal0-owned" — including `EROFS` (which the comment itself names) and `ENOSPC`. Sudo does not fix a read-only mount or a full disk. Gating the hint on `isinstance(exc, PermissionError)` and letting everything else fall through to the plain `cannot seed {path}: {exc}` would keep the advice honest. Acknowledged this is consistent with the existing ownership branch, which widened to `OSError` on the same reasoning.

**c. The test's first assertion doesn't test what its message claims.** `assert result.exit_code != 0, "a permission-denied seed dir must abort, not traceback"` passes on the *unfixed* code too — `CliRunner` catches the escaping `PermissionError` and reports `exit_code=1`. Only the `"sudo" in result.output` assertion is load-bearing. Adding `assert isinstance(result.exception, SystemExit)` would make the no-traceback claim real and guard against a future regression that dies for the wrong reason.

All three are follow-up-sized; none of them justify holding the fix.

Issue #703, opened 2026-06-11
GPUMemoryView: own GPU-memory quirks (UMA/GTT split, forced-high artifact) in one view


thinmintdev:
## What to build

A `GPUMemorySample` view owning the LIVE GPU memory + utilization surface, ending the private cross-imports and double-derivations. Today: `hardware/stats.py:98-176` max-pools VRAM/GTT via private probe helpers (`_amd_drm_device`, `_read_sysfs_mb`); `api/routes/hardware.py:410-425` re-imports those SAME private helpers to un-do the pooling for the GTT/VRAM dashboard split; `is_uma` is derived twice (probe detect-time max-pool + route's per-request `vram > ram*0.5` heuristic); and NOTHING interprets the forced-high artifact (gpu-compute.service pins perf level → `gpu_busy_percent` flat 100 regardless of load).

**Interface (locked 2026-06-12 grilling):**

```python
# hardware/gpu_view.py (new)
@dataclass(frozen=True)
class GPUMemorySample:
    vendor: str                       # "amd" | "nvidia" | "unknown"
    is_uma: bool
    vram_total_mb: float | None
    gtt_total_mb: float | None
    total_mb: float | None            # max-pool (current stats semantics)
    vram_used_mb: float | None
    gtt_used_mb: float | None
    used_mb: float | None             # max-pool
    gpu_busy: float | None            # 0..1, reported RAW even when forced-high
    util_is_forced_high: bool

def sample() -> GPUMemorySample
```

## Locked decisions

1. **Scope = readers 1+2 only.** `stats.py` gpu_* methods delegate to the view; `routes/hardware.py` consumes typed fields and its sysfs re-read block (410-425) is deleted. `probe.py` REMAINS the one-time detection owner (the view uses its readers as internals — no more private imports crossing module boundaries). **Out of scope:** `capacity.py` per-slot attribution (cgroup accounting, different concern) and `routes/comfyui.py` (third-party's own HTTP accounting) — both noted as optional future consumers of `total_mb` to kill their hardcoded ceilings.
2. **`util_is_forced_high` = factual sysfs read:** `vendor=="amd" and read(drm/"power_dpm_force_performance_level").strip()=="high"`. No heuristics, no value rewriting — `gpu_busy` stays raw; the flag tells consumers not to trust it. (Dashboard can then caption the util gauge instead of rendering the flat-100 lie as load.)
[...]

Issue #697, opened 2026-06-11
SlotConfigStore + ChangeSet: own capabilities ⇄ slot reconciliation


thinmintdev:
## What to build

Introduce a `SlotConfigStore` that owns **both** `capabilities.toml` selections and `slots/*.toml` as one reconciled truth, ending the drift between them. Today reconciliation is an *unconditional rewrite* inside `capabilities/orchestrator.py:apply()` whose own comment admits the hazard: a half-finished apply, a manual edit, or a migration seed can leave the two files disagreeing (recurring orchestrator-drift bug).

Resolved interface (CONTEXT.md, architecture-deepening section, 2026-06-11):

- `apply(selection) -> ChangeSet` is **compute-only** — it produces `ChangeSet{before, after}` (snapshots of the on-disk slot config) and writes nothing.
- The store also exposes `commit(cs)` (atomic write) and `revert(cs)`.

Keeping `apply()` pure is the point: the write becomes an explicit, observable, reversible step instead of a hidden rewrite. The orchestrator's defensive unconditional rewrite is replaced by `apply()` → `commit()`. Reconciliation uses `model_meta.device_to_backend` / `is_resolvable` (#695) rather than re-deriving them. Existing direct writers of `slots/*.toml` are migrated onto the store so there is a single write path.

ARCHITECTURE.md's "thin overlay" description of the capabilities layer should be updated: the overlay UX stays, but the reconciliation seam is now a deep module with a testable invariant — note this without re-litigating the overlay decision.

## Acceptance criteria

- [ ] `SlotConfigStore` exists with compute-only `apply(selection) -> ChangeSet`, plus `commit(cs)` and `revert(cs)`.
- [ ] `capabilities/orchestrator.py:apply()` delegates to the store; the unconditional-rewrite workaround is removed.
- [ ] Direct `slots/*.toml` writers are routed through the store (single write path).
- [ ] Drift is covered as an invariant: `disk == cs.after` after `commit`; `disk == cs.before` after `revert`; a failed mid-apply leaves disk at `before` (never half-reconciled).
- [ ] Reconciliation imports `model_meta` (no re-derived device→backend / resolvability logic).
- [ ] ARCHITECTURE.md "thin overlay" wording updated to reflect the deepened reconciliation seam.
- [ ] User-facing capability/slot behaviour is unchanged.

## Blocked by

- #695 (model_meta)

PR #1920, opened 2026-08-17
fix: rc.6 polish rollup — capabilities status, synthetic hal0 reason, live uptime, peers AttributeError

### LINKED ISSUE #1905: rollup: 4 small rc.6 residues — capabilities list drops Status column, synthetic hal0 slot logs/reason wrong, frozen uptime, agent peers AttributeError
**Severity:** cosmetic (rollup) · **Box:** ct151-cpu-fresh (items 1-3), ct150-update (item 4).

Small residues that survived rc.6 adversarial verification, individually below filing threshold, collected together so they are not lost. Report per-item, not as one verdict.

### 1. `hal0 capabilities list` drops the `status` field its own API payload carries
A failed capability apply renders `Enabled: yes` with no Status column (`src/hal0/cli/capabilities_commands.py` ~96-118). The `enabled=true` persistence itself is by-design — see `known-issues.yaml: capability-apply-persists-intent-on-lifecycle-failure` (#697).
```
# fail a capability apply, then:
hal0 capabilities list   # no Status column
```

### 2. `GET /api/slots/hal0/logs` 404s with the same `slot.not_found` envelope as a nonexistent slot
The synthetic `hal0` composite entry's `_synthetic_reason` string ("Backed by remote upstream; install a local slot of the same name to take over") is factually wrong for the `hal0` composite (`slot_view/__init__.py:861` vs. the UI's own copy).
```
curl -s $API/api/slots/hal0/logs ; curl -s $API/api/slots/hal0 | jq ._synthetic_reason
```

### 3. `uptime_s` in `/api/hardware` and `/api/stats/hardware` is host uptime frozen at hardware.json write time
The dashboard host row renders it as live uptime forever — it can exceed or undercut real uptime by weeks, worst on upgraded boxes which never re-probe. ct150 reported 4110s on a box up 326,305s.
```
cat /proc/uptime; curl -s $API/api/stats/hardware | jq .uptime_s   # frozen = defect
```

### 4. `hal0 agent peers` crashes with AttributeError on ct150
`agent_commands.py:1372`: `'str' object has no attribute 'get'` when a card's `hal0_state` metadata comes back as a JSON string instead of a dict (seen on the upgraded box). Spin-off found while cross-checking `agents-bank-boot-writes-lost`.
```
hal0 agent peers   # on a box with published identity cards
```

### Expected
Each corrected independently.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `rc6-polish-rollup`.

thinmintdev:
## Summary

Four small, independently-verified residues from rc.6 adversarial validation, rolled into issue #1905 and fixed here as one branch since it's one filed issue (each item is still independently tested):

1. **`hal0 capabilities list` drops the Status column** (`src/hal0/cli/capabilities_commands.py`) — the API payload already carries a per-selection `status` field; the CLI table just never rendered it, so a failed capability apply looked identical to a healthy one.

2. **Synthetic `hal0` composite slot has the wrong `_synthetic_reason` + its `/logs` route 404s like a nonexistent slot** (`src/hal0/slot_view/__init__.py`, `src/hal0/api/routes/slots.py`) — the reason string ("Backed by remote upstream; install a local slot...") was hardcoded for every synthetic entry regardless of kind, which is factually wrong for the local `hal0` composite. Fixed by branching on `u.kind`. Separately, `GET /api/slots/hal0/logs` raised the typed `slot.not_found` envelope even though `hal0` is a legitimate, listable synthetic entry (not a real slot, but not "not found" either) — it now 200s with an explanatory hint, mirroring how `get_slot`'s detail view already falls through to the synthetic entry. A genuinely unknown slot name still 404s.

3. **`uptime_s` frozen at `hardware.json` write time** (`src/hal0/hardware/uptime.py` new, `src/hal0/api/routes/hardware.py`) — `/api/hardware` and `/api/stats/hardware` served the cached probe's `uptime_s` as if it were live; on a box that's never re-probed since an upgrade this drifts by weeks (ct150 reported 4110s on a host up 326305s). Split the read into a standalone `hal0.hardware.uptime.read_uptime_s()` — deliberately outside `hal0.hardware.probe` so the API route layer doesn't pull in that module's private GPU/NPU probe internals (an existing test, `test_route_module_has_no_private_probe_imports`, guards the #703 architecture boundary that keeps GPU/NPU state flowing through the coalesced `HardwareStats` singleton instead of ad-hoc route-level probing) — and call it fresh on every serve.

4. **`hal0 agent peers` AttributeError** (`src/hal0/cli/agent_commands.py`) — crashed with `'str' object has no attribute 'get'` when a card's `metadata.hal0_state` came back from the memory API as a JSON-encoded string instead of an already-parsed dict (observed on an upgraded box). `md.get("hal0_state") or {}` doesn't catch this since a non-empty string is truthy. Added a small `_as_dict` coercion applied to `metadata`, `endpoint`, and `hal0_state` alike (same storage path, same failure mode) that parses a JSON string back to a dict and degrades anything else — including malformed JSON — to `{}` so the row still renders with "—".

## Test plan

- [x] Each item has its own regression test that fails on the pre-fix code and passes after.
- [x] Confirmed the `agent peers` regression by reverting just that hunk and re-running its tests — reproduces the original `AttributeError` crash.
- [x] `uv run pytest -q` — 10313 passed, 16 skipped, 1 xfailed, 0 failed
- [x] `uv run ruff check` — clean
- [x] `uv run ruff format --check` — clean

## Out of scope

- `capability-apply-persists-intent-on-lifecycle-failure` (#697) — the `enabled=true` persistence-on-failure behavior itself is by-design per `known-issues.yaml`; this PR only adds visibility (the Status column), it doesn't change that behavior.
- Did not touch CHANGELOG.md per the fix-wave process rule (consolidated CHANGELOG PR lands at the end of the wave).

Closes #1905

### DIFF
--- a/src/hal0/api/routes/hardware.py
+++ b/src/hal0/api/routes/hardware.py
@@ -218,9 +218,11 @@ def _with_live_uptime(flat: dict[str, Any]) -> dict[str, Any]:
     can be off by weeks (under- or over-reporting). The dashboard renders
     this as if it were live, so every serve of this shape must recompute
     it. The read is a cheap sysfs stat (not the heavy subprocess-fanout
-    probe), so doing it on every request is fine.
+    probe), so doing it on every request is fine. If /proc/uptime is
+    unreadable (``read_uptime_s`` returns 0), keep the cached value —
+    stale-but-plausible degrades better than a hard 0.
     """
-    flat["uptime_s"] = read_uptime_s()
+    flat["uptime_s"] = read_uptime_s() or flat.get("uptime_s", 0)
     return flat
 
 
--- a/src/hal0/api/routes/hardware.py
+++ b/src/hal0/api/routes/hardware.py
@@ -24,6 +24,7 @@ from hal0.config import paths
 from hal0.config.loader import load_hardware_info
 from hal0.hardware import gpu_view
 from hal0.hardware.gpu_view import GPUMemorySample
+from hal0.hardware.uptime import read_uptime_s
 from hal0.providers import podman_introspect
 from hal0.providers.podman_introspect import PodmanImagesResult
 
@@ -209,25 +210,40 @@ async def _gpu_sample(request: Request) -> GPUMemorySample | None:
         return None
 
 
+def _with_live_uptime(flat: dict[str, Any]) -> dict[str, Any]:
+    """Overwrite ``uptime_s`` with a fresh ``/proc/uptime`` read.
+
+    #1905: ``hardware.json``'s ``uptime_s`` is frozen at the last ``hal0
+    probe`` write — on a box that's never re-probed since an upgrade it
+    can be off by weeks (under- or over-reporting). The dashboard renders
+    this as if it were live, so every serve of this shape must recompute
+    it. The read is a cheap sysfs stat (not the heavy subprocess-fanout
+    probe), so doing it on every request is fine.
+    """
+    flat["uptime_s"] = read_uptime_s()
+    return flat
+
+
 @router.get("/hardware")
 async def get_hardware(request: Request) -> dict[str, Any]:
     """Return cached /etc/hal0/hardware.json, falling back to a fresh probe.
 
     The probe is heavy enough (subprocess fanout) that we prefer the
     cached snapshot; ``POST /api/hardware/probe`` forces a re-run.
+    ``uptime_s`` is always re-read live (#1905) regardless of cache state.
     """
     gpu = await _gpu_sample(request)
     target = paths.hardware_json()
     if target.exists():
         try:
             info = load_hardware_info().model_dump(mode="python")
-            return _flatten_for_ui(info, gpu)
+            return _with_live_uptime(_flatten_for_ui(info, gpu))
         except Exception:
             pass
     # Cache miss → probe now.
     probe = request.app.state.hardware_probe
     info = (await probe.probe_async()).model_dump(mode="python")
-    return _flatten_for_ui(info, gpu)
+    return _with_live_uptime(_flatten_for_ui(info, gpu))
 
 
 @router.post("/hardware/probe")
--- a/src/hal0/api/routes/slots.py
+++ b/src/hal0/api/routes/slots.py
@@ -1483,14 +1483,20 @@ def _find_synthetic_entry(request: Request, name: str) -> dict[str, Any] | None:
 
 
 def _synthetic_logs_hint(entry: dict[str, Any]) -> str:
-    """Explain why a synthetic entry has no journal, worded by kind.
+    """Explain why a synthetic entry has no journal, worded per entry.
 
-    #1905 asked about the ``hal0`` composite, but remote upstreams reach
-    this path too — calling one of those a "composite" would be wrong,
-    so mirror the kind split ``_synthetic_reason`` already makes.
+    #1905 asked about the ``hal0`` composite, but remote upstreams and
+    (in stale-registry states) ordinary container-slot registrations
+    reach this path too — calling those a "composite" would be wrong, so
+    mirror the classification ``slot_view._synthetic_reason`` makes: the
+    composite hint needs BOTH the reserved name and ``kind="slot"``.
     """
+    from hal0.upstreams.registry import RESERVED_UPSTREAM_NAME
+
     if entry.get("kind") == "slot":
-        return "synthetic composite slot has no journal of its own"
+        if entry.get("name") == RESERVED_UPSTREAM_NAME:
+            return "synthetic composite slot has no journal of its own"
+        return "upstream registration without a configured slot; no journal to read"
     return "backed by a remote upstream; no local journal to read"
 
 
--- a/src/hal0/api/routes/slots.py
+++ b/src/hal0/api/routes/slots.py
@@ -1588,6 +1588,15 @@ async def slot_logs_stream(
                 + json.dumps({"message": _synthetic_logs_hint(entry)})
                 + "\n\n"
             )
+            # Hold the stream open after the degraded frame: the client's
+            # degraded listener only records the message, while its onerror
+            # closes and RECONNECTS (ui/src/api/hooks/useLogs.ts) — so a
+            # stream that ends here would put the viewer in a silent retry
+            # loop against this endpoint forever. Keepalive cadence matches
+            # the journal streams (#1472) so proxy idle timeouts are covered.
+            while True:
+                await asyncio.sleep(15)
+                yield ": keepalive\n\n"
 
         return StreamingResponse(
             synthetic_source(),
--- a/src/hal0/api/routes/slots.py
+++ b/src/hal0/api/routes/slots.py
@@ -1485,11 +1485,29 @@ async def slot_logs(
     never started, returns an empty string with a hint. The UI tolerates
     that (renders "No logs available") rather than treating it as an error.
     The journalctl tail lives in :func:`hal0.slots.logs.read_tail`.
+
+    #1905: a synthetic upstream entry (e.g. the ``hal0`` composite) isn't
+    a real slot and has no ``hal0-slot@{name}.service`` journal of its
+    own, but it also isn't "not found" — it's a legitimate, listable
+    entry. Distinguish the two: a genuinely unknown name still 404s with
+    the typed ``slot.not_found`` envelope, but a synthetic entry gets a
+    200 with an explanatory hint instead, mirroring how :func:`get_slot`
+    already falls through for its detail view.
     """
     sm = _get_slot_manager(request)
     # Validate slot exists so unknown names get the typed slot.not_found
     # envelope instead of an empty 200.
-    await sm.status(name)
+    try:
+        await sm.status(name)
+    except Exception:
+        for entry in _synthesize_slots_from_upstreams(request):
+            if entry["name"] == name:
+                return {
+                    "name": name,
+                    "logs": "",
+                    "hint": "synthetic composite slot has no journal of its own",
+                }
+        raise
 
     text, hint = await _logs.read_tail(f"hal0-slot@{name}.service", lines, quiet)
     if hint is not None:
--- a/src/hal0/api/routes/slots.py
+++ b/src/hal0/api/routes/slots.py
@@ -55,6 +55,7 @@ from hal0.slots import metrics_collect as _metrics_collect
 from hal0.slots import port_alloc as _port_alloc
 from hal0.slots import voices as _voices
 from hal0.slots.manager import Slot, SlotManager
+from hal0.slots.state import SlotNotFound
 
 log = structlog.get_logger(__name__)
 
@@ -1473,6 +1474,26 @@ async def swap_slot(name: str, request: Request) -> dict[str, object]:
 # the ``StreamingResponse``.
 
 
+def _find_synthetic_entry(request: Request, name: str) -> dict[str, Any] | None:
+    """Return the synthetic upstream-backed entry for ``name``, if any."""
+    for entry in _synthesize_slots_from_upstreams(request):
+        if entry["name"] == name:
+            return entry
+    return None
+
+
+def _synthetic_logs_hint(entry: dict[str, Any]) -> str:
+    """Explain why a synthetic entry has no journal, worded by kind.
+
+    #1905 asked about the ``hal0`` composite, but remote upstreams reach
+    this path too — calling one of those a "composite" would be wrong,
+    so mirror the kind split ``_synthetic_reason`` already makes.
+    """
+    if entry.get("kind") == "slot":
+        return "synthetic composite slot has no journal of its own"
+    return "backed by a remote upstream; no local journal to read"
+
+
 @router.get("/{name}/logs")
 async def slot_logs(
     name: str, request: Request, lines: int = 200, quiet: bool = True
@@ -1496,17 +1517,21 @@ async def slot_logs(
     """
     sm = _get_slot_manager(request)
     # Validate slot exists so unknown names get the typed slot.not_found
-    # envelope instead of an empty 200.
+    # envelope instead of an empty 200. Catch ONLY SlotNotFound: every
+    # loaded container slot registers a same-name ``kind="slot"`` upstream
+    # (SlotManager._register_container_upstream), so a broad except here
+    # would convert a real slot's operational failure (e.g. SlotConfigError
+    # from a corrupt state.json) into a fake synthetic 200.
     try:
         await sm.status(name)
-    except Exception:
-        for entry in _synthesize_slots_from_upstreams(request):
-            if entry["name"] == name:
-                return {
-                    "name": name,
-                    "logs": "",
-                    "hint": "synthetic composite slot has no journal of its own",
-                }
+    except SlotNotFound:
+        entry = _find_synthetic_entry(request, name)
+        if entry is not None:
+            return {
+                "name": name,
+                "logs": "",
+                "hint": _synthetic_logs_hint(entry),
+            }
         raise
 
     text, hint = await _logs.read_tail(f"hal0-slot@{name}.service", lines, quiet)
@@ -1542,7 +1567,33 @@ async def slot_logs_stream(
     import shutil
 
     sm = _get_slot_manager(request)
-    await sm.status(name)  # 404 fast if unknown
+    try:
+        await sm.status(name)  # 404 fast if unknown
+    except SlotNotFound:
+        # #1905: same synthetic fall-through as the one-shot ``/logs`` route
+        # above — the dashboard log viewer and ``hal0 slot logs --follow``
+        # both drive THIS route, so fixing only the one-shot twin left the
+        # user-facing flow 404ing (EventSource.onerror → endless reconnect
+        # loop). Emit a terminal 'degraded' frame with the hint instead; the
+        # client already listens for 'degraded' (see B13 below). Only
+        # SlotNotFound falls through — other failures on a real slot must
+        # keep surfacing as errors.
+        entry = _find_synthetic_entry(request, name)
+        if entry is None:
+            raise
+
+        async def synthetic_source() -> Any:
+            yield (
+                "event: degraded\ndata: "
+                + json.dumps({"message": _synthetic_logs_hint(entry)})
+                + "\n\n"
+            )
+
+        return StreamingResponse(
+            synthetic_source(),
+            media_type="text/event-stream",
+            headers={"Cache-Control": "no-cache", "X-Accel-Buffering": "no"},
+        )
 
     async def event_source() -> Any:
         if shutil.which("journalctl") is None:
--- a/src/hal0/cli/_shared.py
+++ b/src/hal0/cli/_shared.py
@@ -108,16 +108,28 @@ def api_stream(
         yield resp
 
 
-def _iter_sse_payloads(resp: httpx.Response) -> Iterator[Any]:
-    """Parse an SSE response's ``data: ...`` lines into JSON (or raw text)."""
+def _iter_sse_payloads(resp: httpx.Response) -> Iterator[tuple[str | None, Any]]:
+    """Parse an SSE response into ``(event_name, payload)`` pairs.
+
+    ``event_name`` is the preceding ``event:`` field for named frames
+    (e.g. ``degraded``), ``None`` for plain ``data:`` frames. Payloads are
+    JSON-decoded when possible, raw text otherwise.
+    """
+    event: str | None = None
     for raw in resp.iter_lines():
-        if not raw or not raw.startswith("data:"):
+        if not raw:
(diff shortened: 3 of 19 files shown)


thinmintdev:
## Review — changes requested

Verified by checking out `5925fb18`, running the new tests, then reverting **only** `src/` (`git checkout 5925fb18~1 -- src/`) and re-running.

**Test honesty: good.** 7 of the 8 new tests go red without the fix; the 1 that stays green is the deliberately-labelled "precondition" case. No fake tests here.

```
FAILED tests/cli/test_agent_peers.py::test_peers_survives_hal0_state_as_json_string
FAILED tests/cli/test_agent_peers.py::test_peers_malformed_hal0_state_string_degrades_to_dash
FAILED tests/cli/test_capabilities_commands.py::test_list_renders_status_column
FAILED tests/slot_view/test_aggregator.py::TestSyntheticEntries::test_synthetic_reason_differs_by_kind
FAILED tests/api/test_slots_routes.py::test_logs_synthetic_hal0_composite_returns_hint_not_404
FAILED tests/api/test_hardware_routes.py::TestLiveUptime::test_get_hardware_reports_live_uptime_not_cached_value
FAILED tests/api/test_hardware_routes.py::TestLiveUptime::test_stats_hardware_also_reports_live_uptime
7 failed, 1 passed
```

**Item 1 (capabilities Status column) and item 3 (live uptime) both look correct and complete.** For item 3 I checked the other consumers: `/api/system-info` goes through `get_hardware` so it inherits the fix, and the `hal0 system-info` CLI runs a fresh `HardwareProbe().probe()` rather than reading `hardware.json`, so it was never stale. No gaps.

Items 2 and 4 need work. Three blocking findings below, two of which are the Codex bot's — both of its claims verified and **confirmed**.

---

### 1. BLOCKING — the broad `except Exception` in `slot_logs` converts real slot failures into a fake synthetic 200 (`src/hal0/api/routes/slots.py:1500`)

Codex flagged this; it is correct, and worse than it reads. `SlotManager._register_container_upstream` (`src/hal0/slots/manager.py:471-497`) registers a `kind="slot"` upstream **named after the slot** for *every loaded container slot*. So the fallback loop almost always finds a same-name entry for a real, loaded slot — the escape hatch is not narrow.

Reproduced against a real slot `chat` with a corrupt `state.json` (`read_state` raises `SlotConfigError`), same upstream registration `_register_container_upstream` performs on load:

```
# 5925fb18~1 (before this PR)
STATUS: 400 BODY: {"error":{"code":"slot.config_error","message":"state.json at .../chat/state.json is not valid JSON: Expecting property name enclosed in double quotes: line 1 column 3 (char 2)", ...}}

# 5925fb18 (this PR)
STATUS: 200 BODY: {"name":"chat","logs":"","hint":"synthetic composite slot has no journal of its own"}
```

A precise, actionable config error becomes a 200 that tells the operator their real GPU slot is a synthetic composite. Please catch `SlotNotFound` specifically (`hal0.slots.state.SlotNotFound`) rather than bare `Exception`.

### 2. BLOCKING — the streaming twin is untouched, and it is the only path the dashboard uses (`src/hal0/api/routes/slots.py:1519`)

Also Codex's; confirmed. `slot_logs_stream` still does a bare `await sm.status(name)  # 404 fast if unknown`, so `GET /api/slots/hal0/logs/stream` keeps 404ing.

This is not a theoretical second route. `ui/src/dash/extras.jsx:65` drives the log viewer exclusively through `useSlotLogsStream` → `ENDPOINTS.slotLogsStream` (`ui/src/api/endpoints.ts:53`) — there is no one-shot fallback — and the slot picker's options come from `slotNames = (slotsQuery.data || []).map(s => s.name)` (`extras.jsx:42`), i.e. `/api/slots`, **which includes the synthetic `hal0` entry**. So an operator can select `slot: hal0` in the log viewer, and post-merge it still fails — and fails via `EventSource.onerror`, so the UI shows "Log stream disconnected — reconnecting…" in a retry loop rather than the `degraded` frame that would explain it. `hal0 slot logs hal0 --follow` (`src/hal0/cli/slot_commands.py:414`) is broken the same way.

The issue's literal `curl` repro is fixed; the behaviour a validator actually clicks on is not. rc-validate will re-find this. Please mirror the handling into `slot_logs_stream` as a terminal `event: degraded` frame carrying the hint (the client already listens for `degraded`).

### 3. BLOCKING — item 4 stops the crash but does not recover the data, and its test asserts a wire shape the real path never produces

The root cause is one line: `src/hal0/memory/hindsight_provider.py:956`

```python
metadata={k: str(v) for k, v in meta.items()},
```

`str()` on a dict yields a **Python repr with single quotes**, not JSON. (`pgvector_provider.py:132` keeps the dict intact — this is Hindsight-specific, which is exactly why it showed up on an upgraded box and not everywhere.)

So the actual wire value on ct150 is `"{'registered_at': '2026-08-10T00:00:00Z', 'bootstrap_version': 1}"`. `_as_dict` calls `json.loads` on that, gets `JSONDecodeError`, and returns `{}`. The new tests use `json.dumps(...)` — double-quoted, valid JSON — a shape `str(v)` can never emit, so they pass while the reported box stays broken.

Feeding the real shape through the patched command:

```
WIRE SHAPE: "{'registered_at': '2026-08-10T00:00:00Z', 'bootstrap_version': 1}"
EXIT: 0
┃ Agent ID     ┃ Display name ┃ Roles                  ┃ Endpoint ┃ Registered ┃
│ hermes-ct150 │ —            │ [, ', h, o, m, e, l,   │ —        │ —          │
│              │              │ a, b, -, a, d, m, i,   │          │            │
│              │              │ n, ', ,,  , ', g, e,   │          │            │
│              │              │ n, e, r, a, l, i, s,   │          │            │
│              │              │ t, -, c, h, a, t, ', ] │          │            │
```

Two problems visible in one row:

- **Endpoint and Registered still render `—`.** No crash, but no data either — the peer registry is unreadable on exactly the box the issue was filed from.
- **`roles` was left uncoerced.** The PR's own rationale ("same storage path, same failure mode") applies to it identically, and `", ".join(<str>)` splits it per character. The AttributeError used to fire on `endpoint` first, so this garbage render is *newly visible* after the fix.

Suggested shape: fix `hindsight_provider.py:956` to `json.dumps(v) if isinstance(v, (dict, list)) else str(v)` so the write path stops destroying structure, keep `_as_dict` as the defensive read for already-written cards (adding an `ast.literal_eval` fallback so existing ct150 cards decode), and add an `_as_list` for `roles`. At minimum, retarget the regression test at the repr shape — a test that only covers a shape the system cannot produce is not a regression test.

---

### Nits (non-blocking)

- **`_synthetic_reason` branches on `u.kind == "slot"`, which is broader than "the hal0 composite"** (`src/hal0/slot_view/__init__.py:862`). Per `manager.py:487`, every loaded container slot registers a `kind="slot"` upstream, so all of them now claim to be the "Composite /v1 endpoint that fronts every chat model". Real-slot-wins masks this on the happy path, but it leaks through `get_slot`'s synthetic fall-through (`slots.py:1064`) and through `/api/status` when `real_entries = []` after a SlotManager error (`routes/health.py:203-227`). `u.name == RESERVED_UPSTREAM_NAME` says what you actually mean. (The string itself is right — it matches `ui/src/dash/data.jsx:68` verbatim, which is what the issue asked for.)
- **The new `hint` says "synthetic composite" for remote upstreams too.** `GET /api/slots/haloai/logs` goes from 404 to a 200 calling a remote upstream a composite — a behaviour change for a caller the issue never mentioned. Word it by `kind`, same as the reason string.
- **`_with_live_uptime` overwrites unconditionally** (`routes/hardware.py:223`), so a `read_uptime_s()` that returns `0` (unreadable `/proc/uptime`) clobbers a cached value that was at least plausible. `flat["uptime_s"] = read_uptime_s() or flat.get("uptime_s", 0)` degrades better.
- **Two implementations of the same `/proc/uptime` read** (`probe._read_uptime_s` and `uptime.read_uptime_s`), kept apart only to preserve `probe`'s `_read_text` test seam. The new module imports nothing from `hal0`, so `probe` can just delegate; a test seam is a thin reason to carry a duplicate.
- **`hal0/hardware/uptime.py` has no direct test** — both new tests monkeypatch it away, so the parse/`OSError`/`ValueError` branches are uncovered.

### Process

CHANGELOG.md correctly untouched. No scope creep — the `probe.py` hunk is docstring-only. CI is green.

### Codex triage

Both bot comments verified against the code and **confirmed** — see findings 1 and 2. Neither dismissed.

thinmintdev:
Review round addressed in fca5fb35 (rebased onto current main first — #1909/#1913/#1915/#1919 landed underneath).

### Blocking finding 1 — broad `except Exception` masked real slot failures
`slot_logs` now catches only `hal0.slots.state.SlotNotFound` before the synthetic fall-through. New regression test `test_logs_real_slot_config_error_not_masked_as_synthetic` reproduces your exact scenario: real slot `chat` with a same-name `kind="slot"` upstream (the shape `_register_container_upstream` creates) and a corrupt `state.json` — asserts the `slot.config_error` envelope surfaces instead of a synthetic 200.

### Blocking finding 2 — streaming twin untouched
`slot_logs_stream` gets the same `SlotNotFound`-only fall-through and returns a 200 SSE response with a single terminal `event: degraded` frame carrying the hint — the client already listens for `degraded` (B13), so the log viewer renders the explanation instead of an `EventSource.onerror` reconnect loop. Unknown names still 404 with the typed envelope. Test: `test_logs_stream_synthetic_hal0_composite_emits_degraded_frame`.

### Blocking finding 3 — data not recovered, test asserted an impossible wire shape
Root cause fixed at the write path: `hindsight_provider.py` retain now serializes nested dict/list metadata with `json.dumps` (scalars stay `str()`), so new cards round-trip. Read path: `_as_dict` tries `json.loads` then falls back to `ast.literal_eval` so the repr-shaped cards already written on ct150 decode, and a new `_as_list` gives `roles` the same coercion (no more per-character explosion). The regression test is retargeted at the real wire shape — `str({'registered_at': ...})` / `str([...])` — and asserts Endpoint, Registered, and Roles all recover their data, not just that the command exits 0. Plus `test_add_serializes_nested_metadata_as_json_not_repr` locks the write path.

### Nits
- `_synthetic_reason` now keys the composite string off `u.name == RESERVED_UPSTREAM_NAME`; non-reserved `kind="slot"` upstreams get their own honest string ("Registered upstream for container slot …; the real slot entry is authoritative."), remote keeps the install-a-local-slot copy. Aggregator test extended to cover all three.
- Logs `hint` is now worded by kind — remote upstreams are no longer called composites.
- `_with_live_uptime` keeps the cached value when `read_uptime_s()` returns 0 (test added).
- `probe._read_uptime_s` now delegates to `hal0.hardware.uptime` (duplicate deleted); `tests/hardware/test_uptime.py` covers the parse/missing/garbage/empty branches directly.

### Codex triage
Both bot comments were real; both fixed as above (findings 1 and 2).

Verified: `make lint`, `ruff format --check`, `scripts/check_sunset.py`, and targeted suites (`tests/api/test_slots_routes.py`, `tests/api/test_hardware_routes.py`, `tests/cli/test_agent_peers.py`, `tests/cli/test_capabilities_commands.py`, `tests/memory`, `tests/slot_view`, `tests/hardware`, `tests/cli/test_slot_commands.py`) — all green locally.

thinmintdev:
Second Codex round (on fca5fb35) triaged — all three were real, fixed in 1a2453a0:

- **Reconnect loop after terminal degraded frame** — confirmed against `useLogs.ts` (degraded listener records; `onerror` closes + reschedules `connect()`). The synthetic degraded stream now stays open with `: keepalive` ticks on the standard 15 s cadence (#1472) after the frame, so EventSource never sees an error. Test consumes the generator directly (same pattern as `test_state_stream_emits_transition_event`).
- **Nested metadata not decoded on general read paths** — `_fact_to_item` now inflates JSON-string dict/list metadata via `_decode_metadata` (only values that parse as containers; scalars keep their historical string wire shape). Test: `test_fact_to_item_decodes_nested_metadata_strings`.
- **One-shot CLI swallowed the hint** — `hal0 slot logs <name>` now prints `hint` when `logs` is empty. Test: `test_slot_logs_one_shot_prints_hint_when_logs_empty`.

Lint/format/targeted suites green locally.

thinmintdev:
Third Codex round triaged — two fixed in 464a6865, one declined with reasoning:

**Fixed — CLI follow hang on terminal degraded frame.** `_iter_sse_payloads` now yields `(event, payload)` pairs and `follow_sse_logs` treats `degraded` as terminal: prints the message and returns. The server keeps the stream open for browsers (EventSource reconnect semantics); the CLI stops. Test: `test_slot_logs_follow_terminates_on_degraded_frame` (bounded fake stream that fails loudly if the CLI reads past the frame).

**Fixed — explicit remote `hal0` misclassified.** `_synthetic_reason` now requires the reserved name AND `kind="slot"` for the composite string; an operator-configured `kind="remote"` `hal0` entry gets the remote-backed reason. Test: `test_synthetic_reason_explicit_remote_hal0_classified_by_kind`.

**Declined — encoding marker for JSON-looking scalar strings.** The ambiguity is real but narrow: it only affects a caller that stores a *string* whose content parses to a JSON dict/list, and only on the Hindsight engine (whose metadata is flat string→string — some flattening convention is forced; pgvector keeps native dicts and is unaffected). A proprietary marker would (a) fail to decode every card already written in the plain-JSON shape, including the ones this PR just fixed the writer to produce, and (b) leak a hal0-only encoding into the `shared` Hindsight bank that non-hal0 agents (Pi, Hermes) read directly. Decoding container-shaped strings is the standard convention for flat-metadata stores, and `_decode_metadata` only converts values that fully parse as containers — everything else passes through byte-identical. Trade-off documented in the helper's docstring; if a concrete consumer ever needs verbatim JSON-shaped scalar strings, that is a new contract discussion, not a regression from this PR (the previous behavior destroyed ALL nested metadata unconditionally).

thinmintdev:
Fourth Codex comment (composite log hint applied to every kind="slot" entry) — real, fixed in adf86e1f: `_synthetic_logs_hint` now mirrors `_synthetic_reason`'s classification — composite hint only for the reserved `hal0` name with `kind="slot"`; other slot-kind registrations get "upstream registration without a configured slot; no journal to read". Lint/format/targeted tests green.

thinmintdev:
## Re-review — approve

Verified against head `adf86e1f` in a clean worktree (`git reset`, `git diff HEAD` empty, no phantom index).

### Blocking findings — all three resolved at the root, each confirmed red without the fix

I reverted **only** the specific hunk for each finding and re-ran its regression test.

**1. Broad `except Exception` masked real slot failures** — fixed: `slot_logs` now catches only `hal0.slots.state.SlotNotFound`.
Reverted `except SlotNotFound:` → `except Exception:` and re-ran:
```
FAILED tests/api/test_slots_routes.py::test_logs_real_slot_config_error_not_masked_as_synthetic
```
The test reproduces my exact scenario (real slot `chat` + same-name `kind="slot"` upstream, corrupt `state.json`) and asserts the `slot.config_error` envelope. Green on the branch.

**2. Streaming twin untouched** — fixed: `slot_logs_stream` gets the same `SlotNotFound`-only fall-through and returns 200 with a terminal `event: degraded` frame, then holds open with 15 s keepalives.
Reverted the `try/except` back to the bare `await sm.status(name)` and re-ran:
```
FAILED tests/api/test_slots_routes.py::test_logs_stream_synthetic_hal0_composite_emits_degraded_frame
```
Checked the semantics too: the two `degraded` emitters in `slots.py` (1593, 1619) are both terminal, so `follow_sse_logs` treating `degraded` as terminal is consistent, and `/api/logs/stream` (the only other `follow_sse_logs` caller, `doctor_commands.py:307`) emits no `degraded` frame — no collateral early-return.

**3. Data not recovered; test asserted an impossible wire shape** — fixed at the write path *and* the read path.
Reverted `json.dumps(...)` → `str(v)` in `hindsight_provider.retain` and dropped the `ast.literal_eval` fallback in `_parse_stringified`:
```
FAILED tests/cli/test_agent_peers.py::test_peers_recovers_python_repr_shaped_card
FAILED tests/memory/test_hindsight_provider.py::test_add_serializes_nested_metadata_as_json_not_repr
```
The retargeted test now feeds the real `str({...})` / `str([...])` repr shape and asserts Endpoint, Registered **and** Roles recover their data — no per-character `roles` explosion. `_decode_metadata` in `_fact_to_item` (1969) is the only read adapter that returns metadata, so the fix covers every REST/MCP consumer, not just the CLI.

### No regression on the original #1905 fix

Ran the composite path end-to-end against the live app fixtures (ad-hoc, not committed):
```
REASON: Composite /v1 endpoint that fronts every chat model — not a lifecycle slot.
STATUS: 200 HINT: synthetic composite slot has no journal of its own
```
The `_synthetic_reason` tightening to `name == RESERVED_UPSTREAM_NAME and kind == "slot"` still matches the `_UpstreamsWithHal0Composite` stand-in (`slots.py:235` — `kind="slot"`, `name="hal0"`), so item 2 of the issue is still fixed, with the wrong-kind leak closed. Items 1 (Status column) and 3 (live uptime) untouched by the revision beyond the `read_uptime_s() or cached` degradation and the `probe._read_uptime_s` delegation, both covered.

### Nits I raised last round

All five addressed: name-based composite classification (plus the `kind` conjunct for an explicit remote `hal0`), kind-worded logs hint, `_with_live_uptime` zero-degradation, `probe` delegating to the new module instead of duplicating, and `tests/hardware/test_uptime.py` covering the parse/OSError/garbage branches directly.

### Codex triage

All nine inline comments across four rounds triaged; eight fixed, one declined. I re-read the declined one (encoding marker for JSON-looking scalar strings) and **agree with the decline**: a proprietary marker would break every card the fixed writer just produced and would leak a hal0-only encoding into the `shared` bank that Pi/Hermes read directly. `_decode_metadata` only converts values that fully parse as containers; the residual ambiguity (a caller deliberately storing a container-shaped *string*) is documented in the helper docstring. Codex has not re-reviewed the head SHA `adf86e1f` (its last pass was `464a6865`) — the one-line change there is a straight mirror of an already-reviewed classification.

### Verification

- `ruff check src tests` clean, `ruff format --check` clean (1148 files), `scripts/check_sunset.py` clean (192 <= 192 baseline).
- `tests/api/test_slots_routes.py tests/cli/test_agent_peers.py tests/memory/test_hindsight_provider.py tests/api/test_hardware_routes.py tests/cli/test_capabilities_commands.py tests/slot_view/ tests/hardware/ tests/cli/test_slot_commands.py tests/cli/test_cli_auth_streamtest.py` — **427 passed**.
- CI green on `adf86e1f`: all 8 check-runs `completed/success` (verified via `commits/<sha>/check-runs`, not just the glyphs).
- CHANGELOG.md correctly untouched. No scope creep.

### Non-blocking nit (follow-up, not for this PR)

`_synthetic_logs_hint` (`routes/slots.py`) now duplicates `_synthetic_reason`'s three-way classification, and `adf86e1f` shipped that mirror without a test — the non-composite hint branches ("upstream registration without a configured slot", "backed by a remote upstream") are uncovered, and `test_logs_synthetic_hal0_composite_returns_hint_not_404` only asserts the hint is truthy. Two copies of the same rule will drift. Worth folding the classification into one shared helper when something next touches it.

thinmintdev:
Merge-train note — conflict with #1912, resolved without changing this PR's reviewed content.

After #1907/#1908/#1910/#1911/#1912/#1914/#1916/#1917 landed, this branch went `dirty`. The only conflict was in `src/hal0/cli/agent_commands.py`: **#1912 fixed the same #1905 `hal0 agent peers` stringified-metadata `AttributeError`**, and landed first with a superset implementation (its `_as_list` also turns a bare non-empty string into a single-element list; ours returned `[]`).

Resolution: merged current `main` into the branch (`762f48a9`) and took `main`'s helpers verbatim, deleting this branch's duplicate `_as_dict` / `_parse_stringified` / `_as_list` block. `src/hal0/cli/agent_commands.py` is now byte-identical to `main` (`git diff main -- src/hal0/cli/agent_commands.py` is empty).

Verification that nothing else moved: for every other file in this branch's diff, `git diff main -- <file>` is byte-identical to the reviewed `git diff add97eb3 adf86e1f -- <file>` — 17/17 files identical, plus `src/hal0/cli/slot_commands.py`, whose auto-merge carries exactly the reviewed hunk on top of #1923's changes.

This branch's own regression tests (`tests/cli/test_agent_peers.py`) still pass against `main`'s implementation, alongside #1912's (`tests/cli/test_agent_peers_metadata.py`):

```
595 passed in 80.17s
(tests/cli/test_agent_peers.py test_agent_peers_metadata.py test_slot_commands.py
 test_capabilities_commands.py tests/api/test_hardware_routes.py test_slots_routes.py
 tests/hardware tests/memory tests/slot_view)
```

`ruff check` and `ruff format --check src tests` are clean.

Issue #1922, opened 2026-08-17
slots: add an output-sanity gate to slot readiness ('The capital of France is' → 'Paris')


thinmintdev:
## Why

Split out of the #1888 GA-blocker (see its "Missing output-sanity gate" section). Filed separately per the fix-wave scope fence: #1888's PR retires the Vulkan lane and makes `/dev/kfd` an explicit requirement, but deliberately does NOT touch `slots/manager.py` / `slots/watchdog.py`.

On ct151 during rc-validate v1.0.0-rc.6, a box produced **zero usable language for two hours** while every readiness surface hal0 owns read green:

- slot state `ready`
- `/api/health` green
- `hal0 doctor` "OK gpu"
- the install-time provision smoke test
- the SSE `done` frame
- 60–120 tok/s throughput

Five independent defects were reported by five validation lanes (`hermes -z` hangs forever; install-time `chat_completions` smoke "timed out"; hindsight fact extraction burned the GPU for 96 minutes; …) — every one of them was the same underlying defect. **A one-second coherence probe at the end of slot load would have collapsed all five into one loud failure at install time.**

## What

Add an **output-sanity gate** to slot readiness: after a `llm`-type slot's health check passes, run one greedy completion and require coherent output before the slot is allowed to reach `ready`.

### Probe design (from #1888)

```
POST http://127.0.0.1:<port>/completion
{"prompt": "The capital of France is", "n_predict": 12, "temperature": 0}
```

Pass condition: the response text contains `Paris` (case-insensitive).

**Do NOT key on the garbage string.** The garbage form varies with argv — `--jinja` slots emit runs of `)`, bare `/completion` emits empty content. A positive assertion on the expected token is the only stable check.

### Requirements / open design points

- **Scope:** `llm`-type slots only. Embedding/rerank/TTS/STT/image slots need their own probes (or none) — do not block them on a chat completion.
- **Placement:** at the end of slot load, after the health probe, before the slot is published as `ready`. Likely `slots/manager.py` with the watchdog respecting the same verdict on restart.
- **Failure mode:** slot goes to `error` with an actionable message naming the probe and the expected token — not a silent degrade, and not a retry loop.
- **Budget:** must be cheap and bounded (a few seconds max) so it does not extend the already-tight slot lifecycle budget; needs its own short timeout, and a timeout must be a FAIL, not a pass.
[...]

PR #1923, opened 2026-08-17
fix(gpu): require /dev/kfd and retire the Vulkan LLM lane

thinmintdev:
> ⚠️ **NEEDS OPERATOR REVIEW — release-shape change. Do not merge, do not enable automerge.**
> This changes fresh-install behaviour (a GPU install without `/dev/kfd` now stops) and retires an advertised backend lane. It must be validated on a real box before it lands.

## What

Closes out GA-blocker #1888 by taking the operator-approved direction — **drop/relabel the Vulkan lane** rather than repin or rebuild the runner image.

### 1. `/dev/kfd` is now an explicit, loud requirement

- `installer/lib/preflight.sh` — new classification `HAL0_GPU_RC_NO_KFD(5)`: an AMD render node visible with no ROCm compute node. New test seams `HAL0_GPU_KFD_PATH` / `HAL0_GPU_AMD_OVERRIDE`. Doctor mode stays soft/advisory; only the install gate returns the code. The Proxmox remedy line now says `dev1: /dev/kfd` is **REQUIRED**, not optional.
- `installer/install.sh` — hard-stops on that code with the remedy, with the same explicit CPU-only opt-in shape as the existing no-device case (`HAL0_ALLOW_CPU_ONLY=1` or a `/dev/tty` confirm).
- `src/hal0/providers/_gpu.py` — new `kfd_present()` / `require_kfd_for_gpu_slot()` / `GpuPreflightError`, with an `HAL0_ALLOW_VULKAN_FALLBACK=1` escape hatch that warns rather than silently passing.
- `src/hal0/providers/container.py` — `load_sync` refuses a `gpu-rocm`/`gpu-vulkan` slot with no `/dev/kfd`, alongside the existing NPU loud-fail guard. Enforced on the **load** path, not in `container_spec`, so unit rendering / previews / status renders stay host-independent.

### 2. The `device = "gpu-vulkan"` mislabel is gone from every llama.cpp surface

Both GPU device ids resolve to the same ROCmFPX image for llama.cpp slots, so `gpu-vulkan` never named the backend a slot executed on — it was ROCm wherever `/dev/kfd` existed, and garbage wherever it did not.

- `src/hal0/config/data/seed_stacks.toml` — the three `gpu-vulkan` utility/agent slots (saber, forge, pi) → `gpu-rocm`; stack descriptions no longer say "Vulkan utility helper" / "light Vulkan agent".
- `src/hal0/api/routes/stacks.py` — the stack-apply fallback for an entry with no device: `gpu-vulkan` → `gpu-rocm`.
- `src/hal0/install/profile_derive.py` — an AMD GPU with no ROCm compute derives **`cpu`**, not `gpu-vulkan`. Non-AMD vulkan-capable GPUs keep the lane (see open question).
- `src/hal0/hardware/recommend.py` — `_backend_for`'s AMD ladder: ROCm when `compute_capable` **or** `/dev/kfd` is present (the "AMD UMA / Strix Halo class" rationale string is preserved), else CPU. This branch used to return `vulkan` for the entire Strix Halo class — it is the root of the mislabel.
- `src/hal0/cli/slot_commands.py` — `--hardware` auto-detect: AMD always → `rocm`; the no-probe fallback moves from `vulkan` to `rocm`.
- `src/hal0/capabilities/catalog.py` — the llama.cpp backend fan-out suppresses `gpu-vulkan` entirely on a host that advertises `gpu-rocm`, so the picker cannot offer a lane whose slot reads ready and returns garbage. `gpu-vulkan` is untouched for Kokoro / whisper.cpp / ComfyUI, which run genuinely-Vulkan images.

### 3. The bench harness stops publishing the lane

- `src/hal0/bench/harness.py` — `vulkan_radv` leaves `default_lanes` (GPU tier is now `["rocm"]`), and is stamped in a new `UNSUPPORTED_LANES` / `lane_is_supported()`. The `LaneSpec` is retained so historical records and an explicit `--backends vulkan_radv` still resolve.
- `src/hal0/bench/planner.py` — the registry's `"vulkan"` backend hint now maps to the **rocm** lane, not `vulkan_radv`. An explicit `vulkan_radv` still resolves to itself.
- `installer/bench/suites/lane-matrix.toml` / `roster.toml` — shipped suites no longer sweep the lane.

### 4. Docs stop recommending it

- `docs/hal0-install-migration-guide.html` — the "**Vulkan** recommended on Strix Halo / v1.0 standardizes shared APU models on Vulkan" card is replaced with a "`/dev/kfd` required on AMD" card.
- `docs/concepts/strix-halo.mdx` — "ROCm or Vulkan on the iGPU" → "ROCm is the LLM lane on the iGPU", with a caution box explaining the silent fallback.
- `docs/reference/hardware-matrix.mdx` — device table + decision tree updated to the new ladder.
- `docs/guides/manage-slots.mdx` — `--hardware` guidance and the example command.

## Verification

```
uv run ruff format --check src tests   →  clean
make lint (ruff check src tests)        →  All checks passed!
```

Targeted suites (`env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest ... -q`):

```
tests/providers tests/capabilities tests/config tests/install tests/installer
tests/bench tests/hardware tests/registry tests/api/test_stacks_routes.py
tests/cli/test_slot_commands.py
  → 3105 passed, 10 skipped   (after fixing the 4 unit-render tests below)
tests/providers                → 549 passed
```

Full suite run recorded in the PR thread.

**Red-on-main proof** (same tests, with `src/` + `installer/` stashed back to `main`):

```
FAILED tests/config/test_seed_stack_devices.py::test_no_seed_stack_slot_declares_gpu_vulkan
FAILED tests/config/test_seed_stack_devices.py::test_stack_apply_defaults_to_rocm_when_an_entry_omits_device
FAILED tests/installer/test_preflight_gpu_gate.py::test_gate_amd_render_node_without_kfd_stops
FAILED tests/install/test_profile_derive.py::test_chat_on_amd_box_without_rocm_picks_cpu_not_vulkan
FAILED tests/hardware/test_recommend.py::test_seeded_slot_carries_container_runtime_and_profile
FAILED tests/hardware/test_recommend.py::test_amd_uma_recommends_rocm_not_vulkan
FAILED tests/hardware/test_recommend.py::test_amd_without_rocm_falls_back_to_cpu_not_vulkan
FAILED tests/hardware/test_recommend.py::test_amd_without_rocm_smi_but_with_kfd_is_rocm
8 failed, 50 passed
```
(`tests/providers/test_gpu_kfd_preflight.py` and `tests/bench/test_harness.py` fail at import on main — the symbols they assert on do not exist there.)

New tests: `tests/providers/test_gpu_kfd_preflight.py` (helper + `load_sync` guard, device scoping, escape hatch), `tests/config/test_seed_stack_devices.py` (no shipped seed names the lane), plus four new cases in `tests/installer/test_preflight_gpu_gate.py` (gate stops without kfd, proceeds with it, AMD-scoped, doctor mode stays soft).

Four pre-existing `load_sync` unit-render tests (`test_container_mmproj.py`, `test_container_chat_template.py`) now patch `require_kfd_for_gpu_slot` alongside the `resolve_gpu_device_paths` patch they already carried — they exercise unit rendering, not host GPU validity, and the CI box has no `/dev/kfd`.

## Not in scope

The **output-sanity slot-readiness gate** (`'The capital of France is'` → `'Paris'` probe) — the one-second check that would have collapsed five separately-reported rc.6 defects into one loud install-time failure. It drags `slots/manager.py` and `slots/watchdog.py` into the blast radius and would collide with #1911 / #1915, so it is filed with the full probe design as **#1922**.

Also out of scope by direction: repinning or rebuilding the runner image.

## Open questions for the operator

1. **Non-AMD Vulkan fallbacks.** `_backend_for` still returns `vulkan` for NVIDIA-without-CDI and Intel iGPU, and `derive_device` still returns `gpu-vulkan` for non-AMD vulkan-capable GPUs. Those slots resolve to the **same** ROCmFPX image, so they are plausibly equally broken — but the defect is only characterised on the AMD/HIP build, and routing them to CPU without evidence would regress a population that may be fine. Left alone deliberately. Want a follow-up issue to A/B a non-AMD box?
2. **Fresh-install blast radius.** A Strix Halo box whose probe reports `compute_capable=False` and has no `/dev/kfd` now installs **CPU-only** where it previously installed (garbage-serving) GPU slots. That is the intended behaviour change, but it is a visible one.
3. **Existing installs.** Slots already on disk with `device = "gpu-vulkan"` are not migrated — they will now refuse to load with the preflight message rather than being silently rewritten. Should an updater migration relabel them to `gpu-rocm`?

Refs #1888
Refs #1922

### DIFF
--- a/docs/concepts/strix-halo.mdx
+++ b/docs/concepts/strix-halo.mdx
@@ -19,9 +19,10 @@ A Strix Halo machine gives hal0 three places to run a model:
 
 <CardGrid>
   <Card title="Radeon iGPU" icon="seti:default">
-    The integrated GPU runs LLM and image workloads through either ROCm
-    (`gpu-rocm`) or Vulkan (`gpu-vulkan`). It's detected as an AMD `amdgpu`
-    device; the marketing name comes from the system at probe time.
+    The integrated GPU runs LLM workloads through ROCm (`gpu-rocm`, which
+    requires `/dev/kfd`) and image/audio workloads through Vulkan
+    (`gpu-vulkan`). It's detected as an AMD `amdgpu` device; the marketing
+    name comes from the system at probe time.
   </Card>
   <Card title="XDNA NPU" icon="seti:config">
     The AMD XDNA NPU is driven by FastFlowLM (FLM) on the `npu` device. hal0
@@ -41,26 +42,28 @@ NVIDIA silicon.) A slot's `device` field is what picks one of these — see
 [Capabilities & profiles](/docs/concepts/capabilities-and-profiles) for why
 that ownership lives on the slot, not the profile, as of v1.0.
 
-## ROCm or Vulkan on the iGPU
+## ROCm is the LLM lane on the iGPU
 
-The iGPU can run through two backends, and hal0 treats them as distinct devices
-because they behave differently:
+- **ROCm** (`gpu-rocm`) — AMD's compute stack, and the only supported backend
+  for llama.cpp (LLM) slots on AMD hardware. It needs the `/dev/kfd` compute
+  node to be visible to hal0 and to the slot containers.
+- **Vulkan** (`gpu-vulkan`) — retained for the non-llama.cpp runtimes that
+  genuinely ship Vulkan images (Kokoro TTS, whisper.cpp, ComfyUI). It is **not**
+  a supported LLM backend.
 
-- **ROCm** (`gpu-rocm`) — AMD's compute stack. hal0 tests for it by probing for
-  the ROCm tooling; the GPU is marked compute-capable when it's present. ROCm
-  can deliver the best throughput, but it depends on a matching kernel and
-  user-space combination the installer can't easily verify on its own.
-- **Vulkan** (`gpu-vulkan`) — the broad-compatibility path via Mesa, which ships
-  with the `amdgpu` driver. It works on essentially any AMD GPU without extra
-  setup, which makes it the safe default for unified-memory APUs.
+<Aside type="caution">
+hal0's LLM slots all run the same unified ROCmFPX runner image, which is a
+single HIP + Vulkan build. llama.cpp picks ROCm when `/dev/kfd` is visible and
+silently falls back to that image's Vulkan backend when it is not — and that
+Vulkan backend emits invalid tokens for every model it serves, at full nominal
+speed, while HTTP 200, container health, `hal0 doctor` and the streaming `done`
+frame all read green.
 
-<Aside type="note">
-On Strix Halo-class unified-memory hardware, hal0's hardware recommender leans
-toward Vulkan as the safe default — ROCm requires a verified kernel and
-user-space stack the probe can't confirm on its own. You can still choose
-`gpu-rocm` explicitly when your stack supports it; the seed stacks
-([Stacks](/docs/concepts/stacks)) generally pair a ROCm agent with a Vulkan
-utility model as a middle ground.
+So `/dev/kfd` is a hard requirement, not an optimisation. The installer refuses
+to complete a GPU install without it, and a GPU LLM slot refuses to load. In a
+Proxmox LXC, forward it with a `dev1: /dev/kfd` line in
+`/etc/pve/lxc/<CTID>.conf`, then `pct stop` / `pct start`. Without it, run
+CPU-only.
 </Aside>
 
 ## Unified memory is the tuning target
--- a/docs/guides/manage-slots.mdx
+++ b/docs/guides/manage-slots.mdx
@@ -53,7 +53,7 @@ memory extraction); a custom slot is reachable only by its own name.
 hal0 slot create embed \
   --type embedding \
   --model nomic-embed-text-v1.5-q8_0 \
-  --hardware vulkan \
+  --hardware rocm \
   --ctx-size 8192
 ```
 
@@ -67,9 +67,15 @@ hal0 slot create embed \
 
 3. **`--model` / `-m`** (required) — the initial model ref to assign.
 
-4. **`--hardware`** — the compute backend: `vulkan`, `rocm`, or `cpu`.
-   When omitted it reads the cached hardware probe: `vulkan` if no probe
-   has run yet, `cpu` if the probe found no GPU. This sets the slot's
+4. **`--hardware`** — the compute backend: `rocm`, `cpu`, or `vulkan`.
+   When omitted it reads the cached hardware probe: `rocm` on any AMD GPU
+   (and if no probe has run yet), `cpu` if the probe found no GPU. `vulkan`
+   is **not a supported LLM backend** — llama.cpp slots run the unified
+   ROCmFPX runner image, whose Vulkan build emits invalid tokens for every
+   model, so it is reserved for the non-llama.cpp Vulkan runtimes and for
+   non-AMD GPUs. `rocm` requires the `/dev/kfd` compute node; a slot
+   without it refuses to load rather than serving garbage. This sets the
+   slot's
    `device` field directly (`vulkan → gpu-vulkan`, `rocm → gpu-rocm`,
    `cpu → cpu`) — as of v1.0, **the slot owns its hardware grid**
    (`device`, `n_gpu_layers`, `threads`, `binary`, `image_pin`), not the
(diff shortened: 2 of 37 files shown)


thinmintdev:
## Codex review triage (commit `42eec7f7`)

Six findings were real and are fixed; two more addressed in kind; two left with reasoning.

**Fixed**

| # | Finding | Verdict |
|---|---|---|
| P2 | Static slot seeds still `gpu-vulkan` (`installer/etc-hal0/slots/*.toml`) | **Real and the most important one.** Those six files are what a fresh install actually receives — the same seeds ct151 got. Relabelled, plus a regression test that walks that directory (the seed-stack test genuinely did not cover it). |
| P1 | KFD gate fires on non-AMD `gpu-vulkan` slots | **Real.** Scoped: `gpu-rocm` always (the device name *is* the ROCm claim), `gpu-vulkan` only when `amdgpu` is bound. New `host_is_amd_gpu()` + `amd_host` param. |
| P1 | `strix-halo` branch made the AMD→CPU fallback unreachable | **Real, and it would have broken the CPU-only opt-in path this PR adds.** ROCm now requires `compute_capable` **or** a reachable `/dev/kfd`. |
| P2 | Mixed AMD+Intel host forced to CPU | **Real.** The non-AMD Vulkan check now runs before the AMD→CPU fallback. |
| P1 | Picker suppression keyed off the `gpu-rocm` catalog row | **Real, and it left the trap exactly where it mattered** — the no-kfd box advertises `gpu-vulkan` but not `gpu-rocm`. Suppression is keyed off the AMD host itself. |
| P1 | Stack-apply `gpu-rocm` constant breaks Intel/NVIDIA | **Real.** `_default_stack_slot_device()` resolves from the host probe, `DEFAULT_DEVICE` as last resort. |

**Addressed in kind**

- **`kfd_present` existence vs usability** — right that a mis-mapped gid is a false pass; it now also requires read+write access. A genuinely *functional* ROCm probe is the job of the output-sanity readiness gate (#1922), not the slot-load hot path.
- **`lane_is_supported` never called** — right; the planner now warns once per unsupported lane at plan time, so an opt-in `vulkan_radv` run cannot be planned silently.

**Not changed, with reasoning**

- **Boot / `systemctl restart` bypasses `load_sync`.** Largely already closed by the generated unit: a GPU slot's Quadlet carries `AddDevice=/dev/kfd`, and podman refuses to start a container whose `--device` path is missing — so a unit rendered while the node was present cannot start after it disappears. The residual is a unit rendered by a *pre-fix* hal0 (device `gpu-vulkan`, no kfd `AddDevice`), which is the existing-install migration question raised as open question 3 in the PR body — an updater migration is the right home for it, not this PR.
- **Vulkan bench hints → `rocm` lane on non-AMD.** Real in principle, but pre-existing and not a regression from this change: `default_lanes(TIER_NVIDIA)` already returned the rocm lane before this PR, and the bench harness is built around the ROCmFPX image. Worth its own issue if the operator wants non-AMD bench support; flagging rather than widening this PR's blast radius.

thinmintdev:
## Final verification (`89e0ea48`)

**CI: all green.** `python (3.12)` pass · `γ-suite (chromium)` pass · `sunset` pass · `ui` pass · CodeQL ×4 pass.

**Full local suite:** `10342 passed, 16 skipped, 1 xfailed` in 21m54s (`env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests -q`). `ruff format --check src tests` and `ruff check src tests` clean.

One γ-suite flake noted honestly: run `95540220943` failed on `voice-page-provider-copy-v3.spec.ts` (TTS voice `af_bella` → `— unset —`). It is not caused by this branch — the immediately preceding commit `42eec7f7` differs from `89e0ea48` only by the wording of one docstring, and γ-suite passed there (`95537917344`, 7m58s). Nothing in this PR touches the TTS voice picker (Kokoro fans out through `_RUNTIME_TO_HOST_BACKENDS`, not the llama.cpp branch this PR changed). Re-run passed in 7m44s.

**Still awaiting operator review — do not merge, automerge not enabled.** The three open questions in the PR body (non-AMD Vulkan fallbacks, the fresh-install CPU-only behaviour change, and migration for existing `gpu-vulkan` slot TOMLs) are the ones that need a decision before this lands.

Issue #1896, opened 2026-08-15
hal0 doctor perms: exits 1 with DRIFT on a freshly installed box and --fix can never converge


thinmintdev:
**Severity:** minor · **Box:** ct151-cpu-fresh.

### Summary

`hal0 doctor perms` exits 1 with DRIFT rows on a freshly installed box, and its own remediation (`sudo hal0 doctor perms --fix`) cannot produce a durably green audit — three drift classes regenerate after the installer's own `--fix` backstop:

1. `STATE.md` rewritten `0644` by `hermes_provision._atomic_write` (tmp-write, no chmod, `UMask=0022`) vs. the table's `0664`.
2. `hal0-agentenv` re-tightens `secrets/` to `0700` on every merge-secrets, vs. the table's `0755` — the two shipped components disagree (the contradiction is the defect, not a leak: `*.env` stays `0600` either way — see `known-issues.yaml: doctor-perms-secrets-dir-0755-is-declared`).
3. Every daemon-created dir under recursive rows lands `2755` vs. `2775`.

A self-audit that can never be green trains operators to ignore it. Do not assert a drift ROW COUNT — it grows with every slot loaded and model pulled.

### Repro

```
hal0 doctor perms >/dev/null 2>&1; echo rc=$?          # 1 on a pristine install = defect
COLUMNS=200 hal0 doctor perms | grep DRIFT
# date the drift against the installer's backstop by ctime, not mtime
```

### Expected

On a fresh install, `hal0 doctor perms` is green, and stays green after a slot load, a model pull, and a hermes `STATE.md` render.

### Fix

Localised: chmod after `os.replace` in `_atomic_write` (or tolerate `0644` for `STATE.md`), reconcile agentenv `0700` vs. `perms.py` `0755` in one direction, and mkdir state dirs with an explicit mode.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `doctor-perms-never-converges`.

thinmintdev:
Fixed by PR #1942 (merged as 39a3901e), after two independent review rounds (final: approve, 0 blocking, per-item red-first evidence).

- secrets/ + secrets/agents converge at 0700 root:root (posture upheld by both reviews: every declared reader is root; the hermes wrapper's non-root read is blocked one level down by 0600 either way; 0755 would have required widening a sudoers-reachable root wrapper).
- Dev-mode installs no longer abort: secrets birth is dev-mode-safe (install -d -m 0700 + tolerant chown).
[...]

Issue #1546, opened 2026-07-30
fresh install: /var/lib/hal0/hal0.db is root-owned, so the service can never write the registry — every model pull downloads in full, then fails


thinmintdev:
**A fresh install cannot add a model at all.** Reproduced end to end on CT151 (clean Ubuntu 26.04 LXC, `installer/install.sh` from main, 2026-07-30).

```
$ hal0 model pull qwen3.5-0.8b
Starting pull for qwen3.5-0.8b (unsloth/Qwen3.5-0.8B-GGUF/Qwen3.5-0.8B-UD-Q4_K_XL.gguf)…
  ... 100%  532.9MB / 532.9MB  failed
Error: failed: OperationalError: attempt to write a readonly database
```

The **entire 532 MB downloads successfully**, then registration fails. The file is left on disk, orphaned and unregistered:

```
528M  /var/lib/hal0/models/chat/qwen3.5-0.8b/model.gguf
GET /api/models → {"models":[],"count":0,"filtered_aliases":0}
```

## Root cause

`hal0-api` runs as `User=hal0`. The primary database is owned by root:

```
644 root:root  /var/lib/hal0/hal0.db          ← the registry
644 hal0:hal0  /var/lib/hal0/activity.db      ← correct
644 hal0:hal0  /var/lib/hal0/.hermes/state.db ← correct
644 hal0:hal0  /var/lib/hal0/.hermes/kanban.db ← correct
2775 hal0:hal0 /var/lib/hal0/{,registry,models}  ← dirs correct
```

Confirmed directly as the service account:

```
$ sudo -u hal0 python3 -c "sqlite3.connect('/var/lib/hal0/hal0.db').execute('CREATE TABLE …')"
NOT WRITABLE: OperationalError attempt to write a readonly database
```

SQLite needs write access to the **file** (and its directory, for `-wal`/`-journal`). The directory is fine; the file is not. It is created by root during install and never chowned, while its siblings are.

## Why this is release-blocking

- **The golden path is dead on arrival.** A new user's first substantive action — get a model — cannot succeed. The failure arrives *after* a full download, so it also wastes the user's bandwidth and disk.
- **`install.sh` reports success**, and its own perms engine prints `✓ ownership table applied (37 path(s) reconciled)` / `✓ ownership clean`.
[...]

PR #1933, opened 2026-08-18
fix(install): add ownership rows for model-pull-jobs/ and activity.db

### LINKED ISSUE #1895: installer: root-owned model-pull-jobs dir breaks pull-status persistence for hal0-api across restarts

### Summary

The installer's root-run brain-model pull lazily `mkdir`s `/var/lib/hal0/model-pull-jobs` under root's umask, so on every fresh install it ends up `root:root 0755` with a root-only `0600` snapshot inside. `hal0-api` (`User=hal0`) then fails every pull-job snapshot write (`model.pull_job_persist_failed`, WARNING only) and cannot read the existing one, so the `#626` restart-survival fallback is dead in both directions — `GET /api/models/<id>/pull/status` 404s after a restart. `install/perms.py` has no row for the directory, so `doctor perms --fix` cannot heal it, and `install.sh`'s O13 chown list omits it. Pulls themselves still work; this only breaks status persistence across a restart.

The fail-soft WARNING and the installer-scoped "writable paths" doctor line are by-design (see `known-issues.yaml: pull-job-persist-fail-soft-scope`).

### Repro

```
ls -ld /var/lib/hal0/model-pull-jobs && sudo -u hal0 test -w /var/lib/hal0/model-pull-jobs \
  || echo "hal0 CANNOT write"
journalctl -u hal0-api -b 0 | grep -c model.pull_job_persist_failed
systemctl restart hal0-api && curl -s $API/api/models/<seeded id>/pull/status   # 404 = defect
```

### Expected

`/var/lib/hal0/model-pull-jobs` is hal0-writable after a fresh install, `perms.py` carries a row for it (and for `activity.db` + `-wal`/`-shm`), and a pull's status survives an `hal0-api` restart.

### Test to promote

A pytest over the perms table covering every path `hal0-api` writes.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `model-pull-jobs-root-owned`.

thinmintdev:
## Summary

Fixes #1895: `hal0-api` (`User=hal0`) failed every pull-job snapshot write (`model.pull_job_persist_failed`, WARNING-only fail-soft) and could not read an existing snapshot after a restart, because `/var/lib/hal0/model-pull-jobs` had no row in the `OwnershipStore` declarative table (`src/hal0/install/perms.py`). The installer's root-run brain-model pull (`installer/install.sh`'s bundle-tier auto-pull, before line 2809's `doctor perms --fix --force` backstop) is the first writer on a fresh install, lazily `mkdir`-ing the dir `root:root 0755` under root's umask — the same O13 birth-ownership bug class as #1546 (`hal0.db`). No row meant `doctor perms --fix` had nothing to reconcile.

The issue's "Expected" section also calls out `activity.db` (+ its `-wal`/`-shm` WAL siblings, `hal0.activity.AuditStore` runs `PRAGMA journal_mode=WAL`) as missing the identical row — added here too.

- **No `install.sh` change needed.** The bundle-tier brain-model pull (line ~1967) runs well before the `doctor perms --fix --force` backstop (line 2809), so the table-only fix is sufficient: the backstop now has a declared opinion to reconcile the root-owned dir against.
- Confirmed `runner-image-pull-jobs/` (the sibling runner-image pull store) is NOT affected — it's never touched at install time as root, only ever created by the `User=hal0` service itself, so it's born correctly already. Left out of scope.

## What changed

`src/hal0/install/perms.py`:
- `model-pull-jobs/` — `hal0:hal0 2775`, `glob="*.json"` children `0600` (matches `tempfile.mkstemp`'s birth mode, same convention as `registry.toml`/`slots/*/state.json`).
- `activity.db` / `activity.db-wal` / `activity.db-shm` — `hal0:hal0 0644`, mirroring the existing `hal0.db` rows exactly.

`tests/install/test_perms.py`:
- New `test_every_hal0_api_write_path_has_an_ownership_row` — parametrized over every `var_lib()` path `hal0-api` can be the first writer of (the two new ones + `hal0.db`/`registry/`/`slots/`/`models/` as regression anchors), so a future service-writable path can't go unrowed the same way `model-pull-jobs/` did. Confirmed RED against `main` (5 failures) before the `perms.py` change, GREEN after.
- New `test_model_pull_jobs_dir_snapshot_files_get_hal0_owned_child_rows` — asserts the `model-pull-jobs/` row's `child_mode` matches `persist_pull_job`'s actual `tempfile.mkstemp` birth mode (`0600`).

## Sequencing note

Kept minimal and confined to the missing rows + the coverage test — #1896 (`doctor perms` never-converges) also touches `perms.py` in unrelated areas (`STATE.md`, `secrets/`, recursive-mkdir modes) and is deliberately held until this lands, so it can rebase cleanly.

## Test plan

- [x] `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/install tests/installer -q` → 748 passed, 1 skipped
- [x] Confirmed the 2 new pytest functions are RED on `main` (5 parametrized failures) before the `perms.py` fix, GREEN after
- [x] `make lint` → clean
- [x] `uv run ruff format --check src tests` → clean
- [ ] Operator: re-run `rc-validate` repro (`ls -ld /var/lib/hal0/model-pull-jobs && sudo -u hal0 test -w ...`) on a fresh install to confirm the end-to-end fix

### DIFF
--- a/src/hal0/install/perms.py
+++ b/src/hal0/install/perms.py
@@ -401,6 +401,64 @@ def ownership_table(
             0o644,
             role="hal0.db-shm (WAL shared-memory index)",
         ),
+        # activity.db — the durable audit trail (``paths.activity_db()`` ->
+        # ``var_lib/"activity.db"``, :class:`hal0.activity.AuditStore`). Same
+        # O13 birth-ownership class as ``hal0.db`` above: no row existed, so a
+        # box where the file's first writer was root (an install-time /
+        # doctor-invoked path) left it unreadable/unwritable to the
+        # User=hal0 daemon with no way for ``doctor perms --fix`` to heal it.
+        # ``AuditStore.__init__`` sets ``PRAGMA journal_mode=WAL``, so the
+        # ``-wal``/``-shm`` siblings need their own rows for the identical
+        # reason the ``hal0.db`` journal siblings do above.
+        PermRow(
+            var_lib / "activity.db",
+            state_owner,
+            service_group,
+            0o644,
+            role="activity.db (durable audit trail)",
+        ),
+        PermRow(
+            var_lib / "activity.db-wal",
+            state_owner,
+            service_group,
+            0o644,
+            role="activity.db-wal (WAL journal)",
+        ),
+        PermRow(
+            var_lib / "activity.db-shm",
+            state_owner,
+            service_group,
+            0o644,
+            role="activity.db-shm (WAL shared-memory index)",
+        ),
+        # model-pull-jobs/ — the durable pull-job snapshot store
+        # (``hal0.registry.pull._pull_jobs_dir()`` -> ``var_lib/"model-pull-jobs"``,
+        # the #626/#MR-1 restart-survival fallback). Same O13 birth-ownership
+        # class as ``models/`` above: the installer's root-run brain-model
+        # pull (bundle-tier auto-pull) calls ``persist_pull_job`` -> lazy
+        # ``path.parent.mkdir(parents=True, exist_ok=True)`` as root on a
+        # fresh install, so the dir is born ``root:root 0755`` with a
+        # root-only ``0600`` snapshot inside. ``hal0-api`` (``User=hal0``)
+        # then fails every pull-job snapshot write
+        # (``model.pull_job_persist_failed``, WARNING-only fail-soft — a
+        # write failure here must never break the pull itself) and cannot
+        # read the existing one, so ``GET /api/models/<id>/pull/status``
+        # 404s after a restart even though the pull succeeded (#1895). No
+        # row meant ``doctor perms --fix`` could not heal it either.
+        # ``persist_pull_job`` writes each snapshot via ``tempfile.mkstemp``
+        # + ``os.replace`` — mkstemp always creates its tempfile ``0600``
+        # regardless of umask, matching the ``registry.toml`` /
+        # ``slots/*/state.json`` convention above.
+        PermRow(
+            var_lib / "model-pull-jobs",
+            state_owner,
+            service_group,
+            0o2775,
+            glob="*.json",
+            child_mode=0o600,
+            optional=False,
+            role="model-pull-jobs/ (durable pull-job snapshots)",
+        ),
         # registry/ — the model registry, also born root:root from the same
         # install.sh mkdir and also written by the User=hal0 daemon. Same O13
         # birth-ownership class as slots/ above; heal the dir. registry/ is
--- a/tests/install/test_perms.py
+++ b/tests/install/test_perms.py
@@ -764,3 +764,101 @@ def test_upstreams_toml_is_not_world_readable(tmp_hal0_home: str) -> None:
     assert (row.owner, row.group) == ("hal0", "hal0")
     assert row.mode & 0o007 == 0, f"upstreams.toml is world-readable: {row.mode:o}"
     assert row.mode == 0o640
+
+
+# ── hal0-api write-path coverage (#1895) ───────────────────────────────────────
+#
+# ``hal0-api`` runs ``User=hal0``. Every ``var_lib()``-rooted path it can be
+# the FIRST writer of on a fresh install must carry a row here, or a
+# root-run installer step (a bundle-tier model pull, a schema migration, a
+# health check — anything that touches the path before the daemon's first
+# request) leaves it root-owned with no way for `doctor perms --fix` to
+# heal it (the O13 class: #1546 for hal0.db, #1895 for model-pull-jobs +
+# activity.db). This table is the durable form of that check: add a row
+# HERE whenever a new lazily-created ``var_lib()`` path is wired into
+# ``hal0-api``, matching the row(s) added to ``ownership_table`` in the
+# same change, so a future writer can't be missed the way model-pull-jobs
+# was.
+# Labels only at collection time — the concrete ``Path`` each label resolves
+# to depends on ``HAL0_HOME``, which ``tmp_hal0_home`` sets PER TEST FUNCTION
+# (after collection already ran). Resolving paths here instead of inside the
+# test body would race that fixture: the parametrize list would freeze
+# whatever ``HAL0_HOME`` was active at collection time, while
+# ``ownership_table()`` inside the test resolves against the per-test tmp
+# dir — two different tmp trees that would never compare equal.
+_HAL0_API_WRITE_LABELS = [
+    # primary database + WAL siblings (#1546)
+    "hal0.db",
+    "hal0.db-wal",
+    "hal0.db-shm",
+    # durable audit trail + WAL siblings (#1895)
+    "activity.db",
+    "activity.db-wal",
+    "activity.db-shm",
+    # durable pull-job snapshot store (#1895, the #626/#MR-1 fallback)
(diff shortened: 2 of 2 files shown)


thinmintdev:
## Independent review — APPROVE (0 blocking, 4 non-blocking notes)

I did not write this code and verified every claim in the PR body against the tree rather than taking it on trust.

### 1. Does the diff fix #1895's exact repro?

Yes. Traced the whole path:

- `hal0.registry.pull.persist_pull_job` (pull.py:442) does `path.parent.mkdir(parents=True, exist_ok=True)` then `tempfile.mkstemp(dir=path.parent)` — lazy, umask-driven birth, exactly as the issue describes.
- It is reached from `run_pull` (pull.py:1166, 1662), which the installer's brain-model pull enters as root via `hal0.install.brain_model` → `hal0.install.orchestrate.run_pull_and_activate` (brain_model.py:200). So root really is the first writer on a fresh install.
- `install/perms.py` had no row for `var_lib/"model-pull-jobs"`, so `plan()`/`commit()` had nothing to reconcile — the missing-declaration root cause, not a symptom.

After the change the backstop chowns the dir to `hal0:hal0 2775` and any existing `*.json` snapshot to `hal0:hal0 0600`, which is precisely what `sudo -u hal0 test -w /var/lib/hal0/model-pull-jobs` and the post-restart `GET /api/models/<id>/pull/status` need.

### 2. Root cause vs symptom

Fix is at the cause (the declarative table is the single source of truth; the O13/#1546 class), not a one-off `chown` bolted into `install.sh`. Correct call.

### 3. Revert-and-confirm-red (independently run, not taken from the PR body)

Detached worktree at `034c1633`, `git checkout HEAD~1 -- src/hal0/install/perms.py` (tests left at PR state):

```
FAILED tests/install/test_perms.py::test_every_hal0_api_write_path_has_an_ownership_row[activity.db]
FAILED tests/install/test_perms.py::test_every_hal0_api_write_path_has_an_ownership_row[activity.db-wal]
FAILED tests/install/test_perms.py::test_every_hal0_api_write_path_has_an_ownership_row[activity.db-shm]
FAILED tests/install/test_perms.py::test_every_hal0_api_write_path_has_an_ownership_row[model-pull-jobs/]
FAILED tests/install/test_perms.py::test_model_pull_jobs_dir_snapshot_files_get_hal0_owned_child_rows
5 failed, 6 passed, 34 deselected in 0.16s
```

Exactly the 5 parametrized failures claimed. Restoring `perms.py`:

- `tests/install` → **259 passed**
- `tests/installer` → **489 passed, 1 skipped** (shellcheck absent)
- other perms-touching suites (`tests/agents/test_hermes_security_deliverables.py`, `tests/config/test_loader.py`, `tests/security/test_api_env_mode.py`) → **89 passed**
- `ruff format --check src tests` → `1147 files already formatted`; `ruff check` on both changed files → `All checks passed!`

The tests fail for the right reason (missing table row / `KeyError` on the row lookup), not incidentally.

### 4. Modes

- `model-pull-jobs/` `2775` matches the sibling runtime-tree convention (`slots/`, `registry/`, `models/` are all `0o2775` setgid with `state_owner`/`service_group`).
- `child_mode=0o600` for `*.json` matches the actual birth mode: `persist_pull_job` uses `tempfile.mkstemp`, which is always `0600` regardless of umask, then `os.replace`. Same rationale already documented for `registry/registry.toml` and `slots/*/state.json`. Only the daemon reads the snapshots, so `0600` is functionally sufficient too — the row matches the writer instead of fighting it.
- `activity.db{,-wal,-shm}` `0644` mirrors the `hal0.db` trio byte-for-byte, and `0644` is `sqlite3.connect`'s birth mode with a single writer. `paths.activity_db()` (config/paths.py:184) resolves to `var_lib/"activity.db"`, so the row targets are right.

### 5. `runner-image-pull-jobs/` out of scope

Verified, claim holds. `_pull_jobs_dir()` for runner images is `runner_pull.py:126`; its only caller chain is `registry/runner_pull_jobs.py` → `api/routes/runner_images.py` (plus the API lifespan shutdown hook). Nothing in `installer/install.sh`, the CLI, or any root-run install step touches it, so it is born under `User=hal0`. Leaving it out is defensible.

### 6. `install.sh` needs no edit

Verified the ordering directly:

- brain-model pull: `installer/install.sh:1993` (`python -m hal0.install.brain_model`), inside the `DEV_MODE -eq 0` branch
- backstop: `installer/install.sh:2809` (`"${HAL0_BIN}" doctor perms --fix --force`), also gated on `DEV_MODE -eq 0`

Pull strictly precedes the backstop, and both are skipped together in dev mode, so the table-only fix is sufficient. Claim confirmed.

### 7. Scope discipline

Diff is confined to `src/hal0/install/perms.py` (rows only, appended in the existing var_lib block) + `tests/install/test_perms.py`. No `install.sh`, no `STATE.md`/`secrets/`/recursive-mkdir churn — #1896 should rebase cleanly.

---

### Non-blocking notes

1. **The "durable coverage" test does not enforce the invariant its docstring claims.** `_HAL0_API_WRITE_LABELS` is a hand-maintained list, so it can only fail for paths someone already remembered to add — i.e. it cannot catch the next unrowed path, which is the exact failure mode it says it prevents. Concretely, these `var_lib()` paths are written by `hal0-api` today, have **no** ownership row, and the new test stays green anyway: `update-jobs/` (`api/routes/updater.py:221`), `dashboard-layout.json` (`dashboard/layout_store.py:36`), `gpu_arbiter.json` (`slots/manager.py:3289`), `stacks/state.json` (`config/paths.py:490`), `state/agents/<name>/provision.json`, `runner-image-pull-jobs/`. Suggest either softening the docstring to what it actually is (a regression anchor for known paths) or deriving the list from a real registry of write paths. Not a blocker for #1895.

2. **`optional=False` is inert.** `PermRow.optional` is never read by `_expand_row`, `plan`, `commit`, or `audit_rows` — absent paths are already never `changed` (`PermDiff.changed` short-circuits on `before.exists`) and are reported as a dim `absent` row. Setting it matches the sibling rows' style, which is fine, but `test_model_pull_jobs_dir_snapshot_files_get_hal0_owned_child_rows` asserting `row.optional is False` is asserting a no-op field. Worth knowing before anyone reads the comment as load-bearing.

3. **`regressions.yaml` entry left untouched.** The file header says a repro promoted into pytest should `set promoted_to: and drop it in the same PR`, and `model-pull-jobs-root-owned` is `promote_ready: true` with "Promote: a pytest over the perms table covering every path hal0-api writes" — which is what this PR adds. The new test only covers the table half (not the on-box restart/404 half), so keeping the entry is arguably correct — but that decision should be explicit (either set `promoted_to:` or say why the entry stays).

4. **Nit, pre-existing:** `mkstemp` temp files are named `.<id>.json.XXXX.tmp`, so they match neither the new `*.json` perms glob nor `sweep_pull_jobs`'s `*.json` glob. A crashed root-run pull can leave a root-owned `0600` stray that nothing heals or reaps. Cosmetic only (the daemon can unlink it once it owns the dir); out of scope here.

Verified with `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest ...` in an isolated detached worktree at `034c1633`.

Issue #1924, opened 2026-08-17
update: migrate existing device="gpu-vulkan" slot TOMLs to gpu-rocm — post-#1923 they refuse to load


thinmintdev:
Follow-up to #1888 / PR #1923 (Vulkan lane retirement).

PR #1923 deliberately does not migrate slot TOMLs already on disk: an existing install with `device = "gpu-vulkan"` slots will refuse to load after updating, with the new preflight message, until the operator relabels them. Quadlet units rendered by the new code carry `AddDevice=/dev/kfd`, so only pre-fix units are exposed.

Proposal: an updater migration that relabels `gpu-vulkan` to `gpu-rocm` on AMD hosts where /dev/kfd is present (and to `cpu` with a warning where it is not), with a journal breadcrumb. Needs to respect the #1867 rails (scope migrations narrowly, log every mutation).

Raised as open question 3 in PR #1923's report. Refs #1888.

thinmintdev:
Fixed by PR #1934 (merged as 99024df2): post-activation migration relabels gpu-vulkan to gpu-rocm (kfd present) or cpu (loud warning), scoped to llama.cpp-backed slots via _spec_provider_for — non-llama Vulkan slots (kokoro/whisper/ComfyUI) survive byte-identical. Review approved 0-blocking after one revision. The non-llama kfd-absent guard over-fire is #1941.

Issue #1926, opened 2026-08-18
ci: γ-suite timeout-minutes:12 has no real headroom — green runs hit 10-11.3 min and slow runners randomly block PRs


thinmintdev:
Found during the rc.6 fix-wave merge train. `.github/workflows/playwright.yml:26` sets `timeout-minutes: 12` with a comment claiming 8-min budget + headroom, but observed green runs are 10m08s-11m16s (85-95% of cap). γ-suite (chromium) is a required status check, so any slow runner trips the cap and blocks the PR — this cost 3 attempts on PR #1913 (two kills at 12m15s mid-run with 0 failure markers, 576/653 and 578/653 progressing; third attempt green at 8m26s; measured per-test pace showed a 40% slower runner). Raise the cap or shard the suite.

PR #1934, opened 2026-08-18
fix(updater): migrate gpu-vulkan slot TOMLs to gpu-rocm/cpu

thinmintdev:
## Summary

PR #1923 retired the Vulkan LLM lane (#1888) and made `container.py`'s `load_sync` refuse to load a GPU slot whose `device` is `gpu-vulkan` or `gpu-rocm` when `/dev/kfd` is not visible. It deliberately relabeled only the installer's *seed* slot TOMLs (`installer/etc-hal0/slots/*.toml`) from `gpu-vulkan` to `gpu-rocm`, and explicitly left slot TOMLs already materialised on an existing install untouched — so any pre-#1923 install with `device = "gpu-vulkan"` slots refuses to load them after updating, until an operator manually relabels the TOML (raised as open question 3 in #1923's report, filed as #1924).

This PR adds an updater migration that does that relabel automatically.

## Design

`relabel_stale_vulkan_slots` is a new sixth pass in `run_post_activation_migrations` (`src/hal0/updater/updater.py`) — the single sequence both `Updater.commit()` (self-update) and `install.sh`'s repair/upgrade-in-place path call, so both upgrade routes converge on the same on-disk state.

- **Trigger for `gpu-rocm`**: `/dev/kfd` present and usable (`hal0.providers._gpu.kfd_present()`, added in #1923) — matches PR #1923's own seed relabel exactly.
- **Trigger for `cpu`**: `/dev/kfd` not present. This is a genuine operator-visible behavior change (the slot drops off the GPU and gets much slower), so it is logged as its own distinct WARNING-level breadcrumb (`updater.slot_vulkan_relabeled_cpu_fallback`, note prefixed `BEHAVIOR CHANGE:`) — never folded silently into the routine relabel log line (`updater.slot_vulkan_relabeled_rocm`).
- **AMD-host scope**: only fires when the amdgpu kernel driver is bound on this host (`hal0.providers._gpu.host_is_amd_gpu()`), exactly matching `require_kfd_for_gpu_slot`'s own scope. A non-AMD host (Intel iGPU, NVIDIA without CDI) has no `/dev/kfd` by design, was never gated by #1923's guard, and is a full no-op here — no relabel, no log.
- **Runtime scope (llama.cpp-backed slots only)**: `device = "gpu-vulkan"` is not exclusively an llama.cpp label — `capabilities/catalog.py` deliberately keeps it for the non-llama runtimes (Kokoro TTS, whisper.cpp/Moonshine STT, ComfyUI), which run genuinely-Vulkan images. Every candidate slot is resolved through `hal0.providers.container._spec_provider_for` — the same discriminator `load_sync` itself uses — and is skipped ENTIRELY (no relabel, no warning, no log, on either kfd axis) unless it resolves to `None` (the default llama-server GPU provider). The kfd-absent non-llama case (`require_kfd_for_gpu_slot` itself over-firing for non-llama runtimes) is a separate, already-filed bug tracked as #1941 and is deliberately NOT addressed here.
- **Narrow scope (#1867 rails)**: raw `tomllib` load / `write_toml_atomic` dump, same convention as the existing `retag_stale_slot_images` / `clear_stale_mtp_overrides` passes in this module. Only the `device` key's *value* is ever changed (top-level or nested under `[slot]`, matching the two TOML shapes those passes already handle); every other field, and every slot whose `device` is not literally `"gpu-vulkan"` or that isn't llama.cpp-backed, is left untouched. Note this is narrow-scope-by-key, not byte-for-byte file preservation — like every other pass in this module, a touched file's comments are dropped on rewrite (`write_toml_atomic` reserializes the whole parsed document via `tomli_w`).
- **Idempotency**: once a slot's `device` reads `"gpu-rocm"` or `"cpu"` it no longer matches the `"gpu-vulkan"` guard, so a second run touches nothing and emits no new log lines (covered by `test_relabel_is_idempotent`, `test_relabel_is_idempotent_on_cpu_fallback`, `test_idempotent_second_run_logs_nothing_new`).
- **Journal breadcrumb**: every mutation logs `slot`, `old` (`"gpu-vulkan"`), `new` (`"gpu-rocm"` or `"cpu"`), and `job_id`, via `structlog` `log.warning` — same shape as the other post-activation passes.
- Non-fatal in the aggregate sequence, like the other four data-cleanup passes: a failure here logs and does not block the other migrations or the caller's activation.

This should ship in the **same release as #1923** — otherwise any pre-#1923 install with `gpu-vulkan` LLM slots hits the new load-time refusal with no automatic recovery path.

## Test plan

- [x] `tests/updater/test_vulkan_slot_migration.py` (new file, 22 tests) — TDD throughout, including the runtime-scope fix (RED first against the pre-fix code, confirmed the non-llama tests failed, then implementation). Covers: kfd-present → gpu-rocm relabel (flat + nested `[slot]` shape), kfd-absent → cpu relabel with a loud distinct breadcrumb, AMD-host scoping (non-AMD host is a full no-op, real-probe default), runtime scoping (Kokoro/ComfyUI/transcription-shaped slots survive completely untouched on both kfd axes, alongside a genuine llama.cpp slot that still relabels), narrow scope (other keys/slots untouched), unreadable-TOML is skipped not fatal, idempotent re-run (both branches, including "logs nothing new"), and the default (no override) probes the real `kfd_present()`/`host_is_amd_gpu()`.
- [x] `tests/updater/test_post_activation_migrations.py` (4 existing tests edited, stub signature updated for `amd_host`) to stub and assert the new sixth pass runs in order, alongside the others, and is skipped when the schema migration fails first.
- [x] `installer/install.sh`'s post-activation-migrations transcript/log updated to name the vulkan-slot pass alongside the other four.
- [x] `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/updater tests/install -q` → 538 passed
- [x] `uv run ruff format --check src tests` → all formatted
- [x] `make lint` → all checks passed
- [x] `python scripts/check_sunset.py` → scar markers at baseline

Refs #1924 #1888. Related: #1941 (tracks `require_kfd_for_gpu_slot` itself over-firing for non-llama Vulkan runtimes on a kfd-absent AMD host — out of scope for this migration).

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -2417,22 +2417,27 @@ if status["due"]:
 PYEOF
 fi
 
-# ── post-activation migrations (schema + seed-profile/mtp/image-pin/extra-args) ─
+# ── post-activation migrations (schema + seed-profile/mtp/vulkan/image-pin/extra-args) ─
 # One shared sequence with Updater.commit()'s self-update path (GH #1475):
 # hal0.toml schema migrations, profiles.toml virtual-seed pruning (seeds live
 # in code and are overlaid on every load — an older install that materialised
 # them into profiles.toml froze stale definitions; this prunes them so the
 # code definition wins again), stale mtp=true slot-override clearing (a
 # force-on pointing at a model with no MTP heads crashes llama-server at load
-# once the unit re-renders under post-separation code), stale runner-image
-# pin retagging, and defaults.extra_args sanitizing. Every pass after the
-# schema migration is independently best-effort — a single pass failing never
-# aborts the others. Previously this block ran only the seed-profile and
-# stale-MTP passes directly, so a box upgraded by re-running install.sh (the
-# documented repair/upgrade path) kept a stale meta.schema_version, stale
-# runner-image pins, and unsanitised defaults.extra_args that `hal0 update`
-# would have fixed — two boxes on the same version with different on-disk
-# state. Operator edits and non-seed profiles are always left untouched.
+# once the unit re-renders under post-separation code), retired
+# device="gpu-vulkan" slot relabeling (#1924 — llama.cpp-backed GPU slots
+# left over from before #1923 retired the Vulkan LLM lane now refuse to load
+# under the new /dev/kfd preflight guard; relabeled to gpu-rocm/cpu, AMD
+# hosts only, non-llama runtimes like Kokoro/ComfyUI left untouched), stale
+# runner-image pin retagging, and defaults.extra_args sanitizing. Every pass
+# after the schema migration is independently best-effort — a single pass
+# failing never aborts the others. Previously this block ran only the
+# seed-profile and stale-MTP passes directly, so a box upgraded by
+# re-running install.sh (the documented repair/upgrade path) kept a stale
+# meta.schema_version, stale runner-image pins, and unsanitised
+# defaults.extra_args that `hal0 update` would have fixed — two boxes on the
+# same version with different on-disk state. Operator edits and non-seed
+# profiles are always left untouched.
 #
 # Ran through hal0_migration_step (not a bare heredoc): the schema migration
 # used to be able to abort this whole script under set -euo pipefail before
@@ -2445,7 +2450,7 @@ fi
 if [[ "${DEV_MODE}" -eq 1 ]]; then
     info "dev mode — skipping post-activation migrations (no system writes)"
 else
-    info "running post-activation migrations (schema + seed-profile/mtp/image-pin/extra-args)"
+    info "running post-activation migrations (schema + seed-profile/mtp/vulkan-slot/image-pin/extra-args)"
     hal0_migration_step "post-activation migrations" <<'PYEOF'
 from hal0.config.migrations.v2 import PROFILE_CATALOG_SCHEMA_VERSION
 from hal0.updater.updater import run_post_activation_migrations
@@ -2459,7 +2464,7 @@ if target != source:
     print(f"  hal0.toml schema migrated {source} -> {target}")
 else:
     print(f"  hal0.toml schema already at {target}")
-print("  seed-profile / stale-MTP / runner-image / extra-args cleanup passes ran (see log for any per-item detail)")
+print("  seed-profile / stale-MTP / vulkan-slot / runner-image / extra-args cleanup passes ran (see log for any per-item detail)")
 PYEOF
 fi
 
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -2864,7 +2864,7 @@ def retag_stale_slot_images(*, job_id: str | None = None) -> int:
 def relabel_stale_vulkan_slots(
     *, job_id: str | None = None, kfd_present: bool | None = None
 ) -> int:
-    """Relabel legacy ``device = "gpu-vulkan"`` slot TOMLs (upgrade migration).
+    """Relabel retired ``device = "gpu-vulkan"`` slot TOMLs (upgrade migration).
 
     #1888/PR #1923 retired the Vulkan LLM lane and switched ``load_sync`` to
     hard-refuse loading a GPU slot when the ROCm compute node (``/dev/kfd``)
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -2895,7 +2895,8 @@ def relabel_stale_vulkan_slots(
     On an AMD host, for every slot TOML with a literal ``device =
     "gpu-vulkan"`` (top-level or nested under ``[slot]``, the same two-shape
     handling :func:`retag_stale_slot_images` and
-    :func:`clear_stale_mtp_overrides` use), relabel ``device`` to:
+    :func:`clear_stale_mtp_overrides` use) that is ALSO llama.cpp-backed
+    (see runtime scope below), relabel ``device`` to:
 
     * ``"gpu-rocm"`` — when :func:`hal0.providers._gpu.kfd_present` reports
       the ROCm compute node present and usable. This is the same target PR
@@ -2906,12 +2907,40 @@ def relabel_stale_vulkan_slots(
       explicitly, so this is logged as its own WARNING-level breadcrumb,
       never folded silently into the routine "migrated" line.
 
-    Only the ``device`` key is ever written — narrow scope per the #1867
-    rails (no other field, comment, or slot is touched, and a slot whose
-    ``device`` is not literally ``"gpu-vulkan"`` is skipped entirely).
+    Runtime scope — llama.cpp-backed slots ONLY: ``device = "gpu-vulkan"``
+    is not exclusively an llama.cpp label. ``capabilities/catalog.py``
+    deliberately keeps it for the non-llama runtimes (Kokoro TTS,
+    whisper.cpp/Moonshine STT, ComfyUI), which run genuinely-Vulkan images —
+    #1923's Vulkan-lane retirement and its ``require_kfd_for_gpu_slot`` guard
+    are about llama.cpp's unified ROCmFPX runner specifically, not those.
+    Relabeling a non-llama slot here would be actively wrong on both kfd
+    axes: ``gpu-rocm`` is a label its image can't honor (and
+    ``gpu_visibility_env`` would silently swap
+    ``GGML_VK_VISIBLE_DEVICES`` for ``HIP_VISIBLE_DEVICES``, dropping any
+    ``gpu_index`` pin), and ``cpu`` would strip ``/dev/dri`` from a slot that
+    was working fine. So every candidate slot is resolved through
+    :func:`hal0.providers.container._spec_provider_for` — the same
+    authoritative runtime-family discriminator ``load_sync`` itself uses —
+    and skipped ENTIRELY (no relabel, no warning, no log line, on either kfd
+    axis) unless it resolves to ``None`` (the default llama-server GPU
+    provider). The kfd-absent non-llama case (``require_kfd_for_gpu_slot``
+    itself over-firing for non-llama runtimes) is a separate, already-filed
+    bug (#1941) and is deliberately NOT addressed by this migration.
+
+    Only the ``device`` key is ever changed in VALUE — narrow scope per the
+    #1867 rails (no other field or slot is touched, and a slot whose
+    ``device`` is not literally ``"gpu-vulkan"``, or that isn't
+    llama.cpp-backed, is skipped entirely). Note that "narrow scope" is
+    about which *keys* get new values, not byte-for-byte file preservation:
+    like every other pass in this module, the write goes through
+    :func:`~hal0.config.loader.write_toml_atomic`, which reserializes the
+    whole parsed document via ``tomli_w`` — so any comments in a touched
+    slot's TOML are dropped on that rewrite, exactly as they are for
+    :func:`retag_stale_slot_images` and :func:`clear_stale_mtp_overrides`.
     Idempotent: once a slot's ``device`` reads ``"gpu-rocm"`` or ``"cpu"`` it
     no longer matches the ``"gpu-vulkan"`` guard, so a second run touches
-    nothing and logs nothing new.
+    nothing and logs nothing
+    new.
 
     Args:
         job_id: Optional breadcrumb for structured-log tracing.
@@ -2932,6 +2961,7 @@ def relabel_stale_vulkan_slots(
     from hal0.config.paths import slots_config_dir
     from hal0.providers._gpu import host_is_amd_gpu as _probe_host_is_amd_gpu
     from hal0.providers._gpu import kfd_present as _probe_kfd_present
+    from hal0.providers.container import _spec_provider_for
 
     is_amd = _probe_host_is_amd_gpu() if amd_host is None else amd_host
     if not is_amd:
@@ -2956,6 +2986,22 @@ def relabel_stale_vulkan_slots(
         elif isinstance(raw.get("slot"), dict) and raw["slot"].get("device") == "gpu-vulkan":
             holder = raw["slot"]
         if holder is None:
+            continue
+        try:
+            provider = _spec_provider_for(holder)
+        except Exception as exc:
+            log.warning(
+                "updater.vulkan_migration_slot_runtime_unresolved",
+                slot=slot_name,
+                error=str(exc),
+            )
+            continue
+        if provider is not None:
+            # A non-llama runtime (Kokoro/Moonshine/ComfyUI/FLM/Qwen3TTS)
+            # genuinely runs a Vulkan image on this device string — #1923's
+            # guard is scoped to llama.cpp's unified ROCmFPX runner, not
+            # these. Leave completely untouched (see docstring; #1941 tracks
+            # the separate kfd-absent over-firing issue for this class).
             continue
         new_device = "gpu-rocm" if have_kfd else "cpu"
         holder["device"] = new_device
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -2961,6 +2961,11 @@ def relabel_stale_vulkan_slots(
     from hal0.config.paths import slots_config_dir
     from hal0.providers._gpu import host_is_amd_gpu as _probe_host_is_amd_gpu
     from hal0.providers._gpu import kfd_present as _probe_kfd_present
+
+    # Private import, deliberate: this must be the EXACT discriminator
+    # load_sync uses to pick a provider, not a re-implementation of it — do
+    # not "clean this up" into a public wrapper or a local re-derivation,
+    # that would risk drifting from the real dispatch logic it mirrors.
     from hal0.providers.container import _spec_provider_for
 
     is_amd = _probe_host_is_amd_gpu() if amd_host is None else amd_host
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -2862,7 +2862,10 @@ def retag_stale_slot_images(*, job_id: str | None = None) -> int:
 
 
 def relabel_stale_vulkan_slots(
-    *, job_id: str | None = None, kfd_present: bool | None = None
+    *,
+    job_id: str | None = None,
+    kfd_present: bool | None = None,
+    amd_host: bool | None = None,
 ) -> int:
     """Relabel retired ``device = "gpu-vulkan"`` slot TOMLs (upgrade migration).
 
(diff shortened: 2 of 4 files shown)


thinmintdev:
Triaged the Codex review comments:

1. **P1 "Restrict relabeling to affected AMD llama.cpp slots"** — valid, fixed in b8640843. The migration was relabeling any `device = "gpu-vulkan"` slot based solely on `kfd_present()`, but `require_kfd_for_gpu_slot` (added in #1923) only ever gates a `gpu-vulkan` slot on an AMD host (`host_is_amd_gpu()`). On Intel/NVIDIA hardware `/dev/kfd` never exists by design, so a working non-AMD Vulkan slot was never broken by #1923, and the unscoped migration would have silently downgraded it to `cpu`. Added the same `amd_host` gate `require_kfd_for_gpu_slot` uses — on a non-AMD host the migration is now a full no-op. New tests cover this (`test_non_amd_host_is_a_full_noop`, `test_non_amd_host_logs_nothing`, `test_default_probes_real_amd_host`).

   On the narrower point about Kokoro/whisper.cpp/ComfyUI slots specifically: those providers don't branch their container spec on the `gpu-rocm` vs `gpu-vulkan` string (they resolve GPU device paths generically via `resolve_gpu_device_paths()`/`resolve_gpu_group_ids()`), and `require_kfd_for_gpu_slot` itself gates *any* `device="gpu-vulkan"` slot on an AMD host regardless of provider type — so on an AMD box without `/dev/kfd`, those slots are equally blocked by #1923's own guard and equally need the relabel. This mirrors PR #1923's own precedent, which relabeled the `rerank`/`embed` seed slots (non-LLM types) the same way.

2. **P1 "Run the new migration from the activated release"** — accurate, but this is a pre-existing architectural property of `run_post_activation_migrations` shared by all four existing passes (`ensure_seed_profiles`, `clear_stale_mtp_overrides`, `retag_stale_slot_images`, `sanitize_model_extra_args`), not something introduced by this PR: `Updater.commit()`'s step 7 (migrations) runs before step 8 (`seam.activate()`, the symlink swap + venv re-pip), using the already-imported *old* process's code. So on the specific `hal0 update` self-update transaction that installs the release containing this fix, it runs one cycle late — the same known limitation every prior post-activation migration has always had. `install.sh`'s repair/upgrade-in-place path is unaffected (it always runs a fresh `python -c` invocation against the already-pip-installed target code). Restructuring `Updater.commit()`'s activation ordering to close this gap for all five passes at once is a materially larger, separately-scoped change (the code has an explicit comment on why migrations currently run before the profile-catalog-reset gate check) — out of scope for this narrowly-scoped #1924 fix. Noting it here rather than silently dropping it; happy to file a tracking issue if the team wants it addressed as its own PR.

3. **P2 "Probe KFD access as the service account"** — real edge case, skipping for this PR. The migration's `kfd_present()` probe runs as whatever user executes the migration (root during `install.sh`, or the `hal0 update` caller's process). The *actual* enforcement (`require_kfd_for_gpu_slot` in `container.py`'s `load_sync`) runs later, in the real hal0-api service process, as the real `User=hal0` account, and re-probes `/dev/kfd` access independently at slot-load time. So a root/service-account permission mismatch (the exact LXC gid-mapping failure mode #1923's own preflight gid-check exists to catch) can't produce a silently-broken slot: if the migration picks `gpu-rocm` optimistically and the service account genuinely can't reach `/dev/kfd`, the slot still fails to load with the same clear `GpuPreflightError` message it would have hit as `gpu-vulkan` — just one hop later, and diagnosable by the same remedy (fix the render/kfd gid mapping). Probing as the service account from inside an install-time/root migration would need a privilege-drop or subprocess-as-user, which is a meaningfully bigger and more invasive change than this narrowly-scoped migration warrants.

thinmintdev:
## Independent review — verdict: **request changes** (1 blocking, 3 non-blocking)

Reviewed at `b8640843` in a detached worktree. Migration-grade scrutiny per the #1867 rails.

### What holds up

- **Issue #1924's requirement is met** for the case it describes: AMD-scoped (`host_is_amd_gpu`), `gpu-rocm` when `kfd_present()` else `cpu`, distinct per-mutation journal breadcrumbs, idempotent, wired as a best-effort pass in `run_post_activation_migrations` so both `Updater.commit()` and install.sh's repair path converge.
- **Only `device` is ever written.** Verified by reading: `holder` is bound only from a literal `device == "gpu-vulkan"` (top-level or `[slot]`-nested), and the sole assignment is `holder["device"]`. Same two-shape convention as `retag_stale_slot_images` / `clear_stale_mtp_overrides`.
- **`write_toml_atomic` semantics** — yes, it fully reserializes through `tomli_w`, so key *values* survive but comments/formatting do not. That is identical to the tradeoff `retag_stale_slot_images` and `config/migrations/hw_slot_ownership.py` already accept on the same files, so it is precedent, not a new regression (but see finding 2). Ownership is not a hazard here: the pass runs in-process as the service account under `commit()`, and install.sh's root-run path has the P3-perms backstop after the migration block.
- **Triage item 2 (pre-swap gap) is genuinely pre-existing.** `Updater.commit()` step 7 calls `run_post_activation_migrations` before `seam.activate()`, and *all five* prior passes live inside that same function — there is no other call site except install.sh's `python -c` block (which runs against already-pip-installed target code). Documented-not-fixed is the right call here; it should get its own issue.
- **Triage item 3 (kfd probe as caller vs service user)** — reasoning checks out. `require_kfd_for_gpu_slot` re-probes independently in `container.py:load_sync` as `User=hal0` at slot-load time, so an optimistic `gpu-rocm` can only ever produce the same loud `GpuPreflightError`, never a silent bad lane.
- **Interaction with #1923's guard after a `cpu` relabel:** verified safe. `require_kfd_for_gpu_slot` returns immediately for `device="cpu"` (not `gpu-vulkan`, not `gpu-rocm`), and `_effective_backend_and_device_class` treats `device` as sole truth, so the slot renders as a CPU lane and loads on a kfd-less box.
- **Idempotency test** proves both halves: return `0` **and** `len(calls)` unchanged on the re-run (`test_idempotent_second_run_logs_nothing_new`). "No writes" is structural rather than asserted (the `holder is None: continue` precedes the write) — acceptable.

### Revert-and-confirm-red

Six mutations against the merge head, running `tests/updater/test_vulkan_slot_migration.py` + `tests/updater/test_post_activation_migrations.py` (19 collected — note the PR description's "24 new tests" is 13 new + edits to 3 existing):

| Mutation | Result |
|---|---|
| `is_amd = True` (drop AMD scope) | **3 failed** / 16 passed |
| kfd-absent branch → `gpu-rocm` instead of `cpu` | **3 failed** / 16 passed |
| drop the nested `[slot]` shape handler | **1 failed** / 18 passed |
| match any `device` (break idempotency) | **4 failed** / 15 passed |
| cpu fallback logs the routine rocm breadcrumb (silent behavior change) | **1 failed** (`test_cpu_fallback_logs_a_loud_warning`) |
| remove the pass from `run_post_activation_migrations` | **2 failed** / 17 passed |

Baseline restored → 19 passed. Full suite `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/updater tests/install -q` → **531 passed**. The tests are load-bearing, not decorative.

---

### 1. BLOCKING — the migration relabels non-llama.cpp `gpu-vulkan` slots, which are a *live, currently-advertised* configuration

`gpu-vulkan` was retired for the **llama.cpp** lane only. `capabilities/catalog.py` says so explicitly, in the comment right above the suppression this PR builds on (L429-431):

> `gpu-vulkan` survives for the non-llama runtimes below (kokoro / whisper.cpp / ComfyUI), which run genuinely-Vulkan images, and for non-AMD GPUs.

and `_RUNTIME_TO_HOST_BACKENDS` (L73-79) still fans whisper.cpp / kokoro / vibevoice / ComfyUI out to `gpu-vulkan` on an AMD host today. So a whisper.cpp or Kokoro-GPU slot created by the picker *this week* on a Strix box carries `device = "gpu-vulkan"` legitimately. This migration rewrites all of them:

- **AMD host, kfd present** (the ordinary Strix Halo box): that slot loads fine today — `require_kfd_for_gpu_slot` returns early because kfd is present. #1923 never broke it. The migration rewrites it to `gpu-rocm` anyway: a label its runtime image cannot honour, outside the device list its own picker offers, and it silently swaps `gpu_visibility_env`'s `GGML_VK_VISIBLE_DEVICES` for `HIP_VISIBLE_DEVICES` on any slot with a `gpu_index` pin (`providers/_gpu.py:365-368`), so the multi-GPU pin is quietly lost on a Vulkan runtime.
- **AMD host, kfd absent**: those slots *are* currently refused (the guard is provider-agnostic — `load_sync` passes only the `device` string), so the author's triage is right that they need *something*. But `cpu` is the wrong something: whisper.cpp's and ComfyUI's Vulkan images work fine without `/dev/kfd`, and because `_effective_backend_and_device_class` makes `device` the sole hardware truth, `cpu` strips `/dev/dri` passthrough entirely. A working GPU image-gen / STT slot becomes a CPU one. The `BEHAVIOR CHANGE` breadcrumb makes it loud, but the #1867 rails are about not making the mutation in the first place, not only about logging it.

The triage on Codex's P1 answered the AMD/non-AMD axis but not this one: "those slots are equally blocked by #1923's own guard" is true only in the kfd-absent branch, and the PR applies the relabel in both.

Suggested resolution — either:
- **(a)** scope the pass to llama.cpp-backed slots (the actual #1888 defect surface: llama-server provider / `llm`+`embedding`+`reranking` types), leaving genuinely-Vulkan runtimes alone; and file the "#1923's guard over-fires on non-llama providers" case as its own issue, since demoting them to `cpu` is a workaround for an over-broad guard rather than a fix; or
- **(b)** if the team wants (b) anyway, at minimum drop the `gpu-rocm` branch for non-llama runtimes — relabeling a slot that was never refused is a pure unforced mutation.

Either way this needs a test asserting a whisper.cpp/ComfyUI-shaped `gpu-vulkan` slot's outcome, since today's suite has none.

### 2. Non-blocking — the docstring overstates what survives the rewrite

> Only the `device` key is ever written … (no other field, **comment**, or slot is touched…)

`write_toml_atomic` dumps the whole `tomllib`-parsed dict through `tomli_w`: **every comment and all formatting in the file are destroyed**, and the shipped seed slot TOMLs (`installer/etc-hal0/slots/*.toml`) are comment-heavy — including the #1888 rationale comments an operator would want to still be there after this exact migration. It matches the existing convention and I am not asking to change the mechanism, but a migration docstring claiming comments are untouched is the kind of claim the #1867 rails exist to keep honest. Please reword (and the PR body's "narrow scope" bullet with it).

### 3. Non-blocking — install.sh's operator-facing text was not updated

`installer/install.sh:2448` and `:2462` still enumerate "seed-profile / stale-MTP / runner-image / extra-args", and the block comment at 2426-2435 still describes four passes. An operator running the documented repair/upgrade path gets a transcript that never mentions the vulkan pass — the one pass in the sequence that can make a slot slower.

### 4. Nit — stub signature drift

`tests/updater/test_post_activation_migrations.py`'s `_vulkan_migration(*, job_id=None, kfd_present=None)` omits `amd_host`. Harmless today (production passes only `job_id`), but the stub no longer mirrors the real signature.

---

### CI

`python (3.12)`, `ui`, `sunset`, CodeQL/Analyze all green. `γ-suite (chromium)` pending — known #1926 flake, a rerun is in flight; nothing in this diff touches the UI surface.

thinmintdev:
Addressed all four findings from the changes-requested review, pushed in f5155430.

## Blocking — runtime scoping

Fixed. The migration now resolves every `device = "gpu-vulkan"` candidate slot through `hal0.providers.container._spec_provider_for(holder)` — the same authoritative runtime-family discriminator `load_sync` itself uses to pick a provider (`None` = the default llama-server GPU provider; non-`None` = Kokoro/Moonshine/ComfyUI/FLM/Qwen3TTS). A slot only gets relabeled when `_spec_provider_for` returns `None`; otherwise it's skipped **entirely** — no relabel, no warning, no log line, on either kfd axis. If `_spec_provider_for` itself raises (an unrecognized `runtime_family`), the slot is also left untouched and a distinct `updater.vulkan_migration_slot_runtime_unresolved` warning is logged (conservative: we can't safely conclude it's llama.cpp-backed, so we don't touch it).

Per your note, the kfd-absent non-llama case (i.e. `require_kfd_for_gpu_slot` itself over-firing and blocking a genuinely-Vulkan Kokoro/ComfyUI/whisper.cpp slot on an AMD box with no `/dev/kfd`) is explicitly **not** touched by this migration — that's #1941's problem, not this one's.

New TDD coverage in `tests/updater/test_vulkan_slot_migration.py`: `test_kokoro_slot_survives_untouched_kfd_present/absent`, `test_comfyui_slot_survives_untouched_kfd_present/absent`, `test_transcription_slot_survives_untouched`, `test_non_llama_slots_log_nothing_on_either_kfd_axis`, and `test_llama_slot_still_relabels_alongside_untouched_non_llama_slot` (a genuine llama.cpp slot in the same directory as a Kokoro slot — the llama slot relabels, its neighbor doesn't). Confirmed these 7 tests were RED against the pre-fix code (kokoro/comfyui/transcription-shaped slots were being wrongly relabeled) before implementing the `_spec_provider_for` gate.

## Non-blocking

2. **Comment-preservation claim** — fixed. The docstring's "no other field, comment, or slot is touched" line implied byte-level preservation, which is false (`write_toml_atomic` reserializes the whole document via `tomli_w`, same as every other migration in this module). Reworded to "narrow scope by key, not byte-for-byte file preservation," with an explicit note that comments are dropped on any touched file's rewrite — matching `retag_stale_slot_images`/`clear_stale_mtp_overrides`'s actual (undocumented but real) behavior. Also corrected the equivalent claim in the PR body.

3. **install.sh transcript** — fixed. `installer/install.sh`'s post-activation-migrations block (comment header, `info` line, and the printed summary line) now names the vulkan-slot pass alongside seed-profile/mtp/image-pin/extra-args, in the same order `run_post_activation_migrations` actually runs them.

4. **Test stub signature** — fixed. `test_post_activation_migrations.py`'s `_vulkan_migration` stub now accepts `amd_host` (added `*, job_id=None, kfd_present=None, amd_host=None`), matching the real function's signature after the earlier AMD-scoping fix (b8640843).

5. **PR body test-count claim** — corrected. The PR body now states the actual counts: `tests/updater/test_vulkan_slot_migration.py` is a new file with 22 tests (all new relative to `main`, including the 7 runtime-scope tests added in this round); `tests/updater/test_post_activation_migrations.py` has 4 existing tests edited (one renamed `test_runs_all_five_passes` → `test_runs_all_six_passes`), no new test functions there.

## Verification

- `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/updater tests/install -q` → **538 passed**
- `uv run ruff format --check src tests` → all formatted
- `make lint` → all checks passed
- `python scripts/check_sunset.py` → scar markers at baseline (192 ≤ 192)
- `bash -n installer/install.sh` → syntax OK

Not merging or enabling automerge — awaiting review.

thinmintdev:
## Re-review of `f5155430` — verdict: **approve**, 0 blocking (3 nits, none gating)

Re-reviewed in a fresh detached worktree at `f5155430`. The blocking finding is genuinely fixed, not papered over.

### Blocking finding — resolved

`_spec_provider_for` is the right discriminator: it is literally the function `load_sync` calls to pick a provider, so the migration's "is this llama.cpp-backed" question is answered by the same authority that decides what actually launches. Verified by reading `container.py:1123-1177`:

- `RuntimeFamily` is a closed literal (`llama-server, flm, kokoro, qwen3tts, moonshine, comfyui`); every non-llama member has a dispatch branch returning a provider instance, so `None` really does mean "the default llama-server GPU provider".
- The gate catches the **profile-family** path, not only the `type` string — I probed a `type = "llm"` slot carrying `profile = "qwen3-tts"`: skipped, `device` unchanged. That is the case the new tests don't cover (they exercise `type=tts|image|transcription`), and it works.
- whisper.cpp has no runtime family of its own; a whisper-shaped slot resolves through `family == "moonshine" or slot_type == "transcription"` → Moonshine provider → skipped. Covered.
- No false-skip for real llama slots: `llm` / `embedding` / `reranking` types with no special profile fall through to `None` and still relabel (`test_llama_slot_still_relabels_alongside_untouched_non_llama_slot` proves it with a Kokoro neighbor in the same directory).

**Exception branch** — verified conservative, as documented. I probed it directly by making `_spec_provider_for` raise: the slot file is **byte-identical** afterwards on *both* kfd axes, `relabel_stale_vulkan_slots` returns `0`, and exactly one `updater.vulkan_migration_slot_runtime_unresolved` warning fires per pass, carrying the slot name. Not a silent skip — logged, and at the same key granularity as the sibling `updater.vulkan_migration_slot_unreadable` breadcrumb (both omit `job_id`, consistent with the existing per-slot warnings in this function). Correct call: an unresolvable runtime family cannot be *concluded* to be llama.cpp-backed, so not touching it is the safe default.

Deferring the kfd-absent non-llama case to #1941 is the right split — that is `require_kfd_for_gpu_slot` over-firing, not a migration bug.

### Revert-and-confirm-red on the new gate

Four mutations at `f5155430`, each run against `tests/updater/test_vulkan_slot_migration.py` (22 collected):

| Mutation | Result |
|---|---|
| remove the `_spec_provider_for` block entirely (pre-fix behavior) | **7 failed** / 15 passed |
| keep the call, drop only the skip (`if False:`) | **7 failed** / 15 passed |
| invert the gate (`if provider is None:` — skip llama, relabel non-llama) | **18 failed** / 4 passed |
| exception branch falls through to relabel instead of skipping | **22 passed** ← see nit 1 |

Exactly 7 red on revert, matching the claim. Baseline restored → 22 passed; full `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/updater tests/install -q` → **538 passed**.

### Re-verified post-revision

- **Device-key-only** still holds, now proven by value rather than by reading: I round-tripped a slot carrying `runtime`, `port`, `n_gpu_layers`, `gpu_index`, `[model]` and `[server].env`, deleted `device` from both sides, and asserted the parsed documents are equal. Only `device` changed.
- **Idempotency** still holds and is now stronger than the suite asserts: the second run leaves the file **byte-identical** (not merely "device unchanged" / "no new log lines") — the `continue` precedes the write, so no rewrite occurs.
- **Non-AMD no-op** unchanged: still returns before the directory is globbed.

### Non-blocking follow-ups (2, 3, 4 from the first pass — all landed)

- Docstring comment claim: **fixed**, and better than asked — it now names `write_toml_atomic`'s `tomli_w` reserialization and states that comments are dropped, plus notes the same is true of `retag_stale_slot_images` / `clear_stale_mtp_overrides`.
- install.sh transcript: **fixed** in all three places (block comment, `info` line, printed summary), in the order the passes actually run.
- Stub signature: **fixed** (`amd_host` added).
- PR body test-count claim: **corrected** (22 new + 4 edited).

### Nits — not gating

1. **The exception branch is untested.** Mutating it to relabel-instead-of-skip leaves all 22 tests green. It behaves correctly (I verified out-of-band), but nothing guards it against a future edit. A ~10-line test monkeypatching `hal0.providers.container._spec_provider_for` to raise, asserting file-unchanged + the `runtime_unresolved` event, would close it. Same for the profile-family path (`type=llm` + a non-llama `profile`), which is the discriminator's more interesting half and is currently only covered by `type`-shaped slots.
2. **Private cross-module import.** `from hal0.providers.container import _spec_provider_for` couples the updater to a private name. Justified here (parity with `load_sync` is the whole point, and a public alias would be its own churn), but worth a one-line note at the import so a future rename doesn't silently change this migration's scope.
3. **Docstring formatting artifact** — `"…logs nothing\n    new."` in `relabel_stale_vulkan_slots` wrapped mid-phrase. Cosmetic.

### CI

All green on this head, including `γ-suite (chromium)` (11m58s pass — the #1926 flake did not recur).

No blocking objections. Approving.

thinmintdev:
Pushed the two nit tests + import comment in b7600d75.

- **`test_runtime_resolution_error_leaves_slot_untouched_and_logs`** — monkeypatches `_spec_provider_for` to raise, asserts the slot TOML is byte-identical after the run, the pass returns `0` (no crash), and `updater.vulkan_migration_slot_runtime_unresolved` is logged with `slot`/`error`. Confirmed this test catches a mutation: temporarily changed the `except` branch to fall through as `provider = None` (i.e. treat an unresolvable runtime as llama.cpp and relabel it) — the new test failed (`assert result == 0` → got `1`), reverted, re-ran to confirm green again.
- **`test_qwen3tts_profile_slot_survives_untouched`** — pins the profile-family-shaped case (`type="llm"`, `profile="qwen3-tts"`, `device="gpu-vulkan"` — the seed profile name resolves to the `qwen3tts` runtime_family via `ProfileCatalog`, matched before the generic `type="tts"` → Kokoro fallback in `_spec_provider_for`) as a regression guard on both kfd axes.
- Added a one-line comment where `_spec_provider_for` is imported from `container.py`, noting the parity-with-`load_sync` is intentional.

Verification: `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/updater tests/install -q` → **540 passed**; `ruff format --check` clean; `make lint` clean; `check_sunset.py` at baseline.

Not merging — leaving for operator merge.

Issue #1889, opened 2026-08-15
api: image_status is always "missing" and actual_image is always null for running slots — #663 image-drift detector permanently inert


thinmintdev:
**Severity:** major · **Box:** ct151-cpu-fresh, also reproduces on ct150-update and ct105-prod (rc.5) — pre-existing, not an rc.6 regression.

### Summary

`ContainerProvider.image_present()` / `running_image()` shell out to bare `podman` as the `hal0` user against **root's** image store (hal0 has no subuid ranges at all), so on every standard install:

- `GET /api/slots` reports `image_status: "missing"` for every running, healthy slot.
- `GET /api/slots/{name}/pull/status` reads "missing" for a present image.
- `actual_image` is always `null`.
- The `#663` image-drift detector is permanently inert — it can never fire.

The fix already exists in-tree for a sibling call site: commit `cbc8e94d` added `podman_introspect.py` + the `hal0-podman-ro` sudo seam and wired it into `/api/system-info`, which answers correctly. `providers/container.py`'s `image_present`/`running_image`/`container_command` were never migrated onto the same seam.

No UI surface renders the wrong value directly; the lying surfaces are the JSON payloads, `hal0 slot show`, and any agent reading them.

### Repro

```
curl -s $API/api/slots | jq -c '.[] | select(.container_status=="running")
  | {name, image_status, actual_image}'
ssh <box> podman image inspect <the slot image> >/dev/null && echo present-in-root-store
```

### Expected

For any slot with `container_status=running`, `image_status` is `"present"` (never `"missing"`), and `actual_image` is populated so `image_mismatch` can ever fire.

### Fix

Route `image_present()` / `running_image()` / `container_command()` through `podman_introspect` when `is_hal0_service_user()`, same as `/api/system-info` already does.

### Test to promote

A pytest on the `is_hal0_service_user()` gate in `providers/container.py` asserting `image_status == "present"` for a slot whose container is running.

---
Found by `rc-validate` v1.0.0-rc.6. Regression register key: `image-status-wrong-podman-store`.

thinmintdev:
[...]

Issue #1465, opened 2026-07-30
installer: no doctor/preflight check covers the sudo seam all slot lifecycle depends on; install.sh only warns when the grant fails to install


thinmintdev:
**Severity: HIGH** · area: doctor / install seams · surface: installer

Audit finding title: No doctor/preflight check covers the sudo seam that all slot lifecycle depends on; install.sh only warns when the grant fails to install

## Evidence

[...]

Issue #663, opened 2026-06-08
Cleanup: retire /proc actual_backend; deterministic image-tag mismatch


thinmintdev:
## Parent
#652

## What to build
For container slots, drop the fragile /proc/<pid>/exe actual_backend introspection. The running backend IS the image tag (podman inspect); 'mismatch' = running image tag ≠ slot's profile image. Update the dashboard mismatch chip accordingly.

## Acceptance criteria
- [ ] No /proc introspection for container slots
- [ ] actual_backend = image tag via podman inspect; mismatch computed from it
- [ ] lemond slots keep existing behavior until they migrate

## Blocked by
- #657
- #662

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1937, opened 2026-08-18
fix(api): route slot image reads through the rootful podman seam

thinmintdev:
> [!IMPORTANT]
> **SUDOERS + INSTALLER CHANGE — operator review required.** This PR widens a root sudo seam (`hal0-podman-ro`) from zero caller-supplied argv to three verbs that each accept one validated positional operand. Please review the validation regexes and the argument doctrine in the wrapper header before merging. **Do not automerge.**

Refs #1889.

## The bug

`ContainerProvider.image_present` / `running_image` / `running_argv` shelled out to a bare `podman` as the unprivileged `hal0` service user, which has no subuid ranges at all, so they read hal0-api's **own rootless** image store. Slots run **rootful** podman (Quadlet units under `/etc/containers/systemd/`, root's store), so that store by construction never holds a slot image.

On every standard install (reproduced on ct151, ct150 and ct105/rc.5 — pre-existing, not an rc.6 regression):

- `GET /api/slots` reported `image_status: "missing"` for every running, healthy slot
- `GET /api/slots/{name}/pull/status` read "missing" for a present image
- `actual_image` was always `null`, so the **#663 image-drift detector could never fire**

The `hal0-podman-ro` seam and its `is_hal0_service_user()` gate already existed — `cbc8e94d` wired them into `/api/system-info`, which answers correctly — but the wrapper exposed only the argument-free `images` verb, and its header doctrine forbade caller args reaching podman's argv. The three call sites that need a *specific* image ref or container name therefore could not use it.

## The privilege boundary

The doctrine is **refined, not abandoned**, and lands exactly where `hal0-systemctl`'s slot/agent verbs have had it since P3-perms: a caller-supplied value may reach podman's argv only as a **single positional operand validated on the root side against a closed regex**, for a verb whose podman subcommand, flags and `--format` string are literals in the wrapper.

| verb | operand | root-side validation |
|---|---|---|
| `image-exists <ref>` | OCI image reference | `^([A-Za-z0-9]+([.-][A-Za-z0-9]+)*(:[0-9]{1,5})?/)?[A-Za-z0-9]+([._-][A-Za-z0-9]+)*(/…)*(:tag)?(@sha256:[0-9a-f]{64})?$`, length ≤ 512 |
| `container-image <slot-token>` | slot **instance token** | `^[A-Za-z0-9_-]{1,64}$` — byte-identical to `hal0-systemctl`'s `validate_slot_id` |
| `container-argv <slot-token>` | slot **instance token** | same |

Deliberate design points:

- **The container verbs do not take a container name.** They take the bare instance token and the wrapper assembles `hal0-slot-<token>` itself, root-side. The caller can therefore only ever address a hal0 slot container — never an arbitrary one, and never a podman option.
- A validated ref cannot contain whitespace, a shell metacharacter, a leading `-` (no flag smuggling), `..`, or a second argv word. Each verb rejects a second argv word outright.
- Exec arrays only. No shell, no `eval`, no word splitting, no wildcards, no caller-supplied `--format`. `rm`/`run`/`build`/`exec`/`pull` remain unreachable.
- The image-ref regex is **structure-strict, case-permissive**: the security properties come from the structure, and podman itself rejects uppercase repo names, so folding case here would only convert a podman error into a wrapper rejection — while risking a false reject, which is *literally the bug this PR fixes* (a rejected ref degrades to `image_status: "missing"` again).
- The **sudoers grant is unchanged** (`hal0 ALL=(root) NOPASSWD: /usr/lib/hal0/bin/hal0-podman-ro`). A bare command path in sudoers already permits any argv; the wrapper is the control surface. Enumerating verbs in sudoers instead would be a second, silently-drifting copy of the verb list, and sudoers wildcards are a known footgun. Only the explanatory comment changed.

### Exit-code contract

`0` = podman answered (empty stdout on a container verb is a real "no such container"); `64` = wrapper rejected the argument or verb; `65` = podman absent. This is what lets the Python side tell **"the seam did not answer"** from **"the answer is negative"**. The named-object reads deliberately do **not** silently fall back to the rootless store: a rootless answer about a *named* image is not a stale answer, it is an answer about a *different object* — that conflation is #1889. The fallback survives only for the dev/CI case where the process is not the hal0 service user (no grant exists there, and the operator's own store is the right one).

## Files changed

| file | change |
|---|---|
| `installer/wrappers/hal0-podman-ro` | 3 new read verbs + 2 side-effect-free validator probes (`check-image-ref`, `check-slot-token`), 2 validators, exit-code contract, rewritten argument doctrine |
| `packaging/sudoers/hal0-podman-ro` | comment only — documents the new surface and why argv is not enumerated in sudoers |
| `installer/install.sh` | seam #5 comment updated to the new surface |
| `installer/lib/preflight.sh` | `hal0-podman-ro` gains a `help` probe — the grant is now load-bearing for slot status, so presence-only checking was no longer enough |
| `src/hal0/system/seam_check.py` | same probe, kept in lock-step (the `#1465` inventory) |
| `src/hal0/providers/podman_introspect.py` | `image_exists` / `container_image` / `container_argv` (tri-state), `is_valid_image_ref` / `is_valid_slot_token` mirrors |
| `src/hal0/providers/container.py` | the three methods route through the seam first; shared `_decode_argv_json` helper |
| `tests/installer/test_podman_ro_validation.py` | **new** — 220 tests |
| `tests/providers/test_container_podman_ro_routing.py` | **new** — 10 tests |
| `tests/providers/test_podman_introspect.py` | +26 tests |

## Test evidence

The wrapper tests run the **real bash wrapper** (no root, no sudo, no podman, no provisioned box — the `hal0-systemctl` drop-in suite's posture) over ~40 malicious argv shapes: command chaining (`;`/`&&`/`||`/`|`), substitution (`$(…)`/backticks), flag smuggling (`--rm`, `-v/:/host`, `--format={{.Config}}`, bare `-`), traversal (`../../../etc/shadow`, `foo/../../etc/passwd`), malformed digests, control bytes and RTL-override unicode, plus 12 legitimate refs that must **not** be rejected. They also assert **parity** between the wrapper and the Python mirrors — a mirror that drifts looser turns a fast rejection into an opaque `rc 64`; one that drifts stricter silently re-creates #1889.

```
tests/installer/test_podman_ro_validation.py ....... 220 passed
tests/providers/test_podman_introspect.py ........... 47 passed
tests/providers/test_container_podman_ro_routing.py . 10 passed

tests/installer tests/providers tests/slot_view tests/install tests/system
                                                     1811 passed, 1 skipped
tests/api/test_slots_container_state.py tests/api/test_slots_image_pull.py
tests/api/test_system_info_route.py tests/updater/test_wrapper_refresh.py
                                                       48 passed

make lint                                            All checks passed
uv run ruff format --check src tests                 1149 files already formatted
uv run python scripts/check_sunset.py                scar markers 192 <= baseline 192
```

The issue's "test to promote" is `test_slot_view_reports_present_for_a_running_slot`, which drives the real `ContainerProvider` through `slot_view`'s TTL cache helper — the exact path that produces `image_status`. The seam-routing tests booby-trap `subprocess.run` so a test cannot pass by silently falling through to the rootless read.

## Not covered / follow-ups

- The full `tests/api` suite was not run locally (it exceeds a 10-minute local budget); the relevant slot/system-info files were. CI covers the rest.
- `shellcheck` is not installed on this box, so `tests/installer/test_platform_gate_hardening.py` skipped its shellcheck arm as it does on main. `bash -n` passes and is asserted by a test.
- Verifying the fix end-to-end on a provisioned box (`image_status: "present"` for a running slot on ct151) needs a deploy and is left to RC validation.

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1735,11 +1735,14 @@ PYEOF
         fi
     fi
 
-    # Privileged seam #5 (O12): hal0-podman-ro covers READ-ONLY podman
-    # introspection (image presence today) against ROOT's podman store — the
-    # store slots actually populate via Quadlet, NOT hal0-api's own rootless
-    # store. Narrow + hardcoded (no shell, no wildcards, no operator-supplied
-    # podman flags) — see the wrapper source.
+    # Privileged seam #5 (O12, extended in #1889): hal0-podman-ro covers
+    # READ-ONLY podman introspection — the local repo set, per-image presence,
+    # and a running slot container's image + argv — against ROOT's podman
+    # store, the store slots actually populate via Quadlet, NOT hal0-api's own
+    # rootless store. Narrow + hardcoded (no shell, no wildcards, no
+    # operator-supplied podman flags or --format); the three verbs that take
+    # an argument validate it root-side against a closed regex before exec.
+    # See the wrapper source for the full argument doctrine.
     PODMAN_RO_SRC="${REPO_ROOT}/installer/wrappers/hal0-podman-ro"
     if [[ -f "${PODMAN_RO_SRC}" ]]; then
         install -d "${LIB_DIR}/bin"
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1339,6 +1339,10 @@ _hal0_seam_probe() {
     case "$1" in
         hal0-systemctl) printf 'help\n' ;;
         hal0-update)    printf 'check\n' ;;
+        # #1889: podman-ro is now the only source of truth for a running
+        # slot's image_status/actual_image, so a silently-missing grant stops
+        # being cosmetic. `help` prints usage and touches nothing.
+        hal0-podman-ro) printf 'help\n' ;;
         *)              return 1 ;;
     esac
 }
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1341,8 +1341,12 @@ _hal0_seam_probe() {
         hal0-update)    printf 'check\n' ;;
         # #1889: podman-ro is now the only source of truth for a running
         # slot's image_status/actual_image, so a silently-missing grant stops
-        # being cosmetic. `help` prints usage and touches nothing.
-        hal0-podman-ro) printf 'help\n' ;;
+        # being cosmetic. NOT `help` — the pre-#1889 wrapper implements that
+        # too, so a failed wrapper refresh would probe green while every new
+        # verb was rejected. `check-slot-token` is release-specific and
+        # side-effect-free (validates, prints the name it would build, never
+        # calls podman). Keep in lock-step with src/hal0/system/seam_check.py.
+        hal0-podman-ro) printf 'check-slot-token hal0probe\n' ;;
         *)              return 1 ;;
     esac
 }
@@ -1382,11 +1386,17 @@ _preflight_seam() {
 
     local probe
     if probe="$(_hal0_seam_probe "${name}")" && (( rc == 0 )); then
+        # Probes are hardcoded above, never caller input, so splitting the
+        # line into argv words is safe — and #1889 needs a probe that takes an
+        # argument (a verb-only probe cannot tell a stale wrapper from a
+        # current one).
+        local -a probe_argv=()
+        read -r -a probe_argv <<< "${probe}"
         # The grant is written for the hal0 user, so the only honest test runs
         # AS that user. -n keeps it non-interactive: a missing grant fails
         # immediately instead of prompting.
-        if ! sudo -n -u hal0 sudo -n "${bin}" "${probe}" >/dev/null 2>&1; then
-            "${report}" "seam ${name}: 'sudo -n ${name} ${probe}' failed as the hal0 user — the ${grant} grant does not apply"
+        if ! sudo -n -u hal0 sudo -n "${bin}" "${probe_argv[@]}" >/dev/null 2>&1; then
+            "${report}" "seam ${name}: 'sudo -n ${name} ${probe}' failed as the hal0 user — the ${grant} grant does not apply, or the wrapper is stale"
             rc=1
         fi
     fi
--- a/installer/wrappers/hal0-podman-ro
+++ b/installer/wrappers/hal0-podman-ro
@@ -49,12 +49,23 @@
 # answer is negative" from "the seam itself is unusable", because the caller
 # must fall back to its rootless read only in the second case:
 #
-#   * rc 0   — podman ran. `image-exists` printed `present` or `missing`;
-#              `container-image` / `container-argv` printed the value, or
-#              nothing at all when no such container is running. An empty
-#              stdout on rc 0 is a real answer, not an error.
+#   * rc 0   — podman ran and gave a DEFINITIVE answer. `image-exists`
+#              printed `present` or `missing`; `container-image` /
+#              `container-argv` printed the value, or nothing at all when no
+#              such container exists. An empty stdout on rc 0 is a real
+#              answer, not an error.
 #   * rc 64  — this wrapper rejected the argument (validation) or the verb.
 #   * rc 65  — podman is not installed / not executable here.
+#   * rc 66  — podman ran but FAILED operationally (store corruption, lock
+#              contention, permission error, …). NOT a negative answer.
+#
+# rc 66 exists because conflating an operational podman failure with "the
+# image is absent" reproduces #1889 with extra steps: the API would report a
+# locally present image as missing and the caller would never know to
+# degrade. That is why the presence probes below use `podman image exists` /
+# `podman container exists`, whose contract is exactly rc 0 = yes, rc 1 = no,
+# anything else = error — `podman inspect` collapses "not found" and every
+# other failure into a single rc 125 and cannot make this distinction.
 #
 # A caller that gets rc 0 must trust stdout; anything else means "seam did
 # not answer" (including sudo's own rc 1 when the grant is missing).
@@ -88,8 +99,15 @@ SLOT_TOKEN_RE='^[A-Za-z0-9_-]{1,64}$'
 # would only convert a podman error into a wrapper rejection while risking a
 # false reject on a legitimate ref; a false reject is precisely the failure
 # mode #1889 is about (it degrades to image_status="missing" again).
-_REF_HOST='[A-Za-z0-9]+([.-][A-Za-z0-9]+)*(:[0-9]{1,5})?'
-_REF_PATH='[A-Za-z0-9]+([._-][A-Za-z0-9]+)*(/[A-Za-z0-9]+([._-][A-Za-z0-9]+)*)*'
+# The separator set is the distribution-reference grammar's, verbatim:
+#   separator := "." | "_" | "__" | "-"+
+# Getting this wrong is not a hardening nit — a legitimate ref that this
+# rejects never reaches the rootful store, so the caller degrades to the
+# rootless one and #1889 comes straight back for that image. `model__gpu` and
+# `model--gpu` are both legal repository names.
+_REF_SEP='(__|[._]|-+)'
+_REF_HOST="[A-Za-z0-9]+(([.]|-+)[A-Za-z0-9]+)*(:[0-9]{1,5})?"
+_REF_PATH="[A-Za-z0-9]+(${_REF_SEP}[A-Za-z0-9]+)*(/[A-Za-z0-9]+(${_REF_SEP}[A-Za-z0-9]+)*)*"
 _REF_TAG='(:[A-Za-z0-9_][A-Za-z0-9._-]{0,127})?'
 _REF_DIGEST='(@sha256:[0-9a-f]{64})?'
 IMAGE_REF_RE="^(${_REF_HOST}/)?${_REF_PATH}${_REF_TAG}${_REF_DIGEST}\$"
@@ -113,6 +131,10 @@ require_podman() {
   [[ -x "$PODMAN" ]] || { echo "hal0-podman-ro: podman not found at $PODMAN" >&2; exit 65; }
 }
 
+# An operational podman failure — distinct from a negative answer. See the
+# exit-code contract in the header.
+podman_failed() { echo "hal0-podman-ro: $1 (rc=$2)" >&2; exit 66; }
+
 # Run podman read-only, tolerating a non-zero rc (a missing image/container is
 # a legitimate NEGATIVE answer, not a seam failure). Sets PODMAN_RC/PODMAN_OUT.
 # stderr is dropped: podman emits benign device warnings under LXC.
@@ -120,6 +142,25 @@ run_podman() {
   PODMAN_OUT=""
   PODMAN_RC=0
   PODMAN_OUT="$("$PODMAN" "$@" 2>/dev/null)" || PODMAN_RC=$?
+}
+
+# Read one LITERAL --format field off a hal0 slot container.
+#   $1 = the format string (a literal from the case arm below, never caller
+#        input); $2 = the container name, already assembled from a validated
+#        token on this side of the boundary.
+# Prints the value and exits 0; prints nothing and exits 0 when the container
+# does not exist; exits 66 when podman itself failed.
+container_read() {
+  local fmt="$1" name="$2"
+  run_podman container exists -- "$name"
+  case "$PODMAN_RC" in
+    0) ;;
+    1) return 0 ;;   # no such container — a real negative answer
+    *) podman_failed "podman container exists failed" "$PODMAN_RC" ;;
+  esac
+  run_podman inspect --format "$fmt" -- "$name"
+  (( PODMAN_RC == 0 )) || podman_failed "podman inspect failed" "$PODMAN_RC"
+  printf '%s\n' "$PODMAN_OUT"
 }
 
 cmd="${1:-help}"; shift || true
@@ -133,11 +174,16 @@ case "$cmd" in
     [[ $# -eq 1 ]] || die "image-exists takes exactly one argument"
     validate_image_ref "$1"
     require_podman
-    # --format {{.Id}} is a LITERAL here: it keeps the inspect output to one
-    # short line instead of the full JSON manifest, and the value is
-    # discarded — only podman's rc decides presence.
-    run_podman image inspect --format '{{.Id}}' -- "$1"
-    if (( PODMAN_RC == 0 )); then echo present; else echo missing; fi
+    # `image exists` — NOT `image inspect`. Its contract is rc 0 = present,
+    # rc 1 = absent, anything else = operational failure, which is the only
+    # way to keep "podman is broken" from being reported as "the image is
+    # missing". `inspect` returns 125 for both.
+    run_podman image exists -- "$1"
+    case "$PODMAN_RC" in
+      0) echo present ;;
+      1) echo missing ;;
+      *) podman_failed "podman image exists failed" "$PODMAN_RC" ;;
+    esac
     ;;
 
   container-image)         # arg: slot token -> prints the running image ref
@@ -146,18 +192,19 @@ case "$cmd" in
     require_podman
     # #663 backend-of-record: the running container's image IS the backend.
     # The container NAME is built here from the validated token.
-    run_podman inspect --format '{{.ImageName}}' -- "hal0-slot-$1"
-    # A `(( … )) && printf` one-liner would make the "no such container" case
-    # exit non-zero under `set -e`, which the caller reads as "seam unusable".
-    if (( PODMAN_RC == 0 )); then printf '%s\n' "$PODMAN_OUT"; fi
+    #
+    # Two podman calls in ONE sudo hop: `container exists` first, because it
+    # is the only call that can say "absent" without also meaning "podman
+    # broke", then the inspect whose failure is therefore unambiguously
+    # operational.
+    container_read '{{.ImageName}}' "hal0-slot-$1"
     ;;
 
   container-argv)          # arg: slot token -> prints the live command argv
     [[ $# -eq 1 ]] || die "container-argv takes exactly one argument"
     validate_slot_token "$1"
     require_podman
-    run_podman inspect --format '{{json .Config.Cmd}}' -- "hal0-slot-$1"
-    if (( PODMAN_RC == 0 )); then printf '%s\n' "$PODMAN_OUT"; fi
+    container_read '{{json .Config.Cmd}}' "hal0-slot-$1"
     ;;
 
   check-image-ref)         # arg: image ref — side-effect-free validator probe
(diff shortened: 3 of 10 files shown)


thinmintdev:
## Codex review triage — all three P2s accepted and fixed in 62b7d786

Each finding was real, and each would have re-created #1889 for a subset of slots rather than fixing it. None were dismissed.

**1. `__` / `-`+ separators rejected — CONFIRMED, fixed.** The distribution-reference grammar's separator set is `"." | "_" | "__" | "-"+`; my regex accepted only the single-character forms, so `team/model__gpu:v1` and `team/model--gpu:v1` were rejected before the seam ran. The failure mode is precisely the bug this PR fixes: a rejected ref falls back to the rootless store and reports a present image as missing. Wrapper and Python mirror both now use the grammar's set — with `__` ordered **first** in the Python copy, since `re` is leftmost-first rather than POSIX leftmost-longest and `_|__` would otherwise match only the first underscore. The three refs from the finding are now in both the legitimate-ref fixtures and the wrapper↔mirror parity test.

**2. podman failure reported as a negative — CONFIRMED, fixed.** `podman inspect` collapses "not found" and store-corruption / lock / permission failures into a single rc 125, so the old arm printed `missing` and exited 0 for both. Fixed at the root: the presence probes now use `podman image exists` / `podman container exists`, whose contract is *rc 0 = yes, rc 1 = no, anything else = error* — the only podman verbs that can make this distinction. New wrapper exit code **66 = "podman ran but failed operationally"**, which `_seam_read` already treats as "the seam did not answer", so the caller degrades instead of lying. The container verbs run both podman calls inside **one** sudo hop, so this costs no extra round-trip. Verified against real podman: `podman image exists nosuch` → rc 1, `podman container exists nosuch` → rc 1.

**3. Drift would cry wolf on every correctly running slot — CONFIRMED, fixed.** Good catch on the second-order effect: `_image_mismatch` was a literal string compare, harmless only because `running_image()` returned `None` on every deployed box. Making it reachable would have traded an inert detector for a lying one. Added `canonical_image_ref()` — implicit `docker.io/` (+ `library/` for a bare single component, using the standard "first component is a registry iff it has `.`/`:` or is `localhost`" heuristic) and implicit `:latest` — applied to both sides before comparison. The tag is split on the last colon *after the last slash*, so `localhost:5000/foo` is not amputated into a false match. Digest-vs-tag compares the repository alone: resolving whether a tag and a digest name the same image needs a registry round-trip this hot path will never make, and guessing "drifted" is the cry-wolf failure #663's contract forbids.

### Verification

```
tests/installer/test_podman_ro_validation.py     226 passed  (+6)
tests/providers/...                              857 passed  (+21)
tests/installer providers slot_view install system
  + api/{slots_container_state,slots_image_pull,slots_routes,system_info_route}
                                                1974 passed, 1 skipped
make lint / ruff format --check                 clean
```

New coverage: the three previously-rejected refs, rc 66 → `None` (never `"missing"`) for both `image_exists` and `container_image`, structural assertions that the presence probes use `exists` rather than `inspect` and that no `--format` is ever read from argv, 11 `canonical_image_ref` cases, 5 no-false-drift pairs, 4 real-drift pairs (including the registry-port case), and the digest-vs-tag rule.

Still **not for automerge** — SUDOERS + INSTALLER change, operator review required.

thinmintdev:
## Codex round 2 — both P2s accepted and fixed in 6d6c4dec

**4. Rootless fallback collapsed the rc-66 distinction — CONFIRMED, fixed.** This was the sharpest finding on the PR: the previous commit added rc 66 precisely so an operational podman failure would not read as "missing", and then the fallback in `image_present` immediately threw that away. On a provisioned box the rootless store *definitionally* cannot hold a slot image, so consulting it after a seam failure converts "podman is broken" / "the grant is missing" / "the ref was rejected" back into an authoritative-looking `image_status: "missing"`.

The fallback is now gated on `is_hal0_service_user()` — it runs only **off** the service account, where the operator's own store genuinely is the store slots use — and the seam-failed-as-`hal0` path emits `hal0.podman_ro.image_present_unanswered` at WARNING so an operator gets a diagnosable signal rather than a silent wrong answer. `running_image` / `running_argv` are gated the same way; there the answer was already `None` either way, so it drops a wasted subprocess spawn from the status hot path and removes the chance of answering about a same-named container that is not this slot's.

One honest limitation, called out rather than papered over: `image_present` must still answer `False` on that path, because its contract is a `bool` and the `image_status` enum has no `"unknown"` member. Surfacing a distinct unknown state is an **API-schema change** and is deliberately out of scope for a fix PR — flagging it here as a follow-up rather than smuggling it in.

**5. Docker Hub's implicit `library/` namespace — CONFIRMED, fixed.** `docker.io/alpine` was treated as already-qualified and canonicalised to `docker.io/alpine:latest`, while podman reports `docker.io/library/alpine:latest` — so a slot declared that way would have been flagged as drifted while running the correct image. The namespace rule now applies to the repository portion *however the registry got there*, sequenced after the registry rule instead of nested inside it, so `alpine`, `docker.io/alpine` and `docker.io/library/alpine` all canonicalise identically.

### Verification

```
tests/installer tests/providers tests/slot_view tests/install tests/system
  + api/{slots_container_state,slots_image_pull,slots_routes,system_info_route}
                                          1984 passed, 1 skipped  (+10)
make lint / ruff format --check           clean
```

New coverage: `no_rootless` (a booby-trapped `subprocess.run` that raises if reached) asserts the rootless store is never touched as the service user for all three methods; the warning is asserted via `caplog`; the dev/CI fallback is re-pinned with an explicit `is_hal0_service_user() == False`; and `docker.io/alpine` / `docker.io/alpine:3.19` are added to the canonicalisation and no-false-drift tables.

Still **not for automerge** — SUDOERS + INSTALLER change, operator review required.

thinmintdev:
## Codex round 3 — all three P2s accepted and fixed in 8290fdfb

**6. The probe could not tell a current wrapper from a stale one — CONFIRMED, fixed.** The best finding of the three, and a self-inflicted one: I added a `help` probe two commits ago specifically because the grant became load-bearing, but `help` exists on the **pre-#1889 wrapper too**. A box whose best-effort wrapper refresh failed would keep the one-verb wrapper, probe green, and report the seam healthy from both `hal0 doctor` and install.sh's post-install assertion — while every new verb was rejected and `image_status` stayed `"missing"`. That is precisely the undiagnosable-green failure `#1465` exists to prevent, so shipping it inside a `#1465`-adjacent inventory would have been embarrassing.

Both inventories now probe `check-slot-token hal0probe`: release-specific (the old wrapper answers rc 64) and still side-effect-free — it validates the token and prints the container name it *would* build, touching neither podman nor the filesystem. `_preflight_seam` now splits its hardcoded probe line into argv words, since a verb-only probe structurally cannot make this distinction. The `seam_check.py` and `preflight.sh` copies stay in lock-step, with a comment on each pointing at the other.

**7. Bracketed IPv6 registries rejected — CONFIRMED, fixed.** `[2001:db8::1]:5000/team/model:v1` is a legal reference host, and the failure mode is #1889 verbatim: the ref never reaches the rootful seam and a present image reports missing. Added as a host alternative in the wrapper and the mirror. The bracket body is `[0-9A-Fa-f:]{2,45}` — hex and colons only — so it cannot carry a path, whitespace, a metacharacter or a second word; four hostile bracket shapes (`[2001:db8::1;id]:5000/foo`, `[../etc]/foo`, `[]/foo`, `[2001:db8::1]extra/foo`) are pinned as rejected, and both IPv6 forms go through the wrapper↔mirror parity test.

**8. Two digest-pinned refs compared as whole strings — CONFIRMED, fixed.** A digest *is* the immutable image id, so `repo:v1@sha256:D` and `repo@sha256:D` name the same image and the tag text alongside it is noise; the old code called that drift. When both sides carry a digest the comparison is now repository-without-tag **plus** digest. The one-sided case is unchanged (repository only — resolving a tag against a digest needs a registry round-trip this hot path will never make).

### Verification

```
tests/installer providers slot_view install system updater
  + api/{slots_container_state,slots_image_pull,slots_routes,system_info_route}
                                          2270 passed, 1 skipped
bash -n installer/lib/preflight.sh        ok
make lint / ruff format --check           clean
```

Prior rounds' CI on 6d6c4dec was fully green (python 3.12, γ-suite, ui, sunset, CodeQL ×4) before these commits landed.

Still **not for automerge** — SUDOERS + INSTALLER change, operator review required.

thinmintdev:
> Independent security review by a second agent (posted as a comment: GitHub will not let this account formally approve its own PR). **Verdict: APPROVE — 0 blocking findings.**

## Independent security review — APPROVE (no blocking findings)

Adversarial review of the privilege boundary only. I did not write this code and re-ran every claim rather than reading the PR body for them. **No validation bypass found.** 4 non-blocking findings, all hardening/plumbing, none of which lets a caller escape the intended object or reach a non-read podman subcommand.

Note this does not clear the PR's own operator gate — it is a sudo-seam change and still needs the operator's sign-off, not automerge.

### Design claims verified

**1. Root-side validation.** ~120 hostile inputs beyond the suite's set, run against the real `installer/wrappers/hal0-podman-ro` (`check-image-ref`, and re-confirmed through the live `image-exists` verb). All rc 64:

- chaining/substitution: `alpine;id`, `alpine&&id`, `alpine|id`, `` alpine`id` ``, `alpine$(id)`, `<(id)`, `>x`
- flag smuggling: `-`, `--`, `-rm`, `--rm`, `-v/:/host`, `--format={{.Config}}`
- whitespace/IFS: leading space, trailing space, embedded space, `\t`, `\n`, `\r`, `$IFS`
- traversal / shape: `..`, `../../../etc/shadow`, `foo/../../etc/passwd`, `foo/..`, `/abs`, `abc/`, `a//b`, empty
- control bytes and unicode: `\x01`, `\x7f`, `\x1b[31m`, U+202E RTL override, invalid UTF-8 `\xc0\x80`
- digest edges: bare `@sha256:<64>`, 2-hex, 63-hex, uppercase hex, `sha512:`, trailing `g`
- IPv6 bracket abuse: `[]/a`, `[/etc/passwd]/a`, `[2001:db8::1]extra/foo`, `[2001:db8::1;id]:5000/foo`

Arity is enforced per verb (`[[ $# -eq 1 ]]`): zero args and a second argv word both rc 64 on all three verbs, and on the two probes. Unknown verb rc 64. `images` ignores trailing argv (literal exec array), so it cannot be widened by extra words.

No shell interpolation anywhere: every podman call is a quoted exec array, `set -euo pipefail`, no `eval`, no unquoted expansion, no globbing. `[[ "$x" =~ $RE ]]` has the RHS correctly unquoted (a quoted RHS would silently become a literal match — worth keeping in mind if anyone "tidies" it later).

**No ReDoS.** `_REF_SEP=(__|[._]|-+)` under a `*` is the classic shape, so I timed the adversarial partitions: `a`+500×`-`, `(a-)×250+!`, `(a__)×160+!`, `(/a)×250+!`, `(a.)×250+!`, 45-char IPv6 body — all ≤ 3 ms. The 512-byte cap is applied *before* the regex (verified: 513 chars rejects without touching the engine). A 1 MB slot token is also safe.

**2. Container verbs cannot address an arbitrary container.** Confirmed. The token charset admits `-`, `--all`, `-rm`, but the value is only ever concatenated into `hal0-slot-<token>` root-side and passed after `--`; `check-slot-token --all` prints `hal0-slot---all`. There is no code path where the token becomes a bare argv word. 64 chars accepted, 65 rejected. `.`/`/`/`:`/`@`/whitespace all rejected, so the assembled name can never be a path, an ID, or a second word.

**3. Sudoers byte-unchanged.** `diff <(git show main:packaging/sudoers/hal0-podman-ro | grep -v '^#') <(git show HEAD:… | grep -v '^#')` is empty — comments only, grant line identical. Agree with the reasoning for not enumerating argv in sudoers.

**4. Exit-code contract + fallback gate.** Wrapper side is right: `image exists`/`container exists` (rc 0/1/other) rather than `inspect` (125 for everything) is the correct primitive, and I confirmed real podman returns 125 for a malformed-but-regex-valid ref (`ALPINE` → wrapper rc 66, not a false "missing"). The `is_hal0_service_user()` gate is asserted in both directions by the routing tests, and the rootless fallback is provably unreachable on the service account (see finding 2 below for the one place the distinction is dropped).

**5. Parity test really runs the wrapper.** Mutated `_REF_IPV6` in the wrapper in a scratch worktree → 4 tests red, including both `test_python_image_ref_mirror_agrees_with_the_wrapper[…]` cases. Not a self-referential test.

**6. Codex round-3 #6 — probe is release-specific.** Confirmed against the real old wrapper from `main`:
```
$ ./w-old check-slot-token hal0probe   -> hal0-podman-ro: bad cmd: check-slot-token   rc=64
$ ./w-new check-slot-token hal0probe   -> hal0-slot-hal0probe                          rc=0
$ ./w-old help                         -> rc=0     # why `help` was insufficient
```

**7. Revert-and-confirm-red.** `git checkout main -- src/hal0/providers/container.py` → `38 failed, 6 passed`, with `test_slot_view_reports_present_for_a_running_slot` among them. The booby-trap is real (`monkeypatch.setattr(container_mod.subprocess, "run", _boom)` raising `AssertionError`), so a seam test cannot pass by falling through to the rootless read.

**8. No hostile ref reaches a non-read subcommand.** The only podman invocations in the file are `images`, `image exists`, `container exists`, `inspect --format <literal>`. No TOCTOU between validation and exec — same in-process string, no filesystem lookup in between. Wrapper stderr (which does echo the rejected operand verbatim) is never logged by the Python side: `_seam_read` reads `stdout` only and preflight discards both, so there is no log-injection path.

Suites: `tests/installer/test_podman_ro_validation.py tests/providers tests/system/test_seam_check.py` → **900 passed**.

---

### Findings (non-blocking)

**1. [low] Root-side regex semantics are caller-influenced by locale, and the test hides it.**
glibc range expressions are collation-based outside the C locale, and sudo's default `env_keep` carries `LANG`/`LC_*` through to the wrapper. So the authoritative root-side validator behaves differently depending on an environment variable the caller supplies:

```
LC_ALL=C           check-image-ref alpiné  -> rc 64
LC_ALL=en_US.UTF-8 check-image-ref alpiné  -> rc 0     # also ＡＢＣ, ⅰmage
```

The Python mirror uses `re` with ASCII-only `[A-Za-z0-9]` and rejects all three, so this is a genuine wrapper-looser-than-mirror parity break on any box whose service environment carries a UTF-8 `LANG`. The suite cannot see it because `_ENV = {"PATH": "/usr/bin:/bin"}` strips the locale, which is exactly why `"alpiné"` sits in `MALICIOUS_REFS` and passes.

I brute-forced all 255 single bytes under both locales and the accepted set is **identical and ASCII-alnum-only** in each, so nothing structural leaks: no metacharacter, no space, no `/`, no leading `-`, no `..`. The blast radius is multibyte letters only, which podman rejects anyway. Not an escalation — but a root-side allow-list whose meaning depends on caller-supplied env is worth closing on principle.

Suggested: `export LC_ALL=C` immediately after `set -euo pipefail` (also stabilises podman's output), plus one parity case that runs the wrapper with `LANG=en_US.UTF-8` in `_ENV` so the property is actually pinned.

**2. [low] rc 64 / 65 / 66 all collapse to `None` one layer up, so the wrapper's careful distinction never reaches the API.**
`_seam_read` returns `None` for every non-zero rc, and on the service account `image_present` then logs and returns `False`. Net effect: an operational podman failure still surfaces as `image_status: "missing"` — the shape Codex flagged. This PR does materially improve it (no rootless conflation, an operator-visible warning, a seam probe in preflight/doctor that catches the missing-grant case), and the docstring is honest that an `unknown` state is an API-schema change. Please file the follow-up so the rc 66 path the wrapper works hard to produce doesn't stay decorative. Evidence: `image-exists ALPINE` → podman 125 → wrapper rc 66 → `image_exists()` `None` → `image_present()` `False`.

**3. [nit] `container_read` uses `podman inspect`, not `podman container inspect`.**
Gated by `container exists`, and the name is always `hal0-slot-*`, so I could not construct an abuse — but `--type container` (or `container inspect`) removes the object-type ambiguity by construction rather than by argument, which is the doctrine the rest of this file follows.

**4. [nit] TOCTOU between `container exists` and `inspect`.**
If the slot is torn down between the two calls, inspect fails → rc 66 → `None` → "no actual_image". The fail-direction is the safe one (unknown, never false drift), so this is a note, not a request.

Assembly of `hal0-slot-<token>` root-side, `--` before every operand, the exec-array-only rule, and the "structure-strict, case-permissive" reasoning are all correct calls. Approving on the security boundary.

thinmintdev:
## Security review hardening — both findings addressed in `73ae6fcf`

Re: the review at https://github.com/Hal0ai/hal0/pull/1937#issuecomment-5323541603 (APPROVE, 0 blocking). Finding 1 was a genuine hole rather than a hardening nit, and the test suite structurally could not have caught it.

**Finding 1 [low] — locale-sensitive validators. Reproduced first, then fixed.**

```
env -i PATH=/usr/bin:/bin                     …/hal0-podman-ro check-image-ref alpiné  → rc 64
env -i PATH=/usr/bin:/bin LC_ALL=en_US.UTF-8  …/hal0-podman-ro check-image-ref alpiné  → rc 0   ← accepted
env -i PATH=/usr/bin:/bin LANG=en_US.UTF-8    …/hal0-podman-ro check-image-ref alpiné  → rc 0   ← accepted
python -c "…is_valid_image_ref('alpiné')"                                              → False
```

bash's `[[ =~ ]]` bracket expressions use the current locale's collation, so under a UTF-8 locale `[A-Za-z0-9]` collates accented letters into the range. That made the wrapper **looser than its mirror** — the dangerous direction of a parity break — and it was reachable in production, since sudo's default `env_keep` passes `LANG`/`LC_*` straight through from the calling process. Your read of why the suite was blind to it is exactly right: `_ENV` is `{"PATH": …}` only, so every existing case ran under an implicit C locale.

`export LC_ALL=C` now sits immediately after `set -euo pipefail`, before any validator can run, so every character class means exactly the ASCII bytes it spells regardless of the caller's environment. Post-fix: rc 64 under both locale vars.

**Finding 3 [nit] — bare `podman inspect`.** Switched to `podman container inspect`. The bare form also resolves images, volumes, networks and pods, so a name collision could have returned a different object's fields; the name is already pinned to the `hal0-slot-` prefix, but type-safety by construction is strictly better. Error text updated to match.

**Tests — 7 new cases, all red before the fix:**

- unicode refs (`alpiné`, `ALPINÉ`, `café/img:tag`) and tokens (`braîn`, `tokén`) must be rejected under `LANG` + `LC_ALL` = `en_US.UTF-8`
- the **entire** ref corpus (legitimate + malicious) must reach an *identical* verdict in both locales, **and** still agree with the Python mirror under UTF-8 — so this cannot regress for one input class while passing for another
- a structural assertion that `export LC_ALL=C` is present and ordered **ahead of the first validator**. This one carries weight: on a box where `en_US.UTF-8` is not generated bash silently falls back to C, so the behavioural cases alone would be a false green on CI
- `test_inspect_calls_are_type_qualified` — every `run_podman … inspect` must name its object type

```
tests/installer/test_podman_ro_validation.py        313 passed  (+7)
tests/installer providers slot_view install system updater
  + api/{slots_container_state,slots_image_pull,slots_routes,system_info_route}
                                                   2341 passed, 1 skipped
make lint / ruff format --check / check_sunset.py  clean
bash -n on the wrapper                             ok
```

Still **not for automerge** — SUDOERS + INSTALLER change, operator merge required.

Issue #843, opened 2026-06-15
Hermes: prevent root from running it / clobbering perms (root-clobber regression)


thinmintdev:
## Problem

Hermes must run **only** as the unprivileged `hal0` user with
`HERMES_HOME=/var/lib/hal0/.hermes`. The systemd unit enforces this
(`hal0-agent@.service` → `User=hal0`), but nothing stops root from running
Hermes — interactively (`sudo hermes …`, a root TUI session) or transitively
during install/bootstrap. When root runs it, two things break (often together):

1. **Split-brain home.** As root, Hermes resolves `~/.hermes` →
   `/root/.hermes` instead of `/var/lib/hal0/.hermes`, writing a competing
   config/state tree the hal0 runtime never reads.
2. **Permission clobber.** Where root writes into `/var/lib/hal0/.hermes`, the
   files land `root:root` (`config.yaml` 0600, the `runtime.json` embed token
   0600, the venv). The `hal0`-user service then hits `EACCES` — or Hermes
   silently falls back to the **default provider** because it cannot read its
   own config.

This is the documented "root-clobber regression"
(`docs/superpowers/plans/2026-06-06-hindsight-engine-swap.md` Task P5H-4). The
plan calls for a self-guarding wrapper that re-execs as hal0; the current
wrapper has only a cwd-guard, not a privilege drop. Past fixes regressed because
they were applied on the live box, not in the repo wrapper/installer.

## Current state (anchors)

- `installer/wrappers/hermes:64-74` — cwd-guard only, **no privilege drop**;
  deliberately does NOT pin `HERMES_HOME` (lines 5-10,40-41).
- `installer/wrappers/hal0-hermes:30,35` — legacy wrapper that DID pin
  `HERMES_HOME`; now symlinks to `hermes`.
- `installer/install.sh:1257` — system gateway install correctly uses
  `env -u HERMES_HOME … hermes gateway install --system --run-as-user hal0`.
- `installer/install.sh:1222` — `hal0 agent install hermes` runs as the calling
  user (root during install) → venv created root-owned.
- `installer/systemd/hal0-agent@.service:31,32,97` — runtime correctly
  `User=hal0`, `ReadWritePaths=… /var/lib/hal0 …`. No `WorkingDirectory=`.
- `src/hal0/agents/hermes_provision.py`: `:126` default home;
  `:447` `_install_venv()` no chown; `:620-626` `_claim_hermes_home()` no chown;
[...]

PR #1942, opened 2026-08-19
fix(perms): make doctor perms converge on a fresh box

thinmintdev:
## ⚠️ OPERATOR REVIEW REQUIRED

This PR touches three things that need a human eye before merge:

- **SECURITY-POSTURE DECISION** — the `secrets/` directory mode moves **0755 → 0700**. See "The decision point" below; this is the operator's call, not the agent's.
- **PERMISSIONS TABLE** — `hal0.install.perms.ownership_table()` is the declarative authority for every path hal0 owns; two rows change value.
- **PRIVILEGED WRAPPER** — the shipped `installer/wrappers/hal0-agentenv` seam (root via `/etc/sudoers.d`) is **NOT modified** by this PR, but it is now *coupled* to the table by a test. The chosen direction was picked specifically so the wrapper stays byte-identical.

## What was broken

`hal0 doctor perms` exited 1 with DRIFT rows on a **freshly installed box**, and its own remediation (`sudo hal0 doctor perms --fix`) could never produce a durably green audit. Three shipped components each declared a different truth for the same paths, so every repair was undone by the next write. A self-audit that can never be green trains operators to ignore it.

The contradiction *is* the defect — this was never a credential leak (`*.env` stays `0600 root:root` in every direction).

## The decision point — secrets/ 0700 vs 0755

Two shipped components disagreed:

| Component | Declared |
|---|---|
| `hal0.install.perms.ownership_table()` | `/var/lib/hal0/secrets` + `secrets/agents` = **0755** |
| `installer/wrappers/hal0-agentenv` (`install -d -m 0700`, run on every `merge-secrets`) | **0700** |

So `--fix` widened to 0755, the next agent secret write re-tightened to 0700, forever.

**Chosen direction: 0700 — move the TABLE to the restrictive side.** Evidence from the code, not preference:

- **systemd** reads the `EnvironmentFile` under `secrets/` **as root**, before dropping to the service user (that is the stated reason the rows are pinned `root:root` in the first place).
- The **`hal0-agentenv` seam** — the only writer of `secrets/agents/<id>.env` — runs **as root** via `/etc/sudoers.d/hal0-agentenv`.
- `hermes_provision._read_secrets_env()` is a **best-effort** read; the files are `0600 root:root`, so an unprivileged caller already failed on the *file*. The directory mode changes nothing for it.
- The one non-root traverser, `installer/wrappers/hermes` (it re-execs as the unprivileged `hal0` user per the #843 run-as-hal0 guard, then sources `secrets/agents/hermes.env`), is already blocked one level down: the file is `0600 root:root`, so `hal0` can `stat`/list the directory but still can't read the file. The directory mode changes nothing for it.

So 0700 costs no reader any access it actually had. What it removes is **local agent-id enumeration** by any unprivileged user on the box — small, but free, and aligned with ADR-0002 (agent credential isolation).

**Consequence of this choice:** the privileged wrapper needs **no change at all** — it was already right. `installer/install.sh` now births both dirs `install -d -m 0700 -o root -g root` instead of `mkdir -p` + leaning on the `--fix` backstop.

If the operator prefers 0755 instead, the reversal is: flip the two `PermRow` modes back, change `install -d -m 0700` → `0755` in the wrapper (**a privileged-wrapper edit**), and invert `test_agentenv_wrapper_and_table_agree_on_the_secrets_mode`. The coupling test keeps either direction honest.

## The three components reconciled

One direction throughout: **the ownership table is the authority; every writer is made to birth what it declares.**

1. **`secrets/` + `secrets/agents/` → 0700** (`perms.py`, `install.sh`). Wrapper untouched. Two follow-ups from review: `installer/install.sh`'s `HF_TOKEN` persistence block births `${SECRETS_DIR}` unconditionally (not `DEV_MODE`-guarded like its prod-only twin in the later "hal0 system user" follow-up) — `install -o root -g root` as an unprivileged `--dev` user aborted the installer under `set -euo pipefail` whenever `HF_TOKEN`/`HUGGING_FACE_HUB_TOKEN` was exported; fixed to the tolerant `install -d` + best-effort `chown` idiom already used a few lines down (review item 1, blocking). And `hermes_provision._merge_env_file`'s bare `path.parent.mkdir(parents=True, exist_ok=True)` would have re-introduced the exact #1896 drift class on a root-run provision hitting an absent `secrets/agents/` — routed through `perms.ensure_shared_dir(..., mode=perms.SECRETS_DIR_MODE)` instead (review item 3). Both secrets `PermRow`s are now `optional=False` so the fresh-install convergence tests actually materialise and check them — previously the default `optional=True` meant `_materialise_fresh_install` skipped both rows and the 0700 change was covered only by the declaration test, never the convergence one (review item 4; `optional` remains inert in `plan()`/`commit()`/`audit_rows()` in production — it only gates the test helper).

2. **`STATE.md` 0644 → declared 0664.** `hermes_provision._atomic_write` did tmp-write + `os.replace`, so the *new inode* was born at the process umask (`0644` under the daemon's `UMask=0022`), silently discarding the `0664` `--fix` had just applied. It now carries the target's existing mode across the rename, or an explicit `mode=`. The render sites pass a shared `RUNTIME_SNAPSHOT_MODE` constant. A rewrite is a content change, not a posture change. **Scope note (review item 7):** the mode-carry applies at all 6 `_atomic_write`/`_atomic_write_if_changed` call sites, not just `STATE.md` — `/etc/hal0/agents/hermes-terminal-tool`, the hermes `config.yaml`, an MCP hosts JSON, and the `SOUL.md`/`AGENTS.md`/`MCP-CLIENTS.md` writers. Previously every rewrite reset to the process umask on each write (a self-healing narrow); now an over-wide *existing* mode on any of those files is perpetuated instead of healed on the next rewrite. Deliberate: none of the six hold credentials, so this is not a leak, and "a rewrite is a content change, not a posture change" is the general principle — it just applies more broadly than the STATE.md example above.

3. **Daemon-created dirs `2755` → declared `2775`.** `Path.mkdir` masks its mode with the umask, and the setgid bit a parent passes down carries group *ownership*, not group-*write*. New `perms.ensure_shared_dir()` mkdirs then chmods **only the components it creates** (so a lazy `mkdir(parents=True)` through a deliberately tighter parent — `secrets/` 0700, `agents/` 0711 — can never widen it) and is **fail-soft on chmod** (an operator's model store may sit on an NFS export that refuses it; a permissions nice-to-have must never break a pull). Applied at every state-dir creation site under `slots/` and `registry/`.

## Known limits of the test coverage (review items 5, 6, 8)

- **Item 5 — one leg of the acceptance test is partly tautological.** `test_stays_green_after_slot_load_model_pull_and_state_render`'s `models/<org>/<repo>` leg calls `perms.ensure_shared_dir` directly rather than a real weights writer (`pull._download_one`'s `final.parent`), so it partly re-proves the helper's own unit test rather than a real call site. The `persist_pull_job` leg IS a real writer. Left as-is: wiring a real multi-GB-shaped download into a unit test isn't a good trade for the marginal coverage.
- **Item 6 — owner/group are out of scope for `_mode_drift`, so "green after mutation" is proved on the mode axis only.** An unprivileged test process can't chown to `hal0`/`root`, so this is a hard ceiling for a unit test, not a gap in this PR. Carried forward explicitly: the rc.7 ct151 validation brief should re-check the owner axis after a slot load and a `STATE.md` render on a real box — `os.replace` births a new inode owned by the writing process, and the table declares `state_owner="hal0"` while `install.sh` comments that `hal0-api` runs as root.
- **Item 8 — fixed.** `perms.ensure_shared_dir`'s out-of-root log line moved WARNING → INFO; a legitimate operator store outside every enumerated mount would otherwise emit one WARNING per directory created on a routine migration/pull, training operators to ignore it (the exact failure mode #1896 itself is about).

## How it was verified

TDD, red first — the red run reproduced the issue's exact drift, e.g. `('…/slots/gpu-rocm-0', '0o2775', '0o2755')`.

New `tests/install/test_perms_converge.py` (12 tests) asserts **green-after-mutation**, which is the issue's own acceptance criterion:

> On a fresh install, `hal0 doctor perms` is green, and stays green after a slot load, a model pull, and a hermes `STATE.md` render.

A fresh-install tree audits clean, then stays clean after a **real** `write_state_atomic` slot load, a **real** `persist_pull_job`, a `models/<org>/<repo>` creation and two `STATE.md` renders — every mutation run under `UMask=0022`, the shipped unit's umask, with the audit re-run after each.

**No drift ROW COUNT is asserted anywhere** (per the issue's trap): that set grows with every slot loaded and every model pulled. The helper reports `(path, declared, observed)` triples so a failure names the offender.

A coupling test parses `hal0-agentenv`'s `install -d -m <mode>` line out of the shipped wrapper and fails if it ever diverges from the table again — the specific way this bug was born.

```
make lint                                  All checks passed!
uv run ruff format --check src tests       1160 files already formatted
scripts/check_sunset.py                    scar markers 192 <= baseline 192; no overdue shims
full pytest suite                          10925 passed, 16 skipped, 1 xfailed (17m21s)
```

## Out of scope / follow-up

- No `CHANGELOG.md` entry — the rc.7 changelog wave collects merged PRs separately.
- `known-issues.yaml: doctor-perms-secrets-dir-0755-is-declared` moves `by-design` → `fixed`, with `still_report_if` inverted to expect 0700, and the entry `id` renamed to `doctor-perms-secrets-dir-0700-is-declared` — the old id asserted the pre-fix truth, and a validation agent pattern-matching it post-fix could mask a genuine 0755 regression as "known". The `regressions.yaml` probe `doctor-perms-never-converges` is left in place as the permanent regression watch.
- Not verified on a live box — this is unit-level evidence only; the rc.7 validation run on ct151 is the real confirmation.

Refs #1896

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1233,7 +1233,7 @@ if [[ -n "${HF_TOKEN_VAL}" ]]; then
     # an explicit rotate signal); with no token set, any existing file here is
     # left untouched — never deleted just because the env var was omitted on
     # a later run.
-    mkdir -p "${SECRETS_DIR}"
+    install -d -m 0700 -o root -g root "${SECRETS_DIR}"  # 0700: see #1896
     HF_SECRETS_TMP="$(mktemp "${HF_SECRETS_ENV}.XXXXXX")"
     cat > "${HF_SECRETS_TMP}" <<EOF
 # HuggingFace token for gated / large model pulls — gathered at install time
@@ -2706,7 +2706,14 @@ else
     # whole /etc/hal0 + /var/lib/hal0 tree to an unprivileged service user so a
     # dropped hal0-api could rewrite config — was removed. hal0-api runs as root,
     # so it writes config/state, applies updates, and restarts services directly.
-    mkdir -p "${ETC_DIR}/agents" "${VAR_DIR}/secrets/agents"
+    mkdir -p "${ETC_DIR}/agents"
+    # secrets/ + secrets/agents are born 0700 root:root, matching BOTH the
+    # ownership table (hal0.install.perms) and the hal0-agentenv seam's
+    # `install -d -m 0700`. A plain `mkdir -p` here births them 0755 under the
+    # installer's umask, which the `doctor perms --fix` backstop then had to
+    # tighten on every run — cosmetic drift the audit reported as a defect
+    # (#1896). Explicit mode = born correct, nothing to reconcile.
+    install -d -m 0700 -o root -g root "${VAR_DIR}/secrets" "${VAR_DIR}/secrets/agents"
 
 fi
 
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1233,7 +1233,15 @@ if [[ -n "${HF_TOKEN_VAL}" ]]; then
     # an explicit rotate signal); with no token set, any existing file here is
     # left untouched — never deleted just because the env var was omitted on
     # a later run.
-    install -d -m 0700 -o root -g root "${SECRETS_DIR}"  # 0700: see #1896
+    # -o/-g dropped from this `install -d`: in --dev this runs as an ordinary
+    # user (dev mode never elevates), so `install -o root -g root` would fail
+    # under `set -euo pipefail` and abort the install whenever HF_TOKEN /
+    # HUGGING_FACE_HUB_TOKEN is exported — the prod-only twin at the
+    # "hal0 system user" follow-up below stays root:root because it sits in
+    # the DEV_MODE=0 branch. Tolerant chown mirrors the HF_SECRETS_TMP idiom
+    # a few lines down: best-effort in --dev, real in a root prod install.
+    install -d -m 0700 "${SECRETS_DIR}"  # 0700: see #1896
+    chown root:root "${SECRETS_DIR}" 2>/dev/null || true
     HF_SECRETS_TMP="$(mktemp "${HF_SECRETS_ENV}.XXXXXX")"
     cat > "${HF_SECRETS_TMP}" <<EOF
 # HuggingFace token for gated / large model pulls — gathered at install time
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -61,7 +61,7 @@ from hal0.agents.anchor_window import (
     resolve_anchor_window,
 )
 from hal0.agents.role_slots import candidate_from_slot_mapping, resolve_role_slots
-from hal0.install.perms import ensure_shared_dir
+from hal0.install.perms import SECRETS_DIR_MODE, ensure_shared_dir
 from hal0.slot_lifecycle_budget import STACK_APPLY_SLOT_ALLOWANCE, slot_lifecycle_timeout_s
 from hal0.slots.state import SlotState, is_dispatchable_state, provider_requires_model
 from hal0.system import seam as _seam
@@ -4822,7 +4822,16 @@ def _merge_env_file(path: Path, updates: dict[str, str]) -> None:
 
     import contextlib
 
-    path.parent.mkdir(parents=True, exist_ok=True)
+    # #1942 review finding 3: a bare `path.parent.mkdir(parents=True,
+    # exist_ok=True)` births an absent parent at the process umask (0755
+    # under UMask=0022) — the sole caller is HERMES_SECRETS_ENV
+    # (secrets/agents/hermes.env), so an absent parent means a fresh box
+    # provisioning before install.sh (or a lost/never-created secrets/agents)
+    # would reintroduce the exact #1896 drift class this PR closed.
+    # `ensure_shared_dir` is umask-proof and only chmod's the components it
+    # creates, so a pre-existing parent (the common case — install.sh seeds
+    # it first) is left untouched.
+    ensure_shared_dir(path.parent, mode=SECRETS_DIR_MODE)
     tmp = path.with_suffix(path.suffix + ".tmp")
     tmp.write_text("\n".join(out_lines) + "\n", encoding="utf-8")
     os.replace(tmp, path)
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -43,6 +43,7 @@ import pwd
 import re
 import shutil
 import sqlite3
+import stat
 import subprocess  # nosec B404 — needed to spawn python -m venv + pip
 import sys
 import tempfile
@@ -60,6 +61,7 @@ from hal0.agents.anchor_window import (
     resolve_anchor_window,
 )
 from hal0.agents.role_slots import candidate_from_slot_mapping, resolve_role_slots
+from hal0.install.perms import ensure_shared_dir
 from hal0.slot_lifecycle_budget import STACK_APPLY_SLOT_ALLOWANCE, slot_lifecycle_timeout_s
 from hal0.slots.state import SlotState, is_dispatchable_state, provider_requires_model
 from hal0.system import seam as _seam
@@ -2723,6 +2725,15 @@ ETC_HAL0_AGENT_SKILLS = ETC_HAL0_DIR / "agent-skills"
 # provision time as root.
 RUNTIME_SNAPSHOT_DIR = Path("/var/lib/hal0")
 
+#: Mode for the runtime snapshots rendered into RUNTIME_SNAPSHOT_DIR. This is
+#: the SAME value the ownership table declares for STATE.md
+#: (``hal0.install.perms.ownership_table``); the two are asserted equal by
+#: ``tests/install/test_perms_converge.py`` so they can never drift apart
+#: again (#1896). Group-writable because the file is written by the hal0
+#: daemon AND by a root-run ``hal0 hermes render-context``, and read by the
+#: session-start hook — one shared group, no owner-dependent access.
+RUNTIME_SNAPSHOT_MODE = 0o664
+
 
 def _latest_env_snapshot(hermes_home: Path) -> dict[str, Any]:
     """Load the env snapshot env_probe wrote (stable ``env.json``, or a legacy
@@ -2753,16 +2764,36 @@ def _render_template(name: str, **vars_: Any) -> str:
     return env.get_template(name).render(**vars_)
 
 
-def _atomic_write(path: Path, content: str) -> str:
-    """Tmp-write + rename for atomicity. Returns the sha256 of content."""
+def _atomic_write(path: Path, content: str, *, mode: int | None = None) -> str:
+    """Tmp-write + rename for atomicity. Returns the sha256 of content.
+
+    The replacement file inherits ``mode`` when given, else the mode of the
+    file it replaces, else the process umask (#1896). A tmp-write + ``rename``
+    otherwise silently RE-MODES its target: the new inode is born at the
+    umask, so the daemon's ``UMask=0022`` rewrote ``STATE.md`` back to ``0644``
+    after every render, discarding the ``0664`` the ownership table declares
+    and ``hal0 doctor perms --fix`` had just applied. That made the self-audit
+    unable to converge — a rewrite is a content change, not a posture change.
+
+    The chmod happens on the TMP file, before the rename, so the target is
+    never observable at the wrong mode.
+    """
     path.parent.mkdir(parents=True, exist_ok=True)
+    if mode is None:
+        with contextlib.suppress(OSError):
+            mode = stat.S_IMODE(path.stat().st_mode)
     tmp = path.with_suffix(path.suffix + ".tmp")
     tmp.write_text(content, encoding="utf-8")
+    if mode is not None:
+        with contextlib.suppress(OSError):
+            os.chmod(tmp, mode)
     os.replace(tmp, path)
     return content_hash(content)
 
 
-def _atomic_write_if_changed(path: Path, content: str) -> tuple[str, bool]:
+def _atomic_write_if_changed(
+    path: Path, content: str, *, mode: int | None = None
+) -> tuple[str, bool]:
     """Atomic write that skips a byte-identical file. Returns ``(sha256, wrote)``.
 
     The convergence signal for the rendered context files: a re-render that
@@ -2774,7 +2805,7 @@ def _atomic_write_if_changed(path: Path, content: str) -> tuple[str, bool]:
                 return content_hash(content), False
         except OSError:
             pass
-    return _atomic_write(path, content), True
+    return _atomic_write(path, content, mode=mode), True
 
 
 def _safe_symlink(target: Path, link: Path) -> bool:
@@ -4185,7 +4216,7 @@ def render_live_context(
     # STATE.md — content-hash gated (ignore the as_of line). Written under the
     # hal0-owned RUNTIME_SNAPSHOT_DIR so render-context works under the User=hal0
     # / ProtectSystem=strict hermes sandbox (#473).
-    RUNTIME_SNAPSHOT_DIR.mkdir(parents=True, exist_ok=True)
+    ensure_shared_dir(RUNTIME_SNAPSHOT_DIR)
     state_path = RUNTIME_SNAPSHOT_DIR / "STATE.md"
     existing = ""
     if state_path.exists():
@@ -4199,7 +4230,7 @@ def render_live_context(
         return out  # state_written=False, hermes_written=False, degraded=True
 
     if _state_body_minus_timestamp(existing) != _state_body_minus_timestamp(new_state):
-        _atomic_write(state_path, new_state)
+        _atomic_write(state_path, new_state, mode=RUNTIME_SNAPSHOT_MODE)
         out["state_written"] = True
     elif reachable:
         # Content unchanged, but we just confirmed it current against a
@@ -4236,7 +4267,7 @@ def render_live_context(
         hpath = RUNTIME_SNAPSHOT_DIR / "HERMES.md"
         out["hermes_path"] = str(hpath)
         if not hpath.exists() or hpath.read_text(encoding="utf-8") != hermes_md:
-            _atomic_write(hpath, hermes_md)
+            _atomic_write(hpath, hermes_md, mode=RUNTIME_SNAPSHOT_MODE)
             out["hermes_written"] = True
     except Exception as exc:  # best-effort; STATE.md already written
         log.warning("hermes_provision.render_live_context_hermes_failed", error=str(exc))
--- a/src/hal0/install/perms.py
+++ b/src/hal0/install/perms.py
@@ -79,6 +79,13 @@ log = logging.getLogger(__name__)
 #: (the daemon, a root-run CLI, the wrapper seams) can write the same tree.
 SHARED_DIR_MODE = 0o2775
 
+#: The mode ``secrets/`` and ``secrets/agents/`` are declared at in the
+#: ownership table below (root:root, root-only traverse+list — #1896). NOT
+#: "shared": callers birthing either dir off the umask-proof path (root-run
+#: provisioners, not just install.sh) pass this explicitly to
+#: :func:`ensure_shared_dir` instead of the group-writable default.
+SECRETS_DIR_MODE = 0o700
+
 
 def ensure_shared_dir(path: Path | str, *, mode: int = SHARED_DIR_MODE) -> Path:
     """``mkdir -p`` whose result the process umask cannot narrow (#1896).
@@ -134,7 +141,14 @@ def ensure_shared_dir(path: Path | str, *, mode: int = SHARED_DIR_MODE) -> Path:
     for created in reversed(missing):
         resolved = created.resolve()
         if not any(_is_within(resolved, root) for root in roots):
-            log.warning(
+            # #1942 review finding 8: INFO, not WARNING — an operator model
+            # store outside every enumerated mount is a supported, expected
+            # configuration (see the CodeQL note above), and a migration or a
+            # multi-directory pull into one can log this once per component
+            # created. WARNING-per-directory on a routine path trains
+            # operators to ignore the log, same failure mode #1896 itself
+            # names for a self-audit that can never be green.
+            log.info(
                 "perms.ensure_shared_dir_outside_hal0_roots path=%s (created, not chmod'ed)",
                 resolved,
             )
@@ -666,24 +680,40 @@ def ownership_table(
         # never be durably green. Reconciled toward the RESTRICTIVE side
         # because nothing outside root has business here: systemd reads the
         # EnvironmentFile as root, the agentenv seam runs as root via
-        # /etc/sudoers.d, and every *.env inside is already 0600 root:root, so
-        # 0700 costs no reader any access it actually had. What it removes is
-        # agent-id ENUMERATION by any local user — small, but free (ADR-0002,
-        # agent credential isolation). See known-issues.yaml
-        # `doctor-perms-secrets-dir-0755-is-declared`, now superseded.
-        PermRow(var_lib / "secrets", "root", "root", 0o700, role="secrets/"),
+        # /etc/sudoers.d. The one non-root traverser, `installer/wrappers/
+        # hermes` (it re-execs as the unprivileged `hal0` user per the #843
+        # run-as-hal0 guard, then sources secrets/agents/hermes.env), is
+        # already blocked one level down — every *.env inside is 0600
+        # root:root, so `hal0` can `stat` the directory but still can't read
+        # the file. 0700 costs no reader any access it actually had. What it
+        # removes is agent-id ENUMERATION by any local user — small, but free
+        # (ADR-0002, agent credential isolation). See known-issues.yaml
+        # `doctor-perms-secrets-dir-0700-is-declared`.
+        PermRow(
+            var_lib / "secrets",
+            "root",
+            "root",
+            SECRETS_DIR_MODE,
+            optional=False,
+            role="secrets/",
+        ),
         # secrets/agents/ — per-agent secret .env files (written by the
         # hal0-agentenv seam, never by the service directly). Pinned root:root
         # like secrets/ itself; the dir mode is 0700 (root-only traverse+list,
         # matching the seam — see above), the per-agent .env files are 0600
-        # (owner-read-only tokens), unchanged by the tightening.
+        # (owner-read-only tokens), unchanged by the tightening. Non-optional
+        # (#1942 review finding 4): the default `optional=True` meant
+        # `_materialise_fresh_install` never created either secrets row, so
+        # the fresh-install convergence tests never exercised the 0700 change
+        # this PR makes — see test_fresh_install_tree_actually_exercises_the_secrets_mode.
         PermRow(
             var_lib / "secrets" / "agents",
             "root",
             "root",
-            0o700,
+            SECRETS_DIR_MODE,
             glob="*.env",
             child_mode=0o600,
+            optional=False,
             role="secrets/agents/ (+ <id>.env)",
         ),
         # benchmarks/ (+ runs/, logs/, server-ab/) — GPU bench artifacts,
--- a/src/hal0/install/perms.py
+++ b/src/hal0/install/perms.py
@@ -95,6 +95,16 @@ def ensure_shared_dir(path: Path | str, *, mode: int = SHARED_DIR_MODE) -> Path:
     through a deliberately tighter parent (``secrets/`` 0700, ``agents/`` 0711)
     can never widen it.
 
+    The ``chmod`` is CONTAINED: it only ever touches a path that resolves
+    inside one of hal0's own roots (:func:`_shared_dir_roots`). Callers derive
+    these paths from request-supplied ids (``persist_pull_job`` ->
+    ``pull_job_file(job.model_id)``, a slot's state dir, ...), and while each
+    caller sanitises its own component, a mode-changing sink must not depend on
+    that being true forever. Same shape as
+    :func:`hal0.config.store.assert_under_store` with ``severity="warn"``: an
+    out-of-tree path is still created (behaviour preserved for callers pointed
+    at an operator's own store) but never re-moded.
+
     Fail-soft on ``chmod``, like :func:`hal0.config.store.finalize_perms`: an
     operator's model store may live on an NFS export that refuses the call, and
     a permissions nice-to-have must never break a pull or a slot load.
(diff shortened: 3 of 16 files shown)


thinmintdev:
## CI note — CodeQL `py/path-injection` (2 new alerts, not a required check)

Triaged, not ignored:

- The **genuinely new sink** the first push introduced was `os.chmod` inside `ensure_shared_dir`. That is now **contained** (7f0dc6c4): each created component is resolved and only chmod'ed if it lands inside `var_lib` / `etc` / `var_log` or a configured model-store mount — the same shape as the existing `config.store.assert_under_store(severity="warn")`. Those 2 alerts are `state: fixed`.
- The **2 remaining** alerts are on `probe.exists()` and `path.mkdir(parents=True, exist_ok=True)`. Those are byte-for-byte the operation each of the 13 call sites already performed on the *identical* path before this helper existed — the taint was relocated, not created. No new capability.
- Containing the `mkdir` as well was **deliberately rejected**: an operator's model store may sit outside every enumerated mount, and hard-failing a multi-GB pull to satisfy a static-analysis rule this tree already tolerates is a worse trade. The reasoning is recorded in-code at `perms.py:112`.
- Context: `main` already carries **71 standing open `py/path-injection` alerts**, and CodeQL is **not** in `main`'s required status checks (`python (3.12)`, `ui`, `γ-suite (chromium)`).

Flagging rather than deciding: if the operator wants zero-new-alerts enforced here, the options are (a) dismiss the 2 as "used in tests / false positive", or (b) contain the `mkdir` too and accept the pull-failure risk. Say which and I'll apply it.

thinmintdev:
## CI status

| check | result |
|---|---|
| `python (3.12)` **(required)** | ✅ pass (25m41s) |
| `ui` **(required)** | ✅ pass |
| `γ-suite (chromium)` **(required)** | ❌ **#1926 timeout, not a regression** |
| `sunset` | ✅ pass |
| `Analyze (python / js-ts / actions)` | ✅ pass |
| `CodeQL` (not required) | ❌ 2 residual alerts — triaged in the comment above |

**γ-suite:** cancelled at `12m17s` with `##[error]The operation was canceled` after `[599/658]` — the known 12-minute wall-clock ceiling tracked as **#1926**. No spec asserted a failure; the run simply ran out of clock. This PR touches zero UI/TS code (`src/hal0/**`, `installer/install.sh`, `tests/**` only), so there is no plausible causal path. Needs a re-run once #1926 lands, or an operator re-run.

**Local evidence** (full suite, twice, on the final tree):
```
10926 passed, 16 skipped, 1 xfailed  (16m14s)
make lint                            All checks passed!
ruff format --check src tests        1160 files already formatted
scripts/check_sunset.py              scar markers 192 <= baseline 192
```

Not merged, automerge not enabled — awaiting the operator on the 0700 security-posture decision.

thinmintdev:
## Independent review — PR #1942 (fix #1896)

Reviewed on `6353a93c` against `99024df2`. Codex was rate-limited, so this is the only deep review; I did not write this code. Everything below was re-derived locally in a detached worktree, not taken from the PR body.

**Verdict: 1 blocking, 8 non-blocking. The core fix is sound and the tests genuinely bind.** The blocker is a one-line installer regression unrelated to the security decision.

---

### ✅ What I verified (not just read)

**Coupling test really parses the shipped wrapper.** Mutated `installer/wrappers/hal0-agentenv` `install -d -m 0700` → `0755` in a scratch tree:

```
FAILED test_agentenv_wrapper_and_table_agree_on_the_secrets_mode
AssertionError: hal0-agentenv installs secrets/ 0755 but the ownership table declares 0700 — that disagreement IS #1896
```

The regex only matches lines carrying `/var/lib/hal0/secrets`, so the two `/etc/hal0/agents` `install -d` lines (94, 111) don't false-match; a switch to a variable mode trips the "no longer creates … with an explicit mode" assert. Good.

**Revert-and-confirm-red.** Neutered `ensure_shared_dir` back to a bare `mkdir(parents=True, exist_ok=True)` and stripped the mode-carry from `_atomic_write`, tests held. Red, with the issue's exact shape:

```
AssertionError: drift after a slot load
assert [('…/var-lib/hal0/slots/gpu-rocm-0', '0o2775', '0o2755')] == []
…
test_atomic_write_preserves_an_existing_files_mode: assert 0o644 == 0o664
```

**No drift-row-count assertion anywhere.** The only `len(...drifted)` in the tree is `tests/install/test_perms.py:67`, a pre-existing single-row synthetic-table test — not the fresh-install set. The trap is respected.

**Suites green on the head:** `tests/install tests/installer tests/cli` → `1867 passed, 1 skipped` (skip = shellcheck absent).

**`_atomic_write` edge probes** (run, not reasoned):

| case | result |
|---|---|
| absent target, no mode, umask 022 / 077 | `0644` / `0600` — umask, as documented |
| absent target, `mode=0664`, umask 077 | `0664` — explicit mode beats a stricter umask ✅ |
| existing `0600`, no mode | `0600` carried ✅ |
| existing `0600`, `mode=0664` | `0664` — widens, but to the *declared* value, and main widened to `0644` anyway |
| target is a symlink | symlink clobbered into a regular file carrying the **link target's** mode; real file untouched. Identical to main (`os.replace` always did this) |

No mode-widening path via the carry.

**`ensure_shared_dir` edge probes:**

| case | result |
|---|---|
| pre-existing `secrets/` 0700, create `agents/` through it | parent stays `0700` ✅ |
| symlink component inside `var_lib` → another in-root dir | resolved, real dir chmod'ed `2775` |
| symlink escaping `var_lib` | created `0755`, **not** chmod'ed, warning logged ✅ |
| dangling symlink as final component | `FileExistsError` — identical to main's `mkdir(exist_ok=True)`, not a regression |
| pre-existing dir at `0755` (upgraded box) | left `0755`, only the created child gets `2775` — healed by `--fix`, correct |

Resolve-then-contain ordering is right: `resolve()` runs after `mkdir` and before the containment test, so a planted symlink is followed *and then* rejected. The residual TOCTOU (swap a just-created dir for a symlink between `mkdir` and `chmod`) is bounded by containment to hal0's own roots and needs pre-existing write access there. Acceptable.

**`install -d` semantics** (coreutils 9.11): `-m` applies to the **final component only** — intermediates get 0755 — so `install -d -m 0700 "${VAR_DIR}/secrets"` can never tighten `/var/lib/hal0` itself. And it **does** re-chmod an already-existing dir, so upgraded rc.6 boxes (secrets/ at 0755) converge on the next `install.sh` run. Both good.

---

### 🔴 BLOCKING — 1

**1. `installer/install.sh:1236` aborts a `--dev` install whenever `HF_TOKEN` is exported.**

```sh
install -d -m 0700 -o root -g root "${SECRETS_DIR}"  # 0700: see #1896
```

This line is **not** inside a `DEV_MODE` guard (the prod-branch twin at 2716 correctly is — it sits in the `else` of the `if [[ "${DEV_MODE}" -eq 1 ]]` at 2655). In `--dev`, `VAR_DIR="${PREFIX}/var/lib/hal0"` under `$PWD/.hal0ai` and the installer runs as an ordinary user. `install -o root -g root` then fails:

```
$ bash -c 'set -euo pipefail; install -d -m 0700 -o root -g root x/sec; echo REACHED'
install: cannot change owner and permissions of ‘x/sec’: Operation not permitted
rc=1        # "REACHED" never printed
```

`install.sh` has `set -euo pipefail` (line 17), so this is a **hard mid-install abort**. The previous `mkdir -p` succeeded for any user. `HF_TOKEN` / `HUGGING_FACE_HUB_TOKEN` exported in the shell is normal for anyone doing ML work, and `bash install.sh --dev` is a documented mode (header line 4), so this is reachable.

Fix — the tolerant idiom is already on the very next lines (1247, `chown root:root … 2>/dev/null || true`):

```sh
install -d -m 0700 "${SECRETS_DIR}"
chown root:root "${SECRETS_DIR}" 2>/dev/null || true
```

or gate on `[[ "${DEV_MODE}" -eq 0 ]]`. Either is a one-liner.

---

### 🟡 NON-BLOCKING — 8

**2. The 0700 evidence chain has one missed non-root traverser — harmless, but the claim needs correcting.**

`installer/wrappers/hermes` re-execs as the **unprivileged `hal0` user** (`hal0_ensure_runas hal0`, the #843 root-clobber guard) and *then* sources `/var/lib/hal0/secrets/agents/hermes.env`. So "No non-root process anywhere in `src/` traverses that tree" is true only because it is scoped to `src/`.

It is genuinely harmless — the file is `0600 root:root` (`_merge_env_file` chmods 0600, `_write_secrets_env` chowns 0,0, and the table pins `child_mode=0o600` root:root), so `[ -r "$HAL0_HERMES_SECRETS" ]` is *already* false for the `hal0` user; the directory mode changes nothing. But the claim as written in the PR body, in the `perms.py:658` comment, and in the new `known-issues.yaml` `why:` is repo-scoped-false. Please reword to something like: *"the one non-root traverser, `installer/wrappers/hermes` after its run-as-hal0 re-exec, is already blocked at the file level by 0600 root:root, so 0700 costs it nothing."* That is the honest version and it survives a future reader grepping `installer/`.

Everything else in the chain checks out:
- (a) `write_gateway_secrets_dropin` is the only thing pointing an `EnvironmentFile=` at the vault, and it targets a **system** unit (`/etc/systemd/system`) — read by PID 1 as root before privilege drop. `hal0-agent@.service` (`User=hal0`) reads `/etc/hal0/agents/%i.env`, **not** the secrets tree. `hal0-systemctl:210` hard-rejects any drop-in `EnvironmentFile` outside the vault. ✅
- (b) `hal0-agentenv` is the only writer of `secrets/agents/*.env` for the unprivileged path (`_privileged_env_write("merge-secrets", …)` over sudoers); the root path writes directly. ✅
- (c) `_read_secrets_env` is genuinely best-effort — bare `except OSError: return {}`, and `ensure_gateway_api_server_key` has an explicit `outcome="unreadable"` branch for exactly the non-root case. ✅
- (d) Repo-wide sweep of `src/`, `installer/`, `scripts/` for the secrets tree: `hal0-systemctl` (validator only), the two hermes wrappers (above), `hal0-agentenv` (root), `hermes_provision` (root or seam), `install.sh` (root), `check-service-env.sh` (operator/root ops script). `/api/secrets` writes `/etc/hal0/api.env`, a different tree entirely. No other reader. ✅

**3. Residual convergence hole of the *same class*, in a file this PR already edits.** `hermes_provision._merge_env_file` (~4823) still does:

```python
path.parent.mkdir(parents=True, exist_ok=True)
```

with no mode. `_write_secrets_env` calls it directly when `os.geteuid() == 0`, so a root-run provision on a box where `secrets/agents` is absent births it **0755 under the installer/daemon umask** — re-creating the exact drift #1896 is about, in the tree this PR just tightened. `install.sh` normally seeds the dir first so it is not the common path, but the seam that was supposed to be the *one* authority (`hal0-agentenv`) is bypassed on the root branch. Worth an `ensure_shared_dir(path.parent, mode=0o700)` (or an `install -d`-equivalent) before this ships, or an explicit note that the root branch is trusted to run only after `install.sh`.

**4. The "fresh-install tree is green" tests never materialise `secrets/` at all.** `PermRow.optional` defaults to `True`, and neither secrets row passes `optional=False`, so `_materialise_fresh_install` (which skips optional rows) never creates them. The 0700 change is covered *only* by the two declaration tests — the convergence tests don't touch it. Not wrong, but the PR body's "A fresh-install tree audits clean" is a narrower claim than it reads.

**5. Step 2 of the acceptance test is partly tautological.** `test_stays_green_after_slot_load_model_pull_and_state_render` calls `perms.ensure_shared_dir(var_lib / "models" / "org" / "repo")` — the helper under test — rather than a real weights writer (`pull._download_one`'s `final.parent`). The `persist_pull_job` leg is a real writer, but it passes `path.parent`, a constant dir. The `models/<org>/<repo>` leg proves only that the helper does what its own unit test already proves.

**6. Owner/group are out of scope in `_mode_drift`, so "green after mutation" is proved on the mode axis only.** The issue's acceptance criterion is `hal0 doctor perms` green, which includes owner and group. This is a real and unavoidable limit of an unprivileged test process, and the PR body does disclaim live-box verification — but it should be carried explicitly into the rc.7 ct151 brief: *re-check the owner axis after a slot load and a STATE.md render*, since `os.replace` births a new inode owned by the writing process and the default table declares `state_owner="hal0"` while `install.sh:2705` comments that `hal0-api runs as root`.

**7. The mode-carry is a posture change beyond STATE.md, unmentioned.** Six call sites now inherit the target's existing mode instead of resetting to umask: `_atomic_write` at 1687 (`/etc/hal0/agents/hermes-terminal-tool`), 2048 (hermes `config.yaml`), 2108 (an MCP hosts JSON), plus the three `_atomic_write_if_changed` callers (SOUL.md, AGENTS.md, MCP-CLIENTS.md). Previously every rewrite *reset* to the umask default — a self-healing narrow. Now an over-wide existing mode is perpetuated forever. None of these files hold credentials, so this is not a leak today, and "a rewrite is a content change, not a posture change" is a defensible principle — but it applies to more than STATE.md and the PR body only frames it as the STATE.md fix.

**8. `ensure_shared_dir`'s containment miss logs at WARNING, per created component.** `execute_migration` to a target not yet in `model_mount_roots()`, or a pull into an operator store outside every enumerated root, will emit one `perms.ensure_shared_dir_outside_hal0_roots` WARNING per directory created — potentially many per pull. Consider `info`/`debug`, or log once per call.

**9. `known-issues.yaml` — semantics fine, the entry *id* is now actively misleading.** `status: fixed` is a new value (the file otherwise uses `by-design` / `open` / `crashed`), but the kit is agent-read prose with no parser, so it's harmless, and the `still_report_if` inversion is exactly right. The problem is the id: `doctor-perms-secrets-dir-0755-is-declared` now asserts the opposite of the truth. `lanes/_shared.md` tells validation agents "do not re-report anything listed there" — an agent that observes 0755 on an rc.7 box and pattern-matches the id could file it as `known` and move on, when 0755 now means **the fix regressed**. Rename to `doctor-perms-secrets-dir-0700-is-declared` (or `-mode-reconciled`).

---

### CodeQL judgment

**Agree with the author's triage; leaving the 2 residual alerts is correct.** No reachable injection:

- The genuinely new sink (`os.chmod`) is contained, and I confirmed containment empirically (escape case above): an out-of-root path is created but never re-moded.
- Of the four `ensure_shared_dir` callers CodeQL flags, `pull.persist_pull_job` and `runner_pull.write_local_marker` pass `path.parent` — a **constant** directory (`_pull_jobs_dir()` / `_runner_images_root()`), with `_sanitise_id()` applied to the id for the leaf filename. Zero taint reaches the helper.
- The taint-carrying ones (`_download_one`'s `final.parent` / `dest.parent`, `execute_migration`'s `dst`) were already `mkdir(parents=True, exist_ok=True)` on the byte-identical path on `main`. Relocated, not created.
- Containing the `mkdir` too would hard-fail a multi-GB pull into an operator store outside every enumerated mount. That's the worse trade; rejecting it was right, and `perms.py:112` records why.

CodeQL is not a required check and `main` carries 71 standing alerts of this rule. No action needed.

---

### Posture opinion for the operator: **0700, keep it**

Independent of the author's reasoning, I'd land on the same answer, for a reason the PR undersells:

1. **Nothing loses access.** Every declared reader is root (systemd PID 1 / root's user manager, the sudoers seam, root-run CLI). The one non-root traverser (finding 2) is already blocked one level down by `0600 root:root`. The access delta is exactly zero.
2. **The alternative edits a privileged root helper to *widen* a credential directory.** Reconciling toward 0755 means changing `hal0-agentenv` — a `/etc/sudoers.d`-reachable root wrapper — so that it stops tightening a secrets dir, purely to satisfy a table. You'd be spending privileged-surface churn to buy a strictly worse posture. 0700 reconciles by touching only declarative Python and one installer line, and leaves the wrapper byte-identical. That asymmetry is the decisive argument.
3. **It closes local agent-id enumeration for free**, which is the thing ADR-0002 is about. Small, but there is no column in which it costs anything.

The 0755 case would only hold if some unprivileged consumer needed to `stat`/list the tree. I looked for one across `src/`, `installer/`, and `scripts/` and there isn't one; the ops helpers that do (`check-service-env.sh`, operator `ls`) run as root.

**Recommend: fix #1 (one line), address #3 or document why not, correct the claim in #2, rename the known-issues id (#9), then merge.** The rest are follow-ups.

thinmintdev:
Thanks for the deep pass — all fixed except items 5/6/8 which were non-blocking by design. Pushed to `42b065d2` (3 commits on top of `6353a93c`).

**1. BLOCKING — installer aborts `--dev` when HF_TOKEN is exported.** Fixed: `install.sh:1236`'s secrets-dir birth dropped `-o root -g root` and now chowns tolerantly (mirrors the `HF_SECRETS_TMP` idiom a few lines down), matching the pattern the prod-only twin already used. Red-first test added to `tests/systemd/test_hf_token_secrets.py::TestPersistence::test_secrets_dir_birth_is_dev_mode_safe` — confirmed it fails against the pre-fix line (`AssertionError: SECRETS_DIR must be created without -o/-g root`), passes after. Commit `0c1233c7`.

**2. "No non-root process traverses secrets/" is repo-scoped-false.** Reworded in the PR body, `perms.py`'s `secrets/` `PermRow` comment, and the `known-issues.yaml` `why:` — now names `installer/wrappers/hermes` (re-execs as `hal0` per the #843 run-as-hal0 guard, then sources `secrets/agents/hermes.env`) as the one non-root traverser, blocked one level down by 0600 root:root. Commit `262f35e4`.

**3. `hermes_provision._merge_env_file` bare `mkdir` reintroduces the drift class.** Confirmed — the sole caller is `HERMES_SECRETS_ENV` (`secrets/agents/hermes.env`), so a root-run provision on a box where that dir is absent would have birthed it 0755 under the daemon's umask. Routed through `perms.ensure_shared_dir(path.parent, mode=perms.SECRETS_DIR_MODE)` (new constant, also now backing both secrets `PermRow`s so the table and this call site can't drift apart). Red-first test in `tests/agents/test_hermes_env_seam.py::test_secrets_env_root_births_absent_parent_at_0700` — failed at 0755 pre-fix, 0700 post-fix. Commit `262f35e4`.

**4. `PermRow.optional` — verified inert in production, not inert for the test helper.** Confirmed: `plan()`/`commit()`/`audit_rows()`/`_expand_row()` never read `.optional` — only the test-only `_materialise_fresh_install` does. So setting `optional=False` on both secrets rows changes zero production behavior, but does make the fresh-install convergence tests actually create and check `secrets/`+`secrets/agents/` — previously only `test_secrets_rows_declare_0700` (a declaration test) covered the 0700 value; the convergence suite silently skipped it. Added `test_fresh_install_tree_actually_exercises_the_secrets_mode`, confirmed red before the `optional=False` flip (`_materialise_fresh_install must create secrets/`), green after. Commit `262f35e4`.

**9. known-issues.yaml id renamed.** `doctor-perms-secrets-dir-0755-is-declared` → `doctor-perms-secrets-dir-0700-is-declared` (your suggested form) — the old id asserted the pre-fix truth and could mask a genuine 0755 regression as "known" under `lanes/_shared.md`'s do-not-re-report rule. Updated the two live cross-references (`regressions.yaml`, `lanes/readonly/cli.md`); left the frozen `reports/v1.0.0-rc.6.md` alone since it's a point-in-time record of rc.6's own (since-superseded) belief, not a live doc. Commit `42b065d2`.

**7. `_atomic_write` mode-carry scope.** Added a sentence to the PR body's item 2: the carry applies at all 6 call sites (`hermes-terminal-tool`, hermes `config.yaml`, an MCP hosts JSON, plus the three `_atomic_write_if_changed` callers for `SOUL.md`/`AGENTS.md`/`MCP-CLIENTS.md`), not just `STATE.md`. Deliberate, no credentials involved, per your framing — documented rather than changed.

**8. WARNING-per-directory log noise — fixed.** `ensure_shared_dir`'s out-of-root log line downgraded to INFO; a routine migration/pull into an operator store outside every enumerated mount shouldn't train operators to ignore the audit, which is the exact #1896 failure mode. Commit `262f35e4`.

**5 & 6 — documented, not changed.** Added a "Known limits of the test coverage" section to the PR body covering both: (5) the `models/<org>/<repo>` acceptance leg partly re-proving `ensure_shared_dir`'s own unit test rather than a real weights writer — left as-is, wiring a real download into a unit test isn't a good trade; (6) owner/group being out of `_mode_drift`'s scope is a hard ceiling for an unprivileged test process, carried forward as an explicit note for the rc.7 ct151 validation brief (re-check the owner axis after a slot load + STATE.md render on a real box).

**Verification:** `make lint`, `uv run ruff format --check src tests`, and `env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/install tests/installer tests/cli tests/agents -q` all green (1868 + 781 passed, 3 skipped for missing shellcheck/venv, 1 pre-existing xfail unrelated to this PR). `scripts/check_sunset.py` clean (scar markers at baseline).

thinmintdev:
## Re-review of 42b065d2 — APPROVE (0 blocking)

Independent verification of the three revision commits (`0c1233c7`, `262f35e4`, `42b065d2`) against the prior review. Ran everything in a throwaway detached worktree; the branch worktree was left untouched.

### 1. The blocker — `install.sh` secrets-dir birth in `--dev` — **FIXED**

Reproduced the prior probe as an ordinary user:

```
$ bash -c 'set -euo pipefail; install -d -m 0700 -o root -g root /tmp/secprobe/old; echo REACHED'
install: cannot change owner and permissions of ‘/tmp/secprobe/old’: Operation not permitted
old-shape rc=1                       # ← the prior failing shape

$ bash -c 'set -euo pipefail; install -d -m 0700 "$SECRETS_DIR"; chown root:root "$SECRETS_DIR" 2>/dev/null || true; echo REACHED'
REACHED
rc=0    →  700 mint:mint             # ← the new shape
```

Both end states check out on a real flow:

* **prod** — `install.sh:273` aborts a `DEV_MODE=0` run that is not uid 0, so the HF block's `install -d -m 0700` + `chown root:root` lands **0700 root:root**; the prod twin at `2724` (`install -d -m 0700 -o root -g root`) sits in the `DEV_MODE=1 ? : else` branch, and the `doctor perms --fix --force` backstop at `2813` (also `DEV_MODE=0`-gated) reconciles the whole table before first daemon start.
* **dev** — `VAR_DIR="${PREFIX}/var/lib/hal0"` (a user-local prefix), so the dir is born **0700 owned by the invoking user**. `root:root` is unattainable and meaningless in a dev tree — the tolerant `chown` is the right idiom, and the table is only ever applied on the prod path.

`bash -n installer/install.sh` → OK.

**Red-first confirmed.** Reverted the hunk to the old shape:

```
FAILED tests/systemd/test_hf_token_secrets.py::TestPersistence::test_secrets_dir_birth_is_dev_mode_safe
```

The test is a shape assertion on the shipped script text rather than a dev-mode install harness — but no dev-mode install harness exists in this repo, and the failure mode is purely lexical (`-o root -g root` under `set -euo pipefail`), so the shape assertion binds exactly the thing that broke. Accepted.

### 2. The "no non-root process traverses `secrets/`" claim — **CORRECTED, and the conclusion holds**

PR body, `perms.py:683-691`, and `known-issues.yaml:821-826` all now name `installer/wrappers/hermes` as the counterexample. Verified the substance at the source: the wrapper guards with `[ -r "$HAL0_HERMES_SECRETS" ]` (`installer/wrappers/hermes:57`), which is **false for the `hal0` user at 0755 and at 0700 alike** (the file is 0600 root:root), so the wrapper silently skips the source in both directions. 0700 genuinely costs it no access it had.

*Nit, non-blocking:* the replacement phrasing — "`hal0` can `stat`/list the directory but still can't read the file" — is itself imprecise post-fix. At 0700 the `hal0` user can `stat` the `secrets/` **inode** (`/var/lib/hal0` is 2775) but can neither list it nor traverse into `agents/`. The argument doesn't depend on it; worth a word if the file is touched again.

### 3. `_merge_env_file` no longer births at umask — **FIXED, red-first confirmed**

Routed through `ensure_shared_dir(path.parent, mode=SECRETS_DIR_MODE)`. Reverted to the bare `path.parent.mkdir(parents=True, exist_ok=True)`:

```
FAILED tests/agents/test_hermes_env_seam.py::test_secrets_env_root_births_absent_parent_at_0700
```

### 4. The secrets rows are genuinely bound by a convergence test — **FIXED**

Two independent mutations, both red:

```
# SECRETS_DIR_MODE 0o700 → 0o755
FAILED tests/install/test_perms_converge.py::test_secrets_rows_declare_0700
FAILED tests/install/test_perms_converge.py::test_agentenv_wrapper_and_table_agree_on_the_secrets_mode
FAILED tests/install/test_perms_converge.py::test_fresh_install_tree_actually_exercises_the_secrets_mode
FAILED tests/agents/test_hermes_env_seam.py::test_secrets_env_root_births_absent_parent_at_0700

# optional=False removed from both secrets rows (back to the default)
FAILED tests/install/test_perms_converge.py::test_fresh_install_tree_actually_exercises_the_secrets_mode
```

So the 0700 declaration is now bound at the *materialised tree*, not only at the declaration table — which is what the #1933 finding asked for.

Also re-verified the PR body's inertness claim independently: `grep -rn '\.optional' src/hal0/` returns only two unrelated hits in `comfyui/orchestrate.py`. `PermRow.optional` is read **nowhere** in `plan()`/`commit()`/`audit_rows()` — only by the test helper `_materialise_fresh_install`. `optional=False` therefore has zero production behaviour change; no risk of `doctor perms` newly reporting an absent `secrets/` as drift.

### 5. Known-issue id rename — **OK**

`doctor-perms-secrets-dir-0700-is-declared`, `status: fixed`, `still_report_if` inverted to expect 0700. Grepped the whole kit for both ids: `regressions.yaml:1127`, `lanes/readonly/cli.md:55`, and `perms.py:691` all point at the new id. The only remaining hit on the old id is `reports/v1.0.0-rc.6.md:243` — a frozen historical report of what was true at rc.6, and the kit has no machine cross-reference validator (`README.md:19` makes `known-issues.yaml` the authority agents read). Correctly left alone.

### 6. No regressions / CI

```
$ env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest \
    tests/install tests/installer tests/cli \
    tests/systemd/test_hf_token_secrets.py tests/agents/test_hermes_env_seam.py -q
1905 passed, 2 skipped, 14 warnings in 37.20s
```

(the 2 skips are `shellcheck not installed` locally — CI has it and both suites pass there.)

Check-runs on `42b065d2`:

| check | conclusion |
|---|---|
| `python (3.12)` | success |
| `ui` | success |
| `γ-suite (chromium)` | success |
| `sunset` | success |
| `Analyze (python / javascript-typescript / actions)` | success |
| `CodeQL` | failure |

The `CodeQL` failure is the alert gate, and the open alerts on this head are exactly the two already-triaged, non-required ones:

```
110  py/path-injection  src/hal0/install/perms.py:138
107  py/path-injection  src/hal0/install/perms.py:132
```

No new alerts.

---

**Verdict: approve — 0 blocking, 1 wording nit (item 2).** Items 5, 6 and 8 from the prior review are accurately carried as documented limits in the PR body, and item 6's owner-axis check is correctly deferred to the rc.7 ct151 run. Cleared to land.

(Posted as a comment rather than a formal approval — GitHub refuses `--approve` from the authoring account.)

PR #1943, opened 2026-08-19
ci: give the required jobs realistic wall-clock bounds

thinmintdev:
Fixes the two CI-config gaps filed from the rc.6 fix-wave merge trains.

**γ-suite (playwright.yml): `timeout-minutes` 12 → 18.** Observed green runs span 8m26s–11m58s on shared runners — the 12-minute cap had no real headroom, and because γ-suite is a **required** status check, a merely slow runner cancelled healthy runs mid-suite with zero test failures. This blocked #1913 (3 attempts), #1934 (2), and #1942 (2). 18 still bounds a genuinely wedged run.

**python (ci.yml): add `timeout-minutes: 45`.** The job previously had no bound; a 41-minute `apt-get` infrastructure stall observed on #1915 would have run to GitHub's 6-hour default without a manual cancel. 45 covers the observed 22–26-minute full-suite runs with slow-runner margin.

Verification: both files pass workflow-syntax validation implicitly on push; the change is two `timeout-minutes` keys and comments — no step, matrix, or required-check semantics touched.

**CI-configuration change — operator review required, no automerge.**

Closes #1926
Closes #1927

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1683, opened 2026-08-08
field-info popup can clip inside overflow:hidden panels on narrow viewports


thinmintdev:
## What

`.field-info-pop` (the `(i)` description popup used in slot/model drawers and settings pages) is `position: absolute` relative to its `.field-info-wrap` icon. On viewports ≤720px it drops below the icon, horizontally centered via `left: 50%; transform: translateX(-50%)`, with `max-width: min(260px, calc(100vw - 48px))`.

That sizing is viewport-aware but not *panel*-aware. Several settings panels (`.s-panel`) set `overflow: hidden`. An icon positioned near the left/right edge of one of these panels (e.g. after a long label like "ROCm / compute capable") can have its popup clipped horizontally by the panel bounds even though it fits the viewport. Likewise, an icon in the panel's last `.s-row` can have its popup clipped at the panel's bottom edge, since the popup renders below the icon within the same clipped box.

## Where

- `ui/src/dashboard.css` — `.field-info-pop` base rule (~line 3163) and the `@media (max-width: 720px)` override (~line 3186)
- Consumers: any `.field-info-wrap` inside an `.s-panel` (settings pages) or other `overflow: hidden` container

## Why not fixed inline in #1638

#1638 widened the popup from a collapsed one-word-per-line box to `width: max-content` (capped at 280px) — a real bug fix, but it makes this pre-existing clipping pattern more likely to trigger since the box is now bigger. A correct fix needs the popup to escape its clipped ancestor, which means either:
- rendering it through a portal positioned via `getBoundingClientRect()` (JS), or
- switching to `position: fixed` with a computed anchor point,

both bigger than the copy/layout scope of #1638. Flagged by Codex review on that PR (2026-08-08); filing here as the tracked follow-up rather than scope-creeping the layout PR.

## Repro

1. Narrow viewport to ≤720px.
2. Open a settings panel with a `.field-info-wrap` icon near the right edge of an `.s-panel` (e.g. a long label like "ROCm / compute capable"), or in the panel's last row.
3. Focus/hover the icon.
4. Popup is visually clipped by the panel's `overflow: hidden` bounds.

PR #1949, opened 2026-08-20
fix(ui-e2e): key the voice-page copy specs on labelled rows, not ordinals

### LINKED ISSUE #1944: flaky γ-suite: voice-page-provider-copy-v3.spec.ts fails intermittently on unrelated PRs (kokoro voice renders '— unset —')
Two documented sightings on PRs that cannot affect the voice page:

1. PR #1923 (backend-only Vulkan retirement), run 95540220943 — `voice-page-provider-copy-v3.spec.ts` failed with TTS voice `af_bella` rendering `— unset —`; passed on re-run; commit-adjacent evidence ruled out the diff (docs-only delta between the failing and passing heads).
2. PR #1943 (workflow-YAML timeout bump only), run 32209679770 — `voice-page-provider-copy-v3.spec.ts:91 'kokoro selection keeps …' ` failed on retry1, 640/658 passed, 7.4m runtime. Trace/screenshot attached on the run.

Pattern suggests a data-load race in the voice picker (voice list not yet resolved when the assertion samples the copy) rather than a product defect — but two sightings in one week on a required check is enough to fix the spec's wait condition or the underlying load race.

Refs #1926 (the other γ-suite reliability tax).

thinmintdev:
Closes #1944

## Root cause

The flake and a second, larger defect share one cause: `voice-page-provider-copy-v3.spec.ts` addressed the TTS panel's controls **by position** and asserted **without waiting for the capabilities probe**.

```js
const voiceOption = ttsPanel.locator('select option').first()
```

`ModelRow` renders a free-text `<input>` while `sel.catalogItems` is empty and a `<select>` once `GET /api/capabilities` delivers the catalog. So that locator points at two different elements depending on timing:

| probe state | first `<select>` in the panel | its first `<option>` |
|---|---|---|
| in flight | Default voice picker | `— use engine default (af_bella) —` |
| resolved | **Model** picker | `— unset —` |

The spec's only wait was `expect(h2).toHaveText('AI Capabilities')`, and `CapabilitiesPage` renders that `<h2>` unconditionally — it is not a data-load gate. Whether the assertion sampled before or after the probe was a pure race, which is why the check went red on #1923 (backend-only) and #1943 (workflow YAML only) and green on re-run.

### Reproduced

Adding a real sync point (`Model` select holding `kokoro-v1`) before the existing assertion reproduces the CI failure verbatim, including the reported string:

```
Locator: …locator('select option').first()
Expected substring: "af_bella"
Received string:    "— unset —"
```

### The bigger half: the guard was not guarding

The same element swap made the qwen3 test **permanently vacuous** — after the probe lands it only ever asked whether `— unset —` contains `af_bella`. Verified by reverting the #1470 fix:

- hardcode `defaultVoiceLabel = "— use engine default (af_bella) —"` → old spec **9/9 green**
- additionally hardcode the speed hint to `Kokoro clamps to 0.5–2.0` and the sample-rate hint to the Kokoro sentence → old spec **6/6 green**

Two of those assertions were vacuous for a second reason: since #1683 the sub-copy is portalled to `<body>`, so a row's own `textContent` can never contain `Kokoro` / `Kokoro clamps`. Those two lines were never updated for the portal.

So the #1470 regression guard could be fully reverted without turning the suite red.

## What changed

**`ui/tests/e2e/specs/voice-page-provider-copy-v3.spec.ts`** — rules the file now follows:

- every control is reached through its labelled `.s-row`, never by ordinal;
- `capabilitiesApplied()` syncs on the Model picker holding the mocked id before any provider-keyed assertion (the provider is `""` until then, so the copy under test is not on screen yet);
- sub-copy is read through `fieldInfoPopup`, not row text;
- every "must not claim X" is paired with the positive statement of what the copy must say instead, so a blank or unrendered element cannot satisfy it.

Two cases added: the qwen3 **live** voice list (the only shape in which the qwen3 placeholder `<option>` renders at all — without it nothing asserts the #1470 default-voice label), and the non-moonshine STT case the old test's title claimed but never exercised.

**`ui/src/dash/settings/pages/capabilities/TtsPanel.jsx`** — the load race the issue suspected is real, just not the cause of the CI flake:

```js
const kokoroish = !sel.model || isKokoro
```

`!sel.model` stood in for "unconfigured slot", but it is equally true while the probe is in flight or has failed. A qwen3tts box therefore rendered Kokoro's seed pack and its `(af_bella)` default for the whole load window — #1470's defect scoped to the load gap. Now gated on `providerKnown = !sel.loading && !sel.errored`; an unconfigured slot still gets the seed pack once the probe has answered. This matches `SttPanel`, which already keys every string off `resolvedProvider` and says "select an STT model to see its language support" when it is unknown.

## Verification

- New in-flight test is **red before** the `TtsPanel` change (`Expected 1 / Received 10` options — the seed pack) and green after. The route is held behind a promise gate, so the in-flight state is a deterministic held state, not a timing window.
- New spec catches all three sabotages the old one missed: default-voice label → `Expected "(Ryan)" / Received "(af_bella)"`; sample-rate hint → `Expected "Qwen3-TTS model at startup"`; speed hint → `Expected "the engine clamps to 0.5–2.0"`.
- `npx playwright test specs/voice-page-provider-copy-v3.spec.ts --repeat-each=6` → 36/36 passed.
- Full γ-suite: **644 passed, 17 skipped, 0 failed** (4.0m) — includes `ui-sweep-a-v3` ("bundled voices (Kokoro v1)") and `capability-catalog-pickers-v3` (failed-probe branch), both of which the `TtsPanel` change touches.
- `npm run lint`, `npm run typecheck`, `npm run test:unit` (127 passed), `npm run build` — all green.
- `ruff check src tests` + `ruff format --check src tests` — clean (no Python touched).

## Out of scope

- No backend files, per the planning blast radius. Nothing here needs hermes; the γ-suite runs against mocked routes only.
- `CHANGELOG.md` untouched (#1545 — one consolidated entry lands at the end of the wave).
- #1926, the other γ-suite reliability tax, is a separate cause and is untouched.
- The same by-ordinal pattern (`panel.locator('select option')` without a row scope) appears in `capability-catalog-pickers-v3.spec.ts:136/174/210`. Those panels have a single select each today, so they are not currently ambiguous — left alone rather than widening this PR.

### DIFF
--- a/ui/src/dash/settings/pages/capabilities/TtsPanel.jsx
+++ b/ui/src/dash/settings/pages/capabilities/TtsPanel.jsx
@@ -75,16 +75,29 @@ export function TtsPanel({ registry }) {
   const resetAll = () => { sel.reset(); setVoice(origVoice); setSpeed(origSpeed); setFormat(origFormat) }
 
   const liveVoices = ttsVoicesQuery.data?.source === "live" ? (ttsVoicesQuery.data.voices || []) : []
-  const kokoroish = !sel.model || isKokoro
-  const options = liveVoices.length > 0
-    ? liveVoices.map(v => {
-        const seed = KOKORO_VOICES.find(k => k.id === v)
-        return { id: v, label: seed ? seed.label : v }
-      })
-    : (kokoroish ? KOKORO_VOICES : null)
-  const srcNote = liveVoices.length > 0
-    ? "voices reported live by the tts slot"
-    : (kokoroish ? "bundled voices (Kokoro v1) · slot offline — list is the seed pack" : "model-specific voice id")
+  // #1944: `!sel.model` used to stand in for "unconfigured slot", but it is
+  // equally true while GET /api/capabilities is still in flight (or has
+  // failed) — so a qwen3tts box rendered Kokoro's seed pack and its
+  // "(af_bella)" default for the whole load window. That is #1470's defect
+  // (Kokoro facts asserted as engine truth) scoped to the gap before the
+  // probe lands. Claim nothing until the provider is actually known; SttPanel
+  // already keys every string off `resolvedProvider` this way. An
+  // unconfigured slot still gets the seed pack once the probe has answered.
+  const providerKnown = !sel.loading && !sel.errored
+  const kokoroish = providerKnown && (!sel.model || isKokoro)
+  const options = !providerKnown
+    ? []
+    : liveVoices.length > 0
+      ? liveVoices.map(v => {
+          const seed = KOKORO_VOICES.find(k => k.id === v)
+          return { id: v, label: seed ? seed.label : v }
+        })
+      : (kokoroish ? KOKORO_VOICES : null)
+  const srcNote = !providerKnown
+    ? (sel.errored ? "capability probe failed — voice list unavailable" : "loading the tts slot's voice list…")
+    : liveVoices.length > 0
+      ? "voices reported live by the tts slot"
+      : (kokoroish ? "bundled voices (Kokoro v1) · slot offline — list is the seed pack" : "model-specific voice id")
   const defaultVoiceLabel = kokoroish
     ? "— use engine default (af_bella) —"
     : isQwen3 ? "— use engine default (Ryan) —" : "— use engine default —"
--- a/ui/tests/e2e/specs/voice-page-provider-copy-v3.spec.ts
+++ b/ui/tests/e2e/specs/voice-page-provider-copy-v3.spec.ts
@@ -14,9 +14,34 @@
  * These specs pin the qwen3tts selection getting qwen3tts-appropriate copy
  * (not Kokoro's), and the kokoro selection keeping the original Kokoro
  * facts as a regression guard.
+ *
+ * #1944 — two defects in this file, one cause. Every control was addressed
+ * by its ordinal inside the panel (`ttsPanel.locator('select option').first()`)
+ * and every assertion ran without waiting for the capabilities probe:
+ *
+ *   - The Model row renders a free-text <input> until GET /api/capabilities
+ *     delivers the catalog, and a <select> afterwards. So that one locator
+ *     pointed at the Default-voice picker before the probe landed (first
+ *     option "— use engine default (af_bella) —") and at the Model picker
+ *     after it landed (first option "— unset —"). The kokoro guard therefore
+ *     passed or failed purely on whether Playwright's first poll beat the
+ *     mocked probe — the intermittent CI red on #1923 and #1943.
+ *   - The same swap made the qwen3 guard permanently vacuous: post-probe it
+ *     only ever asked whether "— unset —" contains 'af_bella'. Reverting the
+ *     #1470 default-voice fix left the suite green.
+ *   - Two more qwen3 assertions were vacuous for a different reason: since
+ *     #1683 the sub-copy lives in a popup portalled to <body>, so a row's
+ *     own text can never contain "Kokoro" / "Kokoro clamps". Reverting those
+ *     two hints to hardcoded Kokoro copy also left the suite green.
+ *
+ * Rules this file now follows: address a control through its labelled row,
+ * never by ordinal; sync on the probe having been applied before asserting;
+ * read sub-copy through fieldInfoPopup; and pair every "must not claim X"
+ * with the positive statement of what the copy must say instead, so a blank
+ * or unrendered element cannot satisfy it.
  */
 import { test, expect, json } from '../fixtures/apiMock'
-import type { Locator } from '@playwright/test'
+import type { Locator, Page } from '@playwright/test'
 
 // #1683: FieldInfoIcon's description now portals to document.body (so
 // overflow:hidden panels can't clip it), so it's no longer a DOM descendant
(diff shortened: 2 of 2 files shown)


thinmintdev:
## Review — approve

Verified independently in a throwaway worktree at the PR head (`8eda4279`), `ui/node_modules` symlinked from another checkout.

### 1. Does it fix the defect #1944 reports? Yes, and the diagnosis is exact.

Confirmed the element-swap mechanism by reading the components rather than trusting the description:

- `ModelRow` (`shared.jsx`) renders `<input>` while `items.length === 0` and `<select>` once the catalog arrives. In `TtsPanel` the Model row precedes the Default-voice row, so `ttsPanel.locator('select option').first()` resolves to the Default-voice placeholder pre-probe and to Model's `<option value="">— unset —</option>` post-probe.
- That is precisely the string the issue reports (`kokoro voice renders '— unset —'`), and it explains the intermittency: the old spec's only wait (`h2` = "AI Capabilities") is rendered unconditionally by `CapabilitiesPage`, so whether Playwright's first poll beat the mocked probe was a coin flip under CI load.

Not an adjacent path — same locator, same assertion, same message.

### 2. Root cause, not symptom. Both halves.

The spec change removes the ordinal dependency and adds a real sync point (`capabilitiesApplied`). The `TtsPanel` change removes the `!sel.model` conflation that made "no model configured" and "probe hasn't answered" the same state. Aligning on `providerKnown` matches what `SttPanel` already does with `resolvedProvider`, so the two voice panels now agree.

### 3. Does the new test fail without the fix? Verified by running it, not by reasoning alone.

Reverted `TtsPanel.jsx` to `9e7c7a48` at the PR head's spec:

```
✘ the TTS voice picker claims no engine facts while the capabilities probe is in flight (#1944)
  Expected: 1
  Received: 10          ← placeholder + the 9-entry Kokoro seed pack
5 passed, 1 failed
```

Also verified the "the guard was not guarding" claim, which is the more valuable half of this PR. Sabotaging the #1470 fix (`defaultVoiceLabel` hardcoded to `— use engine default (af_bella) —`):

- **old spec** → `9 passed` (`--repeat-each=3`). Fully vacuous.
- **new spec** → 2 failed, with the real diff:
  ```
  Expected: "— use engine default (Ryan) —"
  Received: "— use engine default (af_bella) —"
  ```

So the regression guard for #1470 was dead and this PR resurrects it. That is worth more than the flake fix.

The route promise-gate makes the in-flight state a held state rather than a timing window — `/api/capabilities` is `networkFirst: true` in `MOCK_ALLOWLIST`, so the `page.route` handler really does own the response and the gate holds. No `__hal0MockPassthrough` opt-out needed here (the 200 is never substituted).

### 4. Regressions

Only `TtsPanel` changed; nothing else imports it. Full local γ-suite at the PR head: **630 passed / 17 skipped**, plus 14 `board-render-v3` failures that are an artifact of my `node_modules` symlink (Vite `serving allow list` 404s trip that spec's console-error assertion) — unrelated to this diff, and CI's γ-suite is green. `voice-page-provider-copy-v3` `--repeat-each=8` → **48/48**. `npm run typecheck` clean.

### 5. Scope

Two files, both on-topic. No `CHANGELOG.md` hunk. No drive-by refactors. The issue text itself authorises the product-side half ("fix the spec's wait condition **or** the underlying load race"), so `TtsPanel.jsx` is in scope.

### 6. Codex review bot

No reviews and no comments on this PR — the bot posted nothing. Recording that explicitly: there are no bot claims to triage here, and its silence is not an approval.

---

## Nits — none blocking

1. **`sel.errored` blanks a list we already know.** `providerKnown = !sel.loading && !sel.errored`, and TanStack Query keeps `data` while flipping status to error on a *failed refetch*. `useCapabilityApply` invalidates `['capabilities']`, so a transient failure on that post-save refetch now empties the voice picker even though the provider is still known from cached data. Save is already gated on `errored` (#1467), so nothing is lost — but `providerKnown` could key off "we have caps data" rather than query status.

2. **Live voices are suppressed during the probe.** `/api/slots/tts/voices` is provider-independent, but `options = !providerKnown ? [] : …` hides a list that may already have resolved, and renders a `<select>` with only a placeholder — so for that window the operator cannot type a voice id, and then the control flips to a free-text `<input>` for a non-Kokoro slot. Keeping `liveVoices` while neutralising only `defaultVoiceLabel`/`srcNote` would be tighter. Relatedly, `"loading the tts slot's voice list…"` mis-attributes the wait: what is loading is the capabilities probe, not the voices endpoint.

3. **`[]` vs `null` is load-bearing and uncommented.** `options ? <select> : <input>` — the empty array is truthy, which is the only reason the in-flight state renders a select at all. A later cleanup pass that "simplifies" the loading branch to `null` would silently flip the control and quietly change what the new test pins. One line of comment on the `[]` would fix that.

4. **Follow-up worth filing rather than widening this PR.** Agreed with leaving `capability-catalog-pickers-v3.spec.ts` alone, but it carries the same class of defect — `:125` `await expect(ttsRow).not.toContainText('no installed TTS models')` is an unpaired negative with no probe sync, which is exactly the shape that made three assertions here vacuous. Suggest an issue so it isn't re-found by the next validation run.

PR #1950, opened 2026-08-20
fix(updater): relay full child stderr to the journal on the failure path too

### LINKED ISSUE #1940: updater: failure-path stderr keeps only a 4000-char tail — a post-cosign failure with a long cleanup tail loses the verification breadcrumbs
Follow-up from PR #1935's review (non-blocking finding 1): the success path now relays full child stderr to the journal, but rc!=0 relays nothing and keeps only `full_stderr[-4000:]` in the raised error details. A failure occurring AFTER cosign passed (e.g. during activate) with a long reaping tail can push `sha256_ok`/`cosign_verify_ok` out of the retained evidence — the run where that evidence matters most.

Suggested shape from the reviewer: relay full stderr to the journal on failure too, keep the bounded tail in error details — different sinks, not a double-record. Note the existing test_failed_stage_does_not_double_relay_stderr_as_breadcrumbs cements today's asymmetry and will need a deliberate update.

Review: https://github.com/Hal0ai/hal0/pull/1935#issuecomment-5323509110. Refs #1901.

thinmintdev:
## What changed

`UpdateSeam._invoke` (the sudo-seam wrapper around `stage`/`activate`/`discard`) now replays the root helper's **full** captured stderr into the parent's journal via `_relay_child_log` on **both** outcomes, not just success. The bounded `full_stderr[-4000:]` tail that's folded into the raised `UpdateError`'s `details["stderr"]` is unchanged — it's a separate, quick-triage sink, not the durable audit trail.

## Why

PR #1935 fixed the happy-path gap where the child's structured log (its `sha256_ok`/`cosign_verify_ok`/`prepare_ok`/`extract_ok` breadcrumbs) was captured but never relayed anywhere. The reviewer flagged (non-blocking, #1940) that the fix only covered success: a failure kept just the last 4000 chars in the raised error's details, with no journal relay at all. A failure occurring *after* cosign passed (e.g. during `activate`) with a long cleanup/reaping tail can push those verification breadcrumbs out of the retained 4000-char window — exactly the run where the evidence matters most, since it's the one an operator will actually go looking for.

The root cause was that `_relay_child_log(...)` was called only in the success branch, after the `if proc.returncode != 0: raise ...` block returned early. The fix moves the relay call ahead of that branch so it always runs, regardless of outcome.

## Verification

- Reproduced the gap first: `test_failed_stage_also_relays_full_stderr_to_the_journal` and `test_failed_activate_relay_survives_a_long_cleanup_tail_after_cosign` both fail against the pre-fix code (confirmed by stashing the source change and re-running) and pass after it.
- Rewrote `test_failed_stage_does_not_double_relay_stderr_as_breadcrumbs`, which cemented the old asymmetry, into `test_failed_stage_also_relays_full_stderr_to_the_journal` — it now asserts the failure path relays to the journal *and* still carries the bounded tail in the error details (different sinks, not a double-record of the same content).
- Added `test_failed_activate_relay_survives_a_long_cleanup_tail_after_cosign`, mirroring the existing success-path long-tail regression test but for a failure after a >4000-char cleanup tail following the sha256/cosign breadcrumbs.
- `uv run --extra dev pytest tests/updater/ -q` → 296 passed.
- `uv run --extra dev pytest -q` (full suite, `HAL0_HOME` override) → 10932 passed, 16 skipped, 1 xfailed.
- `uv run --extra dev ruff check src tests` → all checks passed.
- `uv run --extra dev ruff format --check src tests` → all files already formatted.
- `uv run --extra dev mypy src/hal0/updater/privileged.py` → no issues.

## Out of scope

CHANGELOG.md is deliberately untouched per the fix-wave convention (#1545) — a consolidated CHANGELOG PR lands at the end of the wave.

Closes #1940

### DIFF
--- a/src/hal0/updater/privileged.py
+++ b/src/hal0/updater/privileged.py
@@ -155,14 +155,19 @@ class UpdateSeam:
 
         ``-n`` is load-bearing: a missing or broken grant fails immediately with
         a non-zero exit instead of prompting for a password inside a daemon.
-        The child's stderr (its structured log) is folded into the raised error
-        so a failure is diagnosable from the job result alone. On success the
-        same stderr is replayed into the parent's own log (see
-        :func:`_relay_child_log`) — the child's structlog output is captured,
-        not inherited, so without this replay every prepare/verify breadcrumb
-        (``sha256_ok``, ``cosign_verify_ok``, ``prepare_ok``, …) the root
-        helper emits is silently discarded on the happy path and never reaches
-        the journal (#1901).
+        The child's stderr (its structured log) is captured, not inherited, so
+        on BOTH the success and failure paths it is replayed in full into the
+        parent's own log (see :func:`_relay_child_log`) — without this replay
+        every prepare/verify breadcrumb (``sha256_ok``, ``cosign_verify_ok``,
+        ``prepare_ok``, …) the root helper emits is silently discarded and
+        never reaches the journal (#1901). A failure occurring AFTER those
+        breadcrumbs (e.g. during ``activate``, after a long cleanup/reaping
+        tail) still needs them in the durable journal, not just in a raised
+        error's details (#1940). The failure path additionally folds a
+        *bounded* stderr tail into the raised error so a failure is
+        diagnosable from the job result alone — that tail is a separate,
+        truncated sink for quick triage, not a substitute for the full
+        journal relay.
         """
         try:
             proc = self._run(  # nosec B603 — fixed argv; every arg re-validated as root
@@ -181,10 +186,20 @@ class UpdateSeam:
             raise _remediation(f"could not execute `sudo -n {self._seam_bin}`: {exc}") from exc
 
         full_stderr = proc.stderr or ""
+        # Relay the FULL stderr to the journal on EVERY outcome, not a
+        # truncated tail and not success-only: a run (successful or not) can
+        # emit many post-verification cleanup lines (stale quarantine/staging
+        # reaping in `_extract_tarball`) after the sha256/cosign breadcrumbs,
+        # and a 4000-char tail can push those breadcrumbs out of the relayed
+        # data — silently reopening the very audit gap this replay closes. A
+        # failure occurring after those breadcrumbs (e.g. during `activate`)
+        # is exactly the run where the evidence matters most (#1940).
+        _relay_child_log(full_stderr, verb=parts[0], job_id=self._job_id)
         if proc.returncode != 0:
-            # Bounded tail only for the error path — a raised error's details
-            # are a diagnostic breadcrumb, not the audit trail this exists to
-            # protect, so truncation here is fine.
+            # The bounded tail here is a SEPARATE sink from the journal relay
+            # above: a quick-triage breadcrumb folded into the raised error
+            # itself, not the durable audit trail. Truncating it is fine
+            # because the full record already reached the journal.
             raise UpdateError(
                 f"privileged update seam failed: {parts[0]} (rc={proc.returncode})",
                 details={
@@ -194,12 +209,6 @@ class UpdateSeam:
                     "seam_bin": self._seam_bin,
                 },
             )
-        # Relay the FULL stderr, not a truncated tail: a successful run can
-        # emit many post-verification cleanup lines (stale quarantine/staging
-        # reaping in `_extract_tarball`) after the sha256/cosign breadcrumbs,
-        # and a 4000-char tail can push those breadcrumbs out of the relayed
-        # data — silently reopening the very audit gap this replay closes.
-        _relay_child_log(full_stderr, verb=parts[0], job_id=self._job_id)
         return _parse_result(proc.stdout or "", verb=parts[0], stderr=full_stderr[-4000:])
 
     # ── preflight ──────────────────────────────────────────────────────────────
--- a/tests/updater/test_privileged_seam.py
+++ b/tests/updater/test_privileged_seam.py
@@ -247,19 +247,71 @@ def test_successful_stage_relay_survives_a_long_cleanup_tail() -> None:
     assert any("cosign_verify_ok" in e["line"] for e in relayed)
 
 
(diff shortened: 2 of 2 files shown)


thinmintdev:
## Review — approve

Verified independently against the issue's repro, not the PR description.

### Does it fix #1940?

Yes, and at the exact path the issue names. `_relay_child_log(full_stderr, ...)` moved ahead of the `if proc.returncode != 0: raise` block, so the relay is now unconditional. That is the root cause — the relay call sitting *after* an early-returning branch — not a symptom patch. There is no deeper cause to chase: the child already emits the breadcrumbs, `_pin_logs_to_stderr` already routes them to stderr, and the parent already captures them; only the replay was gated on rc==0.

The two sinks stay correctly separated: full stderr to the journal, `full_stderr[-4000:]` into the raised `UpdateError`'s `details["stderr"]`. That matches the shape the #1935 reviewer suggested.

### Do the new tests actually fail without the fix?

Confirmed by experiment, not by reading the PR body. Checked out `3eb78820`, reverted **only** `src/hal0/updater/privileged.py` back to the merge base `39a3901e`, and re-ran:

```
FAILED tests/updater/test_privileged_seam.py::test_failed_stage_also_relays_full_stderr_to_the_journal
FAILED tests/updater/test_privileged_seam.py::test_failed_activate_relay_survives_a_long_cleanup_tail_after_cosign
2 failed, 37 passed
```

`assert any("sha256_ok" in e["line"] for e in relayed)` is the failing line in the second one — the exact assertion the issue's scenario turns on. Restored, all 39 pass. The long-tail test uses `activate` with a >4000-char reaping tail *after* the sha256/cosign breadcrumbs, which is the issue's scenario verbatim, and it pins both halves of the contract (journal keeps the breadcrumbs, error detail is allowed to have lost them).

Rewriting `test_failed_stage_does_not_double_relay_stderr_as_breadcrumbs` was necessary, not convenient — it asserted `assert not [...]` on the relay events and cemented the asymmetry the issue asks to remove. The replacement is strictly stronger (asserts the relay *and* the bounded tail), so no coverage was traded away.

### Regressions

`_invoke` has three callers (`stage`, `activate`, `discard`); `updater.privileged_child_log` has exactly one emitter and no consumer anywhere in `src/` — nothing parses or counts those events, so there is no downstream record that can now overflow. The only observable delta for a caller is extra journal lines on a failed routed run. The unrouted path is untouched.

Volume is bounded in practice: `activate_release` runs pip with `capture_output=True`, so pip's output never lands on the child's stderr — the relayed stream is the child's own structlog lines plus one `updater.privileged_failed` line, not a compiler dump.

- `tests/updater/` → 296 passed
- full suite, `HAL0_HOME` override → **10932 passed, 16 skipped, 1 xfailed** (matches the PR's claim exactly)
- `ruff check` / `ruff format --check src tests` → clean (1160 files already formatted)

### Scope

Clean. Two files, 84/23. No CHANGELOG.md hunk (correct for this wave, #1545), no drive-by refactor, no unrelated edits. The comment/docstring churn is all on the lines whose behaviour changed.

### Codex bot

No Codex review, no review comments, no issue comments on this PR — the bot posted nothing. Recording that as silence, not approval; nothing to triage.

---

### Non-blocking

**1. The timeout path still has the exact defect #1940 describes, and worse.**

`_ACTIVATE_TIMEOUT` is 1800s. A wedged `activate` after cosign passed raises from the `except subprocess.TimeoutExpired` arm, whose `details` are `{"verb", "timeout"}` — no relay *and* no stderr at all, so 100% of the evidence is lost rather than all-but-4000-chars. Same class, same "run where the evidence matters most" argument, and it will be re-found by the next validation pass.

Out of the literal scope of #1940 (which says "rc!=0 relays nothing"), so not blocking this PR. Worth its own issue. The data is available — `subprocess.run(capture_output=True)` populates `TimeoutExpired.stderr`, verified:

```
>>> subprocess.run([...], capture_output=True, text=True, timeout=1)
stderr on timeout: b'cosign_verify_ok\n'
```

Note it comes back as **bytes even with `text=True`** — `run()` does not decode on the timeout path — so a fix needs a decode guard, which is why it deserves a deliberate change rather than a one-line copy of this PR's call.

**2. `_relay_child_log`'s own docstring is now stale.** It still reads "so on success it was going nowhere" and closes with "a durable prepare/verify trail exists on the happy path and not only inside a failure's error details." Both clauses describe the success-only behaviour this PR just removed. `_invoke`'s docstring was updated; this one wasn't. In a module that carries this much explanatory prose, the stale version is the one a future reader hits first when they follow the `:func:` reference.

**3. Cosmetic:** `test_failed_stage_also_relays_full_stderr_to_the_journal` exercises `discard`, not `stage`. Inherited from the old test's name, so it isn't a regression — but the rename was the moment to fix it.

thinmintdev:
Holding this PR: the red `python (3.12)` check is **not caused by this branch**.

`tests/api/test_updater_routes.py::test_commit_convergence_is_durable_before_the_restart_that_can_kill_us` (added by #1961) is flaky and is currently red on `main` itself:

- `main` @ `3cb0a71` (#1961's own merge commit), `python (3.12)` job `96404036058` — FAILED, `convergence persisted by then was [None]`. None of this branch's code is present there.
- this branch, `python (3.12)` job `96409328768` — FAILED the same way, `[None, None]`.

Locally it passes: full suite on merged main `49e2cf48` = 10965 passed; full suite on this branch (`b2d06eb6`) = 10966 passed, 0 failed; the test standalone passes 5/5. `pytest-randomly` and `xdist` are both absent, so collection order is deterministic — this is a wall-clock race, not an ordering interaction.

The production ordering #1961 introduced is correct (`job["convergence"] = ...` then `_persist_job(job)` then the restart). The race is in the test harness: `job_id_holder["id"]` is only assigned *after* the POST returns, but `fake_run` — invoked from the background commit job — reads it and falls back to `None` when the job wins the race, appending `None` regardless of what is actually on disk.

Not re-running this hoping it goes green. Tracked with the #1961 author; this PR will merge once `main` is green again.

Issue #1883, opened 2026-08-15
hal0 update via any channel URL fails at manifest verification: the sibling cosign bundle is not proxied (releases.hal0.dev/<channel>.json.bundle → 404)


thinmintdev:
## Summary

`hal0 update` (and a fresh install via `installer/bootstrap.sh` with a channel URL) fails at release-manifest verification for **every channel**, because `releases.hal0.dev` does not serve the manifest's sibling Sigstore bundle, and manifest verification is mandatory with no bypass.

## Evidence

Live proxy state (2026-08-15):

```
$ curl -sS -o /dev/null -w '%{http_code} %{content_type}\n' https://releases.hal0.dev/nightly.json
200 application/json; charset=utf-8
$ curl -sS -o /dev/null -w '%{http_code} %{content_type}\n' https://releases.hal0.dev/nightly.json.bundle
404 application/json; charset=utf-8
$ curl -sS -o /dev/null -w '%{http_code} %{content_type}\n' https://releases.hal0.dev/preview.json.bundle
404 application/json; charset=utf-8
```

Local empirical repro, driving the real client code against the live endpoint:

```python
await _fetch_verified_release_manifest("nightly")
# RAISED: UpdateDownloadError: download returned HTTP 404
# details: {'url': 'https://releases.hal0.dev/nightly.json.bundle', 'status': 404}
```

## Affected code paths (both clients, correct as written)

- `src/hal0/updater/updater.py:942` — `bundle_url = _release_manifest_bundle_url(url)` → `_download(bundle_url, ...)` → `_verify_cosign(...)`. The docstring (~line 861) states verification is always mandatory, no bypass. The 404 raises before cosign is even invoked.
- `installer/bootstrap.sh:452-455` — `fetch_sidecar "release manifest signature bundle" "$(release_manifest_bundle_url ...)"` fetches the same URL.

The tarball's own `bundle_url` (a manifest field, used at `updater.py:3008`) points at GitHub release assets and is **not** affected. Only the manifest's sibling bundle is.

## Root cause

[...]

PR #1956, opened 2026-08-20
docs(release-validation): kit v5 — pre-rc.7 brief hardening

thinmintdev:
Pre-Phase-3 hardening of the validation kit, from the read-only sanity sweep the GA session ran against kit v4 before its first full exercise on rc.7.

## What changed

**Four new checks** closing gaps where recent fixes would otherwise reach rc.7 CI-verified only:

| Brief | Check | Verifies |
|---|---|---|
| `lanes/update/post-upgrade.md` 2b | #1934 gpu-vulkan relabel migration | relabel fired on real stale TOMLs, device-key-only, non-llama untouched, breadcrumbs, idempotency |
| `lanes/update/upgrade.md` 2b | #1883 channel-URL path | `releases.hal0.dev/<channel>.json` + `.json.bundle` both 200, bundle byte-matches, cosign verifies via proxy — the box's pinned GitHub-asset `HAL0_RELEASES_URL` bypasses this path entirely, so without the check rc.7 would never exercise #1883 |
| `lanes/stateful/slots.md` 8b | #1922 output-sanity gate | the gate itself fires (slot lands in `error`, never `ready`), as a product gate distinct from the agents' own coherence canary |
| `lanes/stateful/hermes-e2e.md` 9 | #1942 doctor-perms convergence | rc=0 re-probed after slot load + pull + STATE.md render, plus manual owner/group `stat` — the review proved green-after-mutation on the mode axis only |

**Corrections:**
- `upgrade.md` check 4: now expects the #1935-restored audit trail (populated `job_id`, prepare/verify breadcrumbs); the old text documented their absence as designed behavior.
- `rc-validate.js`: explicit `model:` pins on every phase to match `kit.toml [defaults]` — previously preflight, all lane agents, judgment regressions, and curation silently inherited the invoking session's tier (an opus/fable session would run the whole 25-30-agent sweep at that tier). File-issues deliberately still inherits; README documents this.
- `rc-validate.js` verify prompt: dropped the stale "fresh boxes run the Vulkan lane only" fleet note (stale since #1923).
- `boxes.toml` ct151 gotcha 5: post-rollback `systemctl start hal0.target`, so preflight blockers can name the fix instead of just "API unreachable".
- `kit_version` 4 → 5 per the versioning policy; README changelog entry added.

## Verification

`uv run --extra dev pytest tests/release -q` → 106 passed (includes `test_workflow_contract.py`, which mirrors the kit registry against the workflow script). `node --check` clean on the workflow script.

Out of scope, deliberately: the `lane: preflight` stray value in `regressions.yaml` (harmless, single entry — noted for the next curation pass).

### DIFF
--- a/.claude/workflows/rc-validate.js
+++ b/.claude/workflows/rc-validate.js
@@ -227,7 +227,7 @@ own CONTEXT file.
    Write it for an agent that knows nothing about this fleet. Every later agent depends on it.
 
 Return the schema. Set ready=false only for something that genuinely invalidates the run.
-`, { label: `preflight:${boxId}`, phase: 'Preflight', schema: PREFLIGHT })))
+`, { label: `preflight:${boxId}`, phase: 'Preflight', schema: PREFLIGHT, model: 'sonnet' })))
 
 const ok = preflights.filter(Boolean)
 const blocked = ok.filter((p) => !p.ready)
@@ -267,7 +267,7 @@ YOUR LANE: ${key} (${kind}) on box ${box}
 YOUR BRIEF: ${briefPath(kind, key)} — read it now and work through it.
 ${extra || ''}
 Write your raw notes to ${RUN}/lanes/${box}-${key}.md as you go, then return the schema.
-Set box="${box}" and lane="${key}".`, Object.assign({ label: `${kind}:${key}`, phase: 'Sweep', schema: LANE_RESULT }, opts || {}))
+Set box="${box}" and lane="${key}".`, Object.assign({ label: `${kind}:${key}`, phase: 'Sweep', schema: LANE_RESULT, model: 'sonnet' }, opts || {}))
 }
 
 // serialised chain: each stage is handed the previous stage's box_state_on_exit verbatim
@@ -302,7 +302,7 @@ if (updateBox) {
 // intent judgement stays on the workhorse tier.
 const REG_BATCHES = [
   { tier: 'mechanical', model: 'fable', desc: 'entries with tier: mechanical' },
-  { tier: 'judgment', desc: 'entries with tier: judgment' },  // no model override: inherit session tier
+  { tier: 'judgment', model: 'sonnet', desc: 'entries with tier: judgment' },  // kit.toml regression_judgment tier
 ]
 if (freshBox && wanted('regressions')) {
   for (const b of REG_BATCHES) {
@@ -393,7 +393,7 @@ Rules:
   - If an entry's repro no longer matches the current CLI or API surface, say so.
 
 Return one result object per entry you ran.`,
-    { label: 'regressions:serialized', phase: 'Sweep', schema: REGRESSION_RESULT })
+    { label: 'regressions:serialized', phase: 'Sweep', schema: REGRESSION_RESULT, model: 'sonnet' })
   serializedRegressionResults = (serialResult && serialResult.results) || []
 }
 const laneResults = swept.filter(Boolean).flatMap((r) => (Array.isArray(r) ? r : (r.results ? [] : [r])))
@@ -479,8 +479,9 @@ relitigates it.
 
 Also judge \`other_hardware_applicability\`: would a GPU box, a privileged container, or an
 upgraded (rather than freshly installed) box hit this? No box in the fleet is BOTH a fresh
-install AND has /dev/kfd visible (the fresh boxes lack /dev/kfd and run the Vulkan lane only;
-the boxes with /dev/kfd — update and prod — are not fresh installs), so a read-only cross-check
+install AND has /dev/kfd visible (the fresh boxes lack /dev/kfd, so post-#1923 their LLM slots
+run the CPU lane — the Vulkan LLM lane is retired; the boxes with /dev/kfd — update and prod —
+are not fresh installs), so a read-only cross-check
 against the production box is the best available evidence — use it when the finding touches
 backend selection, profile flags, hardware probing, or native tool calling.
 
@@ -594,7 +595,7 @@ written. Moving work out of this kit and into CI is a success, not a loss.
 
 Return a summary of every file you changed and what you changed in it, plus the promotion list.
 Do not commit — the operator reviews the diff.`,
-  { label: 'curate', phase: 'Curate' })
+  { label: 'curate', phase: 'Curate', model: 'sonnet' })
 
 return {
   version: VERSION,
--- a/.claude/workflows/rc-validate.js
+++ b/.claude/workflows/rc-validate.js
@@ -479,9 +479,11 @@ relitigates it.
 
 Also judge \`other_hardware_applicability\`: would a GPU box, a privileged container, or an
 upgraded (rather than freshly installed) box hit this? No box in the fleet is BOTH a fresh
-install AND has /dev/kfd visible (the fresh boxes lack /dev/kfd, so post-#1923 their LLM slots
-run the CPU lane — the Vulkan LLM lane is retired; the boxes with /dev/kfd — update and prod —
-are not fresh installs), so a read-only cross-check
+install AND has /dev/kfd visible (the fresh boxes lack /dev/kfd, so post-#1923 their
+gpu-rocm LLM slots REFUSE to start — the Vulkan LLM lane is retired and there is no CPU
+fallback for a gpu-device slot; LLM coverage on those boxes means slots explicitly set to
+device=cpu. The boxes with /dev/kfd — update and prod — are not fresh installs), so a
+read-only cross-check
 against the production box is the best available evidence — use it when the finding touches
 backend selection, profile flags, hardware probing, or native tool calling.
 
--- a/tests/release-validation/README.md
+++ b/tests/release-validation/README.md
@@ -66,7 +66,7 @@ lessons*), and monitors each PR to a conclusion.
 | 1 | Read-only sweep | 1 per read-only lane | sonnet | no (pipelines into triage) |
 | 2 | Stateful lane | 1 per stateful lane, **serialised** | sonnet, effort high | serialised by construction |
 | 2b | Update lane | 1–2 on the update box | sonnet, effort high | runs concurrently with 2 (different box) |
-| 3 | Regression probes | 1 per `regressions.yaml` batch | fable (mechanical) / sonnet (judgment) | no |
+| 3 | Regression probes | 1 per `regressions.yaml` batch | fable (mechanical) / sonnet (judgment) | no — except the `serialize: true` subset, which runs sequentially after the whole sweep (its repros mutate box-wide state) |
 | 4 | Triage + dedup | 1 | opus | yes — needs every finding at once |
 | 5 | Adversarial verify | 1 skeptic per candidate | opus, effort high | no |
 | 6 | Synthesis + report | 1 | opus | yes |
@@ -79,6 +79,11 @@ brief that should have contained it. Its output is a diff for the operator to re
 silent commit.
 
 A full two-box run is roughly 25–30 agents. Restrict `lanes` for smaller passes.
+
+Tier labels are enforced in the workflow script via explicit `model:` pins (kit v5 — before
+that, only mechanical/triage/verify/report were pinned and every other phase silently inherited
+the invoking session's tier). The one deliberate exception: File issues (phase 7) still
+inherits the session tier.
 
 ## Process lessons (carried forward, do not relearn)
 
@@ -124,6 +129,24 @@ known-issues, or regressions, and add a line to the changelog below. A report re
 `kit_version` it ran under, so an old report can always be read against the rules it ran with.
 
 ### Kit changelog
+
+* **5** (2026-08-20) — pre-rc.7 brief hardening from the GA-plan handoff, applied before the
+  kit's first v-4-era full run. Four new checks close verification gaps for fixes that would
+  otherwise ship CI-verified only: `post-upgrade.md` 2b (the #1934 gpu-vulkan→gpu-rocm slot
+  relabel on the update box, device-key-only, non-llama untouched, journal breadcrumbs,
+  idempotency), `upgrade.md` 2b (the #1883 channel-URL + `.json.bundle` proxy path, which the
+  update box's pinned GitHub-asset `HAL0_RELEASES_URL` otherwise bypasses entirely),
+  `slots.md` 8b (#1922's output-sanity readiness gate must be verified as a product gate, not
+  merely worked around by the coherence canary), and `hermes-e2e.md` 9 (doctor-perms
+  convergence re-probed after all mutation, plus a manual owner/group `stat` — #1942 proved
+  green-after-mutation on the mode axis only). `upgrade.md` check 4 inverted to expect the
+  #1935-restored audit trail (populated `job_id`, prepare/verify breadcrumbs) instead of
+  documenting their absence as designed. Model tiers are now enforced: every phase's `agent()`
+  call carries an explicit `model:` pin matching `[defaults]` (previously preflight, all lanes,
+  judgment regressions, and curation silently inherited the invoking session's tier). The
+  verify prompt's fleet note no longer claims kfd-less fresh boxes "run the Vulkan lane"
+  (stale since #1923). ct151 gains gotcha 5: post-rollback services need
+  `systemctl start hal0.target`.
 
 * **4** (2026-08-15) — curation after the `v1.0.0-rc.6` run. Landed as mode=report, so the 17
   new entries first carried `issue: null`; the filing PR then filed all of them (plus one more
--- a/tests/release-validation/README.md
+++ b/tests/release-validation/README.md
@@ -70,7 +70,7 @@ lessons*), and monitors each PR to a conclusion.
 | 4 | Triage + dedup | 1 | opus | yes — needs every finding at once |
 | 5 | Adversarial verify | 1 skeptic per candidate | opus, effort high | no |
 | 6 | Synthesis + report | 1 | opus | yes |
-| 7 | File issues (`mode: file`) | 1 | sonnet | — |
+| 7 | File issues (`mode: file`) | 1 | inherits session tier | — |
 | 8 | Kit curation | 1 | sonnet | — |
 
 Phase 8 is what makes the kit compound: it writes back the new known-issues, promotes every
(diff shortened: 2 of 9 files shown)


thinmintdev:
## Independent review — `changes-requested`

Reviewed against the merged fixes on `main` (#1934 `99024df2`, #1935 `31a34606`, #1937 `3fae9b1b`, #1942 `39a3901e`, #1883 closed, #1922 open), the workflow script, and the kit contract. Reproduced the PR's verification: `node --check .claude/workflows/rc-validate.js` clean, `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/release -q` → **106 passed**. Scope is clean — 8 files, all kit/workflow, **no `CHANGELOG.md` hunk**.

The v5 direction is right and several of these checks are exactly the gaps that let #1888-class defects through. But four of the new/edited checks assert something the code does not do, and because this kit *is* the rc.7 ship/no-ship gate, each one converts into either a false pass on a path we specifically wanted covered, or a false finding filed against a fix that is fine.

No Codex bot comments on the PR at review time — noting that explicitly, since silence is not approval.

---

### Blocking

**1. `hermes-e2e.md` 9 — `/var/lib/hal0/hermes/STATE.md` is not a real path.**
STATE.md lives at `/var/lib/hal0/STATE.md`: `RUNTIME_SNAPSHOT_DIR = Path("/var/lib/hal0")` (`src/hal0/agents/hermes_provision.py:2726`), and the ownership table declares `PermRow(var_lib / "STATE.md", …)` (`src/hal0/install/perms.py:745`). There is no `hermes/` subdirectory under `var_lib` anywhere in the tree. The `stat` in the check errors on that argument and exits non-zero, and the agent is left deciding whether a missing STATE.md after a render is itself the finding. Fix: drop `hermes/`.

**2. `post-upgrade.md` 2b — the "before" snapshot it diffs against does not exist.**
The check opens with *"The upgrade stage's `before` snapshot recorded every slot TOML's `device` value — diff against it."* `upgrade.md`'s **Before you touch anything** list records *"every slot: name, assigned model, state, port, profile"* — no `device`, and `hal0 slot list`'s columns are Name/State/Model/Backend/Port/Kind (`src/hal0/cli/slot_commands.py:199-204`), so nothing incidentally captures it either. The primary assertion of the new check is unexecutable as specified. Fix: add `device` to the upgrade lane's snapshot list, or better, `cp -a /etc/hal0/slots/ <run>/before-slots/` so 2b can diff the actual TOMLs (which is what "every other key byte-identical modulo re-serialization" needs anyway).

**3. `post-upgrade.md` 2b — the #1934 relabel will not fire on an rc.6 → rc.7 `hal0 update`, so "it must now read `gpu-rocm`" fails on the release under test.**
`run_post_activation_migrations` is called in-process at Step 7 (`src/hal0/updater/updater.py:3796`), **before** Step 8's `seam.activate` symlink swap (`:3849`). The calling interpreter is the pre-upgrade daemon, and the codebase states this explicitly — `_rerender_units_after_swap`'s docstring (`:3538-3542`): *"the caller's own interpreter is still running the PRE-swap code (Python doesn't hot-swap already-imported modules)"*, which is precisely why the unit re-render was moved to a subprocess and the migrations were not. rc.6 has no `relabel_stale_vulkan_slots`, so on this upgrade the pass does not exist. ct150's `boxes.toml` entry confirms the check is otherwise live (AMD + `/dev/kfd`, and *"slot TOMLs say device=gpu-vulkan"*), so an agent running 2b as written will find `gpu-vulkan` unchanged, no journal lines, and file a "migration didn't run" regression against a correctly-implemented fix.

Two things needed here: (a) the brief must tell the agent what to expect and how to distinguish "migration absent because it runs under the outgoing tree" from "migration broken" — e.g. re-run the activation/`install.sh` post-activation-migration step under the new tree to exercise it; and (b) this looks like a genuine product gap worth its own issue: as shipped, #1934 does not help anyone upgrading *into* rc.7, only out of it, which is not what #1924 asked for. That call is above this PR, but the brief shouldn't encode the optimistic reading.

**4. `upgrade.md` 2b — the verification step cannot verify anything; both halves are no-ops.**
- `hal0 update --check` does **not** cosign-verify. `/api/updates/check` → `check_updates()` (`src/hal0/api/routes/updater.py:624`) → `fetch_release_manifest()` (`src/hal0/updater/updater.py:544`), which only fetches and JSON-decodes. The bundle download + `_verify_cosign` live in `_fetch_verified_release_manifest` (`:915`), reached only from `prepare()` (`:3154`) and `Updater.check()` (`:3632`) — and `Updater.check()` is called from `PUT /api/updates/channel` (`routes/updater.py:987`), never from `GET /check`. So `--check` returning 200 against the channel URL proves the `.json` is reachable and nothing about #1883's `.json.bundle` path.
- Setting `HAL0_RELEASES_URL` in the agent's shell has no effect. `hal0 update` is a thin client over the daemon (`src/hal0/cli/update_commands.py:612`, `api_get("/api/updates/check")`); the value is read from the daemon's environment and, for the privileged stage path, from `$HAL0_HOME/api.env` (`tests/installer/test_hal0_update_wrapper.py:141-152`). As written the agent re-tests the box's pinned GitHub-asset URL and records a pass on the path the check exists to exercise.

Fix: point `HAL0_RELEASES_URL` at the channel URL **in `/etc/hal0/api.env` and restart `hal0-api`**, then drive a path that actually verifies — `hal0 update --channel <ch>` (→ `Updater.check()`) or a real `--target` prepare/stage. Worth also noting in the brief that `releases_url()` appends `?channel=<ch>` to an http override for any non-`stable` channel (`updater.py:492-495`), so the URL the client fetches is not byte-identical to the `curl` target.

**5. Verify prompt: the stale claim was replaced with a different inaccurate one.**
New text: *"the fresh boxes lack /dev/kfd, so post-#1923 their LLM slots run the CPU lane"*. They don't run at all. `require_kfd_for_gpu_slot` gates `gpu-rocm` **unconditionally** (`src/hal0/providers/_gpu.py:129-133` — `gpu-vulkan` is the AMD-scoped case, `gpu-rocm` is *always* gated), every seeded LLM slot ships `device = "gpu-rocm"` (`installer/etc-hal0/slots/{brain,agent,coder,embed,rerank,utility}.toml`), and `install.sh:447` says it outright: *"No /dev/kfd — proceeding CPU-only …; GPU LLM slots will refuse to start."* Only slots explicitly labeled `device = "cpu"` run. This matters because the verify agent uses this note to judge `other_hardware_applicability`: told "fresh boxes run the CPU lane", it will read a refused-slot finding as expected CPU behaviour. Suggested: *"…the fresh boxes lack /dev/kfd, so post-#1923 their seeded GPU LLM slots refuse to load (`require_kfd_for_gpu_slot`) — the Vulkan LLM lane is retired and only `device = "cpu"` slots run…"*

**6. `boxes.toml` still carries the same stale Vulkan claim, in the higher-leverage place.**
The PR fixes the verify prompt but leaves it verbatim in `boxes.toml:38` (ct151: *"It has NO /dev/kfd, so llama.cpp runs the VULKAN lane here"*) and twice in the fleet note — `:174` (*"it only ever exercises the Vulkan lane"*) and `:179` (*"Every box class runs the ROCm lane when /dev/kfd is visible and the Vulkan lane otherwise"*). `boxes.toml` feeds preflight → `CONTEXT.md` → every lane agent, which the README itself says *"is what lane agents obey"*. The PR already edits this file (gotcha 5), so it's in scope, and the changelog's claim that the stale note is gone isn't true until these go too.

---

### Nits

**7.** README self-contradiction on phase 7: the phase table row still reads `File issues | 1 | sonnet`, while the new paragraph three lines below says File issues *"still inherits the session tier"*. `kit.toml [defaults]` declares no tier for it. In a PR whose whole point is making the tier claims true, pick one — pin `file-issues` to sonnet, or change the table row to "inherits session tier".

**8.** `hermes-e2e.md` 9's rationale is factually wrong. `hal0 doctor perms` audits **owner and group and mode** — `PermDiff.changed` is `before.owner != owner or before.group != group or before.mode != mode` (`src/hal0/install/perms.py:830-838`), and `audit_rows` renders `is <owner>:<group> <mode>, want …` (`:1062-1070`). Ownership drift with correct modes reads **red**, not green. What #1942 proved on the mode axis only was its *green-after-mutation tests*, not the tool's coverage. The manual `stat` is fine as independent evidence — keep it — but restate the reason, because as written it teaches the agent a false model of `doctor perms` that will bias how it reads that tool everywhere else.

**9.** `slots.md` 8b degrades gracefully on the "can't force a bad case" axis (good — the explicit *"coverage gap rather than a pass"* is the right instinct), but not on the "#1922 didn't land" axis: an agent finding no gate will file it as a new finding rather than recognising an open GA-blocker. Add: *"if the gate is absent, confirm #1922 is still open and record it as known-open, not a new finding."* Also, #1922 contemplates scoping the gate to `llm`-type slots / models declaring `chat` capability, plus a default-ON opt-out — so the suggested synthetic bad case ("a degenerate/**non-instruct** model file") is exactly the false-positive class the implementation may deliberately exclude. Naming the opt-out env/config as something to locate and check would make 8b robust to either shape.

**10.** `upgrade.md` 4 is correct for the path the lane takes — `hal0 update` drives `/api/updates/prepare` + `/commit`, both now `Updater(channel=channel, job_id=job_id)` (`routes/updater.py:409`, `:456`), and `commit()` threads `self.job_id` into the migrations. Only 10 error-path `updater.*` lines lack `job_id` at all. But `/api/updates/rollback` builds `Updater(channel=channel)` with no job by construction (`routes/updater.py:918`) — so after check 6 the journal legitimately contains `job_id=None` lines. Half a sentence ("rollback has no job; `job_id=None` there is expected") prevents a false `update-audit-trail-gaps` re-open.

**11.** The real event name is `updater.slot_vulkan_relabeled_cpu_fallback`. `_cpu_fallback` parses as a suffix so a grep still works, but for a verbatim-grep instruction spell it out.

**12.** Nothing enforces v5's central claim. `tests/release/test_workflow_contract.py` mirrors the lane registry against the script but does not assert `kit.toml [defaults]` against the script's `model:` pins — so the pins can silently drift right back. ~10 lines of test would make this stick, and it's the difference between "we fixed it once" and "it stays fixed".

**13.** (pre-existing, surfaced by the new README wording) The update-box regression batches (`rc-validate.js:340-360`) have no *"SKIP any entry whose `serialize: true`"* rule — only the fresh-box batch does — and the serialized runner is `freshBox`-only with no lane filter. Today only one `serialize: true` entry exists (`regressions.yaml:1093`) and it isn't update-lane, so it's latent; but the new phase-3 barrier note now reads as if the serialized subset is handled globally.

---

### Verified correct (for the record)

- `kit_version` 4 → 5 per the versioning policy; changelog entry accurately describes what changed.
- `Object.assign({ …, model: 'sonnet' }, opts || {})` — explicit opts win, and `chain()` passes only `{ effort: 'high' }`, so **no conflict with the pins**. Every pin matches `kit.toml [defaults]` (preflight/readonly/stateful/update/regression_judgment/curation = sonnet, mechanical = fable, triage/verify/synthesis = opus).
- *"before that, only mechanical/triage/verify/report were pinned"* — accurate against `main`.
- The phase-3 serialized-barrier README note matches the script: the serialized batch runs after `await parallel(work)`.
- #1883's URL shape is exactly right — `releases_url()` → `https://releases.hal0.dev/{channel}.json` (`updater.py:499`), `_release_manifest_bundle_url()` → `<path>.bundle` (`:502-507`) — and the byte-match-against-the-GitHub-asset assertion is precisely the fix the issue asked hal0-web for.
- #1937 correctly gets no new brief check: it is already covered by `regressions.yaml:983` (`image-status-wrong-podman-store`) plus `readonly/api.md` 6 and `stateful/slots.md` 25.
- ct151 gotcha 5 matches the observed post-rollback behaviour and names the right unit (`hermes-gateway.service`).

thinmintdev:
## Re-review of `6b7f362e..5de011b8` — `changes-requested`, **1 blocking left (one-word fix)**

Re-verified every delta against source. **5 of 6 blocking resolved.** The sixth (B2) is fixed in intent but the replacement names the wrong directory, which lands the check back where it started. Everything else here is ready.

Reproduced the verification: `node --check` clean, `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/release -q` → **111 passed**, `ruff format --check` + `ruff check` clean on the new test. Scope still clean — 9 files, all kit/workflow/test, **no `CHANGELOG.md` hunk**.

---

### Blocking (1)

**B2 — `upgrade.md:18` points at the state directory, not the config directory.**

```
value, read from the TOMLs under `/var/lib/hal0/slots/` directly
```

Slot config TOMLs live in **`/etc/hal0/slots/`**: `slots_config_dir()` is `etc() / "slots"` with the docstring *"Return the slot config directory (/etc/hal0/slots/)"* (`src/hal0/config/paths.py:147-149`), and `relabel_stale_vulkan_slots` — the very migration 2b diffs — globs exactly that (`updater.py:2980`, `slots_dir = slots_config_dir()`). `/var/lib/hal0/slots/` is the per-slot **state** dir: `PermRow(var_lib / "slots", …)` is the recursive `slots/<id>/state.json` row (`src/hal0/install/perms.py:464`, and its comment block above says so). There are no `device` keys under it.

The kit's own corpus already has this right — `known-issues.yaml:105` (*"a model bound in /etc/hal0/slots/agent.toml"*), `:235`, `regressions.yaml:506`, `:563` — and `slots.md:39` correctly uses `/var/lib/hal0/slots/<name>/` for residue, so the two are already used distinctly everywhere else.

As written the agent globs a directory with no TOMLs, records an empty `device` snapshot, and post-upgrade 2b has nothing to diff — the exact failure the B2 fix was for. Change `/var/lib/hal0/slots/` → `/etc/hal0/slots/` and B2 is closed. (`post-upgrade.md` 2b names no directory itself, so this one line is the whole fix.)

---

### Confirmed fixed

**B1 ✓** `/var/lib/hal0/STATE.md` is correct (`hermes_provision.py:2726`, `perms.py:745`). The reframed rationale is now accurate — *"doctor perms does audit all three axes (`PermDiff`)"* matches `perms.py:830-838` — and the independent-evidence-on-a-live-box argument is a better justification than the original one. Good rewrite, not just a correction.

**B3 ✓** #1960 filed, and its body holds up against source: the Step 7 (`updater.py:3796`) vs Step 8 (`:3849`) ordering, the `_rerender_units_after_swap` docstring at `:3538-3542`, and *"only two call sites"* (updater + `install.sh:2467`) all check out. The 2b gate is the right shape — desired end state stated first, then routing, with the important inverse spelled out (*"if it is, the unfired migration is a fresh regression of its own"*) so a fixed #1960 doesn't silently absorb a real failure.

**B4 ✓** and the mechanism is exactly right. `PUT /api/updates/channel` genuinely is the only non-`prepare` path into the cosign-verified fetch: `_fetch_verified_release_manifest` (`updater.py:915`) has two callers — `prepare()` (`:3154`) and `Updater.check()` (`:3632`) — and `Updater.check()`'s sole caller is `set_channel` (`routes/updater.py:987`). Journal evidence exists to confirm against (`updater.cosign_verify_start` / `_ok`, `:886` / `:912`). The api.env copy-aside + byte-for-byte restore + report-it instruction is the right discipline for mutating a live box mid-lane.

**B5 ✓** Matches `require_kfd_for_gpu_slot`'s actual scope (`_gpu.py:129-133`) and `install.sh:447`.

**B6 ✓** ct151's blurb and the fleet-coverage note now describe the post-#1923 posture, and the non-llama carve-out (*"TTS/whisper/ComfyUI may still use Vulkan"*) is correct per `capabilities/catalog.py` and #1934's runtime scoping. The kept sentence about ct151 still exercising *"the refusal path itself"* is a genuinely useful reframe of what that box now covers.

**Nits 7, 9, 12 ✓** README row 7 now reads "inherits session tier"; `slots.md` 8b routes a missing gate to open #1922; the contract test does pin `kit.toml [defaults]` against the script's pins.

---

### Remaining nits (none blocking)

**a.** `ENV_ALLOW_VULKAN_FALLBACK` (`_gpu.py:136-145`) downgrades the refusal to a warning and lets the slot start. The new "REFUSE, no fallback" wording is right for the default, but a box with that env set serves #1888 garbage while the brief says refusal is guaranteed. One clause in the ct151 blurb — "unless `HAL0_ALLOW_VULKAN_FALLBACK` is set, which warns and proceeds" — closes it, and it is worth having precisely because that env is the #1888 failure mode.

**b.** (re-raised, not taken) `releases_url()` appends `?channel=<ch>` to an http override for any non-`stable` channel (`updater.py:492-495`), so the URL the daemon fetches is not byte-identical to the `curl` target in the first half of 2b, and the journal will show the query form. One parenthetical saves a confused correlation.

**c.** (re-raised, not taken) `upgrade.md` 4's *"parent-side `updater.*` lines must carry a populated `job_id`"* — `/api/updates/rollback` builds `Updater(channel=channel)` with no job by construction (`routes/updater.py:918`), so after check 6 the journal legitimately holds `job_id=None` lines. Half a sentence prevents a false `update-audit-trail-gaps` re-open.

**d.** `test_rc_validate_kit_contract.py` pins the tiers that exist but not *completeness* — a newly added unpinned `agent()` call passes all five tests. Counting `agent(` occurrences against pinned ones (minus the deliberate `file-issues` exception) would catch the next instance of v5's motivating defect rather than only its current instances. Worth noting the regexes are formatting-coupled (`schema: X, model: 'y'` ordering), but they fail *closed* on a reformat, which is the right direction.

---

**Verdict: `changes-requested` on one line.** Flip `/var/lib/hal0/slots/` → `/etc/hal0/slots/` in `upgrade.md:18` and I'm at approve — no re-review needed for a one-word change, and nits (a)-(d) are fine to take or leave.

thinmintdev:
Re-review blocker + all four nits addressed in the latest push: the snapshot now reads `/etc/hal0/slots/` (config, the glob the migration walks) with the state-dir distinction spelled out; the fleet note records that the kfd refusal is default-only (`ENV_ALLOW_VULKAN_FALLBACK` downgrades it to a warning and re-opens #1888 — its presence on a box is itself a finding); the channel-URL check compares content not URLs (`?channel=` suffix); rollback-window `job_id=None` lines are documented as legitimate; and the contract test now includes a completeness pass — any newly added labelled `agent()` call without a model pin or whitelist entry fails it. `tests/release`: 112 passed. Landing on green CI per the re-review.

Issue #1739, opened 2026-08-09
P0 regression (#1732): doctor perms --fix dereferences model-layout symlinks, rewriting real files under /mnt/ai-models


thinmintdev:
Found by the pre-GA verification review (reproduced against the real hal0.install.perms module).

#1732 added 'hal0 migrate model-layout --apply' to install.sh Service-start, right before 'hal0 doctor perms --fix --force'. The migration plants symlinks under /var/lib/hal0/models/... → real files under /mnt/ai-models. perms.py walks the models/ PermRow (glob=*, recursive) via rglob, matches those symlinks, and commit()._apply_one calls os.chown/os.chmod with follow_symlinks=True (chmod has no lchmod on Linux) — dereferencing and rewriting the TARGET file owner→hal0:hal0 mode→0644, OUTSIDE hal0's declared tree, on every install/upgrade.

Contradicts migrate_commands.py invariant ('does not write to /mnt/ai-models at all'). No symlink-awareness in perms.py; zero symlink test coverage.

Fix: perms.py commit path must never dereference symlinks (os.chown follow_symlinks=False / lstat / is_symlink skip); add symlink test coverage. Fires unconditionally — no compromise needed.

PR #1958, opened 2026-08-20
fix(install): add perms.py ownership rows for four remaining hal0-api write paths

### LINKED ISSUE #1938: install: four more hal0-api write paths have no perms.py ownership rows — images/cache, dashboard-layout.json, update-jobs, chat_templates
Surfaced during #1895 (PR #1933) write-path enumeration; deliberately left out of that PR to keep its diff minimal for #1896's rebase.

Paths hal0-api can be first-writer of that have no ownership/mode row in src/hal0/install/perms.py:
- images/cache (api/image_cache.py)
- dashboard-layout.json (dashboard/layout_store.py)
- update-jobs (api/routes/updater.py)
- chat_templates store dir (api/routes/chat_templates.py)

Same O13 birth-ownership class as #1895. PR #1933's new parametrized coverage test (test_every_hal0_api_write_path_has_an_ownership_row) is the durable gate — extend it with these four when adding the rows. Should land after #1896 settles the table.

Refs #1895, #1896.

thinmintdev:
## What changed

Adds four missing `ownership_table()` rows in `src/hal0/install/perms.py` for
`var_lib()`-rooted paths `hal0-api` (`User=hal0`) can be the first writer of:

- `images/cache/` (`api/image_cache.py` — the image-gen PNG cache)
- `dashboard-layout.json` (`dashboard/layout_store.py`)
- `update-jobs/` (`api/routes/updater.py` — the durable update-job snapshot store)
- `models/chat-templates/` (`api/routes/chat_templates.py` — the custom chat-template store, nested under the default model store)

Each new row's `child_mode` matches the writer's actual birth mode rather than
fighting it: `0600` for the two `tempfile.mkstemp` + `os.replace` atomic
writers (`dashboard-layout.json`, `update-jobs/*.json`), `0644` for the two
plain-`open()`/`write_text()` writers (`images/cache/*.png`,
`models/chat-templates/*.jinja`) — following the exact convention already
established for `model-pull-jobs/*.json` and `models/*` in this table.

## Why

Same O13 birth-ownership class as #1895: a path with no `ownership_table()`
row is born from whichever process touches it first (often root, e.g. a
future migration or health-check walking `var_lib`), the `User=hal0` daemon
then can't write it, and `doctor perms --fix` has nothing to reconcile
because the table never declared an opinion. These four were surfaced during
#1895's write-path enumeration and deliberately left out of PR #1933 to keep
that diff minimal ahead of #1896's rebase (issue #1938).

The `models/chat-templates/` row deliberately stays a fixed
`var_lib/"models"/"chat-templates"` path (matching every other model-store
row in this table, e.g. `models/`) rather than resolving dynamically through
`paths.model_store_root()` — that avoids coupling `ownership_table()` (a
config-load-free, always-safe-to-call function) to a config-file read, and
matches the table's existing, accepted limitation that an operator-overridden
external model store (`[models].store` pointing outside `var_lib`, the common
production case per the `models/` row's own comment) isn't covered here
either. Not a regression this PR introduces.

## Out of scope

The four writer modules themselves are read-only references — nothing in
`api/image_cache.py`, `dashboard/layout_store.py`, `api/routes/updater.py`,
or `api/routes/chat_templates.py` changed.

## Verification

- Wrote the regression test first: extended
  `test_every_hal0_api_write_path_has_an_ownership_row`'s parametrize list
  with the four new labels, confirmed all four **fail** on the pre-fix table,
  then confirmed they **pass** after adding the rows.
- `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/install/test_perms.py -v` — 49 passed
- `tests/install/test_perms_converge.py` (the #1942 doctor-perms convergence work) + `tests/cli/test_doctor_perms.py` + `tests/security/test_api_env_mode.py` — 91 passed
- Every other test file that references `ownership_table`/`perms.plan`/`doctor perms`/`OwnershipStore` (`test_hal0_toml_lock_scope.py`, `test_diagnosis.py`, `test_doctor_json_diagnoses.py`, `test_loader.py`, `test_config_show_edit_selector.py`, `test_agent_install_hermes.py`, `test_memory_admin_routes.py`) — 201 passed
- Every test file for the four writer modules (`test_dashboard_layout*.py`, `test_chat_templates.py`, `test_updater_routes.py`, `test_updater_slot_drift.py`, `test_updater.py`, `test_chat_template_resolve.py`, `test_container_chat_template.py`) — 240 passed
- `ruff check src tests` — all checks passed
- `ruff format --check src tests` — 1160 files already formatted

The full `pytest tests/ -v` (~11k tests) was not re-run end-to-end locally —
it runs ~55 min serially in this sandbox — but the change is purely additive
(4 new table rows + 4 new parametrize labels, no existing row or writer
touched), and CI runs the full suite on push.

Closes #1938
Refs #1895, #1896

### DIFF
--- a/src/hal0/api/image_cache.py
+++ b/src/hal0/api/image_cache.py
@@ -33,6 +33,7 @@ import uuid
 from pathlib import Path
 
 from hal0.config import paths
+from hal0.install.perms import ensure_shared_dir
 
 log = logging.getLogger(__name__)
 
@@ -55,7 +56,10 @@ _UUID_RE = re.compile(r"^[0-9a-fA-F-]{8,40}$")
 def cache_dir() -> Path:
     """Return ``/var/lib/hal0/images/cache``, creating it on demand."""
     p = paths.var_lib() / "images" / "cache"
-    p.mkdir(parents=True, exist_ok=True)
+    # images/cache/ is declared 2775; a bare mkdir under the daemon's
+    # UMask=0022 births 2755 and re-drifts `doctor perms` after every image
+    # generation (#1896 class, precedent: registry/pull.py persist_pull_job).
+    ensure_shared_dir(p)
     return p
 
 
--- a/src/hal0/api/routes/chat_templates.py
+++ b/src/hal0/api/routes/chat_templates.py
@@ -26,6 +26,7 @@ from pydantic import BaseModel
 
 from hal0.api.middleware.error_codes import BadRequest, Hal0Error
 from hal0.config.paths import model_store_root
+from hal0.install.perms import ensure_shared_dir
 
 router = APIRouter()
 
@@ -131,7 +132,11 @@ async def create_chat_template(body: _TemplateBody) -> dict[str, Any]:
 
     store = _templates_dir()
     try:
-        store.mkdir(parents=True, exist_ok=True)
+        # models/chat-templates/ is declared 2775; a bare mkdir under the
+        # daemon's UMask=0022 births 2755 and re-drifts `doctor perms` after
+        # every custom template write (#1896 class, precedent:
+        # registry/pull.py persist_pull_job).
+        ensure_shared_dir(store)
         (store / f"{body.id}.jinja").write_text(body.content)
     except OSError as exc:
         raise Hal0Error(
--- a/src/hal0/api/routes/updater.py
+++ b/src/hal0/api/routes/updater.py
@@ -47,6 +47,7 @@ from hal0.api.middleware.error_codes import BadRequest, Hal0Error
 from hal0.config import paths
 from hal0.config.loader import hal0_config_txn, load_hal0_config
 from hal0.config.schema import Hal0Config
+from hal0.install.perms import ensure_shared_dir
 from hal0.release.policy import ReleaseKind
 from hal0.updater import (
     Updater,
@@ -236,7 +237,10 @@ def _persist_job(job: dict[str, Any]) -> None:
         return
     path = _job_file(str(job_id))
     try:
-        path.parent.mkdir(parents=True, exist_ok=True)
+        # update-jobs/ is declared 2775; a bare mkdir under the daemon's
+        # UMask=0022 births 2755 and re-drifts `doctor perms` after every
+        # update job (#1896 class, precedent: registry/pull.py persist_pull_job).
+        ensure_shared_dir(path.parent)
         fd, tmp_str = tempfile.mkstemp(prefix=f".{path.name}.", suffix=".tmp", dir=path.parent)
         tmp_path = Path(tmp_str)
         try:
--- a/src/hal0/install/perms.py
+++ b/src/hal0/install/perms.py
@@ -998,8 +998,21 @@ def _expand_row(row: PermRow) -> list[tuple[Path, PermRow]]:
     """
     if row.glob is None:
         return [(row.target, row)]
-    if not row.target.is_dir():
-        return [(row.target, row)]  # the dir itself (absent/optional handled in plan)
+    # #1739-class guard (the PR #1743 fix missed this exact spot): ``Path.is_dir()``
+    # FOLLOWS symlinks, so a row whose ``target`` is itself a symlinked
+    # directory (e.g. an operator symlinking
+    # ``models/chat-templates`` to an external store) would otherwise pass
+    # this check, glob the link's REAL children, and hand them to `commit()`
+    # as ordinary plan entries — `_is_or_is_under_symlink` only guards
+    # children reached THROUGH a symlink *below* `row.target`, it never
+    # checks `row.target` itself, so those children are real paths outside
+    # hal0's declared tree that would get chowned/chmodded. Falling through
+    # to the plain (non-glob) return here means `plan()`/`observe_fn` sees
+    # the symlink via `lstat` and reports it via the existing `is_symlink`
+    # path (never dereferenced, never "changed") — exactly like a declared
+    # row whose own target is a symlink.
+    if not row.target.is_dir() or row.target.is_symlink():
+        return [(row.target, row)]  # the dir itself (absent/optional/symlink handled in plan)
     out: list[tuple[Path, PermRow]] = [(row.target, row)]
     matches = row.target.rglob(row.glob) if row.recursive else row.target.glob(row.glob)
     for child in sorted(matches):
--- a/src/hal0/install/perms.py
+++ b/src/hal0/install/perms.py
@@ -575,6 +575,27 @@ def ownership_table(
             optional=False,
             role="model-pull-jobs/ (durable pull-job snapshots)",
         ),
+        # update-jobs/ — the durable update-job snapshot store
+        # (``api/routes/updater.py`` ``_jobs_dir``/``_persist_job``), which
+        # mirrors the process-local ``app.state.update_jobs`` dict to disk so
+        # a ``hal0-api`` restart mid-apply doesn't 404 the CLI's status poll.
+        # Same O13 birth-ownership class as ``model-pull-jobs/`` immediately
+        # above — surfaced during #1895's write-path enumeration but left out
+        # of that PR to keep its diff minimal (#1938). Optional, unlike
+        # ``model-pull-jobs/``: nothing touches this path during install, so
+        # it is only ever born once an update actually runs. ``_persist_job``
+        # writes each snapshot via ``tempfile.mkstemp`` + ``os.replace``,
+        # which always births the tempfile ``0600`` regardless of umask —
+        # the same convention as ``model-pull-jobs/*.json`` above.
+        PermRow(
+            var_lib / "update-jobs",
+            state_owner,
+            service_group,
+            0o2775,
+            glob="*.json",
+            child_mode=0o600,
+            role="update-jobs/ (durable update-job snapshots)",
+        ),
         # registry/ — the model registry, also born root:root from the same
         # install.sh mkdir and also written by the User=hal0 daemon. Same O13
         # birth-ownership class as slots/ above; heal the dir. registry/ is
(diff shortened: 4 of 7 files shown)


thinmintdev:
## Review: changes requested

Verified against the issue repro, the pre-fix table, and the #1896 convergence contract. The
row declarations themselves are well-reasoned and the new parametrize cases are genuinely red
without the fix (confirmed: checked out `main`'s `perms.py` under this branch's tests → 4
failures, `images/cache/`, `dashboard-layout.json`, `update-jobs/`, `models/chat-templates/`).
Scope is clean: two files, one commit, no CHANGELOG hunk.

But the fix stops at the declaration and never touches the writers, and that reintroduces the
exact non-convergence class #1896 closed. Two blocking items, both of which the Codex bot also
flagged and both of which I confirmed by running the real code.

---

### 1. BLOCKING — `update-jobs/` and `images/cache/` make `hal0 doctor perms` permanently red after first normal use (Codex P2, confirmed)

`ownership_table()` declares both dirs `0o2775`. Their writers create them with a bare `mkdir`:

- `api/routes/updater.py:239` — `path.parent.mkdir(parents=True, exist_ok=True)`
- `api/image_cache.py:58` — `p.mkdir(parents=True, exist_ok=True)`

Under the shipped unit's `UMask=0022` that births `2755` (setgid is inherited, the group-write
bit is not) — which is verbatim component 3 of `tests/install/test_perms_converge.py`'s module
docstring. The repo already has the remedy and the precedent: `registry/pull.py:452` routes
`persist_pull_job` through `perms.ensure_shared_dir` with the comment *"model-pull-jobs/ is
declared 2775; a bare mkdir under the daemon's umask…"*. This PR declares two more 2775 dirs and
skips that step.

Measured, `_mode_drift()` from `test_perms_converge.py` run against a materialised fresh-install
tree under `umask 0o022`, simulating each writer:

```
===== this branch =====
baseline drift:      []
after update-job:    [('/var-lib/hal0/update-jobs',            '0o2775', '0o2755')]
after image gen:     [('/var-lib/hal0/update-jobs',            '0o2775', '0o2755'),
                      ('/var-lib/hal0/images/cache',           '0o2775', '0o2755')]
after chat tpl:      [... , ('/var-lib/hal0/models/chat-templates', '0o2775', '0o2755'),
                            ('/var-lib/hal0/models/chat-templates', '0o2775', '0o2755')]

===== main =====
baseline drift:      []
after update-job:    []
after image gen:     []
after chat tpl:      [('/var-lib/hal0/models/chat-templates',  '0o2775', '0o2755')]
```

So on `main` a `hal0 update apply` and an image generation both leave `doctor perms` green; on
this branch each one leaves a drift row that regenerates after every `--fix`. That is a
user-visible regression on the RC boxes, and it is invisible to CI: `test_perms_converge.py`
passes on this branch (14 passed) because `test_stays_green_after_slot_load_model_pull_and_state_render`
doesn't exercise any of the three new writers.

This is the symptom/root-cause split the wave cares about: declaring a row makes the coverage
test pass, but the birth mode is what actually has to agree. Requested fix:

- route `updater._persist_job`, `image_cache.cache_dir`, and `chat_templates.create_chat_template`
  through `perms.ensure_shared_dir(...)` (the chat-templates one already drifts on `main` via the
  recursive `models/` row — worth fixing in the same pass, it is part of #1938's chat-templates
  item), **and**
- extend `test_stays_green_after_slot_load_model_pull_and_state_render` (or add a sibling) with an
  update-job persist, an image-cache write, and a chat-template write, so this can't come back.

### 2. BLOCKING — the new `models/chat-templates` glob row dereferences a directory symlink (Codex P1, confirmed)

`_expand_row` only guards the *children* (`_is_or_is_under_symlink(child, row.target)`), never
`row.target` itself. `Path.is_dir()` follows the link, so a symlinked `models/chat-templates`
gets globbed and its real, external `*.jinja` files land in the plan as `changed`. Reproduced —
`models/chat-templates` symlinked to a tmp dir containing `custom.jinja`:

```
== this branch ==
  /var-lib/hal0/models/chat-templates            changed=False  (symlink, correctly skipped)
  /var-lib/hal0/models/chat-templates/custom.jinja changed=True  realpath=/tmp/external-…/custom.jinja
== main ==
  (nothing — the recursive models/ row sees the symlink child and drops it)
```

`_apply_one`'s `islink` guard does not help: the *file* isn't a link, only the directory above it
is. So `doctor perms --fix` would chown `hal0:hal0` and chmod `0644` a file in the operator's own
tree — the thing the module docstring (lines 48–54, #1739) promises it never does. Main handles
this case correctly today, so this is a regression the diff introduces, even though the general
form of the hole is pre-existing (a symlinked `var_lib/models` already leaks the same way through
the `models/` row — I verified that too).

Two-line root fix in the file you're already touching, which closes the pre-existing form as well:

```python
if not row.target.is_dir() or row.target.is_symlink():
    return [(row.target, row)]
```

plus a test that a symlinked glob-row target contributes no child diffs. If you'd rather keep this
PR to declarations only, that's defensible — but then it needs its own issue filed before merge,
not a TODO.

---

### 3. nit — the `models/chat-templates` row is functionally a no-op

The recursive `models/` row directly above already plans that directory at `2775` and its
`*.jinja` children at `0644` — identical owner/group/mode. Verified: on `main`, `plan()` already
emits

```
/var-lib/hal0/models/chat-templates             0o2775  models/ (default model store, recursive) :: chat-templates
/var-lib/hal0/models/chat-templates/foo.jinja   0o644   models/ … :: chat-templates/foo.jinja
```

The only behavioural delta the new row produces (beyond finding 2) is a duplicated `doctor perms`
audit line per path. Harmless for `commit`/rollback — both diffs carry the same pre-apply `before`
snapshot — but noisy. The real chat-templates gap is the non-default `[models].store` case, which
the PR body correctly says it does not cover, so for the default case this row exists only to
satisfy an exact-target assertion. Consider making
`test_every_hal0_api_write_path_has_an_ownership_row` accept coverage-by-ancestor (a recursive row
whose target is a parent) instead of forcing a redundant row — otherwise the gate will keep
manufacturing duplicates for every future path under `models/`.

### 4. nit — `var_lib/images` itself still has no row

`cache_dir()`'s `mkdir(parents=True)` creates `images/` too, and the enumeration this issue asks
for leaves it undeclared. Benign at runtime (a root-born `images/ 0755` is still traversable, so
the daemon can write inside a healed `images/cache`), but it's one dir short of the coverage
claim.

---

Codex triage: P1 confirmed (finding 2), P2 confirmed (finding 1). No other bot comments on the PR.

thinmintdev:
## Delta re-review of `7065bf0b` — 1 blocking remains

Re-reviewed the revision commit against the two blocking items from my review of `028e578c`.
Everything below was run, not read: each hunk was reverted in isolation and the relevant test
re-run to confirm it is genuinely red-first, and the highest-risk claim (B2's symlink shape) was
traced all the way to the `chown`/`chmod` seam rather than to the planner.

| item | verdict |
|---|---|
| **B2** — glob row target symlink deref | ✅ **closed**, verified to the actual seam |
| **B1** — declared `2775` dirs born `2755` | ⚠️ **2 of 3 writers fixed**; the chat-templates leg is a no-op in production |
| red-first claims | ✅ independently confirmed, all four assertions |
| scope / CI | ✅ 6 files, 2 commits, no CHANGELOG hunk, nothing unrelated; 8/8 checks green on `7065bf0b` |

---

### ✅ B2 is properly closed — the "one layer down" worry does not materialise

My concern with `return [(row.target, row)]` was that the row is still *planned* against the
symlink path, and `os.chmod` has no `AT_SYMLINK_NOFOLLOW` on Linux — so the dereference could
just have moved one layer down into `commit()`. Traced it with spy seams instead of assuming:

```
planned paths: ['…/models/chat-templates']
  is_symlink=True  changed=False  mode_declared=0o2775  mode_observed=0o777
drifted:        []
commit applied: []
seam calls:     []                      # ← no chown, no chmod, ever
real dir mode after:  0o700             # untouched
real file mode after: 0o600             # untouched

-- force _apply_one directly on the symlink path --
seam calls when forced: []              # islink hard-refusal holds
```

Two independent layers, both verified:

1. `observe()` is `lstat`-based → `is_symlink=True` → `PermDiff.changed` returns `False`
   unconditionally (`perms.py:920`) → the diff never enters `plan_.drifted` → `commit()` never
   reaches it.
2. `_apply_one`'s `islink` guard (`perms.py:1090`) refuses even when the tuple is forced through
   directly — the #1743 last-line guard covers this shape as designed.

A symlinked target that is *not* a directory, or dangling, short-circuits on the existing
`is_dir()` half of the condition, so the guard is total. The recursive `models/` row's
pre-existing form of the same hole is closed by the same line, as claimed.

**Red-first confirmed independently.** Reverting only the guard:

```
>       assert [d.path for d in pl.diffs] == [linked_target]
E       AssertionError: Left contains one more item:
E         …/models/chat-templates/custom.jinja
```

**On your question 3** — the test does go past planning: it runs `perms.commit(pl)` and asserts
`applied == []` plus the real dir's `0o700` and the real file's `0o600` are untouched. So it
*would* catch a layer-down dereference. Caveat worth recording: in the red run the planning
assertion fires first, so the apply-level assertions are never independently exercised red. That
is a cosmetic ordering point, not a hole — I proved the apply path separately above.

---

### ⚠️ 1. BLOCKING (downgradable — see the severity note) — the chat-templates leg of B1 never runs in production

`create_chat_template` is **not** the first creator of `models/chat-templates/`. The daemon is.
`create_app()` calls `seed_chat_templates()` at startup (`src/hal0/api/__init__.py:2606`), and
that function still bare-mkdirs the exact declared path:

```python
# src/hal0/templates/__init__.py:32-33
dst = Path(model_store_root()) / "chat-templates"
dst.mkdir(parents=True, exist_ok=True)          # ← the only remaining bare mkdir on a declared path
```

It is the **only** one left — I enumerated every creation site of all four declared paths:

```
$ grep -rn mkdir  updater.py image_cache.py chat_templates.py layout_store.py templates/__init__.py images.py dashboard_layout.py
src/hal0/templates/__init__.py:33:  dst.mkdir(parents=True, exist_ok=True)     ← BARE
src/hal0/dashboard/layout_store.py:60:  path.parent.mkdir(...)                 ← var_lib itself (declared, install.sh-created)
… the other three are now the ensure_shared_dir comments
```

And `ensure_shared_dir` **cannot** heal it after the fact — by deliberate design it only chmods the
components *that call* created (`perms.py:129-141`, locked down by
`test_ensure_shared_dir_never_widens_an_existing_dir`). Once the startup seed has birthed the dir
`2755`, `create_chat_template`'s `ensure_shared_dir(store)` is a pure no-op.

Reproduced in production order (fresh-install tree, `umask 0o022`, default `[models].store`):

```
baseline drift:                      []
after startup seed_chat_templates(): [('/var-lib/hal0/models/chat-templates', '0o2775', '0o2755')]
after create_chat_template()       : [('/var-lib/hal0/models/chat-templates', '0o2775', '0o2755')]
                                      ^ the new fix fires and changes nothing
```

The new test passes only because it calls `create_chat_template` into a tree where
`models/chat-templates` does not exist yet — an ordering the daemon can never take, since the seed
always runs first at `create_app()`. So `test_stays_green_after_update_job_image_cache_and_chat_template_writes`
is a **false green** for that leg, and will mask this permanently.

Fix is two lines:

```python
# src/hal0/templates/__init__.py
from hal0.install.perms import ensure_shared_dir
...
ensure_shared_dir(dst)          # instead of dst.mkdir(parents=True, exist_ok=True)
```

plus, in the new converge test, call `seed_chat_templates()` **before** `create_chat_template(...)`
so the asserted order matches the daemon's. Without that second half the test still can't see the bug.

**Severity note, stated honestly so you can make the call:** this is an *incomplete* fix, not a
regression this PR introduces — `main` drifts the same path the same way through the recursive
`models/` row (I verified that in the first review). It is also self-limiting: one
`doctor perms --fix` heals it permanently (the seed's `mkdir` is a no-op once the dir exists, so it
does not regenerate), and it only bites boxes on the **default** `[models].store`; a box pointing
`[models].store` at an external mount puts chat-templates outside this row entirely. Landing rc.7
with this as a follow-up issue is defensible. What I would not leave as-is is the test: it asserts a
sequence production never executes.

---

### ✅ 2. B1 is correctly fixed for the other two writers

`ensure_shared_dir` does birth the declared mode regardless of umask — it collects the
not-yet-existing components deepest-first, `mkdir -p`s, then `os.chmod`s each created component
(`chmod` is not umask-masked), containing the chmod to hal0's own roots. Group ownership is not
chowned and does not need to be: every new dir lands under a `2775` setgid parent (`var_lib` /
`models`), so the gid is inherited as `hal0`, which is what the rows declare. A root-created dir
under a non-setgid parent would still land `root:root` — but that is precisely the residual case
the table + `doctor perms --fix` exists to reconcile, and it is unchanged from the
`registry/pull.py persist_pull_job` precedent.

**Each of the three converge assertions is genuinely red-first** — verified by reverting each
writer *individually* (not all three at once), so no assertion is riding on another's failure:

```
reverted updater.py         → E AssertionError: drift after an update-job persist
reverted image_cache.py     → E AssertionError: drift after an image-cache write
reverted chat_templates.py  → E AssertionError: drift after a chat-template write
```

Full head-of-branch run: `tests/install/test_perms.py` + `test_perms_converge.py` → **65 passed**.

---

### ✅ 3. `dashboard-layout.json` was correctly out of B1's scope

It is a FILE row, and `layout_store._write_json_atomic` writes via `tempfile.mkstemp` + `os.replace`
— `mkstemp` always births `0600` regardless of umask, and `os.replace` preserves it. Verified:

```
dashboard-layout.json born: 0o600   declared: 0o600
drift: []
```

Its parent is `var_lib` itself, which is a declared non-optional row created by `install.sh`, so its
bare `path.parent.mkdir` is not reachable in practice. No writer change was needed here — correct
call.

---

### ✅ 4. Scope and CI

Six files, two commits, both on-topic (`028e578c` declarations, `7065bf0b` the review fixes), no
`CHANGELOG.md` hunk, nothing unrelated in the delta. No import cycle introduced
(`hal0.install.perms` only pulls `hal0.config.paths`). `cache_dir()` is on the image read/serve
path and now costs one extra `stat` per call when the dir exists — negligible. Fail-soft is
preserved at all three call sites (`ensure_shared_dir`'s `mkdir` raises `OSError`, which each
caller's existing `except OSError` already handles).

`gh pr checks 1958` on `7065bf0b`: **8/8 green** (python 3.12 full suite 24m6s, γ-suite chromium,
ui, sunset, CodeQL ×4).

---

### Carried forward, unchanged (both were nits in the first review, neither blocks)

- **nit 3** — the `models/chat-templates` row is still functionally a duplicate of the recursive
  `models/` row for the default store; it exists to satisfy
  `test_every_hal0_api_write_path_has_an_ownership_row`'s exact-target assertion and adds one
  duplicated `doctor perms` audit line per path. Coverage-by-ancestor in that gate would stop
  manufacturing these.
- **nit 4** — `var_lib/images` itself still has no row. Slightly changed by this PR in a benign
  direction: `ensure_shared_dir` now births it `2775` rather than `0755`, but it stays undeclared
  so neither mode registers as drift.

thinmintdev:
**GA-session note:** the revision agent completed the seeder fix on disk but the session cycled before it could commit — recovered verbatim from its worktree, verified, and pushed as `929e9f40`. Verification (mine, independent): red run against the unfixed seeder (main's `templates/__init__.py` + the reordered test) fails exactly as prescribed; with the fix, `tests/install/test_perms.py` + `test_perms_converge.py` = 65 passed; ruff check + format clean. The diff matches the delta re-review's prescription exactly (ensure_shared_dir at the startup seed + production-order test), scope is those two files only. Landing on green CI per the re-review's "everything else stands" verdict.

Issue #1948, opened 2026-08-20
runner: restore a correct Vulkan backend and refresh upstreams


thinmintdev:
Tracking issue for the Vulkan-restoration workstream (operator directive 2026-08-20, "I want to include vulkan"). Supersedes the retire-only posture of #1923/#1934 as a *direction*; that retirement stays as the safety floor until the evidence below is complete.

Refs #1888 (the defect), #1936 (zero-device SIGSEGV), #1925 (non-AMD Vulkan unvalidated), #1941 (kfd guard over-fires), #1922 (output-sanity gate — hard dependency for any re-enable), #1946/#1947 (CIRU distribution track).

## Phase A — build system located

| Piece | Resolved |
|---|---|
| Build recipe | CT105 `/root/dflash_build.sh` (+ `dflash_rebuild.sh`, `dflash/stage{,2}/Containerfile`) |
| **Kyuz0 upstream** | `kyuz0/amd-strix-halo-toolboxes` → fork `Hal0ai/amd-strix-halo-toolboxes`; runtime base `:rocm-7.2.4-rocmfp4-server` |
| **charlie upstream** | `charlie12345/ROCmFPX` (active, pushed 2026-08-17) → forks `Hal0ai/Hal0_ROCmFPX` and `ciru-ai/ROCmFPX`. The older `charlie12345/rocmfp4-llama` (→ `Hal0ai/rocmfp4-llama`) is stale since 2026-06-13 |
| Release pin | `src/hal0/config/schema.py:1013` — `DEFAULT_ROCMFPX_IMAGE = "ghcr.io/hal0ai/hal0-rocmfpx:ade07ba"` (not in `manifest.json`) |
| Build shape | Single HIP+Vulkan build: `-DGGML_HIP=ON -DGGML_HIP_FORCE_MMQ=ON -DGGML_VULKAN=ON`, `JOBS=28`; RADV added at runtime via `dnf install mesa-vulkan-drivers` |
| `ade07ba` provenance (image labels) | source `https://github.com/ciru-ai/ROCmFPX`, revision `ade07baee`, "llama.cpp b10297 + ROCmFPX + hal0 patches (preserved-token markers, vulkan FPX dmmv restore)" |

## Phase A/C evidence — the defect is in *this image*, not the hardware, driver, or upstream

[...]

Issue #1936, opened 2026-08-18
runner ade07ba: segfault (SIGSEGV) at model load when the container has zero GPU devices mapped — crashes before HTTP binds


thinmintdev:
Found during #1925 validation on ct151 (evidence: https://github.com/Hal0ai/hal0/issues/1925#issuecomment-5322245237).

Standalone llama-server from `ghcr.io/hal0ai/hal0-rocmfpx:ade07ba` with **no devices mapped** segfaults at model load (exit 139), reproduced 2/2, before the HTTP port binds. Model: hal0-brain-sft-q8-rocmfpx.

Distinct from #1888 (silent garbage with green health): this is crash-on-load. Population: any container/LXC deployment with neither /dev/kfd nor a render node visible — post-#1923 the preflight derives such boxes to CPU slots, so exposure is narrowed to configs that explicitly pin a GPU device on a device-less box, but the runner should fail with a diagnostic, not SIGSEGV.

Note for triage: the same session confirmed renderD128-without-kfd (AMD RADV lane) reproduces #1888's empty-content garbage exactly — no new evidence there, and #1923's preflight now blocks that config.

Refs #1925, #1888.

PR #1954, opened 2026-08-20
fix(gpu): resolve /dev/kfd access by runner identity and device gid

thinmintdev:
Fixes #1953.

## The bug

`kfd_present()` evaluated `/dev/kfd` accessibility as the **calling process**. That is the wrong identity for the question every caller is actually asking.

- `hal0-api` runs as `User=hal0`
- `hal0-slot@.service` carries **no** `User=` → podman runs rootful, containers open the device **as root**, with `--device=/dev/kfd` explicitly passed

A plain LXC `dev` passthrough leaves the two GPU nodes asymmetric:

| node | ownership |
|---|---|
| `/dev/dri/renderD128` | `root:render 0660` |
| `/dev/kfd` | `root:root 0660` |

So ROCm works perfectly in the container, while the API-side probe reports the compute node unusable. Since #1923's guard gates `gpu-rocm` **always** and `gpu-vulkan` on any AMD host, the practical effect is that **no AMD GPU llama.cpp slot can load** on a fully functional box.

Found on a production Strix Halo box immediately after deploying post-rc.6 code; `root:root` is the default for a plain LXC kfd passthrough, so this is not a one-box quirk.

The old diagnostic compounded it — it told the operator to forward a device that was **already forwarded**:

```
GpuPreflightError: ... needs the ROCm compute node /dev/kfd, which is not visible
here. ... Forward the device from the host (Proxmox LXC: add 'dev1: /dev/kfd' to
/etc/pve/lxc/<CTID>.conf, then pct stop/start)
```

That is a production reboot spent on a group-ownership mismatch.

## What changed

**Identity is now explicit.** `kfd_status()` splits `KFD_MISSING` from `KFD_NOT_OPENABLE` — the two shapes need opposite remedies and must not collapse into one bool. `kfd_present()` and `require_kfd_for_gpu_slot()` take the identity that actually opens the device, defaulting to `SLOT_RUNNER_UID` (root). `for_uid=None` still asks about the current process.

[...]

PR #1959, opened 2026-08-20
feat(runner): repin both AMD GPU lanes to hal0-combined:0822

thinmintdev:
Repins **both AMD GPU lanes** to `ghcr.io/hal0ai/hal0-combined:0822`. `rocmfpx` and `vulkanfpx` both resolve `DEFAULT_ROCMFPX_IMAGE`, so one constant moves them together.

**Pure repin.** The default brain does **not** change — `hal0-brain-sft` stays.

Refs #1948, #1888. Operator-merged (ledger row 10). No automerge.

## Why the image changes

`ade07ba`'s Vulkan backend emits invalid tokens for every model (#1888). Bisected on real hardware to **`ciru-ai/ROCmFPX`'s port of ROCmFPX onto newer upstream** — not hal0's build, the driver, gfx1151, or llama.cpp mainline. Four competing hypotheses were each killed by their own test; full evidence on #1948.

`:0822` is built from charlie's tree (whose Vulkan backend is demonstrably correct) plus two ports:

1. the missing `minicpm5` **pre-tokenizer mapping** — the enum and tokenizer behaviour already existed there, only the name→enum lookup was absent, so every minicpm5-pretokenised GGUF failed to load;
2. the MiniCPM5 **XML tool-call parser**, written in that tree's own idiom rather than transplanted.

That makes `:0822` a genuine **superset** of `ade07ba` — correct Vulkan *and* working native tool calls — rather than the tradeoff an earlier revision of this branch was built around.

| | `ade07ba` (current pin) | `:0822` |
|---|---|---|
| Vulkan output correct | ❌ | ✅ |
| MiniCPM5 native tool calls | ✅ | ✅ |

## Validation

Full §3-C matrix on #1948, including the rows that were previously missing. The decisive one is **ct151** — no `/dev/kfd`, so Vulkan is the *only* lane, and the exact configuration that produced #1888:

| Probe (ct151, kfd ABSENT) | Result |
|---|---|
| Shipped FPX brain model, Vulkan0 | ✅ `' Paris, a city that has been the center of French culture…'` |
| Native tool calls, Vulkan0 | ✅ `finish_reason: tool_calls`, `{"city":"Paris"}` |
| ≥256-token generation | ✅ terminates `stop_type=eos`, no `)`-wedge |

ct150 green across both lanes on non-FPX and FPX models, plus #1936's zero-device diagnostic.

## B1 — the retag reaches folded boxes

`ade07ba` is retired into `STALE_ROCMFPX_IMAGE_REFS`, **and** `retag_stale_slot_images` is extended to the typed `image_pin` field.

The bare `image` key is only half the population. `_resolve_image_ref` resolves `image_pin`, and `hal0 slot migrate-hw --apply` folds `image` into it — recording the then-current default as a *deliberate pin* (it was not in the stale set at the time) and dropping the older key. A folded box carries `image_pin = ade07ba`, honoured verbatim and invisible to a bare-key sweep. Without this it would keep serving the broken runner **forever**, behind the release that replaces it.

Matching against the stale set is what keeps it safe: a genuine operator pin is never an exact former default. Pinned by `test_a_deliberate_image_pin_is_never_touched`. The sweep binds `image_pin` before the bare `image` key — the same precedence `_resolve_image_ref` uses — so a slot carrying both shapes is judged on the key the runtime serves (review nit, commit `6c6a84d7`, both-keys tests in each direction).

## Sequencing — this PR alone does NOT deliver ct151

**Land order is #1973 → this PR, and it is mandatory** (proven by execution in the stack review): `run_post_activation_migrations` runs the vulkan relabel BEFORE the retag, so on a kfd-less box the relabel flips `gpu-vulkan` slots to `cpu` first and the retag then resolves the CPU toolbox — the pin never reaches them. With #1973's lane re-enable (and its shared look-through predicate) landed first, the lane survives the relabel and the retag delivers `:0822`. The ct151 validation rows above describe the *chain's* end state, not this diff's standalone effect.

## Scope note — the LFM2.5 brain swap was dropped

Earlier revisions of this branch also switched the default brain to LFM2.5. **That is gone.**

It was chosen by elimination under my own incorrect sizing of the parser port (I inferred difficulty from a ~1000-line `chat.cpp` length delta; charlie in fact already carried the whole PEG subsystem and a sibling XML parser). With the parser restored the swap is unforced, so it belongs to the operator as a separate product decision rather than riding in as a consequence of a pin.

Dropping it removes two review blockers at the root rather than mitigating them:

- **B3** — existing boxes silently losing tool calls on upgrade: impossible now, the new runner parses the old brain.
- **B4** — `<think>` leaking into `content` on every fresh install: impossible now, the brain profile is untouched.
- **B5** — the vacuous provenance tests were collateral from my retargeting. The reviewer's independent mutation check kills **4 tests, one per provenance predicate** (versus 1 on the swap branch) — the right shape; an earlier revision of this body claimed 13, which did not reproduce.

## Verification

```
tests/updater tests/install tests/config tests/registry tests/runners tests/providers tests/slots
3366 passed, 9 skipped
ruff check src/ tests/   clean
make check-sunset        192 <= baseline 192
```

Diff is three files: `schema.py` (pin + stale set), `updater.py` (B1), and the new `test_image_pin_retag.py`.

## Open, tracked elsewhere

- **Perf row — DONE** (#1948 comment 5365532926, clean window): `:0822` shows no ROCm regression (+5.2% pp / −1.1% tg vs ade07ba) and its Vulkan lane now beats ROCm (+14% pp / +20% tg) on the reference box.
- **#1970** — tracked as PR #1972 (build recipe; digest-pin + provenance-note revisions outstanding).

### DIFF
--- a/src/hal0/config/schema.py
+++ b/src/hal0/config/schema.py
@@ -1010,7 +1010,7 @@ MTP_FLAG_BUNDLE = build_mtp_flag_bundle("rocm")
 #: (debug builds, A/B tests, etc.).  See #__hal0_image_control__ for the
 #: phasing: 0.9.5 wires slot.image + DEFAULT_ROCMFPX_IMAGE; 0.9.6 will
 #: drop ``image`` from SEED_PROFILES entirely.
-DEFAULT_ROCMFPX_IMAGE = "ghcr.io/hal0ai/hal0-rocmfpx:ade07ba"
+DEFAULT_ROCMFPX_IMAGE = "ghcr.io/hal0ai/hal0-combined:0822"
 
 #: Historical DEFAULT_ROCMFPX_IMAGE values (and their pre-consolidation
 #: equivalents). A slot-level ``image`` pin equal to one of these is a STALE
@@ -1040,6 +1040,10 @@ DEFAULT_ROCMFPX_IMAGE = "ghcr.io/hal0ai/hal0-rocmfpx:ade07ba"
 #: control-token pieces are stripped before the parser sees them).
 STALE_ROCMFPX_IMAGE_REFS = frozenset(
     {
+        # Former default (rc.6). Its Vulkan backend emits invalid tokens for
+        # every model (#1888) — a slot still pinned here is debris from an
+        # older release, not an opt-out, and must be retagged.
+        "ghcr.io/hal0ai/hal0-rocmfpx:ade07ba",
         "ghcr.io/hal0ai/hal0-rocmfpx:c077206",
         "ghcr.io/hal0ai/hal0-rocmfpx:vulkan-minicpm5",
         "localhost/hal0-rocmfpx:vulkan-minicpm5",
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -2873,16 +2873,21 @@ def retag_stale_slot_images(*, job_id: str | None = None) -> int:
         # retired runner forever — behind the release that replaces it.
         # Matching against the stale set is what keeps this safe: a genuine
         # operator pin is never an exact former default.
+        # PRECEDENCE MATCHES THE RESOLVER (#1959 review nit): ``image_pin``
+        # binds first, because ``_resolve_image_ref`` reads it first and no
+        # longer reads the bare key at all. Binding ``image`` first here made
+        # the sweep retag (or skip on) a value the runtime never serves while
+        # the pin that IS served stayed stale.
         holder: dict | None = None
-        key = "image"
-        if isinstance(raw.get("image"), str):
+        key = "image_pin"
+        if isinstance(raw.get("image_pin"), str):
             holder = raw
-        elif isinstance(raw.get("slot"), dict) and isinstance(raw["slot"].get("image"), str):
-            holder = raw["slot"]
-        elif isinstance(raw.get("image_pin"), str):
-            holder, key = raw, "image_pin"
         elif isinstance(raw.get("slot"), dict) and isinstance(raw["slot"].get("image_pin"), str):
-            holder, key = raw["slot"], "image_pin"
+            holder = raw["slot"]
+        elif isinstance(raw.get("image"), str):
+            holder, key = raw, "image"
+        elif isinstance(raw.get("slot"), dict) and isinstance(raw["slot"].get("image"), str):
+            holder, key = raw["slot"], "image"
         if holder is None or holder[key] not in STALE_RUNNER_IMAGE_REFS:
             continue
         old_ref = holder[key]
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -2861,21 +2861,38 @@ def retag_stale_slot_images(*, job_id: str | None = None) -> int:
         # An image REF is a top-level or [slot]-nested STRING. The ``[image]``
         # TOML table (image-gen settings, #599) shares the key and must be
         # ignored — same trap as providers.container._resolve_image_ref.
+        #
+        # BOTH key shapes, deliberately (#1948 B1). The bare ``image`` string is
+        # only half the population: ``_resolve_image_ref`` now resolves the
+        # TYPED ``image_pin`` field, and ``hal0 slot migrate-hw --apply`` folds
+        # ``image`` into it — recording the then-current default as a
+        # "deliberate pin" (it was not in STALE_RUNNER_IMAGE_REFS at the time)
+        # and dropping the older key. A folded box therefore carries
+        # ``image_pin = <old default>``, honoured verbatim and INVISIBLE to a
+        # sweep that reads only the bare key, so it would keep serving a
+        # retired runner forever — behind the release that replaces it.
+        # Matching against the stale set is what keeps this safe: a genuine
+        # operator pin is never an exact former default.
         holder: dict | None = None
+        key = "image"
         if isinstance(raw.get("image"), str):
             holder = raw
         elif isinstance(raw.get("slot"), dict) and isinstance(raw["slot"].get("image"), str):
             holder = raw["slot"]
-        if holder is None or holder["image"] not in STALE_RUNNER_IMAGE_REFS:
+        elif isinstance(raw.get("image_pin"), str):
+            holder, key = raw, "image_pin"
+        elif isinstance(raw.get("slot"), dict) and isinstance(raw["slot"].get("image_pin"), str):
+            holder, key = raw["slot"], "image_pin"
+        if holder is None or holder[key] not in STALE_RUNNER_IMAGE_REFS:
             continue
-        old_ref = holder["image"]
+        old_ref = holder[key]
         # HW-gated target: rocmfpx on a Strix GPU lane, the lean toolbox
         # elsewhere. When the host/lane default already equals the pin (e.g. a
         # non-Strix box on the vulkan toolbox), it's a no-op — leave it be.
         new_ref = resolve_default_image(_backend_of(raw), _device_class_of(raw))
         if new_ref == old_ref:
             continue
-        holder["image"] = new_ref
+        holder[key] = new_ref
         try:
             write_toml_atomic(toml_path, raw)
         except Exception as exc:
@@ -2886,6 +2903,7 @@ def retag_stale_slot_images(*, job_id: str | None = None) -> int:
             "updater.slot_image_retagged",
             job_id=job_id,
             slot=slot_name,
+            field=key,
             old=old_ref,
             new=new_ref,
             note=(
--- a/tests/updater/test_image_pin_retag.py
+++ b/tests/updater/test_image_pin_retag.py
@@ -1,0 +1,71 @@
+"""#1948 B1 — a folded box pins the runner in `image_pin`, not `image`.
+
+`hal0 slot migrate-hw --apply` folds a bare `image` string into the typed
+`image_pin` field, recording the then-current default as a "deliberate pin"
+because it was not yet in STALE_RUNNER_IMAGE_REFS. A sweep that reads only the
+bare key is blind to that, so the slot would keep serving a retired runner
+forever — behind the release that replaces it.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from hal0.config.schema import DEFAULT_ROCMFPX_IMAGE
+from hal0.updater.updater import retag_stale_slot_images
+
+RETIRED = "ghcr.io/hal0ai/hal0-rocmfpx:ade07ba"
+
+
+@pytest.fixture
+def slots_dir(tmp_path, monkeypatch):
+    d = tmp_path / "slots"
+    d.mkdir()
+    monkeypatch.setattr("hal0.config.paths.slots_config_dir", lambda: d)
+    return d
+
+
+def test_a_stale_image_pin_is_retagged(slots_dir) -> None:
+    (slots_dir / "brain.toml").write_text(
+        f'device = "gpu-rocm"\nimage_pin = "{RETIRED}"\n', encoding="utf-8"
+    )
+    assert retag_stale_slot_images() == 1
+    body = (slots_dir / "brain.toml").read_text(encoding="utf-8")
+    assert RETIRED not in body
+    assert DEFAULT_ROCMFPX_IMAGE in body
+
+
+def test_a_nested_stale_image_pin_is_retagged(slots_dir) -> None:
+    (slots_dir / "agent.toml").write_text(
+        f'[slot]\ndevice = "gpu-rocm"\nimage_pin = "{RETIRED}"\n', encoding="utf-8"
+    )
+    assert retag_stale_slot_images() == 1
+    assert RETIRED not in (slots_dir / "agent.toml").read_text(encoding="utf-8")
+
+
+def test_the_bare_image_key_still_works(slots_dir) -> None:
+    """The original shape must not regress while adding the typed one."""
+    (slots_dir / "code.toml").write_text(
+        f'device = "gpu-rocm"\nimage = "{RETIRED}"\n', encoding="utf-8"
+    )
+    assert retag_stale_slot_images() == 1
+    assert RETIRED not in (slots_dir / "code.toml").read_text(encoding="utf-8")
+
+
+def test_a_deliberate_image_pin_is_never_touched(slots_dir) -> None:
+    """The stale-set match is the whole safety property: an operator's own pin
+    is never an exact former default, so a debug build must survive."""
+    (slots_dir / "brain.toml").write_text(
+        'device = "gpu-rocm"\nimage_pin = "ghcr.io/example/my-debug-build:v9"\n',
+        encoding="utf-8",
+    )
+    assert retag_stale_slot_images() == 0
+    assert "my-debug-build:v9" in (slots_dir / "brain.toml").read_text(encoding="utf-8")
+
+
+def test_it_is_idempotent(slots_dir) -> None:
+    (slots_dir / "brain.toml").write_text(
+        f'device = "gpu-rocm"\nimage_pin = "{RETIRED}"\n', encoding="utf-8"
+    )
+    assert retag_stale_slot_images() == 1
+    assert retag_stale_slot_images() == 0
--- a/tests/updater/test_image_pin_retag.py
+++ b/tests/updater/test_image_pin_retag.py
@@ -69,3 +69,34 @@ def test_it_is_idempotent(slots_dir) -> None:
     )
     assert retag_stale_slot_images() == 1
     assert retag_stale_slot_images() == 0
+
+
+def test_both_keys_present_the_pin_wins_like_the_resolver(slots_dir) -> None:
+    """With BOTH key shapes on one slot, the sweep must judge the key the
+    runtime serves. ``_resolve_image_ref`` reads ``image_pin`` first and no
+    longer reads the bare key at all — so a stale pin behind a current-looking
+    bare ``image`` was invisible to a bare-key-first sweep, and the slot kept
+    serving the retired runner behind the release replacing it (#1959 review
+    nit; the inverse ordering also logged retags of a value nothing reads)."""
+    (slots_dir / "agent.toml").write_text(
+        f'device = "gpu-rocm"\nimage = "{DEFAULT_ROCMFPX_IMAGE}"\nimage_pin = "{RETIRED}"\n',
+        encoding="utf-8",
+    )
+    assert retag_stale_slot_images() == 1
+    after = (slots_dir / "agent.toml").read_text(encoding="utf-8")
+    assert RETIRED not in after
+    assert f'image_pin = "{DEFAULT_ROCMFPX_IMAGE}"' in after
+
+
+def test_both_keys_with_a_deliberate_pin_is_untouched(slots_dir) -> None:
+    """The mirror case: a deliberate (non-stale) pin with stale bare-key
+    debris behind it. The runtime serves the pin, which is fine — retagging
+    the inert bare key would log a retag that changes nothing served."""
+    (slots_dir / "code.toml").write_text(
+        f'device = "gpu-rocm"\nimage = "{RETIRED}"\n'
+        'image_pin = "ghcr.io/example/my-debug-build:v9"\n',
+        encoding="utf-8",
+    )
+    before = (slots_dir / "code.toml").read_text(encoding="utf-8")
+    assert retag_stale_slot_images() == 0
+    assert (slots_dir / "code.toml").read_text(encoding="utf-8") == before


thinmintdev:
**GA-session coordination** — this PR is the decision ledger's rows 9/10 made concrete, which is good news, and it changes two things that are explicitly operator-gated: the default image pin for both AMD GPU lanes, and the default brain model (the tool-call parser tradeoff the track surfaced earlier, resolved here as LFM2.5 — a product default change, not just a build detail). Per the standing plan: **the operator merges this one**; agent sessions review it and get it green, nothing more.

Two asks before it's decision-ready:

1. **Post the full §3-C matrix to #1948** for `hal0-combined:0821`. The PR body covers the Paris probe, the tool-call A/B, and an e2e reply through hal0's API — but the matrix also requires: both boxes (ct150 with kfd AND kfd-less ct151), an FPX gguf AND a plain non-FPX gguf, the ≥256-token long-generation probe (`finish_reason` set, no `)`-repetition wedge), the zero-device readable-diagnostic case (#1936), and perf numbers vs the historical envelope (Vulkan decode ~-10% vs ROCm was the known ade07ba-era gap; >20% regressions need explanation). Whatever ran, post it; whatever didn't, run it — the fold-in call reads that table, not this PR body.
2. **Sequencing:** #1952 (kfd guard scoping) lands first through the fix-wave train, and #1954 has a changes-requested review on it (uid-0 short-circuit). If this PR's branch doesn't depend on either, say so explicitly; if it does, rebase order is #1952 → #1954 → this.

Also noting for the record: the e2e verification ran against CT105's brain slot — that's the production box, which the plan holds read-only. It worked out, but candidate-image serving tests belong on ct150/ct151.

An independent opus review of this diff is being spawned now (same convention as #1954's). Timing-wise this arrives BEFORE the rc.7 cut, so if the full matrix lands clean and the reviews come back green, ledger row 9 (fold into rc.7 vs 1.0.x) becomes a live operator decision earlier than planned — with the rc-validate Vulkan lane briefs as the consequence if fold-in is chosen.

thinmintdev:
## Independent adversarial review — verdict: **changes-requested**

Second pair of eyes; I did not write this code. **This PR is operator-merged** — it changes the
default runner image for both AMD GPU lanes and the default brain model, both operator-gated
decisions. This is a comment only, deliberately not an `--approve` and not a merge.

The direction is right and the bisect on #1948 is genuinely good work. The gap is that the
**upgrade story does not hold on either path**, and one of the two mechanisms the PR body relies
on cannot reach the slots it names.

---

### What I verified clean first

* **No `CHANGELOG.md` hunks.** ✅
* **No lane re-enable smuggled in.** `brain.toml` keeps `device = "gpu-rocm"`, the #1923 preflight
  guard is untouched, and `vulkanfpx` already resolved to `DEFAULT_ROCMFPX_IMAGE` before this diff.
  The Phase-D re-enable still correctly waits on #1922. ✅
* **`context_length=131072` is right.** `LiquidAI/LFM2.5-2.6B/config.json` →
  `"max_position_embeddings": 131072`. The GGUF repo is public + ungated and carries
  `LFM2.5-2.6B-Q8_0.gguf` (2,874,779,648 B). The hermes-floor test resolves the seed ceiling
  65536 ≥ 64000. ✅
* **The image is real.** `ghcr.io/hal0ai/hal0-combined:0821` → anonymous pull HTTP 200,
  `sha256:48fbe1c19085a13a718eaf2f1f7bffc79d248e4cbdfd9e8523b6305b2659eb19`. ✅
* **Rubric-3 flattening is safe.** `brain_model_for_hardware` has exactly two callers
  (`provision_brain_model`, `main`), both in this module. Nothing else consumed the Q8-vs-F16 split.
  The retargeted matrix tests DO pin the flattening — reverting the branch fails
  `test_cpu_only_box_selects_portable_f16` and `test_gpu_present_but_neither_*`. ✅
* **Targeted suite green** on `3c9c53d`:
  `tests/config/test_schema_seeds_c5.py tests/install tests/registry tests/runners tests/updater
  tests/brain tests/providers/test_image_resolution.py tests/bench` → **1603 passed**.
  (CI `python (3.12)` was still pending when I looked.)

---

## Blocking

### B1 — an `image_pin = ade07ba` slot can never be retagged, so #1888 stays live on exactly the boxes that followed the convergence nag
*Not flagged by Codex. This is the one with real teeth.*

`_resolve_image_ref` (`src/hal0/providers/container.py:283-327`) now resolves **only**
`slot_cfg["image_pin"]`; the docstring is explicit that the old `slot.image` / `[slot].image`
string tiers are **DELETED** and "folded into `image_pin` by the migration lane". A non-empty
`image_pin` is *"honored verbatim, never re-resolved"*.

`retag_stale_slot_images` (`src/hal0/updater/updater.py:2718+`) reads and writes **only** the legacy
`image` key (top-level or `[slot]`-nested). It never looks at `image_pin`.

Now trace a box that ran `hal0 slot migrate-hw --apply` while on rc.6 — which the convergence
report (`detect_pending_ownership_migrations`, `updater.py:2469-2520`) actively nags operators to
do. In `hw_slot_ownership.plan_unfold`:

```python
if own_image is not None and own_image not in stale_refs:
    set_image_pin = own_image        # ← "deliberate pin"
```

At that moment `ade07ba` was the *current* default, so it was **not** in `stale_refs` → it was
recorded as a deliberate operator pin, and `drop_slot_image` deleted the legacy key.

Result on upgrade: `image_pin = ghcr.io/hal0ai/hal0-rocmfpx:ade07ba`, honoured verbatim, invisible
to the retag this PR extends. **That slot keeps serving the known-broken-Vulkan runner forever**,
behind a release whose headline is that it fixes exactly that.

Conversely, on a box that has *not* run the fold, the `image = ade07ba` key is already **inert** —
it is no longer a resolution tier — so that slot picks up `:0821` from the post-swap unit re-render
regardless. Which means the STALE-list addition as written is cosmetic TOML hygiene for the
population it *can* see, and blind to the population that actually breaks.

**Fix:** extend `retag_stale_slot_images` to cover the typed `image_pin` field (or add a one-shot
`ade07ba` image_pin migration). At minimum, correct the PR body — "`ade07ba` added to
`STALE_ROCMFPX_IMAGE_REFS` so the updater retags slots still pinned to the old default" is not what
happens on a folded box.

### B2 — the retag does not fire on `hal0 update` at all (#1960)

`run_post_activation_migrations` runs **pre-swap**, in the outgoing interpreter (`updater.py:3796`
via `asyncio.to_thread`, before `seam.activate` at `:3849`). So the `ade07ba` entry this release
adds to `STALE_ROCMFPX_IMAGE_REFS` **never executes on the rc.6 → rc.7 upgrade that ships it** — the
only population it was written for. It works on the `install.sh` re-run path (new code, but no
legacy state) and nowhere else.

This is a cross-PR dependency, not this PR's bug — but the migration story in the PR body is inert
until #1960 lands, and that should be stated. Given B1, fixing #1960 alone is not sufficient.

### B3 — upgraded boxes keep `hal0-brain-sft` on a runner that cannot parse it, with no migration and no runtime recovery
*Agrees with Codex's P1 on `schema.py:1013`; mechanism corrected below.*

This is the rubric-2 answer, traced concretely:

* `hal0 update run` **never** calls `provision_brain_model` — zero brain references anywhere in
  `updater.py`. `/etc/hal0/slots/brain.toml` keeps `[model].default = hal0-brain-sft-q8-rocmfpx`.
* `_rerender_units_after_swap` **does** run post-swap in the just-repipped venv, so an unpinned
  brain slot *does* get `:0821`.
* Net: **new runner + old brain model** — the PR body's own A/B failure row,
  `tool_calls: NONE   content: ' name="get_weather">…'`.

And it cannot self-heal. `_image_native_tools` (`brain/chat.py:1427`) assumes any image not in
`NATIVE_TOOL_INCOMPATIBLE_IMAGE_REFS` is capable, and `:0821` is not listed (correctly — it parses
OpenAI tools fine *for other models*). `_LEARNED_TOOL_INCAPABLE_IMAGES` only learns from
`_TOOL_FORMAT_FAILURE_MARKERS`, i.e. a template-parser **500**. This failure is an HTTP **200** with
XML in `content`, so **the learning layer never fires**. The only residual is
`_tool_intent_artefact` → reroute to `[brain_chat] tool_model` = `hal0/agent`, which ships
**model-less by default** (#1369). On an upgraded box with no agent model bound, the steward loses
its tool surface entirely and answers with the `tools_unavailable` notice.

Two further points worth naming in the PR:

1. **The two upgrade paths diverge.** `install.sh` re-run *does* re-provision (picks LFM2.5, pulls
   ~2.7 GB, stamps `[model].default`); `hal0 update` does not. That is precisely the drift class
   #1475 / `run_post_activation_migrations` was created to eliminate.
2. **The deny list is structurally the wrong shape for this failure.** The incompatibility is an
   *(image, model)* pair — `hal0-brain-sft` × `:0821` — and
   `NATIVE_TOOL_INCOMPATIBLE_IMAGE_REFS` is image-keyed only, so it cannot express it. Either a
   brain-model migration or a model-aware gate is needed; neither exists today.

### B4 — the new default steward ships with visibly broken output
*Agrees with Codex's P1 on `curated.py:404`. Confirmed at source.*

`[profile.brain]` (`src/hal0/config/data/seed_profiles.toml:131`) still launches
`--reasoning off --reasoning-format none`, while this PR's own curated `notes` say the profile
"should carry `--reasoning-format auto` so it lands in `reasoning_content` rather than leaking into
content". The "Honest gaps" section defers this to a separate profile PR.

Shipping a **default** whose raw `<think>` block leaks into `content` on the dashboard's primary
chat surface is a product regression on every fresh install, not a follow-up. The `nano`-sharing
argument is a reason to *split* the profile, not a reason to ship the leak. This should land here or
in a stacked PR merged together with this one.

### B5 — four provenance tests became vacuous (mutation-verified)
*Agrees with Codex's P2. I confirmed it empirically rather than by reading.*

I mutated `already_pulled` to drop every provenance check
(`return dest if dest.exists() else None` — no sidecar, no `curated_id` match, no size check) and
ran `tests/install/test_brain_model.py` on both trees:

| tree | result against the provenance-blind mutant |
|---|---|
| `main` | **4 failed**, 25 passed |
| this PR | **1 failed**, 29 passed |

Surviving the mutant on this branch: `test_a_file_with_no_sidecar_is_not_trusted`,
`test_a_truncated_file_re_pulls`, `test_a_sidecar_from_a_different_model_re_pulls` (plus
`test_provisioning_still_pulls_when_the_prior_file_is_untrustworthy`, whose name no longer describes
what it does). Cause is uniform: `_place(store, BRAIN_MODEL_PORTABLE)` followed by
`assert already_pulled(BRAIN_MODEL_DEFAULT) is None`. The store is id-keyed, so the default's path is
simply absent and the assertion holds no matter what `already_pulled` does. Place and query the
**same** id in each.

---

## Nits

* **N1** — `main()`'s rationale string (`brain_model.py:299-303`) still branches on
  `rocmfpx_capable(hw)`. A CPU-only install now prints
  `brain model: lfm2-5-2-6b-q8-0 (no ROCm/Vulkan device — using the portable F16 build (stock llama.cpp))`.
  Wrong on both branches. (Codex P2 — confirmed.)
* **N2** — the module docstring (`brain_model.py:17-22`) still lists "**Hardware-aware quant
  choice**" as one of four deliberate properties, and explains the custom-tensor-type rationale that
  no longer applies to the default.
* **N3** — shipped, operator-facing doc drift not updated:
  * `installer/etc-hal0/slots/brain.toml:21-27` (the hardware-dependent pick) and `:50-59`
    ("as of runner ade07ba … the brain models are NATIVE tool-callers");
  * `installer/install.sh:2009-2012` — same claim, in the install transcript's own comment block;
  * `[profile.brain]`'s `intent = "Brain steward (1B MiniCPM5 + persona + tool-call routing) · 64K ctx"`
    is surfaced in the dashboard and now names the wrong model, and its sampler rationale
    ("MiniCPM5-1B's own published no-think recommendation") is now applied to a different model
    family with no stated basis.
* **N4** — `BRAIN_HF_REPO` (`brain_model.py:57`) is now referenced nowhere in `src/` and no longer
  imported by any test. Either wire it back (assert the sft family still resolves there) or drop it.
* **N5** — both repo assertions were weakened to `assert curated.hf_repo != "Hal0ai/hal0-brain-sft"`,
  which passes for any typo'd repo string. `assert curated.hf_repo in {BRAIN_HF_REPO,
  BRAIN_DEFAULT_HF_REPO}` keeps the stated intent without re-introducing one-repo-for-all.
* **N6** — `BRAIN_DEFAULT_HF_REPO` is declared and then the curated row hardcodes the same literal:
  two sources of truth on day one.
* **N7** — `size_gb=2.7` vs the actual 2,874,779,648 B (2.87 GB / 2.68 GiB); fine if `size_gb` is
  GiB, worth a consistency pass. `vram_gb_min=4.0` against the seed's 65536 ctx is advisory-only on
  this path, so no action needed — just noting it was checked.
* **N8** — nothing in the test suite pins the image tag literal, so a silent revert of
  `DEFAULT_ROCMFPX_IMAGE` would not be caught. One assertion would close that.

---

## Image provenance (rubric 5) — a release-cut decision, not a merge blocker

* `rocmfpx` and `vulkanfpx` both carry **`manifest_key=None`** (`src/hal0/runners/__init__.py:120-141`),
  so `resolve_runner_image`'s digest-pin tier **never applies to the default runner**. The shipped
  pin is a **mutable tag**.
* `manifest.json:toolbox_images` has no `combined` / `rocmfpx` entry, so
  `scripts/release-check.sh`'s "every toolbox image must have a non-null `digest`" gate does **not**
  cover the one image every AMD GPU slot runs.
* This is pre-existing (`ade07ba` was a tag too), but `:0821` is strictly worse than the prior
  convention: `ade07ba` / `c077206` are git short-shas that identify a source tree; `0821` is a
  **date**. The repo's tag list is already `["0820", "0a59add", "0821"]` — two consecutive date tags
  plus a sha-shaped one — so the churn is real, and re-pushing `:0821` would silently re-point every
  box with no version change.
* **Recommendation for the rc.7 cut:** add a `combined` entry to `manifest.json:toolbox_images`
  pinned to `sha256:48fbe1c19085a13a718eaf2f1f7bffc79d248e4cbdfd9e8523b6305b2659eb19`, give the
  runner rows a `manifest_key` so the digest tier engages, and prefer the sha-shaped tag over the
  date. Also worth confirming `hal0 doctor toolbox-pull`'s anonymous-pull contract covers
  `hal0ai/hal0-combined` (I verified the pull works anonymously; I did not verify doctor knows about
  the repo).

---

## Codex triage

Four bot comments, all verified against source before being repeated here:

| Codex finding | Verdict |
|---|---|
| P1 brain profile `--reasoning-format` | **Confirmed** → B4 |
| P1 upgrade brain-model migration | **Confirmed**, mechanism corrected → B3 (the retag is not what delivers `:0821` to the slot; the post-swap unit re-render is, and B1/B2 mean the retag is weaker than Codex assumed) |
| P2 `already_pulled` test id mismatch | **Confirmed** by mutation testing → B5 |
| P2 installer rationale string | **Confirmed** → N1 |

None dismissed. Codex did not find B1 (`image_pin` blindness) or B2 (#1960 ordering).

---

### Summary

The runner fix and the brain switch are both the right call, and the evidence behind them is solid.
What is missing is the migration layer for boxes that already exist: **B1** leaves a real population
on the broken image permanently, **B2** makes the stated retag inert on the upgrade path that ships
it, and **B3** leaves upgraded stewards without tools with no self-healing path. **B4** is a
user-visible defect in the new default. **B5** is a straightforward test fix.

Recommend addressing B1/B3/B4/B5 in this PR (or a stack merged with it), noting B2's dependency on
#1960 explicitly, and taking the digest pin at the release cut. Then it is a clean operator decision.

thinmintdev:
## Codex review addressed (80f2a849)

All four findings judged valid and fixed. **No automerge** — operator-merge or land-train only.

| # | Finding | Fix |
|---|---|---|
| P1 | Existing brain slots silently lose tool calls on upgrade | `migrate_brain_model_off_minicpm5()`, run immediately **after** the image retag so image and model move together. Narrow by construction: only known MiniCPM5-family ids, only to the curated default |
| P1 | Brain profile still launched `--reasoning off --reasoning-format none` | Now `--reasoning-format auto`. That combination did not suppress anything on a reasoning model — LFM2.5 emits the block regardless, and with no format selected the raw `<think>` landed in `content` |
| P2 | Metadata tests placed PORTABLE, queried DEFAULT | Place and query the same model. The store is keyed by id, so `already_pulled()` was returning `None` because the path was simply **absent** — the missing/truncated/wrong-sidecar logic was never exercised at all |
| P2 | Install transcript rationale still branched on hardware | Rewritten to describe the actual pick. It contradicted itself on both paths — a CPU box was told it was getting "the portable F16 build", a GPU box that the model needed ROCmFPX tensor types |

The **P1 migration** finding is the one that mattered most: I shipped a runner repin that would have left existing stewards loading fine, answering fine, and quietly unable to call any platform tool — with no format error for the learned-incompatible retry to catch. That is exactly the silent-green failure class #1888 was about, and I reintroduced it one layer up.

The **vacuous-test** finding is a defect I introduced while retargeting assertions: I updated the query side and not the placement side, so four regression tests would have passed no matter what the sidecar logic did.

## Verification

```
tests/install tests/config tests/registry tests/runners tests/updater  1795 passed, 9 skipped
tests/updater/test_brain_model_migration.py                            9 passed (new)
ruff check src/                                                        clean
```

thinmintdev:
## B1 fixed + rebased (74499281)

### B1 — the finding I had not addressed, and it was the one that mattered

Confirmed at source: `retag_stale_slot_images` had **zero** `image_pin` references. The review was exactly right — the population the retag was written for is the population it could not see.

`_resolve_image_ref` resolves the typed `image_pin` field, and `hal0 slot migrate-hw --apply` folds a bare `image` string into it, recording the then-current default as a *deliberate pin* (it was not in `STALE_RUNNER_IMAGE_REFS` at the time) and dropping the older key. A folded box carries `image_pin = ade07ba`, honoured verbatim, invisible to a sweep that reads only the bare key — so it would keep serving the known-broken-Vulkan runner **forever**, behind the release that replaces it.

The retag now reads both key shapes, top-level and `[slot]`-nested. The stale-set match is what keeps it safe: a genuine operator pin is never an exact former default, pinned by `test_a_deliberate_image_pin_is_never_touched`. **Confirmed red against the bare-key-only sweep, green with the fix.**

**Correcting the PR body:** "ade07ba added to STALE_ROCMFPX_IMAGE_REFS so the updater retags slots still pinned to the old default" was not true for folded boxes. It is now.

### Also in this push

- Rebased onto main (#1952, #1961 landed).
- `sunset` went red on my own comments: the scar ratchet counts `\blegacy\b`, and four explanatory comments about pre-existing state pushed 192 → 196. Reworded rather than raising the baseline — they describe existing state, they are not new shims.

### Note on the pin

The pin is now `ghcr.io/hal0ai/hal0-combined:0822`, which **restores the MiniCPM5 tool-call parser** — full §3-C matrix on #1948, including the previously-missing ct151 (kfd-less) rows. That makes it a genuine superset of `ade07ba` rather than a tradeoff, which means **B3 and B4 are no longer load-bearing**: the new runner parses the old brain, so nothing silently loses tool calls and no `<think>` leak ships even if the brain default stayed put.

The LFM2.5 swap is therefore now a *product choice*, not a forced consequence of my earlier (incorrect) sizing of the parser port. Both halves are still in this PR; say the word if you'd rather it shrink to a pure repin.

```
tests/updater tests/install tests/config tests/registry tests/runners   1822 passed, 9 skipped
make check-sunset                                                       clean
```

Still operator-merge. Automerge off.

thinmintdev:
## Independent adversarial review (reshaped branch, `d8008586`) — verdict: **comment, one blocking-for-the-decision finding**

Second pair of eyes; I did not write this code. **This PR is operator-merged** (default runner pin for both AMD GPU lanes, GA ledger row 10) — this is deliberately a comment, not an `--approve` and not a merge. Everything below was run, not read.

Re-reviewing against the *original* review's five blockers: **B3 and B4 die at the root** — the LFM2.5 brain swap is gone from the diff, so "existing boxes silently lose tool calls" and "`<think>` leaks into `content`" are no longer reachable. Good call dropping it; a product default belongs to the operator, not to a pin.

---

### 1. Diff is genuinely repin-only ✅

Three files, 97+/4-. `schema.py` (the pin + `ade07ba` into the stale set), `updater.py` (B1), `tests/updater/test_image_pin_retag.py`. **Zero surviving `curated.py` / `brain_model.py` / `seed_profiles.toml` churn** — I diffed `github/main...d8008586` in full. No `CHANGELOG.md` hunk. ✅

Also verified `STALE_ROCMFPX_IMAGE_REFS` is the right set to add to: `runners/__init__.py:225` aliases `STALE_RUNNER_IMAGE_REFS = STALE_ROCMFPX_IMAGE_REFS`, which is what `retag_stale_slot_images` actually reads. ✅

### 2. B1 — verified by execution ✅

Not taking the test file's word for it; I wrote my own probes against `d8008586`:

| Probe | Result |
|---|---|
| Folded box, `image_pin = ...rocmfpx:ade07ba` | retagged → `hal0-combined:0822` ✅ |
| Deliberate pin, `image_pin = ghcr.io/acme/experimental:deadbeef` | `retagged == 0`, file byte-identical ✅ |

B1 is genuinely fixed for the population it names. `tests/updater/test_image_pin_retag.py` — 5 passed.

### 3. B5 — resolved on substance, but the *number* in the PR body does not reproduce (nit)

I re-ran the mutation myself: gutted the provenance block in `already_pulled` (`brain_model.py:157-165` — sidecar presence, `curated_id`/`hf_file` match, `size_bytes` match) down to presence-only, then ran **exactly the scope this PR body quotes**:

```
tests/updater tests/install tests/config tests/registry tests/runners tests/providers tests/slots
→ 4 failed, 3362 passed, 9 skipped
```

The four are `test_a_file_with_no_sidecar_is_not_trusted`, `test_a_truncated_file_re_pulls`, `test_a_sidecar_from_a_different_model_re_pulls`, `test_a_malformed_sidecar_re_pulls` — **one per provenance predicate.** That is the right shape and B5 is resolved: the tests are not vacuous, they place and query the same model and each kills a distinct predicate.

But the body claims **13 failures**, and 13 does not reproduce at any scope I tried (4 at `tests/install tests/registry` alone, 4 at full quoted scope). Please either correct the figure or say what was mutated to get 13 — an overstated mutation score is the same class of problem as the vacuous test it was measuring.

---

## 🔴 Blocking for the merge decision — the repin does not reach the population this PR was validated on

Not a code defect; a **claim/sequencing** defect, and it lands squarely on the evidence the ledger row reads.

The PR body's decisive validation is **ct151, `/dev/kfd` ABSENT, Vulkan the only lane**. On that exact box, `run_post_activation_migrations` runs `relabel_stale_vulkan_slots()` *before* `retag_stale_slot_images()` (`updater.py:1701-1711`). The relabel flips a `gpu-vulkan` slot to `device = "cpu"` when kfd is missing — so by the time the retag runs, `_backend_of`/`_device_class_of` read `cpu` and `resolve_default_image` resolves the **CPU-lane toolbox, not `:0822`**.

Executed on this branch, both passes in the real order:

```
kfd PRESENT → device: gpu-rocm | image_pin: ghcr.io/hal0ai/hal0-combined:0822        ✅
kfd ABSENT  → device: cpu      | image_pin: ghcr.io/hal0ai/amd-strix-halo-toolboxes:vulkan-radv-server   ❌
```

So: on kfd-present boxes the repin works exactly as advertised. On the kfd-less population — the one the §3-C matrix's decisive row covers, the one `:0822` was built to rescue — **this PR alone delivers the CPU fallback.** It is not a regression (that's pre-existing #1923/#1888 behaviour), but the PR reads as though merging it fixes ct151, and it does not. Codex's P1 on `schema.py:1013` ("Re-enable the Vulkan lane for the fixed runner") is the same finding from the guard side and is **valid**.

**Ask:** either (a) state explicitly in the body that this pin is inert on kfd-less boxes until #1973 lands and that the two must ship in the same release, or (b) sequence #1973 first. The operator is merging this on the strength of the ct151 evidence; the body should not let that read as "merged ⇒ ct151 fixed".

## 🔴 Composition with Phase D (#1973) — 4 tests go red in either land order

`feat/vulkan-lane-reenable` is based on `main` and merges textually clean with this branch (`git merge-tree --write-tree` → clean tree, no conflict). But it does **not** compose semantically. #1973 introduces:

```python
VULKAN_FIXED_IMAGE = "ghcr.io/hal0ai/hal0-combined:0822"
VULKAN_CAPABLE_IMAGE_REFS = frozenset({VULKAN_FIXED_IMAGE})
```

and then uses `DEFAULT_ROCMFPX_IMAGE` throughout its tests as a *stand-in for the broken `ade07ba` lineage*:

```python
def test_the_ade07ba_lineage_pin_is_not_capable(self) -> None:
    """#1888's carrier. It is the CURRENT default on main, ..."""
    assert DEFAULT_ROCMFPX_IMAGE not in VULKAN_CAPABLE_IMAGE_REFS
    assert image_serves_vulkan_lane(DEFAULT_ROCMFPX_IMAGE) is False
```

That docstring premise — "it is the CURRENT default on main" — is exactly what this PR falsifies. After the repin, `DEFAULT_ROCMFPX_IMAGE == VULKAN_FIXED_IMAGE`. I merged this branch into #1973 and ran its gate suite:

```
tests/providers/test_gpu_vulkan_lane_gate.py → 4 failed, 39 passed
  TestVulkanCapableImageSet::test_the_ade07ba_lineage_pin_is_not_capable
  TestVulkanLaneGuard::test_a_stale_image_is_refused_at_preflight
  TestVulkanLaneGuard::test_the_stale_image_refusal_honours_the_inspection_opt_in
  TestLoadSyncThreadsTheResolvedImage::test_a_stale_pin_is_refused_through_load_sync   (DID NOT RAISE)
```

The **guard logic is fine** — membership-in-a-capable-set is the right design and stays correct. The break is purely that #1973's tests hardcode `DEFAULT_ROCMFPX_IMAGE` where they mean the literal `ghcr.io/hal0ai/hal0-rocmfpx:ade07ba`. **The fix belongs on #1973** (swap the constant for the literal stale ref in those four tests), but it must happen before whichever of the two lands second, or main goes red. Raising it here because this PR is the one that moves first. Also note #1973's `#: DEFAULT_ROCMFPX_IMAGE is NOT a member while the default pin is still ...` comment becomes false on merge.

---

### 🟡 Nit — B1's key selection is an `elif` chain, but the resolver's precedence is the opposite

`updater.py:2877-2886` binds the holder to the bare `image` key **first** and only falls through to `image_pin`. `_resolve_image_ref` (`container.py:319-326`) does the reverse — `image_pin` wins and the bare `image` string **is not read at all** any more. So when a slot carries both keys, the sweep inspects the one the runtime ignores. Executed:

```
image = <current default>, image_pin = <ade07ba>
  → retagged: 0, effective image after sweep: ghcr.io/hal0ai/hal0-rocmfpx:ade07ba   ❌ still serves the retired runner

image = <ade07ba>, image_pin = <operator debug build>
  → retagged: 1, effective image unchanged   ❌ counts + logs a retag that changed nothing the runtime reads
```

Reachability is genuinely low — `_apply_slot_fold` sets `drop_slot_image = own_image is not None` unconditionally (`hw_slot_ownership.py:282, 334-341`), so `migrate-hw --apply` always drops the bare key when it folds. Both-keys therefore means a hand-edited TOML or a box that hasn't run the (operator-initiated, deploy-window-gated) fold. That's why this is a nit and not a blocker. But the fix is two lines — check `image_pin` **first**, and prefer it as the holder when both are present, mirroring the resolver — and it removes a silent divergence between "what the sweep audits" and "what the container launches". Worth an added case in `test_image_pin_retag.py` either way; the file currently has no both-keys test.

### 🟡 Nit — the pin is a mutable tag, not a digest

`DEFAULT_ROCMFPX_IMAGE = "ghcr.io/hal0ai/hal0-combined:0822"` is a tag. Nothing prevents `:0822` being re-pushed over, and every box that pulls after that silently runs different bytes than the one that passed the §3-C matrix — with `STALE_ROCMFPX_IMAGE_REFS` matching on the tag string, the retag would not even notice. For a **signed release pin** backing a GA ledger row, `@sha256:...` is the shape that makes the validation evidence mean something. Explicitly **non-blocking** and out of scope here — flagging it so it lands somewhere deliberate rather than by omission. #1970/#1972 is the natural home; the anonymous-pull digest for `:0822` was already recorded on an earlier revision of this PR, so the input exists.

### Codex triage

Most inline comments on this PR are against `3c9c53df` / `80f2a849` / `74499281` — the pre-reshape shape — and are **moot** now the brain swap is gone (all the `migrate_brain_model_off_minicpm5` findings, the seed-profile `--reasoning` findings, the installer-rationale finding, "Remove the obsolete MiniCPM migration"). Of the ones still live against `d8008586`, the `schema.py:1013` P1 (**Re-enable the Vulkan lane for the fixed runner**) is **valid and is my blocking finding above**. Recommend the body say which Codex findings are dead-by-reshape rather than leaving a reader to diff commit ids.

### CI

`sunset` ✅ (192 ≤ 192), `ui` ✅, `γ-suite (chromium)` ✅, CodeQL/Analyze ✅. **`python (3.12)` still pending** at the time of writing — that one must be green before the operator merges, since it is the only job that runs the new `test_image_pin_retag.py`. I ran the full quoted scope locally on `d8008586`: **3366 passed, 9 skipped**, matching the body.

---

### Summary

The repin itself is correct, the diff is clean, B1 is fixed and I proved it by execution, and B5 is substantively resolved. Nothing here says don't ship `:0822`. What I am asking for before the operator merge:

1. **Blocking (claim):** say plainly that this pin is inert on kfd-less boxes until #1973, or sequence #1973 first. The ct151 evidence currently oversells what merging this achieves.
2. **Blocking (composition):** #1973's four gate tests must stop using `DEFAULT_ROCMFPX_IMAGE` as the stale-lineage stand-in, or main goes red in either land order.
3. Nit: fix the `elif` precedence inversion + add a both-keys test.
4. Nit: correct or substantiate the "13 failures" figure.
5. Nit: decide the tag-vs-digest question deliberately in #1970/#1972.

thinmintdev:
## Delta re-review `d8008586..6c6a84d7` — verdict: **0 blocking**

Independent verification of a commit I did not write. Everything below was executed, not read. Both of my blocking findings are resolved, and I verified the *remedy* as well as the description — a sequencing note that documents a broken chain would not have cleared this.

---

### 1. Precedence fix — correct, and red-first confirmed independently ✅

The sweep now binds `image_pin` before the bare `image` key, in both shapes, mirroring `_resolve_image_ref`. I reverted the `updater.py` hunk alone and re-ran the test file against the old ordering:

```
2 failed, 5 passed
  test_both_keys_present_the_pin_wins_like_the_resolver
  test_both_keys_with_a_deliberate_pin_is_untouched
```

Exactly the two new tests, failing for the right reason, with the five pre-existing cases still green. The red-first claim is accurate.

Re-running **my own original probes verbatim** against the fix — the two cases that produced the finding:

```
shadowed pin  → retagged: 1, effective: ghcr.io/hal0ai/hal0-combined:0822   (was: still ade07ba)
dead bare key → retagged: 0, effective: ghcr.io/acme/exp:deadbeef           (was: retagged 1, changed nothing served)
```

Both fixed. I also probed four edges the new ordering could plausibly have broken:

| Edge | Result |
|---|---|
| `[image]` **table** (#599 trap) present, no ref | ignored, `retagged == 0` ✅ |
| `[image]` table + sibling stale `image_pin` | table ignored, pin retagged ✅ |
| Nested `[slot].image_pin` stale + top-level bare `image` stale | nested pin wins and is retagged ✅ |
| Bare key only (unfolded box) | still retagged, no regression ✅ |

`3368 passed, 9 skipped` across the quoted scope (+2 = the new tests), `ruff check` clean, `ruff format --check` clean (1165 files), `check-sunset` 192 ≤ 192.

**On the deliberate tradeoff in `test_both_keys_with_a_deliberate_pin_is_untouched`:** not retagging inert bare-key debris is the right call — but I wanted to know whether that debris is *permanent*, since "correct but accumulates forever" would be a new small problem traded for the old one. It isn't. Executed against `plan_unfold`/`_apply_slot_fold`:

```
SlotFold(set_image_pin=None, drop_slot_image=True)
after: {'name': 'code', 'device': 'gpu-rocm', 'image_pin': 'ghcr.io/acme/exp:deadbeef'}
```

The fold drops the inert key and preserves the deliberate pin. The tradeoff is self-healing. Worth a line in the test docstring, since that's the fact that makes it a tradeoff rather than a leak — **nit, non-blocking.**

### 2. Body honesty — accurate, and the remedy itself verified ✅

The new **"Sequencing — this PR alone does NOT deliver ct151"** section states the mandatory `#1973 → #1959` order with the correct relabel-before-retag mechanism, and reframes the ct151 rows as the chain's end state rather than this diff's standalone effect. That is exactly the correction I asked for.

But the section makes a *load-bearing claim of its own* — "with #1973's lane re-enable landed first, the lane survives the relabel and the retag delivers `:0822`". A sequencing note is only worth anything if the sequence works, so I composed the two branches (#1973 @ `85910fba` + this head) and re-ran the probe that produced the original finding:

```
COMPOSED, kfd ABSENT  → device: gpu-vulkan | image_pin: ghcr.io/hal0ai/hal0-combined:0822   ✅
COMPOSED, kfd PRESENT → device: gpu-vulkan | image_pin: ghcr.io/hal0ai/hal0-combined:0822   ✅
```

with `updater.vulkan_migration_slot_lane_supported … not migrated` on the relabel pass. The chain genuinely delivers `:0822` to the kfd-less box. Blocking A is resolved in substance, not just in prose.

**B5 correction** is in the body and is honest about being a correction — "kills **4 tests, one per provenance predicate** … an earlier revision of this body claimed 13, which did not reproduce." Matches my measurement. The **Open** section is refreshed (perf row cleared, #1970 → #1972 with its outstanding revisions named).

### 3. Composition with #1973 @ `85910fba` — now clean ✅

My earlier 4 failures are gone: **408 passed, 0 failed** across `tests/providers/test_gpu_vulkan_lane_gate.py` + `tests/updater` on the composed tree.

I checked that the tests were *fixed rather than deleted*, since "0 failures" is also what removing them looks like: all 36 gate tests are present, all four previously-failing names still exist, and they now key on a literal `ADE07BA_REF = "ghcr.io/hal0ai/hal0-rocmfpx:ade07ba"` (line 50) instead of `DEFAULT_ROCMFPX_IMAGE`. That is the right fix.

### 4. Scope and CI ✅

Delta is 2 files, +43/-7 — `updater.py` (precedence) and `test_image_pin_retag.py` (two tests). No `CHANGELOG.md` hunk. Base is `main`; the rebase picked up merged #1954/#1981 cleanly.

Polled to conclusion — **all 8 checks green**, including the one that was outstanding at my last review:

```
python (3.12)  pass  25m21s
γ-suite        pass  12m29s
sunset ✅  ui ✅  CodeQL ✅  Analyze ×3 ✅
```

---

### One observation for #1973, explicitly **not** gating this PR

`_vulkan_lane_is_loadable` returned `True` for a slot whose `image_pin` was *still* the stale `ade07ba` at relabel time — it looks through to the resolved default rather than judging the literal pin. That is what makes the ordering work, so it reads as deliberate. The narrow consequence: `retag_stale_slot_images` is wrapped non-fatal, so if it ever throws, that slot is left on `gpu-vulkan` holding the broken `ade07ba` pin — the #1888 invalid-token lane — where the pre-#1973 behaviour would have degraded it to CPU (slow, but correct output). Failure-path only, and it belongs to #1973's predicate rather than anything in this diff. Raising it so it's on the record for that PR's own review rather than discovered later.

### Verdict

**0 blocking.** Both findings resolved, the fix is minimal and red-first, the body now describes what the diff actually does, and the sequencing remedy holds under execution. Per the standing convention this is still a comment and not an `--approve` — **this PR is operator-merged** (default runner pin, GA ledger row 10) and the merge is delegated. No objection from the review side once **#1973 lands first** — that order is mandatory, not advisory, and it is the only thing between this and green.

PR #1961, opened 2026-08-20
fix(updater): run post-activation migrations after the symlink swap

### LINKED ISSUE #1960: updater: post-activation migrations run pre-swap in the outgoing tree — a release's new migrations never apply on the upgrade that ships them
Found during independent review of the rc.7 validation-kit PR (#1956, review comment 5353694749), verified against source.

## What happens

`Updater.commit` runs `run_post_activation_migrations` in-process at Step 7 (`updater.py:3796`, via `asyncio.to_thread`) — **before** Step 8's `seam.activate` symlink swap (`updater.py:3849`). The in-process call executes the currently-imported module set, i.e. the **outgoing** version's migrations. The codebase already states this constraint explicitly: `_rerender_units_after_swap`'s docstring (`updater.py:3538-3542`) explains that unit re-render was moved to a subprocess of the just-repipped venv for exactly this reason — 'the caller's own interpreter is still running the PRE-swap code'. Migrations were not given the same treatment.

There are only two call sites of `run_post_activation_migrations`: the updater path above, and `installer/install.sh:2467` (fresh installs — which run the new version's code, but have no legacy state to migrate).

## Consequence

A migration shipped in release N+1 never runs on the N → N+1 in-place upgrade — the population it was written for. Concrete instance about to matter: #1934's gpu-vulkan → gpu-rocm/cpu slot-TOML relabel (shipped for rc.7, answering #1924) will not fire on any rc.6 → rc.7 `hal0 update`. A box like the fleet's update-lane box (stale `device = "gpu-vulkan"` TOMLs, /dev/kfd present) upgrades to rc.7 and keeps its stale TOMLs; the relabel would only run when that box later upgrades *out of* rc.7. The same applies to every future migration under the current ordering.

## What should happen

Migrations must execute the incoming version's code against the activated tree — same pattern as `_rerender_units_after_swap`: a subprocess of the just-repipped venv, after `seam.activate`, with journal breadcrumbs and `job_id` threading preserved (#1935). Alternatively (or additionally, as a safety net) run outstanding migrations at daemon startup of the new version — that also heals boxes that upgraded before the fix.

## Why it gates rc.7

The rc.7 validation run's update lane exists to prove the rc.6 → rc.7 path, and the #1934 migration is one of its named verifications (kit v5, `post-upgrade.md` 2b). Without this fix the migration story shipped in rc.7 is inert for every actual upgrader.

## Environment

main @ 39a3901e. Repro is code-inspection-level: the call ordering in `Updater.commit` plus Python's import semantics; a live repro is the rc.6 → rc.7 update lane on a box with gpu-vulkan TOMLs (validation kit, `boxes.toml` update box).

thinmintdev:
## Summary

Fixes #1960: `Updater.commit()` ran `run_post_activation_migrations` in-process (`asyncio.to_thread`) at Step 7, **before** Step 8's `seam.activate()` symlink swap. Python does not hot-swap already-imported modules, so that in-process call executed the currently-running (OUTGOING) version's migration set, never the incoming release's. Concretely: #1934's gpu-vulkan → gpu-rocm/cpu slot-TOML relabel, shipped for rc.7, would have been inert for every rc.6 → rc.7 `hal0 update` — it would only run the *next* time that box updated. The codebase already documents this exact trap in `_rerender_units_after_swap`'s docstring (unit re-render was moved to a subprocess of the just-repipped venv for precisely this reason); migrations were never given the same treatment.

## What changed

- **`_run_post_activation_migrations_after_swap`** (new, `updater.py`): runs `run_post_activation_migrations` in a subprocess of the just-repipped venv — mirrors `_rerender_units_after_swap` exactly (same `sys.executable -c ...` pattern, same reasoning). Args are passed as a single JSON argv element (not interpolated into the script text) so `job_id` can never be mis-parsed as code.
- **`Updater.commit()`**: the migration *call* moves to after `seam.activate()`, before the unit re-render (so units render from already-migrated config — a relabelled device, sanitised `extra_args`, etc). The `ceiling`/profile-catalog-reset-gating derivation is unchanged, just relocated to where it's still needed. `job_id` threading (#1935) is preserved end to end.
- **Failure-mode decision** (the interesting part): a failed post-swap migration subprocess does **not** abort `commit()`. By the time it runs, the tree is already active — the symlink swap and venv re-pip already succeeded — so raising and discarding (the old pre-swap behaviour) would either be wrong (the release *is* running) or actively destructive (the old code path called `seam.discard()` on the target dir, which post-swap is the **active** `current` tree). Instead the failure is made **loud**: an ERROR-level journal line (`updater.post_swap_migration_failed`, louder than `_rerender_units_after_swap`'s own WARNING for its lower-stakes non-fatal failure), a new `convergence.post_activation_migrations` block (`{ok, from, to, error}`) that flips `convergence.converged` to `False` so `hal0 update` refuses to report a clean success, and a red panel in the CLI (`cli/update_commands.py`'s `_print_convergence`) telling the operator what happened and that it self-heals on the next `hal0-api` restart.
- **`check_outstanding_migrations()`** (new, boot-time safety net, wired into `hal0-api`'s `_boot_slot_reconcile` phase, first in that phase): re-runs the same six migration passes on every `hal0-api` boot. Each pass is a documented idempotent "stale? fix it : no-op" sweep, so a converged box pays for one `hal0.toml` read plus five near-instant no-ops. This heals two populations: boxes that upgraded *before* this fix (permanently missed their migrations — there's no later `commit()` to retry them) and boxes whose post-swap subprocess itself failed. It derives its `ceiling` the same way `commit()` does, via the read-only `profile_reset_status()`, so it cannot race `meta.schema_version` past the one-shot profile-catalog-reset gate outside the consent flow.
- `installer/install.sh`'s call site is untouched — `run_post_activation_migrations`'s signature and behaviour are unchanged; only *how* `commit()` invokes it changed. Fresh installs already run the new code with nothing to migrate.

## Verification

Test-first: added `test_commit_runs_post_activation_migrations_after_the_swap` (asserts exactly one subprocess call whose argv contains the migration marker, and that the symlink already points at the incoming version when it fires) and `test_commit_post_activation_migration_failure_is_non_fatal_but_loud` (asserts the swap stands, `convergence.post_activation_migrations.ok is False`, `converged is False`, and an ERROR-level `updater.post_swap_migration_failed` log line) to `tests/updater/test_updater.py`. Confirmed both **fail** against the pre-fix code (no subprocess call exists at all / no `convergence` key), then implemented the fix and confirmed both pass.

- `tests/updater` (307 tests, was 297 — 10 new unit tests in `tests/updater/test_post_swap_migrations.py` covering `_run_post_activation_migrations_after_swap`'s argument-passing/marker-parsing/error-surfacing and `check_outstanding_migrations`'s ceiling derivation + never-raises contract): **all pass**.
- `tests/cli/test_update_commands.py` (34 tests, 3 new for `_print_convergence`'s new panel/success line): **all pass**.
- `ruff check` + `ruff format --check` on all touched files: clean.
- `mypy` on touched files: pre-existing baseline errors only (verified by line-content inspection — none in the new/changed code); not part of the required gate (`make lint` is ruff-only per CI).
- Sanity swept `tests/api` (touched `_boot_slot_reconcile`'s docstring/body): `test_health_degraded.py`, `test_boot_brain_lane_replay.py`, and a 381-test slot/boot/migration/registry-filtered slice all pass — no regression to app boot.

CHANGELOG.md intentionally untouched (consolidated changelog lands separately, #1545).

Closes #1960

### DIFF
--- a/src/hal0/api/__init__.py
+++ b/src/hal0/api/__init__.py
@@ -1325,12 +1325,33 @@ async def _boot_dispatcher(app: FastAPI, ctx: BootState) -> None:
 async def _boot_slot_reconcile(app: FastAPI, ctx: BootState) -> None:
     """Phase — one-shot slot reconciliation passes + idle-monitor start.
 
-    ORDERING: ``migrate_slot_dir`` (#1369) → ``reconcile_unconfigured_slots``
-    → ``reconcile_npu_trio_slots`` → ``fold_identity`` (folds identity for the
-    trio shadows just reconciled) → ``start_idle_monitor``. Otherwise preserved
-    exactly from the monolithic boot.
+    ORDERING: ``check_outstanding_migrations`` (#1960, safety net) →
+    ``migrate_slot_dir`` (#1369) → ``reconcile_unconfigured_slots`` →
+    ``reconcile_npu_trio_slots`` → ``fold_identity`` (folds identity for the
+    trio shadows just reconciled) → ``start_idle_monitor``. Everything from
+    ``migrate_slot_dir`` on is otherwise preserved exactly from the
+    monolithic boot.
     """
-    # #1369 one-shot sweep, FIRST: every pass below reads ``model.default`` to
+    # #1960 safety net, FIRST: heal a box whose post-activation data
+    # migrations (schema, slot-TOML relabels, registry cleanup) never ran —
+    # either because it upgraded before this fix existed, or because
+    # Updater.commit()'s post-swap migration subprocess itself failed (see
+    # that function's docstring). Every pass check_outstanding_migrations
+    # runs is a documented idempotent "stale? fix it : no-op" sweep, so a
+    # converged box pays for one hal0.toml read plus five near-instant
+    # no-ops. Runs before every pass below because those passes read slot
+    # TOMLs / registry rows this can rewrite. Never raises — see its own
+    # docstring — so no try/except is needed here, but exceptions are boxed
+    # anyway (defensive-in-depth, matching every neighbouring pass in this
+    # phase) in case the import itself fails.
+    try:
+        from hal0.updater.updater import check_outstanding_migrations
+
+        check_outstanding_migrations()
+    except Exception as exc:  # pragma: no cover — defensive
+        log.warning("slot.outstanding_migrations_check_failed", error=str(exc))
+
+    # #1369 one-shot sweep, next: every pass below reads ``model.default`` to
     # decide what is configured, so a slot the operator had switched off with
     # ``enabled = false`` (whose model is therefore still bound on disk) would
     # be reconciled as live for one boot. Idempotent — after the first sweep
--- a/src/hal0/api/routes/updater.py
+++ b/src/hal0/api/routes/updater.py
@@ -460,6 +460,17 @@ async def _run_commit_job(
         job["error"] = str(exc)
         job["error_code"] = type(exc).__name__
     else:
+        # Convergence must land on the job BEFORE the terminal-state persist
+        # below, not after the restart the way it read before (#1960 B3):
+        # the restart a few lines down is the exact #1540 mechanism — it
+        # shells out from inside hal0-api's own cgroup and can SIGTERM this
+        # very process while it's in flight — so anything set only
+        # afterwards may never reach disk. A red (unconverged) post-swap
+        # migration failure has to survive that race, or a CLI that
+        # re-attaches after the restart reads `convergence: None` and
+        # _print_convergence reports a clean success for a run that wasn't.
+        job["convergence"] = (result or {}).get("convergence")
+
         # Terminal state first, then the restart - see the long note in
         # _run_apply_job. commit() is the path #1540 was reported against:
         # the restart kills this process, so "applied" has to already be on
@@ -473,7 +484,6 @@ async def _run_commit_job(
         restarted, restart_error = await asyncio.to_thread(_try_restart_hal0_api)
         job["restarted"] = restarted
         job["restart_error"] = restart_error
-        job["convergence"] = (result or {}).get("convergence")
     finally:
         job["updated_at"] = time.time()
         _persist_job(job)
--- a/src/hal0/cli/update_commands.py
+++ b/src/hal0/cli/update_commands.py
@@ -351,6 +351,18 @@ def _maybe_converge_profiles(status: dict | None, *, yes: bool) -> bool:
     return True
 
 
+#: Human labels for the non-fatal per-pass failure events
+#: run_post_activation_migrations can log (#1960 B2) — keys must match
+#: hal0.updater.updater._NON_FATAL_MIGRATION_FAILURE_EVENTS exactly.
+_MIGRATION_PASS_LABELS: dict[str, str] = {
+    "updater.seed_profiles_prune_failed": "seed-profile cleanup",
+    "updater.mtp_migration_failed": "mtp override cleanup",
+    "updater.vulkan_migration_failed": "gpu-vulkan slot relabel",
+    "updater.image_retag_failed": "runner image retag",
+    "updater.extra_args_sanitize_failed": "managed extra_args sanitisation",
+}
+
+
 def _print_convergence(convergence: dict | None) -> bool:
     """Report how far the updated box is from the v1.0 on-disk shape.
 
@@ -394,6 +406,15 @@ def _print_convergence(convergence: dict | None) -> bool:
         src, dst = post_migrations.get("from"), post_migrations.get("to")
         if src is not None and dst is not None and src != dst:
             console.print(f"[green]✓[/green] on-disk schema migrated v{src} → v{dst}")
+        # A pass that swallowed its own exception into a warning is still
+        # non-fatal by design (see run_post_activation_migrations) — this
+        # does not flip `ok`/`converged` — but it must not be invisible
+        # outside journalctl either (#1960 B2).
+        for event in post_migrations.get("pass_warnings") or []:
+            label = _MIGRATION_PASS_LABELS.get(event, event)
+            console.print(
+                f"[yellow]![/yellow] {label} logged a warning — see journalctl for hal0-api"
+            )
     elif post_migrations:
         console.print(
             Panel(
--- a/src/hal0/cli/update_commands.py
+++ b/src/hal0/cli/update_commands.py
@@ -384,6 +384,30 @@ def _print_convergence(convergence: dict | None) -> bool:
     elif outcome == "error":
         console.print(f"[yellow]![/yellow] profile catalog reset failed: {reset.get('error')}")
 
+    # #1960: post-activation migrations (schema + slot/registry cleanup
+    # sweeps) now run in a subprocess AFTER the symlink swap, so a failure
+    # there does NOT abort the update — the release is already active. It
+    # still has to be reported, loudly, or "hal0 update" would silently hand
+    # back a box whose on-disk data still reflects the previous release.
+    post_migrations = convergence.get("post_activation_migrations") or {}
+    if post_migrations.get("ok"):
+        src, dst = post_migrations.get("from"), post_migrations.get("to")
+        if src is not None and dst is not None and src != dst:
+            console.print(f"[green]✓[/green] on-disk schema migrated v{src} → v{dst}")
+    elif post_migrations:
+        console.print(
+            Panel(
+                "[bold red]Post-update data migrations did not apply.[/bold red]\n"
+                f"{post_migrations.get('error') or 'unknown error'}\n\n"
+                "The new code is installed and running, but on-disk data (slot TOMLs,\n"
+                "hal0.toml schema, registry rows) may still reflect the PREVIOUS release's\n"
+                "shape. hal0-api retries this automatically at its next start — restart the\n"
+                "service now to retry immediately, or wait for the next scheduled restart.",
+                title="Migration incomplete",
+                border_style="red",
+            )
+        )
+
     ownership = convergence.get("ownership_migrations") or {}
     pending = list(ownership.get("pending") or [])
     if not pending:
--- a/src/hal0/updater/updater.py
+++ b/src/hal0/updater/updater.py
@@ -1620,6 +1620,7 @@ def run_post_activation_migrations(
     *,
     job_id: str | None = None,
     ceiling: int | None = None,
+    skip_image_retag: bool = False,
 ) -> tuple[int, int]:
     """Run every post-activation migration pass hal0 ships, in order.
 
@@ -1661,6 +1662,23 @@ def run_post_activation_migrations(
     profile-catalog reset gate). ``None`` (the default, and what
     install.sh's caller always passes) applies no cap.
 
+    ``skip_image_retag`` (#1960 N2) omits ``retag_stale_slot_images``.
+    Every OTHER pass here is either purely additive (seed profiles, mtp
+    clearing) or corrects a state that actively breaks something (a
+    gpu-vulkan slot that refuses to load, an unsanitised managed flag) —
+    running them outside an update is harmless or actively good.
+    ``retag_stale_slot_images`` is different: it rewrites a slot's image
+    pin whenever that pin exactly matches a KNOWN FORMER default, which is
+    a good bet during an actual release (a new default just shipped) but
+    is not correlated with "a release just shipped" at all when called on
+    every daemon boot — an operator who deliberately re-pinned a slot to
+    an older image that happens to be a listed former default would have
+    that choice silently reverted on every crash-restart or reboot, not
+    just when they run an update. :func:`check_outstanding_migrations`
+    passes ``True`` here for exactly that reason; :meth:`Updater.commit`'s
+    post-swap subprocess (an actual update) always passes the default
+    ``False``, so the pass still gets its shot every time a release ships.
+
     Returns the ``(source, target)`` schema-version tuple for breadcrumb
     logging, the same shape ``commit()`` already returned.
     """
@@ -1687,11 +1705,12 @@ def run_post_activation_migrations(
         # Non-fatal: an affected slot keeps refusing to load (the #1923
         # preflight guard) until the operator manually relabels its device.
 
-    try:
-        retag_stale_slot_images(job_id=job_id)
-    except Exception as exc:
-        log.warning("updater.image_retag_failed", job_id=job_id, error=str(exc))
-        # Non-fatal: a stale pin just keeps the previous runner image.
+    if not skip_image_retag:
+        try:
+            retag_stale_slot_images(job_id=job_id)
+        except Exception as exc:
+            log.warning("updater.image_retag_failed", job_id=job_id, error=str(exc))
+            # Non-fatal: a stale pin just keeps the previous runner image.
 
     try:
         sanitize_model_extra_args(job_id=job_id)
@@ -1713,11 +1732,26 @@ def check_outstanding_migrations(*, job_id: str | None = None) -> tuple[int, int
     :func:`_run_post_activation_migrations_after_swap` can itself fail (disk
     full, a bug in one specific pass), and ``commit()`` deliberately does not
     retry it inline — see that function's caller for why. Calling this once
-    at every ``hal0-api`` boot heals both: every one of the six passes
-    ``run_post_activation_migrations`` runs is documented as an idempotent
-    "stale? fix it : no-op" sweep (see each pass's own docstring), so a
-    converged box pays for one ``hal0.toml`` read plus five near-instant
-    no-ops.
+    at every ``hal0-api`` boot heals both: every pass
+    ``run_post_activation_migrations`` runs here is documented as an
+    idempotent "stale? fix it : no-op" sweep (see each pass's own
+    docstring), so a converged box pays for one ``hal0.toml`` read plus a
+    handful of near-instant no-ops.
+
+    ``skip_image_retag=True`` (#1960 N2): unlike every other pass here,
+    ``retag_stale_slot_images``'s heuristic ("this pin exactly matches a
+    known FORMER default") is only actually correlated with staleness right
+    after a release ships a new default — see
+    :func:`run_post_activation_migrations`'s own docstring for the full
+    reasoning. Running it on every boot (crash-restarts and reboots
+    included, not just updates) would silently revert an operator's
+    deliberate re-pin to an old image on every bounce — the same class of
+    bug as #1867 (an automated convergence pass overwriting deliberate
+    operator state), just with a much higher firing frequency. The pass
+    still gets its shot on every actual update, via
+    :meth:`Updater.commit`'s post-swap subprocess, which does not skip it —
+    a box that misses it there only waits until its next real update to
+    converge, same as before this function existed.
 
     Mirrors :meth:`Updater.commit`'s own ``ceiling`` derivation, using
     :func:`profile_reset_status` — a pure read — so a box whose one-shot
@@ -1736,7 +1770,7 @@ def check_outstanding_migrations(*, job_id: str | None = None) -> tuple[int, int
     try:
         reset_outstanding = bool(profile_reset_status().get("due"))
         ceiling = PROFILE_CATALOG_SCHEMA_VERSION - 1 if reset_outstanding else None
-        return run_post_activation_migrations(job_id=job_id, ceiling=ceiling)
+        return run_post_activation_migrations(job_id=job_id, ceiling=ceiling, skip_image_retag=True)
     except Exception as exc:
         log.warning("updater.boot_migration_check_failed", job_id=job_id, error=str(exc))
         return None
@@ -3617,12 +3651,34 @@ async def _rerender_units_after_swap(job_id: str | None) -> None:
         )
 
 
+#: Sentinel key on the single JSON stdout line the migration subprocess
+#: emits. Mirrors :data:`hal0.updater.privileged._RESULT_KEY`'s exact
+#: contract: structured logs go to STDERR (see the script built below) so
+#: stdout carries nothing but this envelope, scanned from the end.
+_MIGRATION_RESULT_KEY = "hal0_migration_result"
+
+#: The five non-fatal per-pass failure events ``run_post_activation_migrations``
+#: can log without raising (#1960 B2). A pass that swallows its own
+#: exception still exits the subprocess with rc=0 — by design, see that
+#: function's docstring — so it would otherwise be invisible outside
+#: journalctl. ``_run_post_activation_migrations_after_swap`` scans the
+#: relayed child log for these exact event names so a swallowed failure
+#: still surfaces in commit()'s convergence report.
+_NON_FATAL_MIGRATION_FAILURE_EVENTS: tuple[str, ...] = (
+    "updater.seed_profiles_prune_failed",
+    "updater.mtp_migration_failed",
+    "updater.vulkan_migration_failed",
+    "updater.image_retag_failed",
+    "updater.extra_args_sanitize_failed",
+)
+
+
 async def _run_post_activation_migrations_after_swap(
     min_data_version: int,
     *,
     job_id: str | None,
-    ceiling: int | None,
-) -> tuple[int, int]:
+    reset_outstanding: bool,
+) -> dict[str, Any]:
     """Run :func:`run_post_activation_migrations` in a subprocess of the
     just-repipped venv, AFTER the ``current`` symlink swap (#1960).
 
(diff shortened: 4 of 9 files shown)


thinmintdev:
## Independent review — `changes-requested`

Reviewed `21734a10` against `main` (`58e2161a`). The core diagnosis and the core fix are right: `run_post_activation_migrations` now executes the **incoming** tree, post-swap, in the just-repipped venv, and I confirmed that mechanically (rubric 1 below). Three things block, all of them about what happens *around* the new subprocess rather than inside it. Both Codex P1s are real; I confirmed them and adopted them as B1 and B3.

---

## Blocking

### B1 — a subprocess timeout or launch failure escapes `commit()` **after** the swap
`updater.py:3675` uses `subprocess.run(..., timeout=300)`. That raises `subprocess.TimeoutExpired`; a launch failure raises `OSError`. Neither is `UpdateError`, so the deliberately non-fatal `except UpdateError` at `updater.py:3975` does not apply. The exception escapes `commit()` → `_run_commit_job` marks the job `failed` → **`_try_restart_hal0_api` is skipped**. Net state: symlink already at the new release, venv already re-pipped, daemon never bounced, `hal0 update` exits 1 "update failed". That is a new failure mode — pre-fix, every failure of this call happened *before* the swap, so aborting was correct.

Confirmed empirically (probe appended to `tests/updater/test_updater.py` on the branch, then reverted):

```
PROBE: commit() RAISED TimeoutExpired: Command '[...run_post_activation_migrations...]'
       timed out after 300 seconds — symlink already at 'hal0-0.0.1'
PROBE: commit() RAISED OSError: [Errno 12] Cannot allocate memory
       — symlink already at 'hal0-0.0.1'
```

`_rerender_units_after_swap` uses a bare `except Exception` for exactly this reason. Catch `Exception` at the call site (or normalize inside the helper to `UpdateError`), and add a regression test — the two new `commit()` tests both drive `subprocess.run` through a fake that returns, never one that raises.

### B2 — the migration subprocess's journal breadcrumbs are captured and discarded (regresses the #1935 contract, and breaks the rc.7 kit's own acceptance check)
`capture_output=True` captures the child's structlog output. The parent scans stdout for the marker line and throws everything else away on success; on failure it keeps a 2000-char stderr tail inside `details`. So every line the migration passes emit — `updater.migrations_applied`, `updater.slot_vulkan_relabeled_rocm` / `_cpu_fallback`, `updater.image_retagged`, `updater.mtp_override_cleared`, `updater.extra_args_sanitized` — each with `job_id` correctly threaded *inside the child* — never reaches the journal.

This is the exact gap #1935 closed for the privileged seam with `_relay_child_log` (`privileged.py:313`), whose docstring spells out the reasoning verbatim ("the child's structlog output is captured, not inherited, so without this replay every … breadcrumb is silently discarded on the happy path"). Its follow-up commit even fixed truncating that relay, because a tail could push breadcrumbs out.

Two concrete consequences, the second of which is why I'm marking it blocking rather than a nit:

1. **It converts loud failures into silent ones.** The five data-cleanup passes each catch their own exception, log a *warning*, and return normally. So a failed `relabel_stale_vulkan_slots` leaves the subprocess exiting 0, the parent reporting `post_activation_migrations.ok = True`, `converged: True` — and the warning discarded. The migration path's only failure signal is now "the schema migration raised".
2. **It fails #1960's own acceptance criterion.** `tests/release-validation/lanes/update/post-upgrade.md` step 2b verifies the #1934 relabel by grepping the post-upgrade journal for `updater.slot_vulkan_relabeled_rocm` / `_cpu_fallback`. With this PR as written, a *correctly working* rc.7 emits none of those lines from the update path — and the boot safety net won't emit them either, because the subprocess already did the work, so the boot pass is a no-op. The kit's documented reading of "TOMLs unchanged, no relabel breadcrumbs" is "first check whether #1960 is fixed in this release". The fix would present as unfixed.

Fix: reuse the `_relay_child_log` pattern — replay the child's captured output into the parent's log with `job_id`. Note structlog's *default* (unconfigured) logger writes to **stdout**, which is the same stream the marker is parsed from; either pin the child to stderr the way `_pin_logs_to_stderr` does and relay stderr, or relay every non-marker stdout line.

### B3 — `convergence` racily never reaches the CLI, so the "loud" contract does not hold on the happy path
`routes/updater.py:463-476`: `_run_commit_job` sets `job["state"]="applied"` and calls `_persist_job(job)` **before** the restart, then assigns `job["convergence"]` **after** `_try_restart_hal0_api()`. That ordering is not accidental — the sibling `_run_apply_job` carries a long #1540 comment explaining that the restart SIGTERMs this very process mid-flight, which is why the terminal state has to be on disk first. `job["convergence"]` is on the wrong side of that line.

So on the normal successful self-restart path the persisted job frequently lacks `convergence`; the reattached CLI passes `None` into `_print_convergence`, which returns `True` at `cli/update_commands.py:361-362` → green "update applied." panel, exit 0. Precisely when a post-swap migration failed, the red panel, the `converged: False` flip and the exit-2 contract this PR is built around can all be lost.

The placement is pre-existing (it already degrades `profile_reset` / `ownership_migrations` reporting), but this PR is the first thing to make that field load-bearing for a *failure* signal, and the fix is two lines: assign `job["convergence"] = (result or {}).get("convergence")` before the first `_persist_job(job)`. The parent's ERROR journal line survives regardless — the operator-facing signal does not.

---

## Non-blocking

**N1 — version skew in the `ceiling` input.** `ceiling` is derived pre-swap from the **outgoing** tree's `PROFILE_CATALOG_SCHEMA_VERSION` (`updater.py:3912`) and handed to the incoming tree's migrations as a bare integer. Inert today (both trees say 2), but the release that bumps the constant will cap a reset-outstanding box at the *old* watermark − 1. Cheap hardening: pass `reset_outstanding` (a bool) and let the child compute `PROFILE_CATALOG_SCHEMA_VERSION - 1` from its own constant — which is what both `check_outstanding_migrations` and `install.sh:2465` already do, so it would make all three call sites structurally identical.

**N2 — the boot net promotes upgrade migrations to every-boot invariant enforcers.** `retag_stale_slot_images` rewrites any slot pin exactly equal to a known former default. That used to happen only at update time; it now happens on every `systemctl restart hal0-api`. A deliberate operator re-pin to a former default — a live pattern on this fleet right now while the #1888/#1948 Vulkan-image bisect is running — is silently reverted at the next bounce, with no opt-out. The pass's own docstring leans on "demonstrably NOT a deliberate operator pin"; running it every boot makes that assumption load-bearing on a much shorter cycle. Worth a deliberate, documented position, or scoping the boot net to the unambiguously repair-only passes.

**N3 — the cost claim understates it.** "one `hal0.toml` read plus five near-instant no-ops" also involves a slot-TOML dir scan, a `profiles.toml` read, and two full model-registry enumerations (`clear_stale_mtp_overrides`, `sanitize_model_extra_args`). All cheap local IO, but it is a synchronous call inside an `async` boot phase, so it sits on the boot critical path. Fine as implemented — just don't let the comment set a false expectation.

**N4 — vacuous half-assertion.** `test_print_convergence_shows_the_migrated_schema_range_on_success` asserts `"v1" in out and "v2" in out`; `"v1"` also matches the pre-existing "converged to the **v1.0** shape" line, so half of it can never fail. Assert the real string (`"on-disk schema migrated v1 → v2"`).

---

## Verified good

**Rubric 1 — the subprocess genuinely executes the incoming tree. ✅** `sys.executable` is the hal0-api venv interpreter (`/usr/lib/hal0/venv/bin/python`); `seam.activate()` performs the symlink swap **and** the venv re-pip as one privileged operation before this runs, so the child imports the incoming `hal0` from that venv's site-packages. No `PYTHONPATH` anywhere in `hal0-api.service`. The inherited cwd cannot shadow it either: `WorkingDirectory=${CURRENT_LINK}` is resolved at daemon start and therefore pinned to the *outgoing* release's inode, but a release tree's root is `src/ docs/ ui/ installer/ …` with no top-level `hal0/` package (verified on CT105), so `-c`'s `sys.path[0]` contributes nothing importable. Argv construction is clean — args ride a single JSON argv element, never interpolated into the `-c` text. Same shape as `_rerender_units_after_swap`.

**Rubric 2 — ordering. ✅** Post-`activate`, pre-unit-re-render is the right slot; rendering units from already-migrated config is a good call and correctly justified. Post-swap-only migrations are also strictly *better* for rollback coherence than the old ordering (pre-fix, a failed swap left new schema + old code). `rollback()` still has no down-migration; correctly out of scope. The non-fatal position is deliberate and well argued — but its *delivery* is undermined by B3, and its failure *detection* by B2.

**Rubric 3 — `job_id` plumbing.** Threaded into the child and used in the child's own log calls; the plumbing itself survives the process boundary correctly. What does not survive is the output. See B2.

**Rubric 5 — the boot safety net.** Idempotent ✅, never raises ✅ (directly tested), heals pre-fix boxes ✅ (on the first boot of the fixed version), heals failed-subprocess boxes ✅. Its ceiling derivation matches `install.sh`'s exactly and correctly uses the read-only `profile_reset_status()`, so it cannot consume the one-shot reset gate outside the consent flow. No lock against a concurrent in-flight update — `write_toml_atomic` prevents corruption but not a lost update; low practical risk, since the boot net runs before serving and `commit()` finishes before the restart it triggers.

**Rubric 6 — `install.sh:2467` untouched. ✅** No `installer/` hunk in the diff; `run_post_activation_migrations`'s signature and behaviour are unchanged.

**Rubric 7 — tests. ✅** On the branch: `tests/updater` + `tests/cli/test_update_commands.py` → **341 passed**. Red-checked by reverting `src/` to `main`: both new `commit()` tests and 2 of the 3 new CLI tests fail, and `test_post_swap_migrations.py` doesn't even import. Mutation-checked the ordering assertion by relocating the whole migration block back above `seam.activate()` on the branch — `test_commit_runs_post_activation_migrations_after_the_swap` goes red on the *symlink* assertion, not merely on call count, so that assertion is genuinely load-bearing. `tests/api/test_startup_slot_seed.py` + `test_updater_routes.py` → 61 passed. `ruff check` on the touched files and `ruff format --check src/ tests/` both clean.

**Rubric 8 — scope. ✅** No `CHANGELOG.md` hunk, no installer hunk, no doc drive-bys. The `api/__init__.py` safety net is scope-adjacent but explicitly sanctioned by #1960's "alternatively (or additionally, as a safety net)".

**Rubric 9 — the rc.6 → rc.7 acceptance criterion.** The *mechanism* is correct: `relabel_stale_vulkan_slots` and `retag_stale_slot_images` both sit inside `run_post_activation_migrations`, which now runs the incoming tree post-swap, and the boot net gives a second shot after the restart. The #1934 relabel and the #1959 STALE-refs retag **will fire** on the upgrade. But per **B2** they will fire *silently*, and the rc.7 kit verifies them by journal grep — so as written this PR satisfies #1960's intent while failing its own release-validation check. B2 has to land before the rc.7 update lane runs.

---

## Codex triage

Both P1s **confirmed**, not dismissed:
- *"Treat migration timeouts as post-swap failures"* → **B1**, reproduced (also covers `OSError` on launch, which the comment doesn't mention).
- *"Persist migration convergence before restarting the API"* → **B3**, confirmed against the #1540 comment in the same file that documents exactly this hazard.

---

**Verdict: `changes-requested`.** B1 and B3 are small, surgical fixes. B2 is the one that actually matters for rc.7 — without the breadcrumb relay, the release-validation lane cannot observe the very migration this PR exists to make run.

<sub>Independent adversarial review; I did not author this change.</sub>

thinmintdev:
Revised per independent review (comment 5354502220). Pushed `ba0ea569`.

**B1 — exception containment.** `subprocess.run(..., timeout=300)` can raise `TimeoutExpired`, and launch failure can raise `OSError` — neither was caught by the old `except UpdateError` at the `commit()` call site, so either propagated straight out of `commit()` post-swap (symlink already swapped, job reported failed, `_try_restart_hal0_api` never reached — reviewer reproduced both). `_run_post_activation_migrations_after_swap` now converts both to `UpdateError`, and the `commit()` call site widens to a bare `except Exception`, matching `_rerender_units_after_swap`'s own containment: once the tree is active, nothing may escape. New tests: `test_commit_survives_a_migration_subprocess_timeout`, `test_commit_survives_a_migration_subprocess_launch_failure` (confirmed red against the pre-fix call site, now green).

**B2 — relay, don't discard.** The child now pins its structured log to stderr (mirrors `hal0.updater.privileged._pin_logs_to_stderr` exactly) so stdout carries only a JSON result envelope; the parent relays the full stderr via `privileged._relay_child_log` on success, restoring #1935's breadcrumb contract and the release-validation kit's journal grep. A non-fatal pass that swallows its own exception into a warning (rc=0 by design — `run_post_activation_migrations`'s five data-cleanup passes have always worked this way) is scanned for in the relayed text and surfaced as `pass_warnings` in `convergence.post_activation_migrations`, and by the CLI as a yellow notice — without flipping `ok`/`converged`, since that best-effort-non-fatal design predates this PR and a swallowed cosmetic warning isn't the same class of problem as the mechanism itself failing. New tests cover the relay, the pass_warnings extraction, and that relay is success-only (mirroring `UpdateSeam._invoke`'s own asymmetry).

**B3 — ordering.** `job["convergence"]` was set AFTER `_try_restart_hal0_api()`, which shells out from inside hal0-api's own cgroup and can SIGTERM this process mid-flight (#1540's exact mechanism, the same file's own docstring). Moved before the terminal-state persist. New test `test_commit_convergence_is_durable_before_the_restart_that_can_kill_us` in `tests/api/test_updater_routes.py`, mirroring `test_apply_persists_applied_before_attempting_the_restart`'s exact "read what's on disk at the moment the restart subprocess is invoked" technique — confirmed red by temporarily reverting the ordering fix, then restored and green.

**N1 — bool over ceiling int.** The wire payload now carries `reset_outstanding: bool`; the child derives `ceiling` from its own imported `PROFILE_CATALOG_SCHEMA_VERSION`, not a value pre-computed by the (outgoing) parent process. Matches `install.sh` and `check_outstanding_migrations`'s existing pattern.

**N2 — decision, with justification (documented in `run_post_activation_migrations` and `check_outstanding_migrations`'s docstrings too).** `check_outstanding_migrations()` now passes `skip_image_retag=True`. `retag_stale_slot_images`'s heuristic ("this pin exactly matches a known FORMER default") is only actually correlated with staleness right after a release ships a new default — that's true whether it runs during an actual update or via `install.sh`'s repair path. Running it on *every* `hal0-api` boot (crash-restarts and reboots included, not just updates the operator consciously triggered) would silently revert a deliberate operator re-pin to an old image that happens to be a listed former default, on every bounce — the same class of bug as #1867 (an automated convergence pass overwriting deliberate operator state), just with a much higher firing frequency than the existing per-update/per-repair-run exposure. Went with skipping the pass in the boot net: it still gets its shot on every real update via the post-swap subprocess (which does **not** skip it), so a box that misses it there only waits until its next real update — no worse than before this function existed. `run_post_activation_migrations` gained a `skip_image_retag: bool = False` parameter for this; every other pass is unaffected (each is either purely additive or already actively self-correcting, so running them outside an update is harmless or good — see the docstring for the full per-pass reasoning).

**N4** — tightened the loose `"v1" in out` CLI assertion to the exact rendered phrase (`"schema migrated v1 → v2" in out`).

Verification: `tests/updater` (317, was 307), `tests/cli/test_update_commands.py` (35, was 34), `tests/api/test_updater_routes.py` (59, +1 new) all pass. `ruff check` + `ruff format --check` clean on all touched files. Sanity-swept the broader `tests/api` slot/boot/migration/registry slice again (381 pass, unchanged) since B3 touched `routes/updater.py`.

thinmintdev:
## Delta re-review (`21734a10..ba0ea569`) — **0 blocking**

All three blocking findings are fixed, and fixed the right way rather than papered over. N1 and N4 are done; N2 has a real decision behind it. I re-ran every check and re-verified my own earlier reproductions against the new code. Detail below, then one non-blocking item I want on the record before this lands.

---

## B1 — contained ✅

`TimeoutExpired` → `UpdateError` and `OSError` → `UpdateError` inside the helper, **and** the `commit()` call site widened to bare `except Exception`. Belt and braces, which is right — the helper's conversion is the contract, the bare except is the containment backstop for the shape nobody predicted. The inline comment justifying the bare except ("by this point the release is already active, so NOTHING may escape") is the correct framing.

Red-checked: with `src/hal0/updater/updater.py` reverted to `21734a10`, `test_commit_survives_a_migration_subprocess_timeout` fails with `subprocess.TimeoutExpired` propagating straight out of `commit()` — the exact shape I reproduced last round.

*Nit, not blocking:* the two new tests assert `commit()` returns instead of raising, plus the convergence shape. They don't directly assert the downstream consequence I demonstrated (job not `failed` → `_try_restart_hal0_api` actually reached), because they sit at the `Updater.apply()` layer where the restart doesn't exist. That's a defensible test boundary and the `else`-branch/restart chain is separately covered — including by the new B3 route test, which drives the restart through that exact path. If you want the consequence pinned end-to-end, one route-level test asserting `state == "applied"` and `restarted is not None` when the migration subprocess times out would close it. Cheap, optional.

## B2 — relayed ✅ (with one note, below)

Correctly done, and better than the minimum I asked for:

- Child pins structlog to stderr via the same `structlog.configure(logger_factory=PrintLoggerFactory(file=sys.stderr))` call as `privileged._pin_logs_to_stderr`, so stdout carries only the envelope.
- Parent relays through the **real** `privileged._relay_child_log` (lazy import — correct, `privileged` imports from `updater`, so a module-level import would be circular).
- Envelope switched from a `KEY=` prefix to a JSON line scanned **from the end** with a `{`-prefix + key-presence guard, mirroring `privileged._parse_result`. Strictly more robust than what I reviewed.
- `pass_warnings` scanning the relayed stderr for `_NON_FATAL_MIGRATION_FAILURE_EVENTS` goes beyond my finding and closes the "swallowed pass failure is invisible" hole directly.

I checked the one hazard this shape could have had: the `structlog.configure` in the child runs *before* `from hal0.updater.updater import …`, so if anything on that import path called `structlog.configure` it would silently undo the stderr pin and send the whole relay to stdout. It doesn't — `structlog.configure` appears only in `cli/main.py` and `privileged.py` (the latter only under its `__main__` guard), neither of which is on this import path. The pin holds.

Mutation-checked: replacing the `_relay_child_log(...)` call with a no-op turns `test_relays_the_childs_log_into_the_parent_journal` red. The assertion is load-bearing, not incidental.

## B3 — durable ✅

`job["convergence"]` now lands before the terminal-state persist, ahead of the restart. The comment correctly names the #1540 mechanism.

Red-first claim verified, not just assessed as plausible: with `src/hal0/api/routes/updater.py` reverted to `21734a10`, the new test fails with

```
AssertionError: convergence was not durable on disk before the restart that can kill this
process; convergence persisted by then was [None, None]
```

The technique (read `_load_persisted_job` from inside the faked restart `subprocess.run`) genuinely captures the snapshot a same-cgroup SIGTERM would leave, and mirrors the existing `test_apply_persists_applied_before_attempting_the_restart`.

## N1 — de-skewed ✅

Exactly the prescribed shape: a `reset_outstanding` **bool** crosses the process boundary and the child re-derives `ceiling = PROFILE_CATALOG_SCHEMA_VERSION - 1` from its own import. I dumped the actual argv the helper builds to confirm rather than reading it off the diff:

```
args = json.loads(sys.argv[1])
ceiling = (PROFILE_CATALOG_SCHEMA_VERSION - 1) if args['reset_outstanding'] else None
...
payload: {"min_data_version": 1, "job_id": "j", "reset_outstanding": false}
```

All three call sites (`commit()`, `check_outstanding_migrations`, `install.sh`) now derive the ceiling from the same tree that runs the migrations.

## N2 — decided, documented, and wired correctly ✅

`skip_image_retag` defaults to `False`; only `check_outstanding_migrations` passes `True`. The reasoning in the docstring is the right one and lands on the right side (`retag_stale_slot_images`' heuristic is correlated with staleness *only just after a release ships a new default*, which is uncorrelated with "a daemon booted"), and the #1867 precedent is apt.

**Explicitly verified your wire-payload question:** the generated child script contains **no** `skip_image_retag` and the JSON payload is `{min_data_version, job_id, reset_outstanding}` only — so the post-swap subprocess calls `run_post_activation_migrations` with the default `False`. **The retag still fires on a real update.** `test_skip_image_retag_omits_only_that_pass` pins the other four passes as unaffected.

Consequence worth stating plainly, since it's the acceptance criterion #1960 exists for: the rc.6 → rc.7 upgrade path runs both `relabel_stale_vulkan_slots` (#1934) **and** `retag_stale_slot_images` (#1959 B1/B2) in the incoming tree, post-swap, with breadcrumbs now reaching the journal. The kit's `post-upgrade.md` step 2b journal grep for `updater.slot_vulkan_relabeled_rocm` / `_cpu_fallback` will find its lines — relayed as `updater.privileged_child_log` entries carrying the original text and `job_id`. That was the thing I blocked on last round; it's resolved.

The one trade-off the docstring correctly owns: a box that upgraded *before* this fix now heals everything except the image retag at boot, and waits for its next real update for that pass. Acceptable, and stated.

## N4 — tightened ✅

`assert "schema migrated v1 → v2" in out`. The vacuous half is gone.

## `pass_warnings` design — judged, and I endorse it

You asked whether surfacing pass warnings without flipping `ok`/`converged` honours my B2 or quietly under-reports. It honours it.

My B2 was *"a failed pass must be visible."* It is now visible in four places: the relayed child log line in the journal (with `job_id`), `convergence.post_activation_migrations.pass_warnings`, the `updater.convergence` parent log line's new `post_activation_migrations_pass_warnings` field, and a yellow CLI line with a human label. That is a real fix, not a token one.

Not flipping `converged` is the correct call, and I'd have pushed back if they had. Those five passes have been independently best-effort since they were written — an affected slot or model simply keeps its previous state until the operator acts. Flipping `converged` would make `hal0 update` exit 2 for a long-standing non-fatal outcome, which turns a visibility fix into a behaviour change and would fire on boxes that are fine. Reclassifying "non-fatal" as "blocks a clean success" is a policy decision that deserves its own issue and its own justification, not a rider on an observability fix. The code comment says exactly this. Agreed as written.

---

## Non-blocking, but please read before landing

**The success-only relay is defensible *here*, but its stated reason is wrong and its new test will fight #1950.**

The code comment says the relay is success-only *"mirroring UpdateSeam._invoke exactly"*, and `test_relay_does_not_happen_on_the_failure_path` cements that. Both are accurate against `main` today — and both are aimed at a target that is already moving: **#1940 is open**, and **PR #1950 is open and authored specifically to invert this**, moving `_invoke`'s relay ahead of the `returncode != 0` branch. #1940's own text even warns that `test_failed_stage_does_not_double_relay_stderr_as_breadcrumbs` "cements today's asymmetry and will need a deliberate update". This PR adds a second test that will need the same deliberate update, at a call site whose comment will then assert a mirroring that no longer exists.

Why this is *not* blocking, and why I'm not simply repeating my B2: the defect that makes #1940 real for `_invoke` has **no analogue at this call site**. There, a failure lands *after* a long successful breadcrumb trail (sha256, cosign, extract), so losing the relay loses real evidence. Here, a non-zero exit can essentially only originate in the **first** pass — `_maybe_run_config_migrations` is the only one that raises; passes 2–6 each swallow their own exception into a warning and cannot produce `rc != 0`. So on the failure path there is no preceding trail to lose. The substance of my B2 is fully addressed. (The one residual is exotic: a failure in the `sys.stdout.write` envelope line, which sits outside the child's `try`, would discard a full six-pass trail. Broken-pipe-only in practice.)

Two ways to close it, either is fine:
1. **Cheapest:** move `_relay_child_log(stderr, …)` above the `if proc.returncode == 0:` branch now, and invert that one test. Three lines, matches where #1950 is taking the sibling, and deletes the follow-up.
2. **Minimum:** leave the behaviour, but reword the comment to state the *actual* reason ("a hard failure here can only come from the first pass, so no preceding breadcrumb trail exists to lose") rather than "because `_invoke` does it", and note in the test that #1950 landing will require revisiting it.

Also worth noting for whoever picks up #1950: the child's `except Exception: print('hal0-migrate: ' + str(exc)); raise SystemExit(1)` discards the traceback, so `details["stderr"]` now holds one line where the previous revision held a full stack. `commit()` drops `exc.details` anyway (`str(exc)` on `Hal0Error` is the message only), so this changes nothing an operator sees today — but it removes the stack from the only sink that still had it.

---

## Verification run

- `tests/updater` + `tests/cli/test_update_commands.py` + `tests/api/test_updater_routes.py` → **412 passed**.
- `ruff check src tests` → all checks passed. `ruff format --check src tests` → 1161 files already formatted.
- Red-checked B1 (revert `updater.py` to `21734a10` → timeout test fails with the raw `TimeoutExpired`) and B3 (revert `routes/updater.py` → durability test fails with `[None, None]`).
- Mutation-checked B2 (no-op the `_relay_child_log` call → relay test red).
- Dumped the real child script + argv payload to confirm N1's bool-not-int and N2's default-`False` retag, rather than inferring from the diff.
- Confirmed nothing on the child's `hal0.updater.updater` import path calls `structlog.configure`, so the stderr pin can't be silently undone.
- No `CHANGELOG.md` hunk, no `installer/` hunk. `install.sh:2467` still untouched (`skip_image_retag` defaults to `False`, so its behaviour is unchanged).

---

**Verdict: `approve` — 0 blocking findings.** Good to land from my side. The one open item above is a comment/test-alignment follow-up that should ride with #1950, not a gate on rc.7.

<sub>Independent adversarial re-review of the delta; I did not author this change. Posting as a comment only — not a formal approval, not merging.</sub>

Issue #1791, opened 2026-08-09
fix(slots): crash-looping slot lifecycle — status stuck 'warming', start-limit leaves the slot permanently down unsurfaced, swap during the loop times out and reapplies the old model


thinmintdev:
Found during v1.0.0-rc.4 fresh-install validation (CT151). Companion to the crash cause in the FPX-on-CPU issue; this one is about how hal0 behaves while any slot crash-loops. Related: #1424 (load/restart error surfacing, reset-failed gap).

Three facets, all observed live:

**1. Status stuck `warming` while the unit is `failed`.** `hal0 status`, `/api/slots`, and `state.json` report `warming` indefinitely (`load_failures: 3`, `health_ok: false`) while `systemctl is-active hal0-slot@brain` says `failed`. Nothing ever transitions the slot to an error state or surfaces the crash loop. Also observed `IllegalSlotTransition: illegal transition warming → starting` warnings in the API journal during retries.

**2. Start-limit exhaustion is permanent and silent.** After 5 systemd restarts: `Start request repeated too quickly ... Failed to start hal0-slot@brain.service`. 25 minutes later the unit had been removed (`Unit hal0-slot@brain.service could not be found`), the slot showed `offline`, and no alert/status hint existed that a previously-loaded slot was lost. Manual `hal0 slot load brain` recovers it — hal0 never does.

**3. `slot swap`/`load` against a crash-looping slot: CLI `ReadTimeout`, and swap re-applied the OLD model.** `hal0 slot swap brain --model qwen3.5-0.8b` timed out client-side; the server proceeded (stop → remove-quadlet → write-quadlet → restart) but the rewritten quadlet still pointed at the old crashing model, so the loop continued. Control: the identical swap on a **healthy** slot works correctly in ~6 s, both directions, quadlet + served model verified — the bug is specific to the crash-loop window.

## Repro

Load any model that crashes the runner (e.g. FPX quant on CPU), then: watch `hal0 status` vs `systemctl`; wait out the start-limit; try `hal0 slot swap <slot> --model <good-model>` during the loop.

Found by automated release-validation agents across two independent sessions; swap behavior settled with a healthy-slot control before filing.

PR #1962, opened 2026-08-20
feat(slots): gate llm slot readiness on an output-sanity probe

### LINKED ISSUE #1922: slots: add an output-sanity gate to slot readiness ('The capital of France is' → 'Paris')

thinmintdev:
## What changed

`type=llm` slots must now prove they produce language before they are published as `ready`.

At the end of a load — after `/health` converges, before the READY transition — a container/llama-server llm slot gets one greedy completion whose answer is known:

```
POST /completion  {"prompt": "The capital of France is", "n_predict": 12, "temperature": 0}
```

A response containing `Paris` (case-insensitive, plus a few non-English spellings) passes. Anything else fails the load: the slot lands in `error` with a message naming the probe, the expected token and what the model actually said, and its unit is stopped.

New: `src/hal0/slots/output_sanity.py` (the probe, the classifier, the scoping and the escape hatches), `SlotOutputSanityFailed` (`slot.output_sanity_failed`), `OUTPUT_SANITY_TIMEOUT_S` in the lifecycle budget, and the gate call in `SlotManager._await_ready`.

## Why — and what the root cause actually was

Every readiness surface hal0 owns is a **transport** proof. `/health` returning 200 means the runner bound its port and mapped its weights; it says nothing about whether the backend can turn those weights into coherent logits. On ct151 during v1.0.0-rc.6 a box served two hours of garbage while slot state read `ready`, `/api/health` was green, `hal0 doctor` said "OK gpu", the SSE stream closed with a `done` frame and throughput measured 60-120 tok/s (#1888). Five validation lanes reported five different symptoms; all five were that one defect.

The root cause is narrower and more embarrassing than "no coherence check exists": the FLM/NPU lane **already had** a one-shot real-inference gate at exactly this point in `_await_ready` (`FLMProvider.verify_inference`). The generic llama-server lane — every GPU llm slot on every box — had none. This PR closes that asymmetry rather than inventing a new mechanism: same place, same one-shot discipline, same reasons it is safe there (slot lock held, slot not yet dispatchable, warming fail-watcher skips `/health`, so no contention and no repetition).

**A second defect surfaced while testing the fix, and is fixed here too.** Stamping ERROR was not enough. `status()`'s inverse-drift branch (and the boot-time upstream reconcile) adopt any ERROR/OFFLINE slot whose unit is *active* straight back to READY — and a slot that fails this gate is exactly that shape: healthy unit, healthy `/health`, garbage output. Without stopping the unit, the very next `/api/slots` poll re-published the garbage slot as dispatchable and the box was green again. Pinned by `test_failed_gate_survives_the_next_status_poll`, which fails on `main`+gate-only.

## Design decisions worth reviewing

- **Positive assertion only.** The garbage form varies with argv (runs of `)` on `--jinja` slots, empty content on a bare `/completion`), so a blocklist would have caught one lane and missed the others.
- **One retry through the chat endpoint before failing.** A wrong answer on raw `/completion` buys exactly one `/v1/chat/completions` attempt with the same prompt. A template-dependent instruct model can answer one well and the other poorly, and false-failing a working slot would make this gate worse than the bug it guards. The retry runs only on the failing path (the load is about to be refused anyway) and does not weaken the gate — the rc.6 garbage reproduced on both endpoints.
- **Scoped to `type="llm"`.** The slot type is derived from the model's declared modalities (`model_meta.modality.slot_type_for`), so it *is* the chat-capability signal. Embedding / rerank / TTS / STT / image slots are never asked to chat.
- **A timeout is a FAIL.** "Inconclusive, so let it through" is the exact silent degrade this issue was filed to remove. The one inconclusive-passes branch is a runner that 404s *both* completion endpoints — not a completion server this gate was designed to judge.
- **Failure raises rather than parking in WARMING.** A wedged loader may converge on a retry; a backend that computes wrong answers never will. Raising routes it through `load`'s ERROR branch, which also gives it crash-loop bookkeeping — loud once, not a retry loop burning the GPU.
- **Escape hatches, default ON:** `HAL0_SLOT_OUTPUT_SANITY=0` box-wide, `output_sanity = false` per slot TOML.
- **Budget:** one probe is bounded at `OUTPUT_SANITY_TIMEOUT_S = 20s`, worst case two; the client-side lifecycle budget charges both, so `slot_lifecycle_timeout_s` still clears the server's worst case.

## How it was verified

- `tests/slots/test_output_sanity_gate.py` — 38 cases. The probe runs against a **real loopback HTTP server** (not a mock transport) playing all four shapes the issue asks for: `Paris`, `)))))`, empty content, and a server that accepts the connection and never answers.
- Red-before-green, confirmed by reverting each hunk in turn: the load-path cases fail without the gate (`DID NOT RAISE SlotOutputSanityFailed`), and `test_failed_gate_survives_the_next_status_poll` fails without the unit teardown.
- `uv run pytest tests` — 10969 passed, 16 skipped, 1 xfailed.
- `uv run ruff check src tests` and `uv run ruff format --check src tests` — clean. `python3 scripts/check_sunset.py` — clean.

Test-suite plumbing: no unit test has a llama-server behind a slot port, so an autouse fixture in `tests/conftest.py` answers the probe with the correct answer by default (the same claim every container-provider double already makes about its unit), and `tests/slots/conftest.py` routes it at `container_stub.sanity_output` so gate behaviour can be driven through `load()`.

## Deliberately out of scope

- **Real-box proof.** The gate is only truly proven on a box without `/dev/kfd` (or forced onto the Vulkan lane), where the slot must land in `error` naming the probe. That belongs to the validation run, not CI.
- **The FLM/NPU sentinel still checks only that `choices` is non-empty**, not the content. Making the NPU lane assert the answer is a real improvement, but it changes the sentinel prompt on hardware this PR cannot exercise; it should be its own change.
- **Adoption is not gated.** `_maybe_adopt_running_slot` runs inside the `/api/slots` status poll, where a real completion per adopted slot would be a hot-path cost — the same reason the FLM gate stays off `health()`. A unit adopted at api start was launched by a load that ran the gate (post-upgrade); a pre-upgrade unit still running from before is the residual gap.
- No `CHANGELOG.md` hunk, per the fix-wave rule (#1545).

Closes #1922
Refs #1888

### DIFF
--- a/docs/reference/config-schema.mdx
+++ b/docs/reference/config-schema.mdx
@@ -245,6 +245,10 @@ Boot start is separate and explicit (`autoload`); a retired `lru` key is still a
 but ignored, with a one-time deprecation warning. See
 [Slot lifecycle → Activation](/docs/reference/slot-lifecycle/#activation).
 
+`output_sanity` (bool, default `true`) is accepted here too, though it is not a typed
+field: `false` exempts this one slot from the
+[output-sanity gate](/docs/reference/slot-lifecycle/#output-sanity-gate).
+
 <Aside type="note">
 A legacy `backend` key on a slot is silently promoted to `device` and dropped on load
 — it is never a schema field itself. On the v1.0 branch (Stream C,
--- a/docs/reference/env-vars.mdx
+++ b/docs/reference/env-vars.mdx
@@ -61,6 +61,12 @@ code paths (`agents/hermes_provision.py`) default to `http://hal0.local:8080`. I
 rely on the default rather than setting it explicitly, verify which code path applies.
 </Aside>
 
+## Slots
+
+| Variable | Default | Controls |
+|---|---|---|
+| `HAL0_SLOT_OUTPUT_SANITY` | unset (gate ON) | `0`/`false`/`no`/`off` disables the [output-sanity gate](/docs/reference/slot-lifecycle/#output-sanity-gate) box-wide, so a `type = "llm"` slot reaches `READY` on the health probe alone. Per-slot equivalent: `output_sanity = false` in the slot TOML. |
+
 ## Hermes agent
 
 | Variable | Default | Controls |
--- a/docs/reference/slot-lifecycle.mdx
+++ b/docs/reference/slot-lifecycle.mdx
@@ -103,12 +103,23 @@ and is not yet dispatchable. A wrong answer on the raw `/completion` endpoint bu
 retry through `/v1/chat/completions` (with room for a reasoning model's `<think>` block)
 before the load is refused.
 
-Failure is terminal for that load: the slot lands in `ERROR` with a message naming the
-probe, the expected token and what the model actually said, its unit is stopped, and it
-will **not** be re-adopted as `READY` by a later `/api/slots` poll or by the boot-time
-reconcile — a passing gate on a subsequent load is what clears the verdict. A probe
-timeout counts as a failure; "inconclusive, so let it through" is the silent degrade this
-gate exists to remove.
+A **wrong answer** is terminal for that load: the slot lands in `ERROR` with a message
+naming the probe, the expected token and what the model actually said, its unit is stopped
+(best-effort — a stop that does not return is reported in the same message), and it will
+**not** be re-adopted as `READY` by a later `/api/slots` poll or by the boot-time
+reconcile. A passing gate on a subsequent load is what clears the verdict.
+
+A **probe that never answers** is treated differently. A timeout says the answer did not
+arrive inside a wall clock, not that the answer was wrong, so the slot resolves to
+`WARMING` — the same policy the health wait already applies to its own ambiguous outcome.
+Nothing is published as `READY`, nothing is stamped, the unit is left alone, and the next
+load re-runs the gate. Slow hardware is judged on its answer; the budget widens to match on
+a `device = "cpu"` slot, which decodes far slower than the accelerated lane.
+
+<Aside type="note">
+"Inconclusive, so let it through" remains the one thing the gate never does: an
+unanswered probe can only produce `WARMING`, never `READY`.
+</Aside>
 
 Embedding, reranking, TTS, STT and image slots are never gated (they cannot serve a chat
 completion), and the FLM/NPU lane keeps its own `verify_inference` sentinel.
--- a/docs/reference/slot-lifecycle.mdx
+++ b/docs/reference/slot-lifecycle.mdx
@@ -18,7 +18,7 @@ the dashboard SSE stream, and the `state.json` persistence layer.
 | `PULLING` | Model files are being downloaded or verified. systemd unit not yet started. |
 | `STARTING` | systemd unit has been started; waiting for the container to come up. |
 | `WARMING` | Container is up; the health probe is returning non-ready responses while the model loads into VRAM/GTT. |
-| `READY` | Passed the full health probe (non-empty `/v1/models` plus a sentinel completion). Ready to serve. |
+| `READY` | Passed the full health probe (non-empty `/v1/models` plus a sentinel completion) **and**, for a `type = "llm"` slot, the output-sanity probe below. Ready to serve. |
 | `SERVING` | An inference request is actively in-flight on this slot. |
 | `IDLE` | Reached two ways, both render identically on the wire: `warming → idle` (process up but `/v1/models` is empty — launched with `--model ""` or a missing file; routers must not treat this as ready), or `ready → idle` (no request for longer than the idle timeout — eviction candidate). |
 | `UNLOADING` | Graceful shutdown in progress. systemd stop issued; waiting for the container to exit. |
@@ -80,6 +80,7 @@ Each is an HTTP-status-bearing error under the `slot.*` namespace:
 | `SlotNotReady` | 503 | — |
 | `SlotSpawnFailed` | 500 | — |
 | `SlotHealthFailed` | 503 | — |
+| `SlotOutputSanityFailed` | 500 | The server is healthy but its output is not language — see [Output-sanity gate](#output-sanity-gate). |
 | `SlotCrashLooping` | 503 | Reload blocked during exponential backoff, or after too many consecutive failures. Carries `retry_after_s`. |
 | `SlotTerminateTimeout` | 504 | Stop didn't return in time — the request unblocks anyway. |
 | `SlotConfigError` | 400 | — |
@@ -88,6 +89,36 @@ Each is an HTTP-status-bearing error under the `slot.*` namespace:
 
 Health-probe resolution on timeout: unresolved health checks resolve to `WARMING`, not
 `ERROR` — an ambiguous outcome is treated as "still loading," not "failed."
+
+## Output-sanity gate
+
+`/health` proves the runner bound its port and mapped its weights. It does not prove the
+backend turns those weights into coherent text — a GPU backend that mis-computes answers
+`/health`, streams tokens and closes the SSE stream cleanly while emitting garbage.
+
+So a `type = "llm"` slot on the container/llama-server lane runs one last check before it
+is published as `READY`: a single greedy completion whose answer is known
+(`The capital of France is` → `Paris`), once per load, while the slot still holds its lock
+and is not yet dispatchable. A wrong answer on the raw `/completion` endpoint buys one
+retry through `/v1/chat/completions` (with room for a reasoning model's `<think>` block)
+before the load is refused.
+
+Failure is terminal for that load: the slot lands in `ERROR` with a message naming the
+probe, the expected token and what the model actually said, its unit is stopped, and it
+will **not** be re-adopted as `READY` by a later `/api/slots` poll or by the boot-time
+reconcile — a passing gate on a subsequent load is what clears the verdict. A probe
+timeout counts as a failure; "inconclusive, so let it through" is the silent degrade this
+gate exists to remove.
+
+Embedding, reranking, TTS, STT and image slots are never gated (they cannot serve a chat
+completion), and the FLM/NPU lane keeps its own `verify_inference` sentinel.
+
+Escape hatches, both default-ON:
+
+| Scope | How |
+|---|---|
+| Whole box | `HAL0_SLOT_OUTPUT_SANITY=0` in the api environment |
+| One slot | `output_sanity = false` in the slot TOML, or via `PUT /api/slots/{name}/config` |
 
 ## Persistence
 
--- a/src/hal0/slot_config/__init__.py
+++ b/src/hal0/slot_config/__init__.py
@@ -261,10 +261,25 @@ def merge_slot_config(base: dict[str, Any], updates: dict[str, Any]) -> dict[str
 #:   - ``default_voice`` / ``default_language`` — TTS engine extras written by
 #:     the dashboard voice settings (PUT /api/slots/tts/config).
 #:   - ``slot``           — the nested on-disk [slot] table shape (hoisted on load).
+#:   - ``output_sanity``  — per-slot opt-out for the llm output-sanity gate
+#:     (#1922, ``hal0.slots.output_sanity.SANITY_CFG_KEY``). The gate's own
+#:     failure message tells the operator to set it, so the write boundary has
+#:     to accept it: an escape hatch reachable only by hand-editing the TOML is
+#:     not one.
 TOLERATED_SLOT_CONFIG_KEYS: frozenset[str] = frozenset(
     # image_pin and binary are declared SlotConfig fields but the running backend may
     # predate them — tolerate so the drawer Save always works.
-    {"type", "default", "lru", "default_voice", "default_language", "slot", "image_pin", "binary"}
+    {
+        "type",
+        "default",
+        "lru",
+        "default_voice",
+        "default_language",
+        "slot",
+        "image_pin",
+        "binary",
+        "output_sanity",
+    }
 )
 
 # ── model-owned slot-write partition (spec-hw-slot-ownership §1) ─────────────
--- a/src/hal0/slot_lifecycle_budget.py
+++ b/src/hal0/slot_lifecycle_budget.py
@@ -51,9 +51,39 @@ OUTPUT_SANITY_TIMEOUT_S = 20.0
 #: answer instead of on the clock.
 OUTPUT_SANITY_CHAT_TIMEOUT_S = 45.0
 
+#: Both gate budgets on a ``device="cpu"`` slot, which is a different machine
+#: from the accelerated lane the two constants above are sized for.
+#:
+#: The 20s/45s pair assumes a GPU that decodes a dozen greedy tokens in about
+#: a second. The fleet's own numbers say the CPU lane is two orders of
+#: magnitude away from that: the ``v1.0.0-rc.5`` validation measured ct151
+#: (8 cores, 16 GB, no GPU passthrough) at 0.12 tok/s for a 0.8B model under
+#: exactly the contention a load happens in, and one brain reply took 3m42s.
+#: A no-GPU box is not an exotic case either — ``install.profile_derive.
+#: derive_device`` resolves EVERY seeded slot to ``cpu`` there, and the brain
+#: slot then binds the unquantized F16 variant precisely because the box is
+#: slow.
+#:
+#: Set to :data:`HEALTH_TIMEOUT_S` rather than to a fresh literal: it is the
+#: wall clock the same box already gets to answer ``/health`` after spawning,
+#: so the gate asks for no more patience than the phase before it. That is
+#: 0.067 tok/s for the raw probe's dozen tokens — under the slowest thing the
+#: fleet has measured — and it also covers the chat fallback, whose 256-token
+#: ask is where a reasoning model on a CPU box would otherwise be judged on
+#: an answer it had no room to write.
+#:
+#: Consumed via ``hal0.slots.output_sanity.probe_budget_s``, which returns it
+#: for a cpu-backed slot and ``None`` (module defaults) for everything else.
+OUTPUT_SANITY_CPU_TIMEOUT_S = HEALTH_TIMEOUT_S
+
 #: What the gate can cost ONE load, worst case: the raw probe, then the chat
 #: fallback. A load that passes on the raw probe pays only the first.
-OUTPUT_SANITY_LOAD_ALLOWANCE_S = OUTPUT_SANITY_TIMEOUT_S + OUTPUT_SANITY_CHAT_TIMEOUT_S
+#:
+#: Charged at the CPU lane's budget because a client timeout must cover the
+#: slowest slot the server might be converging, and the client is handed a
+#: slot name — not its device — by every lifecycle verb. An accelerated slot
+#: spends 65s of this and returns early.
+OUTPUT_SANITY_LOAD_ALLOWANCE_S = 2 * OUTPUT_SANITY_CPU_TIMEOUT_S
 
 #: Sequential unloads a single ``load`` may perform before it spawns
 #: anything: ``preload_evict.admit`` awaits ``host.unload(candidate)`` once per
@@ -90,11 +120,14 @@ EVICTION_UNLOAD_ALLOWANCE = 3
 #:     90   3 x 30, ``preload_evict.admit`` unloading eviction candidates in
 #:          series (EVICTION_UNLOAD_ALLOWANCE)
 #:    180   the post-spawn ``/health`` poll (HEALTH_TIMEOUT_S)
-#:     65   the output-sanity gate (#1922): 20 raw probe + 45 chat fallback
+#:    360   the output-sanity gate (#1922), at its CPU-lane budget: 180 raw
+#:          probe + 180 chat fallback (OUTPUT_SANITY_LOAD_ALLOWANCE_S = 2 x
+#:          OUTPUT_SANITY_CPU_TIMEOUT_S). An accelerated slot spends 20 + 45
+#:          instead, but the client cannot know which it is asking about
 #:     30   the failed gate's teardown, which runs before the ERROR stamp and
 #:          therefore still inside the lock
 #:    ---
-#:    395
+#:    690
 LOAD_LOCK_HOLD_S = (
     TERMINATE_TIMEOUT_S
     + EVICTION_UNLOAD_ALLOWANCE * TERMINATE_TIMEOUT_S
--- a/src/hal0/slot_lifecycle_budget.py
+++ b/src/hal0/slot_lifecycle_budget.py
@@ -34,6 +34,21 @@ HEALTH_TIMEOUT_S = 180.0
 #: ``hal0.slots.manager.SlotManager._terminate_timeout_s``.
 TERMINATE_TIMEOUT_S = 30.0
 
+#: Wall-clock bound on ONE output-sanity completion — the probe a ``type=llm``
+#: load runs after ``/health`` converges (#1922 — the gate that turns "the
+#: port answers" into "the model produces language"). Consumed by
+#: ``hal0.slots.output_sanity.probe``; deliberately small, because the probe
+#: asks for a dozen greedy tokens from an already-warm server, so anything
+#: near this bound is itself the failure the gate reports.
+OUTPUT_SANITY_TIMEOUT_S = 20.0
+
+#: Probes a single load can pay for in the worst case. A wrong answer on the
+#: raw ``/completion`` endpoint buys one retry through
+#: ``/v1/chat/completions`` (a template-dependent instruct model can answer
+#: one well and the other poorly), so the failing path costs two budgets. The
+#: happy path costs one.
+OUTPUT_SANITY_PROBES_PER_LOAD = 2
+
 #: Sequential unloads a single ``load`` may perform before it spawns
 #: anything: ``preload_evict.admit`` awaits ``host.unload(candidate)`` once per
 #: evicted candidate, in series, inside the load path. The true count is
@@ -94,13 +109,19 @@ def slot_lifecycle_timeout_s(*, loads: int = 1, unloads: int = 1, slots: int = 1
     (``/api/updates/restart-slots`` loops ``sm.restart`` over every drifted
     slot). Clamped to at least 1.
 
-    A load charges the health poll plus :data:`EVICTION_UNLOAD_ALLOWANCE`
-    terminates for the evictions ``preload_evict.admit`` may run before the
-    spawn; an unload charges one terminate; the whole request additionally
-    charges one :data:`LOCK_WAIT_ALLOWANCE_S` for queueing behind an in-flight
-    op on the same slot.
+    A load charges the health poll, the output-sanity probes that follow it
+    (:data:`OUTPUT_SANITY_PROBES_PER_LOAD` x :data:`OUTPUT_SANITY_TIMEOUT_S`), plus
+    :data:`EVICTION_UNLOAD_ALLOWANCE` terminates for the evictions
+    ``preload_evict.admit`` may run before the spawn; an unload charges one
+    terminate; the whole request additionally charges one
+    :data:`LOCK_WAIT_ALLOWANCE_S` for queueing behind an in-flight op on the
+    same slot.
     """
-    per_slot = loads * (HEALTH_TIMEOUT_S + EVICTION_UNLOAD_ALLOWANCE * TERMINATE_TIMEOUT_S)
+    per_slot = loads * (
+        HEALTH_TIMEOUT_S
+        + OUTPUT_SANITY_PROBES_PER_LOAD * OUTPUT_SANITY_TIMEOUT_S
+        + EVICTION_UNLOAD_ALLOWANCE * TERMINATE_TIMEOUT_S
+    )
     per_slot += unloads * TERMINATE_TIMEOUT_S
     # The lock allowance is charged per lock-acquiring phase, per slot. A
     # compound verb does NOT hold one lock: ``SlotManager.restart`` releases
--- a/src/hal0/slot_lifecycle_budget.py
+++ b/src/hal0/slot_lifecycle_budget.py
@@ -34,20 +34,26 @@ HEALTH_TIMEOUT_S = 180.0
 #: ``hal0.slots.manager.SlotManager._terminate_timeout_s``.
 TERMINATE_TIMEOUT_S = 30.0
 
-#: Wall-clock bound on ONE output-sanity completion — the probe a ``type=llm``
-#: load runs after ``/health`` converges (#1922 — the gate that turns "the
-#: port answers" into "the model produces language"). Consumed by
-#: ``hal0.slots.output_sanity.probe``; deliberately small, because the probe
-#: asks for a dozen greedy tokens from an already-warm server, so anything
-#: near this bound is itself the failure the gate reports.
+#: Wall-clock bound on the RAW ``/completion`` output-sanity probe — the one a
+#: ``type=llm`` load runs after ``/health`` converges (#1922 — the gate that
+#: turns "the port answers" into "the model produces language"). Consumed by
+#: ``hal0.slots.output_sanity.probe``; deliberately small, because it asks for
+#: a dozen greedy tokens from an already-warm server, so anything near this
+#: bound is itself the failure the gate reports.
 OUTPUT_SANITY_TIMEOUT_S = 20.0
 
-#: Probes a single load can pay for in the worst case. A wrong answer on the
-#: raw ``/completion`` endpoint buys one retry through
-#: ``/v1/chat/completions`` (a template-dependent instruct model can answer
-#: one well and the other poorly), so the failing path costs two budgets. The
-#: happy path costs one.
-OUTPUT_SANITY_PROBES_PER_LOAD = 2
+#: Wall-clock bound on the CHAT fallback, which only runs when the raw probe's
+#: answer was wrong. Wider on purpose: that request asks for
+#: ``output_sanity.SANITY_CHAT_MAX_TOKENS`` (256) rather than a dozen, because
+#: a reasoning model spends its first tokens inside ``<think>`` and a tight cap
+#: fails working slots. 256 tokens inside 45s is ~6 tok/s — under the slowest
+#: warm lane we ship (CPU), so a slow-but-correct model is judged on its
+#: answer instead of on the clock.
+OUTPUT_SANITY_CHAT_TIMEOUT_S = 45.0
+
+#: What the gate can cost ONE load, worst case: the raw probe, then the chat
+#: fallback. A load that passes on the raw probe pays only the first.
+OUTPUT_SANITY_LOAD_ALLOWANCE_S = OUTPUT_SANITY_TIMEOUT_S + OUTPUT_SANITY_CHAT_TIMEOUT_S
 
 #: Sequential unloads a single ``load`` may perform before it spawns
 #: anything: ``preload_evict.admit`` awaits ``host.unload(candidate)`` once per
@@ -71,12 +77,37 @@ OUTPUT_SANITY_PROBES_PER_LOAD = 2
 #: estimate; bounding it server-side is the real fix (#1869).
 EVICTION_UNLOAD_ALLOWANCE = 3
 
+#: Worst wall-clock a single ``SlotManager.load`` can hold the slot lock —
+#: every phase it runs between acquiring and releasing, summed. Keep this
+#: derivation exhaustive: it is the number BOTH the per-load charge and the
+#: lock-wait allowance below are built from, so a phase added to ``load``
+#: without a line here silently re-creates #1832 (client reports a timeout on
+#: an op the server is still converging).
+#:
+#:     30   pre-spawn terminate — the drifted re-converge (#1224), the ERROR
+#:          retry and the stale in-flight branches each stop the unit in place
+#:          before re-spawning
+#:     90   3 x 30, ``preload_evict.admit`` unloading eviction candidates in
+#:          series (EVICTION_UNLOAD_ALLOWANCE)
+#:    180   the post-spawn ``/health`` poll (HEALTH_TIMEOUT_S)
+#:     65   the output-sanity gate (#1922): 20 raw probe + 45 chat fallback
+#:     30   the failed gate's teardown, which runs before the ERROR stamp and
+#:          therefore still inside the lock
+#:    ---
+#:    395
+LOAD_LOCK_HOLD_S = (
+    TERMINATE_TIMEOUT_S
+    + EVICTION_UNLOAD_ALLOWANCE * TERMINATE_TIMEOUT_S
+    + HEALTH_TIMEOUT_S
+    + OUTPUT_SANITY_LOAD_ALLOWANCE_S
+    + TERMINATE_TIMEOUT_S
+)
+
 #: Every lifecycle verb takes the per-slot lock (``SlotManager._lock``), so a
 #: request can sit queued behind whatever is already converging that slot
 #: before doing any of its own work. Charged once per request, at the cost of
-#: the worst thing that can hold the lock — a full load, evictions included,
-#: not just its health poll.
-LOCK_WAIT_ALLOWANCE_S = HEALTH_TIMEOUT_S + EVICTION_UNLOAD_ALLOWANCE * TERMINATE_TIMEOUT_S
+#: the worst thing that can hold the lock — a full load, start to finish.
+LOCK_WAIT_ALLOWANCE_S = LOAD_LOCK_HOLD_S
 
 #: Slots one ``POST /api/stacks/{slug}/apply`` may converge in a single request.
 #: ``StackApplyEngine.converge`` walks the stack's entries sequentially
@@ -109,19 +140,13 @@ def slot_lifecycle_timeout_s(*, loads: int = 1, unloads: int = 1, slots: int = 1
     (``/api/updates/restart-slots`` loops ``sm.restart`` over every drifted
     slot). Clamped to at least 1.
 
-    A load charges the health poll, the output-sanity probes that follow it
-    (:data:`OUTPUT_SANITY_PROBES_PER_LOAD` x :data:`OUTPUT_SANITY_TIMEOUT_S`), plus
-    :data:`EVICTION_UNLOAD_ALLOWANCE` terminates for the evictions
-    ``preload_evict.admit`` may run before the spawn; an unload charges one
+    A load charges :data:`LOAD_LOCK_HOLD_S` — every phase it runs while
+    holding the lock, itemised at that constant; an unload charges one
     terminate; the whole request additionally charges one
-    :data:`LOCK_WAIT_ALLOWANCE_S` for queueing behind an in-flight op on the
-    same slot.
+    :data:`LOCK_WAIT_ALLOWANCE_S` per lock-acquiring phase, for queueing behind
+    an in-flight op on the same slot.
     """
-    per_slot = loads * (
-        HEALTH_TIMEOUT_S
-        + OUTPUT_SANITY_PROBES_PER_LOAD * OUTPUT_SANITY_TIMEOUT_S
-        + EVICTION_UNLOAD_ALLOWANCE * TERMINATE_TIMEOUT_S
-    )
+    per_slot = loads * LOAD_LOCK_HOLD_S
     per_slot += unloads * TERMINATE_TIMEOUT_S
     # The lock allowance is charged per lock-acquiring phase, per slot. A
     # compound verb does NOT hold one lock: ``SlotManager.restart`` releases
--- a/src/hal0/slots/manager.py
+++ b/src/hal0/slots/manager.py
@@ -105,6 +105,7 @@ from hal0.slots.state import (
     SlotConfigError,
     SlotCrashLooping,
     SlotNotFound,
+    SlotOutputSanityFailed,
     SlotPinned,
     SlotSpawnFailed,
     SlotState,
@@ -1757,6 +1758,19 @@ class SlotManager:
                             },
                         )
             except Exception as exc:
+                if isinstance(exc, SlotOutputSanityFailed):
+                    # The gate failed on a unit that is ACTIVE and healthy — it
+                    # just talks nonsense. Stamping ERROR is not enough: the
+                    # inverse-drift branch in ``status()`` (and the boot-time
+                    # upstream reconcile) adopt any ERROR/OFFLINE slot whose
+                    # unit is active straight back to READY, so the very next
+                    # /api/slots poll would undo the gate and re-publish the
+                    # garbage slot as dispatchable. Stop the unit, which also
+                    # stops it burning VRAM producing text nobody can use. A
+                    # later ``load`` re-enters the lifecycle and re-runs the
+                    # gate; recovery is never silent.
+                    with contextlib.suppress(Exception):
+                        await self.terminate(slot_name)
                 # Crash-loop bookkeeping: consecutive failures drive the
                 # backoff/park gate above. ``parked`` rides ``extra`` (not a
                 # new enum value — wire/state-machine compatible) so the UI
@@ -3625,7 +3639,9 @@ class SlotManager:
     async def _await_ready(self, slot_name: str, port: int) -> SlotState:
         """Resolve the slot's final readiness state after spawning.
 
-        Polls GET /health on the slot's container port until 200.
+        Polls GET /health on the slot's container port until 200, then — for a
+        chat-capable (``type=llm``) slot — proves the server can actually
+        produce language before it is published as READY (#1922).
 
         Returns:
             SlotState.READY when the container is serving. On a health-wait
@@ -3633,6 +3649,15 @@ class SlotManager:
             than a lying READY: WARMING keeps the slot retryable so the fail
             watcher / next-request reload governs recovery, instead of live
             traffic being forwarded to a wedged server on a false READY.
+
+        Raises:
+            SlotOutputSanityFailed: the server is healthy but its output is
+                not language (garbage/empty/off-topic, or no completion inside
+                the probe budget). Unlike the health-wait timeout this is NOT
+                ambiguous — retrying cannot fix a backend that mis-computes —
+                so it raises into ``load``'s ERROR branch (red dot, actionable
+                message, crash-loop bookkeeping) instead of parking the slot
+                in a retryable WARMING.
         """
         cfg = await self._maybe_load_config(slot_name)
         if not cfg:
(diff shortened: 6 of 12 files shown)


thinmintdev:
## Review — changes requested

Verified against `aaaedda4`, in a clean worktree. The core of this is right and I want it to land: the gate is in the correct place (`_await_ready`, lock held, pre-dispatch, mirroring the FLM lane), the positive-assertion design is the right call, `provider is None` correctly scopes it to the llama-server lane that was actually the hole, and `heal_missing_llm_type` in `_load_slot_config` means `type` is populated by the time `skip_reason` reads it, so the gate really does fire on the ct151 repro shape (`brain.toml` is `type = "llm"`, `device = "gpu-rocm"`).

**Red-before-green confirmed by reverting each hunk, not by trusting the description:**
- deleting the `if provider is None:` block → 4 load-path tests fail (`DID NOT RAISE SlotOutputSanityFailed`, plus the READY/probe-count case)
- deleting only the failure-path `terminate` → `test_failed_gate_survives_the_next_status_poll` fails alone

Regression sweep: `tests/slots tests/cli/test_slot_commands.py tests/api tests/mcp` → 2672 passed, 1 skipped. `ruff format --check` clean. No `CHANGELOG.md` hunk. No scope creep — every touched file is load-bearing.

Three things need fixing before merge. Two of them I reproduced by execution, not by reading.

---

### 1. BLOCKING — the chat fallback false-fails a working thinking model, and then stops its unit

`output_sanity.py:190` reads only `message.content`, and `output_sanity.py:315` gives the fallback `max_tokens: 12`.

This repo already knows this exact trap. `src/hal0/agents/hermes_provision.py:5313-5320`:

> Thinking-mode models (Qwen3, etc.) burn most of their token budget on a `<think>...</think>` reasoning block before emitting any visible content. A 16-token cap drains entirely into reasoning and the `content` field comes back empty — which falsely flags the wiring as broken.

Its fix was `max_tokens: 256` **and** `haystack = content + reasoning_content`. The new gate does neither.

The precondition is not exotic — it is precisely the case the fallback exists to serve. Raw `/completion` returns an immediate EOS on a heavily instruct-tuned model (`empty_output`), the probe falls through to chat, and a reasoning model spends all 12 tokens in `reasoning_content`. Run against the PR's own `FakeSlotServer`:

```
VERDICT: SanityVerdict(ok=False, status='empty_output', sample='',
                       detail='chat endpoint agreed (empty_output)')
```

…on a slot that answered `"...the capital of France, which is Paris"` in `reasoning_content`.

The consequence is not a warning. `load()` raises, the unit is **stopped**, the slot is stamped ERROR, and crash-loop bookkeeping counts a failure. On a box whose brain slot runs a Qwen3-class model that is a healthy slot taken down by the guard. This PR's own standard is the right one — *"false-failing a working slot would make this gate worse than the bug it guards"* — and this path fails it.

Fix is cheap and can be all three: classify `reasoning_content` as well as `content`; give the fallback a real token budget (the raw probe's 12 is fine, the chat one is not); and/or send `"chat_template_kwargs": {"enable_thinking": false}`. Worth a test with a `reasoning_content`-only chat body.

*(This confirms Codex's P1 on `output_sanity.py:192`.)*

### 2. BLOCKING — a failed teardown silently undoes the gate

`manager.py:1772`, `with contextlib.suppress(Exception): await self.terminate(slot_name)`.

`terminate()` raises `SlotTerminateTimeout` when `systemctl stop` does not return inside 30s — documented in its own docstring as a real, observed behaviour (#1224), and podman container stops are exactly where that bites. When it does, the unit stays ACTIVE, the slot is stamped ERROR, and `status()`'s inverse-drift branch (`manager.py:2085`) adopts ERROR+active straight to READY. Reproduced with `unload_sync` raising:

```
ACTIVE AFTER FAILED TERMINATE: {'chat'}
STATE AFTER POLL: ready   "adopted running slot (container unit active)"
```

That is the identical defect `test_failed_gate_survives_the_next_status_poll` was written to pin, still reachable — the PR diagnosed the cause correctly (adoption has no memory of the verdict) and then fixed it at the symptom (stop the unit so adoption can't see it). The teardown is best-effort by nature, so the verdict must not depend on it.

The durable fix is small and strictly better than what is here: stamp the ERROR record's `extra` with the sanity verdict and have `_maybe_adopt_running_slot` refuse to adopt out of it. Keep the `terminate` — stopping a garbage unit is right on its own merits — but stop making it the load-bearing half. That also closes the "pre-upgrade unit still running from before" gap the PR lists as out of scope, without putting a completion on the `/api/slots` hot path.

*(Confirms Codex's P1 on `manager.py:1773`.)*

### 3. BLOCKING — `LOCK_WAIT_ALLOWANCE_S` was left behind, so the widened load under-budgets

`slot_lifecycle_budget.py:79` is unchanged, and its own docstring says it is *"the cost of the worst thing that can hold the lock — a full load, evictions included."* After this PR that sentence is false:

```
LOCK_WAIT_ALLOWANCE_S              = 270.0
worst load lock hold (gate ok)     = 310.0
worst load lock hold (gate fail)   = 340.0   # + the in-lock terminate from #2
slot_load client budget            = 667.0
server worst, one contender, both failing the gate = 680.0
```

680 > 667: the CLI/MCP client reports a timeout before the server reports the failure — before any process/HTTP overhead. This is the #1832 regression shape, re-introduced by the same function's own edit. `test_restart_budget_charges_both_lock_acquisitions` still passes only because its `load_phase` is hand-built from health+evictions and never learned about the probes.

One line: fold `OUTPUT_SANITY_PROBES_PER_LOAD * OUTPUT_SANITY_TIMEOUT_S` (and the failure-path terminate) into `LOCK_WAIT_ALLOWANCE_S`, and update the floor test so the invariant is actually pinned next time.

*(Confirms Codex's P2 on `slot_lifecycle_budget.py:123`; I re-derived the arithmetic independently and it holds.)*

---

### Non-blocking

**4. The advertised per-slot escape hatch is unreachable through the supported config surface.** `failure_message()` tells the operator to set `output_sanity = false` in the slot TOML, but `output_sanity` is neither a `SlotConfig` field nor in `TOLERATED_SLOT_CONFIG_KEYS`, so both `POST /api/slots` and `PUT /api/slots/{name}/config` refuse it:

```
{'output_sanity': False}           -> BadRequest: unknown slot config key(s): output_sanity
{'slot': {'output_sanity': False}} -> BadRequest: unknown slot config key(s): slot.output_sanity
```

Only a hand edit works. Add it to `TOLERATED_SLOT_CONFIG_KEYS` (or declare it on `SlotConfig`).

Note for the record: the *other* half of Codex's P2 here — that `load_slot_config`/`save_slot_config` round-trips discard the undeclared scalar — is **wrong**, and I checked rather than assumed. `write_slot_toml` preserves unknown top-level keys verbatim, so `_persist_model_default` and friends do not eat the opt-out. Dismissing that half.

**5. Docs.** This adds a new mandatory readiness stage and a new operator-facing env var, and touches neither `docs/reference/slot-lifecycle.mdx` (whose `READY` row still reads "Passed the full health probe") nor `docs/reference/env-vars.mdx` (67 `HAL0_*` vars documented, this one not). Given #4, the failure message is currently the only place the escape hatch exists.

---

### On the deliberate out-of-scope list

Agreed on all of it, with one flag: *"a pre-upgrade unit still running from before is the residual gap"* is the ct151 upgrade path, and fixing #2 at the adoption record rather than at the teardown closes it for free. Worth taking while you are in there.

The gate itself is good work and the reasoning in the module docstring is the kind that survives contact with the next incident. Fix 1–3 and I will re-review promptly.

thinmintdev:
Revised in `a8ba81d`. All three blocking findings fixed red-first; both non-blocking ones taken too.

**1 — chat fallback false-failing a thinking model.** The probe now judges the whole message (`content` + `reasoning_content`/`reasoning`, joined) and the chat turn gets a real budget: `SANITY_CHAT_MAX_TOKENS = 256` inside its own `OUTPUT_SANITY_CHAT_TIMEOUT_S = 45s`, matching `hermes_provision._smoke_chat_completion`. A `<think>` block left inline in `content` by a non-splitting runner reads the same way, so the verdict does not depend on the llama.cpp build.

I did **not** send `chat_template_kwargs.enable_thinking=false`, even though the dispatcher does on production traffic: it is a jinja-only lever and `jinja` is a tri-state model capability (`ModelDefaults.jinja`), so slots launched without `--jinja` would trade this false-fail for a 4xx one on a lane CI cannot exercise. Reading both fields and paying for the tokens works on every build.

Red→green: `test_probe_accepts_a_thinking_model_answering_in_reasoning_content` (was `SanityVerdict(ok=False, status='empty_output')`), `test_chat_fallback_asks_for_more_tokens_than_the_raw_probe`, `test_chat_fallback_gets_its_own_wider_wall_clock`. Property kept: `test_probe_fails_a_thinking_model_that_reasons_in_garbage` — garbage in `reasoning_content` still fails.

**2 — teardown no longer load-bearing.** Took your preferred fix. The verdict is stamped on the state record (`output_sanity_failed` / `output_sanity_status` / `output_sanity_unit_stopped`) and `_maybe_adopt_running_slot` refuses any slot carrying it — one choke point, so it covers both `status()`'s inverse-drift branch and the boot-time upstream reconcile. It reads through to `state.json`, so it survives an api restart as well: that closes the pre-upgrade-unit gap for free, as you suggested. A load that passes the gate clears the mark (`_transition`, alongside `parked`/`load_failures`), so recovery needs no manual step.

The `terminate` stays, but a failed stop is now loud rather than suppressed: error log, explicit upstream deregistration (`terminate` only deregisters on success), and the ERROR message says the container is still resident.

Red→green with `unload_sync` raising `SlotTerminateTimeout`: `test_failed_gate_survives_a_teardown_that_does_not_return` (reproduced your `slot.adopted` line pre-fix), `test_failed_gate_is_not_adopted_by_a_later_api_process`, `test_failed_teardown_is_recorded_not_swallowed`, plus `test_a_passing_gate_clears_the_refusal` guarding the other direction.

**3 — budget recomputed.** The in-lock phases are summed once in `LOAD_LOCK_HOLD_S`, itemised with the arithmetic so the next addition has an obvious place to land:

```
 30  pre-spawn terminate (converge / ERROR retry / stale in-flight)
 90  3 x 30 preload evictions, in series
180  /health poll
 65  output-sanity gate: 20 raw + 45 chat fallback
 30  failed gate's teardown, before the ERROR stamp, still in-lock
---
395
```

`LOCK_WAIT_ALLOWANCE_S = LOAD_LOCK_HOLD_S`, and the per-load charge derives from the same constant. `slot_load` is now 908.5s against a 790s contended worst case (was 667 vs 730). `test_load_budget_covers_a_contended_load_that_fails_the_gate` pins exactly your shape and fails on the pre-fix numbers; `test_restart_budget_charges_both_lock_acquisitions` learned about the probes. Both recompute the components from `providers.container` / `SlotManager` rather than from the budget module.

**4 — escape hatch reachable.** `output_sanity` added to `TOLERATED_SLOT_CONFIG_KEYS`, pinned by `test_the_per_slot_opt_out_is_writable_through_the_api` (flat and nested `[slot]`). Agreed on dismissing the round-trip half — `write_slot_toml` does preserve it.

**5 — docs.** New "Output-sanity gate" section in `slot-lifecycle.mdx` (plus the `READY` row and a `SlotOutputSanityFailed` error row), `HAL0_SLOT_OUTPUT_SANITY` in `env-vars.mdx`, and the tolerated `output_sanity` key noted in `config-schema.mdx`.

Verification: `tests/slots tests/api tests/cli tests/mcp` → 3796 passed, 1 skipped. `tests/config tests/agents/test_hermes_provision.py tests/capabilities tests/docs` → 1024 passed, 9 skipped. `ruff check src tests` and `ruff format --check src tests` clean, `scripts/check_sunset.py` clean. No `CHANGELOG.md` hunk.

thinmintdev:
## Delta re-review of `a8ba81d` — 1 blocking left

Verified in a clean worktree at `a8ba81d`, against the three original blockers. **Two of the three are fully closed and I could not break them.** The third — the reasoning-model false-fail — was fixed on the axis it was reported on (the chat fallback) but the *tighter* budget was left untouched, and the fleet's own validation artifacts say that is the one that bites. Details below; everything else is clean.

CI on `a8ba81d`: all 8 checks green (`python (3.12)` 26m56s, `γ-suite (chromium)` 10m57s, `ui`, `sunset`, `CodeQL` + 3 `Analyze`). Locally: `tests/slots/test_output_sanity_gate.py` → **48 passed** in 0.87s.

---

### ✅ 2 — durability. Closed, and the assertions are not vacuous.

Adversarial sweep of every path that can publish `READY`:

| Path | Verdict |
|---|---|
| `status()` inverse-drift (`manager.py:2134`) | goes through `_maybe_adopt_running_slot` → refused ✓ |
| `status()` no-record branch (`manager.py:2069`) | same choke point ✓ |
| `reconcile_container_upstreams` boot reconcile (`manager.py:564`) | same choke point ✓ |
| explicit `load` | re-runs the gate ✓ |
| `_serving_exit` `SERVING → READY` (`manager.py:3277`) | unreachable — a gated slot is `ERROR` and never dispatchable ✓ |
| `_await_ready` returns `READY` | post-gate ✓ |

One choke point, all four adoption call sites route through it. Confirmed by mutation, not by reading: neutering the refusal to `if False and prior is not None and gate_failed(prior.extra):` turns exactly the two durability specs red —

```
FAILED test_failed_gate_survives_a_teardown_that_does_not_return
FAILED test_failed_gate_is_not_adopted_by_a_later_api_process
2 failed, 46 passed
```

…and the `slot.adopted` line reappears in the log, which is the original defect verbatim.

**Leak across delete/recreate: no.** `delete()` unlinks all four artefacts including `_state_file_for(slot_name)` (`manager.py:2718-2724`), and the read-through in `_maybe_adopt_running_slot` is `self._states.get(...) or read_state(...)` — with both gone the mark cannot resurrect.

**Half-state on the failed-teardown path: no.** `terminate()` does exactly two things — the bounded executor stop, then `self._deregister_container_upstream(slot_name)` (`manager.py:2474`). The failure branch calls that same method with the same argument, and the stop is the only work that did not happen. Identical deregistration, nothing left dangling.

### ✅ 3 — budget arithmetic. Each component charges a real code path.

* **30 pre-spawn terminate** — the three branches at `manager.py:1584 / 1603 / 1626` are `if / elif / elif` on `current`, so **at most one** runs per load and the 30s ceiling is `_terminate_timeout_s = TERMINATE_TIMEOUT_S` (`manager.py:305`), not an assumption. Correct, and correctly charged once.
* **90 / 180 / 65 / 30** — all trace to `EVICTION_UNLOAD_ALLOWANCE × TERMINATE_TIMEOUT_S`, `HEALTH_TIMEOUT_S`, `OUTPUT_SANITY_LOAD_ALLOWANCE_S` and the in-lock failure teardown respectively. Sum 395 ✓.
* Re-derived the client budgets independently: `slot_load` = (395 + 1×395) × 1.15 = **908.5** ✓; `restart` = (395 + 30 + 2×395) × 1.15 = **1397.2** ✓; `stack_apply` = 1215 × 6 × 1.15 = **8383.5** ✓. Contended worst case 790 < 908.5, so #1832's shape is gone with margin.
* **The 2.3h `stack_apply` ceiling is not a synchronous wait anywhere.** Every consumer is a *client read timeout* (`mcp/admin.py:1584`, `cli/*`), and `_CONNECT_TIMEOUT_S = 5.0` is kept separate so a half-open daemon still fails fast. `max_slot_lifecycle_timeout_s()` has no in-tree consumer at all. Nothing blocks for 2.3h; the call returns when the server does.

### ✅ 4 — the Codex P2 dismissal holds (though not for the stated reason).

I checked `load_slot_config` / `save_slot_config` by executing them rather than reading them, and the round-trip **is** lossy — `output_sanity` *and* `type` are both dropped:

```
extra: {}          # lands as a pydantic extra attr, not in cfg.extra
--- round-tripped file ---
[slot]
name = "chat"; port = 8081; device = "gpu-rocm"; provider = "llama-server"; ...
# no `type`, no `output_sanity`
```

But `save_slot_config` has **zero production callers** (`grep -rn save_slot_config src/` → only the export list and two docstrings; every real writer is `write_slot_toml` on a raw dict read from disk — `manager.py:1089/1096/2613/2645/2843/3021`, `profile_adopt.py`, `npu/trio.py`). So the opt-out survives every reachable write path and the dismissal is right on the outcome. Worth a one-line note somewhere that `save_slot_config` is lossy for `TOLERATED_SLOT_CONFIG_KEYS`, since it is a live footgun for the next caller who reaches for it.

### ✅ 5 — docs accurate. ✅ 6 — scope clean.

`slot-lifecycle.mdx` / `env-vars.mdx` / `config-schema.mdx` all check out against the code (the `READY` row, the `SlotOutputSanityFailed` 500 row, the two escape hatches, the FLM carve-out). One nit: "its unit is stopped" now reads as unconditional, but after this revision the teardown is explicitly best-effort — worth "its unit is stopped (best-effort)".

No `CHANGELOG.md` hunk. 12 files, all load-bearing. The only occurrence of "Vulkan" in the true merge-base diff (`39a3901e`) is a comment describing what the gate's sample is *for* — no lane re-enable smuggled in. Branch is 7 commits behind `main` but `mergeStateStatus: CLEAN`.

---

## 🔴 BLOCKING — the false-fail moved to the raw probe, and the CPU lane is where the validation run happens

Fix 1 widened the **chat fallback** to 256 tokens / 45s. But the fallback only ever runs when the raw probe already failed. The gate's *unconditional, every-load, every-box* cost is the raw probe: **12 tokens inside `OUTPUT_SANITY_TIMEOUT_S = 20.0`** — and `probe_timeout` is deliberately **not** in `_SECOND_OPINION_STATUSES`, so it is a terminal `SlotOutputSanityFailed` with no retry, no fallback, plus the new durable `output_sanity_failed` mark. That is the one path where the revision made the consequence *worse* while leaving the budget untouched.

**Three pieces of evidence, all from this repo:**

**(a) `/health` proves nothing about decode rate.** `providers/container.py:2169` is explicit: *"A real-inference sentinel must NOT run"* — the container health probe is `/health` + `/v1/models` only. So the sanity probe is the **first inference that box has ever done**, and nothing upstream has established it can emit 12 tokens in 20s.

**(b) This repo's existing real-inference gate is 12× cheaper and 50% more generous.** `FLMProvider.verify_inference` (`flm.py:611`) asks for `"max_tokens": 1` inside `_HEALTH_INFER_TIMEOUT` read=**30s**; `flm.py:245` keeps a 300s `_INFER_TIMEOUT` for real inference. The new gate asks **12 tokens inside 20s**. And where FLM's gate fails, it returns `ok=False` so the caller *"can hold the slot in the retryable WARMING state rather than a lying READY"* (`flm.py:640-642`) — this one escalates to ERROR + stop + un-adoptable.

**(c) The fleet's own measured CPU numbers do not clear 20s.** The budget module asserts *"256 tokens inside 45s is ~6 tok/s — under the slowest warm lane we ship (CPU)"*. The rc.5 validation report measures the CPU lane at **0.12 tok/s** (`artifacts/hal0-release-validation/1.0.0-rc.5/report.md:57,172`) — a 0.8B model, 8 cores, on ct151 (8 cores / 16 GB, no GPU passthrough), under exactly the contention a load happens in (*"two llama-servers at 653% and 375% CPU on 8 cores"*, line 178). The same box's `brain` slot took **3m42s for one token** (`.../1.0.0-rc.5-pre/lanes/ct151-cpu-fresh-brain.md:56-65`). The 6 tok/s floor is asserted, not measured, and the measurements that do exist are two orders of magnitude under it.

**And `brain` is gated.** `installer/etc-hal0/slots/brain.toml` is `type = "llm"`, `runtime = "container"`, no `provider` key → `provider is None` → the gate runs. On a no-ROCm/Vulkan box install.sh binds `hal0-brain-sft-f16` — the **unquantized F16** variant, chosen precisely *because* the box is slow. 12 tokens needs ≥0.6 tok/s sustained plus prompt-eval plus HTTP; at anything under that the platform steward's slot lands permanently red, VRAM-free but un-adoptable, recoverable only by the documented bypass. That is the module's own rule — *"a wrong FAIL is worse than the bug it guards"* — failing on the one lane the RC gets validated on.

**Suggested fix, and I think it makes the gate strictly better rather than weaker:**

> **Treat `probe_timeout` as ambiguous, not as garbage.** Return `SlotState.WARMING` for it (no raise, no `SANITY_FAILED_KEY` stamp), exactly as `_await_ready`'s own docstring already prescribes for the health-wait — *"an ambiguous outcome is treated as 'still loading,' not 'failed'"* — and exactly as FLM's gate already does. Keep `empty_output` / `incoherent_output` / `no_content` terminal: those are unambiguous.
>
> This costs the #1888 regression net **nothing**, and that is the load-bearing point: rc.6's garbage box streamed at **60–120 tok/s** and closed with a `done` frame. Garbage was *fast*. A timeout is not the #1888 signature and never was — so exempting it removes the entire remaining false-fail surface without touching a single true positive. A slot parked in WARMING is still not dispatchable, still retryable, and still not a lying READY.
>
> Complementary if you want the slow-but-correct box actually *judged*: derive the raw budget from `cfg["device"]` (a `cpu` slot gets the FLM-style 300s-class budget) and fold the larger number into `LOAD_LOCK_HOLD_S` — the constant is now itemised, so there is an obvious line for it.

A test pinning the CPU shape — probe times out, slot lands WARMING/retryable, no `output_sanity_failed` mark, and garbage-at-speed still lands ERROR — would keep the distinction from eroding.

---

Everything else is genuinely done, and the durability fix in particular is better than what I asked for — the state-record stamp plus read-through closes the pre-upgrade-unit gap exactly as hoped. Fix the timeout classification and this lands.

thinmintdev:
## Revision 3 — the blocking finding, closed (`cb9753ff`)

Fixed on the axis you named. `probe_timeout` no longer condemns a slot; the raw probe's budget is now derived from the slot's lane. Everything you signed off on in the delta re-review is untouched.

### 1. `probe_timeout` → `SlotState.WARMING`

`_await_ready` now branches before the raise: a failure whose status is in the new `output_sanity.AMBIGUOUS_STATUSES` returns `WARMING` — **no raise, no `output_sanity_failed` stamp, no unit stop, no crash-loop increment**. `no_content` / `empty_output` / `incoherent_output` are bit-for-bit what they were: terminal, unit stopped, durable un-adoptable mark.

The line the code now draws, and the one the docstrings state: **a completed round-trip whose *answer* fell short is a verdict about the model; a round-trip that did not finish is a fact about the clock.**

Your load-bearing point is in the constant's docstring so it survives the next reader:

```python
#: DO NOT "fix" ``probe_timeout`` back into a terminal failure. The reflex is
#: to read a timeout as "the backend is wedged", but #1888 — the defect this
#: whole module exists for — was never shaped like one: rc.6's garbage box
#: streamed at 60-120 tok/s and closed with a ``done`` frame. Garbage was
#: FAST. A timeout is the opposite signature...
```

**Deliberate non-change:** the other three round-trip failures (`probe_transport_error`, `probe_http_5xx`, `probe_unparseable`) stay terminal. Each is an active, immediate misbehaviour by the runner rather than an expiring clock, and none is the shape a slow-but-correct box produces — so widening past your evidence would trade a measured false-fail for an unmeasured weakening. Flagging it as a judgement call rather than an oversight; happy to widen if you disagree.

### 2. Device-derived budget — taken, and it turned out not to be optional

I took the complement, because without it the fix is only half a fix: a CPU box at 0.12 tok/s never clears 20s **on any retry**, so it would have traded a permanent `ERROR` for a permanent `WARMING`. Still never a lying `READY`, but never a working slot either.

```python
OUTPUT_SANITY_CPU_TIMEOUT_S = HEALTH_TIMEOUT_S   # 180.0
```

Set to `HEALTH_TIMEOUT_S` rather than a fresh literal, and the rationale is statable: **the gate asks for no more patience than the phase before it** — it is the same wall clock the same box already gets to answer `/health` after spawning. For the raw probe's 12 tokens that is a 0.067 tok/s floor, comfortably under the 0.12 tok/s you cited. It covers the chat fallback too, which is where a reasoning model on a CPU box would otherwise be judged on an answer it had no room to write.

Derived via `_cfg_effective_backend(cfg)`, not by reading `device` directly, so a legacy `backend`-only TOML resolves identically and the token cannot diverge from what the load path derives. Applied through `probe()`'s existing `timeout_s` knob, which already overrides both budgets — no new plumbing. And this is the *default* shape on a no-GPU box, not an exotic one: `install.profile_derive.derive_device` resolves **every** seeded slot there to `cpu` (`profile_derive.py:123` — "AMD without a ROCm compute node is NOT a Vulkan lane... CPU is the only honest answer").

Folded into the itemised arithmetic as you asked:

```
    30   pre-spawn terminate
    90   3 x 30, preload_evict.admit
   180   the post-spawn /health poll (HEALTH_TIMEOUT_S)
   360   the output-sanity gate, at its CPU-lane budget: 180 raw + 180 chat
         (an accelerated slot spends 20 + 45, but the client cannot know
          which it is asking about — it is handed a slot NAME)
    30   the failed gate's teardown
   ---
   690
```

**Client-budget deltas** (all derived, no literals to chase — `_SLOT_LIFECYCLE_TIMEOUT_S` is built from `slot_lifecycle_timeout_s()` for every entry):

| | before | after |
|---|---|---|
| `LOAD_LOCK_HOLD_S` | 395 | **690** |
| `slot_load` | 908.5 | **1587.0** |
| `slot_unload` | 483.0 | **828.0** |
| `slot_restart` / `model_swap` / `capability_set` | 1397.2 | **2415.0** |
| `stack_apply` (6 slots) | 8383.5 | **14490.0** (4.03h) |

Worth naming: `stack_apply` goes 2.3h → 4.0h. Your own audit is what makes that acceptable — every consumer is a *client read timeout*, `_CONNECT_TIMEOUT_S = 5.0` is kept separate so a half-open daemon still fails fast, and `max_slot_lifecycle_timeout_s()` has no in-tree consumer. Nothing blocks for 4h; the call returns when the server does. `_server_load_lock_hold_s()` in the parity test is updated to charge the CPU budget too, so it stays a real ratchet rather than a vacuously loose floor.

### 3. Convergence story for `WARMING`

The question is fair and I checked it by execution, not by reading. Three independent routes out, and the slot is non-dispatchable on all of them (`WARMING ∉ _DISPATCHABLE_STATES` → `SlotLoading`, 503 + `Retry-After`):

| Route | Mechanism | Outcome |
|---|---|---|
| **Next request** | `_check_slot_ready_for_dispatch` 503s → dispatcher re-drives `load()` → the stale-in-flight branch (`manager.py:1604`, `PULLING/STARTING/WARMING`) terminates the unit and re-enters from `OFFLINE` | **the gate re-runs, in full** |
| **Nobody asks** | fail-watcher bounds the WARMING dwell at `_WARMING_STALE_AFTER_S = 900s` (`watchdog.py:264`) → `unload` → `load` | **the gate re-runs**; cannot wedge forever |
| **Genuinely dead** | `load` already asks systemd whenever the resolve returns WARMING (`manager.py:1738`, #1791); a unit in `failed` / `start-limit-hit` / gone raises `SlotSpawnFailed` | **ERROR inside the same load**, with the failure counted |

So: **the gate is never skipped on a retry, the slot cannot rest in WARMING, and a dead backend converges to a terminal state within the existing load budget** — no new timer, no new state, no new budget line. Route 3 is pinned by a new test (`test_a_dead_backend_still_converges_to_error_in_the_same_load`), route 1 by `test_a_slot_parked_warming_re_runs_the_gate_on_the_next_load`.

The one honest limit: a unit that stays *active* while its backend never answers loops route 2 indefinitely rather than reaching a terminal state. That is the pre-existing policy for the health-wait timeout, which lands in the same `WARMING` for the same reason — this change joins that policy rather than inventing a second one. Converting it to terminal would re-create the false-fail on the slow box, which is the whole point.

### 4. Red-first evidence

The new WARMING spec drives a **real never-answering socket through a real httpx round-trip** (only the wall clock is shrunk, to 0.3s) — it is the production path, not a stubbed verdict.

Against `a8ba81d`:

```
FAILED test_a_slow_server_parks_the_slot_warming_instead_of_condemning_it
FAILED test_a_probe_timeout_leaves_the_slot_adoptable
FAILED test_a_slot_parked_warming_re_runs_the_gate_on_the_next_load
FAILED test_a_cpu_slot_gets_a_wider_probe_budget
FAILED test_the_load_path_hands_the_probe_the_slot_s_budget
5 failed, 48 passed
```

The 48 that passed include every durability spec and the garbage-in-`reasoning_content` terminal case — the red was exactly the new behaviour, nothing else.

After `cb9753ff`: **54 passed** in `tests/slots/test_output_sanity_gate.py`. The suite gains, alongside the five above, `test_a_dead_backend_still_converges_to_error_in_the_same_load` and `test_garbage_stays_terminal_even_though_a_timeout_does_not` — the last one pins your exact pair (timeout survivable, garbage-at-speed terminal *with* the durable mark) against one real server, so the distinction cannot rot into "any probe failure is survivable".

`tests/slots` + `tests/cli/test_slot_commands.py` + `tests/mcp/test_admin.py`: **972 passed**. Full suite: **10986 passed, 16 skipped, 1 xfailed** (17m36s). `ruff check` and `ruff format --check`: clean over `src/` and `tests/`.

### 5. Docs + your two nits

* `slot-lifecycle.mdx` gains the timeout/`WARMING` policy as its own paragraph, split from the wrong-answer paragraph.
* **"its unit is stopped"** → "its unit is stopped (best-effort — a stop that does not return is reported in the same message)", per your nit.
* The old line *"A probe timeout counts as a failure; 'inconclusive, so let it through' is the silent degrade this gate exists to remove"* would have become false, so it is replaced by an `Aside` that keeps the true half: an unanswered probe can only produce `WARMING`, **never** `READY`.

Not carried in this revision: your note that `save_slot_config` is lossy for `TOLERATED_SLOT_CONFIG_KEYS`. Real footgun, agreed, but it is a pre-existing defect on a function with zero production callers and it is out of this PR's scope — say the word and I will file it rather than smuggle it in.

No `CHANGELOG.md` hunk. Not merging — over to you.

thinmintdev:
## Confirmation pass on `cb9753ff` — 2 blocking

The prescription landed intact and the reasoning around it is better than what I asked for. But CI is red, and there is one composition path where the fix does not reach — reproduced on the production code path, not inferred.

---

### 🔴 BLOCKING 1 — `sunset` is failing. One word.

```
❌ SCAR RATCHET: 193 markers > baseline 192 (+1). Delete shims; a justified new shim
   must carry a HAL0-SUNSET marker, not raise the count.
```

`scripts/check_sunset.py:35` matches `\blegacy\b`, and this revision adds exactly one line containing it — `probe_budget_s`'s docstring:

```
+    ``device`` directly so a legacy TOML carrying only ``backend`` resolves
```

That is the whole diff's only new scar marker (I ran the regex over `git diff a8ba81d cb9753ff -U0`; one `+` hit, no `-` hits to offset it). Rewording to "an older TOML" restores it, verified locally:

```
✅ scar markers: 192 <= baseline 192
✅ no overdue HAL0-SUNSET shims (current v1.0.0-rc.6)
```

No baseline bump needed, nothing to justify — the sentence means the same thing. (Mildly funny that the ratchet fired on a docstring describing legacy-TOML handling, but the check is doing its job: `_cfg_effective_backend` is where that concept belongs and it already says it there.)

**Current CI on `cb9753ff`:** `sunset` **fail**; `ui`, `CodeQL`, `Analyze (actions / javascript-typescript / python)` pass; `python (3.12)` and `γ-suite (chromium)` still pending at time of writing. `mergeStateStatus: BLOCKED`.

### 🔴 BLOCKING 2 — the chat fallback's own timeout is masked, and the CPU lane is where it fires

The fix covers a timeout on the **raw** probe. It does not cover a timeout on the **chat fallback**, because `probe()`'s terminal composition reports the *raw* verdict's status:

```python
# Both endpoints answered badly. Report the RAW verdict ...
return SanityVerdict(
    ok=False,
    status=verdict.status if verdict.status != "probe_absent" else chat.status,
    ...
    detail=f"chat endpoint agreed ({chat.status})",
)
```

So `raw = empty_output` + `chat = probe_timeout` composes to **`empty_output`** — not in `AMBIGUOUS_STATUSES`, therefore terminal: ERROR, unit stopped, durable un-adoptable mark. Reproduced against a real socket through the real `probe()` (server answers `/completion` 200 with empty content, never answers `/v1/chat/completions`):

```
VERDICT: SanityVerdict(ok=False, status='empty_output', sample='',
                       detail='chat endpoint agreed (probe_timeout)')
is_ambiguous(status): False
=> TERMINAL (ERROR + stop + durable mark)
```

**And on the CPU lane this is not an edge case — it is the only outcome that path can have.** The chat fallback asks `SANITY_CHAT_MAX_TOKENS = 256` inside `OUTPUT_SANITY_CPU_TIMEOUT_S = 180`, a **1.42 tok/s** floor. The measurement this whole revision is built on is **0.12 tok/s** on ct151 under load-time contention — 256 tokens there is ~2133s. The widened budget rescues the raw probe (12 tokens → 0.067 tok/s floor, comfortable) and cannot rescue the fallback, because the ask is 21× the tokens against the same clock.

The consequence is sharper than "one uncovered branch". The chat fallback exists *specifically* to protect the template-dependent and thinking models that fail the raw probe — that was revision 2's entire rationale for widening it to 256 tokens. On a CPU box that protection now inverts: reaching the fallback guarantees a timeout, and the timeout is relabelled as the raw verdict's content status and condemned. The slot the fallback was built to save is the slot it kills.

Two smaller things fall out of the same spot:

* The comment `# Both endpoints answered badly` is factually wrong on this branch — only one endpoint answered. And the module's own rule says the other case is not enough: *"failing a working slot on the strength of a single endpoint is the one mistake this gate must not make."*
* `test_garbage_stays_terminal_even_though_a_timeout_does_not` pins the pair I asked for, but both halves complete their round-trip — nothing pins the mixed case, which is why this survived.

**Fix — one condition, consistent with the line the revision already draws:** in the terminal composition, if `is_ambiguous(chat.status)` then the composed verdict is ambiguous too. The second opinion never arrived, so there is no second verdict to condemn on. Worth pairing with a smaller `SANITY_CHAT_MAX_TOKENS` on the cpu lane (256 tokens is unreachable there at any clock this budget can justify) so the fallback can actually *do its job* rather than merely fail safely. A test with `raw=empty_output` + `chat=probe_timeout` asserting WARMING and no stamp keeps it pinned.

---

## Everything else — verified, and it holds

**1. The core fix landed intact.** `_await_ready` branches on `is_ambiguous` *before* the raise and `return SlotState.WARMING`: no raise, so `load`'s except branch is never entered, so **no stamp, no stop, no crash-loop increment** — confirmed by reading the control flow, not the summary. `no_content` / `empty_output` / `incoherent_output` are byte-for-byte the previous terminal path. `AMBIGUOUS_STATUSES = frozenset({"probe_timeout"})` and the DO-NOT-REVERT rationale is in the constant's docstring where the next reader will hit it. Good — that comment is the thing that stops this being re-broken in six months.

**2. The judgment call: I do not block on it, and I do not fully endorse the premise.** You are right that my evidence was timeout-specific and that widening past it trades a measured false-fail for an unmeasured weakening. Terminal stays.

But the premise it rests on — that `/health` 200 strictly precedes `/completion` availability — is **not pinned anywhere**. `ContainerProvider.health` falls back to `/v1/models` when `/health` is non-200 and treats 200 there as healthy; `wait_ready` loops on that; the gate probes immediately after. If the pinned runner (`hal0-rocmfpx:ade07ba`, llama.cpp b10297) answers `/health` non-200 while loading but serves `/v1/models` 200 in that window, `wait_ready` returns early and the gate eats a 503 → `probe_http_503` → terminal, on a healthy slot, worst on the slowest boxes.

I could not settle it from this repo and I am not going to guess: the runner source is not vendored, and `tests/providers/test_container_health_gating.py:107-125` pins the llama-server *ready* shapes (200 non-JSON → ok, 200 JSON without `model_loaded` → ok) plus the FLM 404→`/v1/models` fallback — **no test pins a llama-server LOADING state at all**, and every exercise of that fallback is FLM-shaped. So the answer to your binary is: *neither* — the window is structurally possible and empirically unverified. It is a two-minute curl against a real container mid-load to settle, so I filed it as **#1969** rather than hold this PR on speculation. If it turns out the window is real, the fix is upstream of the gate anyway (scope the fallback to FLM, or check `/health` directly before probing) — not a change to `AMBIGUOUS_STATUSES`.

**3. Device-derived budget.** `OUTPUT_SANITY_CPU_TIMEOUT_S = HEALTH_TIMEOUT_S` — and tying it to "no more patience than the phase before it" is a better justification than the literal I would have written. `probe_budget_s` resolves through `_cfg_effective_backend` (`slots/config_write.py:32`), which promotes a `backend`-only TOML via `map_backend_to_device` before deriving, so legacy configs land identically — the claim checks out. Returns `None` off the cpu lane, so **rocm/vulkan keep 20s/45s** ✓. It reaches both requests because `probe()`'s `timeout_s` already overrides the pair ✓. NPU resolves to `flm`, which never reaches the gate anyway (`provider is None`).

**4. Budget cascade — re-derived independently, all four match.** `LOAD_LOCK_HOLD_S` = 30 + 90 + 180 + 360 + 30 = **690** ✓. `slot_load` (690+690)×1.15 = **1587.0** ✓; `slot_unload` (30+690)×1.15 = **828.0** ✓; `restart` (720+1380)×1.15 = **2415.0** ✓; `stack_apply` ×6 = **14490.0** = 4.025h ✓.

My consumer audit still holds against this diff: no `src/` file outside `slot_lifecycle_budget.py`, `manager.py` and `output_sanity.py` changed, so **nothing new consumes the ceiling** — every consumer remains a client *read* timeout, `_CONNECT_TIMEOUT_S = 5.0` is still separate, `max_slot_lifecycle_timeout_s()` still has no in-tree caller. 4h is a ceiling, not a wait.

**5. Convergence, route (c) — accurate.** `load` does ask systemd on a WARMING resolve: `manager.py:1728-1743` calls `unit_failure_reason(slot_name)` whenever `_await_ready` returns WARMING and raises `SlotSpawnFailed` on a non-empty reason (#1791). The sanity-timeout WARMING inherits that for free — no new plumbing, and a dead backend really does reach ERROR inside the same load. Routes (a) and (b) check out too, and I agree the active-unit/dead-backend limit is honestly stated: it is the pre-existing health-wait policy, and joining it beats inventing a second one.

**6. Evidence — verified, and slightly stronger than claimed.** `tests/slots/test_output_sanity_gate.py` at `cb9753ff`: **54 passed** ✓. `tests/slots` + `tests/cli/test_slot_commands.py` + `tests/mcp/test_admin.py`: **973 passed** (you said 972 — trivial drift, not a discrepancy worth chasing).

Red-first spot check, this revision's tests against `a8ba81d`'s source in a clean worktree:

```
FAILED test_a_slow_server_parks_the_slot_warming_instead_of_condemning_it
FAILED test_a_probe_timeout_leaves_the_slot_adoptable
FAILED test_a_slot_parked_warming_re_runs_the_gate_on_the_next_load
FAILED test_a_dead_backend_still_converges_to_error_in_the_same_load
FAILED test_a_cpu_slot_gets_a_wider_probe_budget
FAILED test_the_load_path_hands_the_probe_the_slot_s_budget
6 failed, 48 passed
```

Six, not five — `test_a_dead_backend_still_converges_to_error_in_the_same_load` goes red too. Red-first confirmed either way. And I checked the WARMING spec is what you say it is rather than a stubbed verdict: it starts a real `FakeSlotServer` with a handler that returns `None`, monkeypatches only the wall clock down to 0.3s, and asserts the full triple — WARMING, no `SANITY_EXTRA_KEYS` in the record, `unload_calls == []`. That is the production path.

**7. Docs.** The split into wrong-answer vs never-answered is right, the `Aside` keeps the true half of the line it replaced, and the "(best-effort — a stop that does not return is reported in the same message)" nit is taken. Accurate against the code. No `CHANGELOG.md` hunk; scope still clean.

**On `save_slot_config`:** you already filed it as **#1967** — that is the right call and nothing more is owed here.

---

Fix the one word and the mixed-verdict branch and I have nothing left. The `AMBIGUOUS_STATUSES` docstring in particular is the kind of comment that earns its tokens.

thinmintdev:
## Revision 4 — both blockers closed (`69e21e79`)

One word, one condition. Plus one part of the prescription I did **not** take, with the arithmetic for why — flagging it prominently rather than quietly skipping it.

### ✅ BLOCKING 1 — the scar ratchet

`legacy` → `an older`. Same sentence, same meaning, and you were right that the concept already lives in `_cfg_effective_backend` where it belongs.

```
✅ scar markers: 192 <= baseline 192
✅ no overdue HAL0-SUNSET shims (current v1.0.0-rc.6)
```

### ✅ BLOCKING 2 — the composed verdict no longer launders a timeout

```python
if is_ambiguous(chat.status):
    # The second opinion never ARRIVED, so there is no second opinion to
    # convict on. ...
    return SanityVerdict(ok=False, status=chat.status, sample=verdict.sample,
                         detail=f"raw probe said {verdict.status}; the chat second opinion never arrived")
```

Your framing is the one in the comment: the fallback exists *because* one endpoint's bad answer is not proof, so no corroboration means no conviction. The raw finding is still carried for the operator — it moves from `status` to `detail`, which is the right way round, since `detail` is prose and `status` is what the machine branches on.

I also fixed the comment you flagged as false: `# Both endpoints answered badly` now guards only the branch where both endpoints actually answered.

**Why this costs #1888 nothing** — and this is the part that makes the carve-out *sufficient* rather than merely safe. Garbage is fast. A mis-computing backend streams at 60–120 tok/s, so it corroborates on **both** endpoints well inside any budget and still composes to a terminal `incoherent_output`. The only slot that can reach the new branch is one too slow to answer twice — which is exactly the slot that must be retried. I pinned that property explicitly rather than leaving it as an argument: `test_garbage_stays_terminal_even_though_a_timeout_does_not` now asserts `seen == ["/completion", "/v1/chat/completions"]`, so if garbage ever stops corroborating, the test says so.

### ⚠️ NOT taken: a smaller cpu-lane `SANITY_CHAT_MAX_TOKENS`

Your premise is correct — 256 tokens inside 180s is a 1.42 tok/s floor and the 0.12 tok/s box cannot clear it. **But the remedy backfires, and it backfires into exactly the failure class this PR exists to remove.**

A truncated generation is not reported as a truncation. A thinking model that spends its whole cap inside `<think>` returns `content=""` with a partial thought in `reasoning_content`, and `_message_text` judges the union. Executed against the real classifier:

```
truncated think block -> incoherent_output | ok= False
is_ambiguous: False
=> TERMINAL (ERROR + stop + durable mark)
```

So shrinking the cap converts a **safe** outcome (`probe_timeout` → ambiguous → WARMING, retryable, unstamped) into a **hard false-fail** — and it does so for precisely the population the fallback was widened to 256 tokens to protect. It trades a free deferral for a condemnation.

The obvious repair — treat a cap-truncated response as ambiguous — is worse still: rc.6's garbage box hits *any* token cap at 60–120 tok/s, so that would blunt #1888 detection on the fallback path. Trading a false-fail for a false-pass on the one defect the gate exists for is not a trade I should make unilaterally.

The timeout is the correct lever and it is now pulled. On the cpu lane the fallback defers instead of convicting, which costs nothing given garbage corroborates fast. Pinned by `test_the_chat_budget_is_not_shrunk_on_the_cpu_lane`, which asserts the trap (truncated think block → `incoherent_output`, not ambiguous) and floors the cap at 256, so nobody walks into this later — including me, next round.

If you still want the cap reduced, the prerequisite is truncation-awareness that can distinguish "still thinking" from "emitting garbage", and I do not think that distinction is available from the response shape. Happy to be shown otherwise.

### 📝 #1969 referenced

Added next to the terminal `probe_*` statuses, since that issue is the premise they rest on:

> That rests on `/health` 200 strictly preceding `/completion` availability — structurally possible to violate (`wait_ready` accepts a `/v1/models` 200 fallback) and not yet empirically settled against the pinned runner: tracked as #1969. If that window turns out to be real the fix is upstream of this set, not inside it.

And thank you for the correction on the red count — **6, not 5**. I miscounted `test_a_dead_backend_still_converges_to_error_in_the_same_load`, which I wrote as a convergence guard and did not expect to be behaviour-dependent; it goes red at `a8ba81d` because the load never reaches the WARMING resolve there. Your number is right. Same for 973 vs 972.

### Red-first, this revision

Both new specs against `cb9753ff` (the mixed case reaching `slot.output_sanity_failed` → ERROR, which is the defect):

```
FAILED test_a_second_opinion_that_never_arrives_does_not_compose_a_verdict
FAILED test_a_mixed_verdict_parks_the_slot_warming_without_condemning_it
2 failed, 54 deselected
```

Both drive a **real socket** — `/completion` answers 200 with empty content, `/v1/chat/completions` hangs — through the real `probe()`, and the load-level one asserts the full triple: `WARMING`, no `SANITY_EXTRA_KEYS` in the record, `unload_calls == []`.

In fairness: `test_the_chat_budget_is_not_shrunk_on_the_cpu_lane` is **green on both sides**. It documents and guards a trap rather than pinning a behaviour change — calling that "red-first" would be a stretch, so I am not.

### Green

| Gate | Result |
|---|---|
| `tests/slots/test_output_sanity_gate.py` | **57 passed** (54 → 57) |
| `tests/slots` + slot-CLI + MCP admin + `tests/providers` | **1619 passed** |
| Full suite | **10989 passed**, 16 skipped, 1 xfailed, 16m09s, exit 0 |
| `scripts/check_sunset.py` | **192 ≤ 192** ✅ |
| `ruff check` / `ruff format --check` | clean over `src/` + `tests/` |

No `CHANGELOG.md` hunk. Budget arithmetic untouched this revision — no new in-lock phase, so `LOAD_LOCK_HOLD_S` stays 690 and the cascade you re-derived still holds. Not merging.

thinmintdev:
## Final pass on `69e21e79` — **0 blocking**

Both blockers are closed, and I am **overruling myself** on the one open question: the refusal is correct and my prescription's second half was wrong. Details below so the reasoning is on the record rather than just the outcome.

### ✅ Blocker 1 — scar ratchet

`sunset` **passes** on `69e21e79`. The reword to "an older TOML" is the whole change, count back to 192. Full CI: **8/8 green** — `python (3.12)` 23m22s, `γ-suite (chromium)` 11m5s, `ui`, `sunset`, `CodeQL`, `Analyze (actions / javascript-typescript / python)`. `mergeStateStatus: CLEAN`.

### ✅ Blocker 2 — the composed verdict

Re-ran my exact repro (real socket: `/completion` 200 with empty content, `/v1/chat/completions` never answers) against `69e21e79`:

```
VERDICT: SanityVerdict(ok=False, status='probe_timeout', sample='',
                       detail='raw probe said empty_output; the chat second opinion never arrived')
is_ambiguous: True
=> WARMING (retryable)
```

Was `empty_output` / terminal, is now `probe_timeout` / ambiguous, and the raw finding is preserved in `detail` rather than lost. The `# Both endpoints answered, and both answered badly` correction is accurate for the branch it now guards.

Red-first against `cb9753ff`, this revision's tests through the real sockets: **2 failed, 55 passed** — exactly `test_a_second_opinion_that_never_arrives_does_not_compose_a_verdict` and `test_a_mixed_verdict_parks_the_slot_warming_without_condemning_it`. Matches your claim. `tests/slots/test_output_sanity_gate.py` at head: **57 passed**.

The sufficiency assertion is the part I want to single out. `assert [path for path, _ in slot_server.seen] == ["/completion", "/v1/chat/completions"]` in the garbage test does real work: it pins that a fast-but-wrong backend *corroborates on both endpoints inside the budget*, which is what makes the carve-out **sufficient** and not merely safe. Without it, "no corroboration, no conviction" would be a claim; with it, it is a property. That is a better test than the one I would have asked for.

---

## Ruling: the refusal stands. I withdraw the second half of my prescription.

I checked the premise by execution rather than accepting the summary, and it is not just correct — it is stronger than stated. Both populations classify **identically**:

```
truncated thinking model -> incoherent_output | ambiguous: False
token salad              -> incoherent_output | ambiguous: False
```

That is the whole argument in two lines. A truncated `<think>` block (`content=""` + partial `reasoning_content`, judged as a union by `_message_text`) and rc.6-class `)))` salad land on the same status, so:

* **Shrink the cap** → thinking models truncate → `incoherent_output` → **terminal**, with the durable un-adoptable mark. My proposal would have converted a *safe* failure (`probe_timeout` → ambiguous → WARMING → retried) into an *unsafe* one, on precisely the population the fallback exists to protect. It would have re-opened the original blocker through a different door.
* **Treat truncation as ambiguous** → salad becomes ambiguous too, because it is the same status → the gate false-passes the #1888 defect it was built for.

You invited an overrule if the shapes are distinguishable from the response. I tried three discriminators and all three fail:

1. **`finish_reason == "length"`** — present on both in my check, and necessarily so: anything that fills the cap reports `length`, reasoning and salad alike. Not a discriminator.
2. **Structural (`<think>` tag present / unclosed)** — where reasoning lands depends on the *runner build*, not on health: a splitting llama-server puts it in `reasoning_content` with no tag, a `--jinja`-less one leaves inline `<think>…`. Keying on it makes the verdict build-dependent, which `_message_text`'s docstring explicitly set out to avoid. And salad can appear inside a think block anyway.
3. **"Is the partial text natural language"** — this is a language-quality judgement, which is design rule #1 inverted ("Positive assertion only. Never key on the garbage string: its shape varies with argv"), and #1888 is the empirical refutation: five lanes, five different garbage shapes, one defect.

The cleanest refutation is your own suite: any discriminator keyed on "reasoning field populated, content empty ⇒ still thinking" would have to fail `test_probe_fails_a_thinking_model_that_reasons_in_garbage`, which pins that garbage in `reasoning_content` must stay terminal. The classifier I would need is the one you already forbade. So there is no classifier, and the 256 floor with the 180s clock is the right pairing: on the cpu lane the fallback fails by **clock** — now ambiguous — never by truncation.

`test_the_chat_budget_is_not_shrunk_on_the_cpu_lane` pinning 256 as a floor with that reasoning in the docstring is the correct artefact. It stops the next reader (me, apparently) making this exact mistake.

**One non-blocking residual**, named only so it is not rediscovered as a surprise: truncation-at-256 is still terminal, so a thinking model whose raw probe fails *and* whose think block exceeds 256 tokens on `"The capital of France is"` would be condemned. Both conditions are needed and the second is unlikely for a question this trivial, and the constant is pinned as a **floor**, so raising it is the available lever if the fleet ever shows it. No action now.

---

**0 blocking, no requested changes.** Everything from the prior two rounds — the adoption choke point and its mutation check, the budget derivation and cascade, the device resolution through `_cfg_effective_backend`, the convergence routes, the docs — I have already verified and am not relitigating. The `#1969` pointer added to the `AMBIGUOUS_STATUSES` docstring is a nice touch: it puts the unverified premise next to the code that depends on it instead of only in a tracker.

This is a good permanent net for the #1888 class. Ship it.

PR #1968, opened 2026-08-21
feat(api): give image_status an `unknown` member and map the seam end-to-end

### LINKED ISSUE #1939: api: image_status has no 'unknown' member — an unanswerable podman-seam read is indistinguishable from a genuinely missing image
Follow-up from #1889 / PR #1937 (open question 1).

`ContainerProvider.image_present` returns a bool and the image_status payload enumerates present/missing only. When the rootful seam is unanswerable as the hal0 service user (rc 66 operational failure, missing grant, wrapper drift), the API now logs `hal0.podman_ro.image_present_unanswered` at WARNING but still reports `missing` — an authoritative-looking answer for an unanswered question. That shape (green/definitive surfaces over an undiagnosable state) is the same failure family as #1888's health surfaces and #1465's seam inventory.

Fix is an API-schema change: add an `unknown` member (or tri-state) to image_status and propagate the seam's tri-state result. Deliberately out of #1937's scope. Worth deciding before GA whether the schema change rides 1.0 or 1.0.x.

Refs #1889.

thinmintdev:
`image_status` could report `missing` about an image nothing had managed to look at.

The `hal0-podman-ro` wrapper spends four exit codes distinguishing "podman answered, negatively" (rc 0 + `missing`) from three flavours of "the seam is unusable" (rc 64 bad operand, rc 65 no podman, rc 66 podman ran and broke), plus sudo's own rc 1 when the grant is absent. All of that collapsed to a bare `None` at the seam, then to `False` in `ContainerProvider.image_present` — whose contract was a bool — and finally to an authoritative-looking `image_status: "missing"` on the wire. Operators, the dashboard, agents reading `/api/slots`, and the release-validation kit all got a confident answer to a question nobody asked.

After this, `missing` means only what it says: **podman was asked and the image is not there.** Everything that prevented the question from being asked is `unknown`.

## Schema decision

This is a **public API-schema change** and it ships **pre-GA, in rc.7**, per decision-ledger row 4 (add `unknown` before GA — a break is cheap now and expensive after). #1939 left 1.0 vs 1.0.x open; that call is made. The change is additive to the payload (a fifth member of an existing enum), so no field is removed or renamed on the wire — but consumers that exhaustively switch on `image_status` must handle it, which is exactly why it lands before the vocabulary is frozen.

Deliberately **not** added: an `image_status_reason` payload field. The reason is real and is preserved, but it goes to the journal (`podman_ro.image_present_unanswered … reason=<name>`), not to the public schema. The payload's job is to stop lying; the journal's job is to say why. That keeps the ratified surface change to exactly the one member the operator approved.

## The chain, site by site

The planner mapped seven sites where the tri-state had to survive. Missing any one leaves the lie in place, so each has its own test:

| # | Site | Before | After | Pinned by |
|---|---|---|---|---|
| 1 | `podman_introspect` seam | rc 64/65/66 + sudo rc 1 → one undifferentiated `None` | `image_exists` → `image_presence`, returning `ImageProbe(presence, reason)`; each rc maps to a named `SeamUnanswered` | `test_image_presence_maps_every_wrapper_rc_to_a_named_unknown` (parametrised over rc 1/64/65/66/125), `…_unknown_on_unexpected_stdout`, `…_unknown_on_subprocess_error`, `…_rejects_bad_ref_without_calling_sudo` |
| 2 | `ContainerProvider.image_present` bool contract | `False` (a lie) + a reasonless WARNING | `bool \| None`, WARNING carries `reason=` | `test_image_present_is_unknown_not_false_when_the_seam_cannot_answer` (parametrised over all five reasons, with the rootless store booby-trapped), `test_image_present_logs_the_seam_reason_when_it_does_not_answer` |
| 3a | `_image_present_cached` | `bool(...)` coerced `None` → `False` | caches the tri-state verbatim | `test_image_present_cache_round_trips_unknown_without_flattening_it` (also asserts the second read is a cache hit) |
| 3b | aggregator except-branch | `"missing"` | `"unknown"` | `test_probe_exception_reports_unknown_not_missing` |
| 3c | aggregator probe-timeout fallback | `"not-configured"` — a claim about config from a probe that read no config | `"unknown"` | `test_one_wedged_slot_cannot_hold_the_list` (extended) |
| 4 | API payload | `GET /api/slots`, `/pull/status`, `/pull/stream` all read `missing` | all three read `unknown`; `inspect_image_state` shares the aggregator's `image_status_for` helper so they cannot drift | `test_image_status_unknown_when_the_probe_cannot_answer`, `test_pull_status_route_reports_unknown_rather_than_missing` |
| 5 | `useSlots.ts` union | 4 members | 5, plus `unknown` added to `ImagePullState` **and** to `IMAGE_PULL_TERMINAL` (otherwise the EventSource hangs on it forever) | compile-time: the vitest file declares a typed `Slot` literal with `image_status: 'unknown'`, and `src/**/*.ts` is typechecked — drop the member and `npm run typecheck` fails |
| 6 | `slots.jsx` rendering | nothing rendered for any image state but `pulling` | `imageStatusChip` → neutral dashed "image ?" chip, on the live `SlotScard` *and* the retired grid card | `image-status-chip.test.ts` (6 cases incl. "cls carries no err/warn/ok token", "tooltip must not claim absence, must name `hal0 doctor`"), plus two γ cases |
| 7 | e2e specs + mock data | — | `slot-card-container-v3.spec.ts` gains 3 cases; the forced-mock seed's `legacy` slot and `mock-data.ts`'s `tts` slot carry `image_status: 'unknown'` | γ: classifier cases + one **DOM** case asserting the chip renders on a real card and that its definitively-answered neighbour has none |

Why neutral and dashed rather than red or amber: this card's colour vocabulary is RED = error, AMBER = transitional/degraded, GREY = not loaded. "We could not read the image store" is none of those, and painting it as one restates the same confident lie in CSS. `missing` deliberately gains no new chip — it has always been backend-only, and adding a red chip here would bury the one state that actually wants an operator's eye.

Two honest-degradation fixes came along because they are the same bug one layer down: `image_present`'s dev-checkout fallback now answers `unknown` when there is no container runtime at all, and when podman itself fails to exec. Neither is an answer about the image.

## Release-validation kit (rides with this PR)

The kit adjudicates the field that changed, so it changes here — `kit_version` 5 → 6, per the kit's own versioning policy ("bump it in the same PR as any change to lane briefs").

`readonly/api.md` check 6, `stateful/slots.md` check 1, and the `image-status-wrong-podman-store` expect clause now distinguish the two failures. `missing` on a running slot is still that regression — a confident wrong answer. `unknown` is the API declining to answer because the seam was unusable: it does **not** re-open the regression, but it is **not a pass either** — it is a seam finding, followed up via the `reason=` on the journal line. A fleet reading `unknown` everywhere means the sudoers grant did not install; `unknown` on a box whose seam is demonstrably healthy is a defect of its own. The #1937 invariant stays meaningful precisely because `missing` is now unreachable except from a real negative.

**Note for the changelog agent:** this PR changes the kit (v6) and adds a public enum member. CHANGELOG.md is untouched here by instruction.

## Verification

Every gate run locally on this branch, at `92d1a0e2` + this commit:

```
pytest tests/api                                   1796 passed, 1 skipped
pytest tests/providers tests/slot_view tests/slots 1577 passed
pytest tests/<everything else>                     7624 passed, 15 skipped, 1 xfailed
pytest <all touched suites, re-run post-edit>      1719 passed
ruff check src tests                               All checks passed!
ruff format --check src tests                      1162 files already formatted
ui: npx eslint .                                   clean
ui: npx tsc --noEmit                               clean
ui: npx vitest run                                 14 files, 135 tests passed
ui: npm run build                                  built in 2.98s
ui: npx playwright test (full γ-suite)             648 passed, 16 skipped
```

The full γ-suite was run rather than just the affected spec, because the change touches the shared forced-mock seed (`src/dash/data.jsx`) that every γ spec renders from.

Closes #1939. Refs #1889, #1937.

### DIFF
--- a/src/hal0/api/routes/slots.py
+++ b/src/hal0/api/routes/slots.py
@@ -1787,7 +1787,8 @@ async def pull_slot_image_stream(name: str, request: Request) -> StreamingRespon
         data: {"slot_name": "...", "image": "...", "state": "pulling",
                "layer": N, "total_layers": M}
 
-    Terminal states: ``completed`` | ``failed`` | ``present`` | ``missing``.
+    Terminal states: ``completed`` | ``failed`` | ``present`` | ``missing`` |
+    ``unknown`` (#1939 — the image store could not be read).
     """
 
     async def _gen() -> Any:
@@ -1795,7 +1796,8 @@ async def pull_slot_image_stream(name: str, request: Request) -> StreamingRespon
         job = slot_pull_jobs.get(name)
 
         if job is None:
-            # No active pull — inspect the image to surface present|missing.
+            # No active pull — inspect the image to surface
+            # present|missing|unknown.
             sm = _get_slot_manager(request)
             image = await _image_pull.resolve_slot_image(sm, name)
             state = await _image_pull.inspect_image_state(image)
@@ -1828,8 +1830,9 @@ async def pull_slot_image_status(name: str, request: Request) -> dict[str, objec
     one-shot poll equivalent for clients that can't hold an EventSource
     open. Returns the in-flight job snapshot when a pull is active,
     otherwise inspects the slot's image and reports ``present`` |
-    ``missing`` — the same terminal vocabulary the stream's no-job branch
-    emits, so a poller and a streamer converge on the same states.
+    ``missing`` | ``unknown`` — the same terminal vocabulary the stream's
+    no-job branch emits, so a poller and a streamer converge on the same
+    states.
 
     Frame shape::
 
@@ -1837,7 +1840,8 @@ async def pull_slot_image_status(name: str, request: Request) -> dict[str, objec
          "layer": N, "total_layers": M, "error": null}
 
     States: ``pulling`` | ``completed`` | ``failed`` | ``present`` |
-    ``missing``.
+    ``missing`` | ``unknown`` (#1939 — the image store could not be read; NOT
+    a report that the image is absent).
     """
     slot_pull_jobs: dict[str, Any] = getattr(request.app.state, "slot_pull_jobs", {})
     job = slot_pull_jobs.get(name)
--- a/src/hal0/providers/container.py
+++ b/src/hal0/providers/container.py
@@ -2597,8 +2597,12 @@ class ContainerProvider(Provider):
         """
         self._run("systemctl", "reset-failed", self._unit_name(_artefact_token(slot)), check=False)
 
-    def image_present(self, image: str) -> bool:
-        """Return True if ``image`` is in the container image store slots use.
+    def image_present(self, image: str) -> bool | None:
+        """Is ``image`` in the container image store slots use? Tri-state.
+
+        ``True`` / ``False`` are answers: something actually looked in the
+        right store. ``None`` means **nobody could look** — the rootful seam
+        was unusable and this process has no other store worth asking.
 
         #1889: asks ROOT's store first, through the ``hal0-podman-ro``
         ``image-exists`` seam. Slots run ROOTFUL podman (Quadlet), while
@@ -2606,46 +2610,52 @@ class ContainerProvider(Provider):
         so the bare ``<runtime> image inspect`` below reads hal0's own
         ROOTLESS store — a different store, which never contains a slot image.
         That is why ``image_status`` was ``"missing"`` for every running,
-        healthy slot on every standard install. The seam returns ``None`` when
-        it cannot answer, and only then do we fall back to the rootless read —
-        and only when this process is NOT the hal0 service user. On a
+        healthy slot on every standard install. The seam reports ``unknown``
+        when it cannot answer, and only then do we fall back to the rootless
+        read — and only when this process is NOT the hal0 service user. On a
         provisioned box the rootless store is definitionally the wrong store,
         so consulting it after a seam failure would collapse "podman is
         broken / the grant is missing" (wrapper rc 66 / sudo rc 1) back into
         an authoritative-looking "missing", which is #1889 with extra steps.
-        We log instead, so an operator gets a diagnosable signal.
 
-        (The seam-failed-as-hal0 case still has to answer ``False`` because
-        this method's contract is a bool and ``image_status`` has no
-        "unknown" member. Surfacing a distinct unknown state is an API-schema
-        change, deliberately out of scope here — see the PR notes.)
+        #1939: and so is answering ``False`` there, which is what this method
+        used to do because its contract was a bool. It is now a tri-state, and
+        every ``None`` is logged at WARNING with the seam's own reason —
+        ``image_status: "unknown"`` tells a reader the answer is untrustworthy,
+        the journal line tells an operator which grant or store to go fix.
 
         In a dev checkout, where hal0-api runs as the operator, the rootless
-        store IS the store slots use, so the fallback is correct there.
+        store IS the store slots use, so the fallback is correct there — but
+        even there, a box with no container runtime at all, or one whose
+        podman will not exec, has not answered the question either.
 
         Runs synchronously — callers must dispatch to a thread executor when
         called from an async context.
         """
-        seam_answer = podman_introspect.image_exists(image)
-        if seam_answer is not None:
-            return seam_answer
+        probe = podman_introspect.image_presence(image)
+        if probe.presence != "unknown":
+            return probe.presence == "present"
         if is_hal0_service_user():
             log.warning(
-                "hal0.podman_ro.image_present_unanswered image=%s — the rootful seam "
-                "did not answer (grant missing, podman failure, or rejected ref); "
-                "reporting missing without consulting the unrelated rootless store",
+                "hal0.podman_ro.image_present_unanswered image=%s reason=%s — the rootful "
+                "seam did not answer; reporting image_status=unknown rather than consulting "
+                "the unrelated rootless store (see `hal0 doctor seams`)",
                 image,
+                probe.reason,
             )
-            return False
+            return None
         try:
             runtime = _container_runtime()
         except RuntimeError:
-            return False
-        result = subprocess.run(
-            [runtime, "image", "inspect", image],
-            capture_output=True,
-            check=False,
-        )
+            return None
+        try:
+            result = subprocess.run(
+                [runtime, "image", "inspect", image],
+                capture_output=True,
+                check=False,
+            )
+        except (OSError, subprocess.SubprocessError):
+            return None
         return result.returncode == 0
 
     def running_image(self, slot: Mapping[str, Any] | str) -> str | None:
--- a/src/hal0/providers/container.py
+++ b/src/hal0/providers/container.py
@@ -2626,8 +2626,18 @@ class ContainerProvider(Provider):
 
         In a dev checkout, where hal0-api runs as the operator, the rootless
         store IS the store slots use, so the fallback is correct there — but
-        even there, a box with no container runtime at all, or one whose
-        podman will not exec, has not answered the question either.
+        only for the one reason that actually means "we never asked because
+        this is not a provisioned box" (``not-service-user``). Any other
+        reason means the seam was tried, or that our own mirror of the
+        wrapper's validator rejected the ref — and a ref this side rejects is
+        one podman would reject too, so the rootless store has nothing to add.
+
+        The fallback probe is ``podman image exists``, NOT ``image inspect``,
+        for exactly the reason the wrapper header gives: ``exists`` has the
+        contract rc 0 = yes / rc 1 = no / anything else = operational failure,
+        while ``inspect`` collapses "not found" and "podman is broken" into a
+        single non-zero rc. Using ``inspect`` here would reintroduce the same
+        conflation on dev boxes that the seam removed on deployed ones.
 
         Runs synchronously — callers must dispatch to a thread executor when
         called from an async context.
@@ -2635,7 +2645,7 @@ class ContainerProvider(Provider):
         probe = podman_introspect.image_presence(image)
         if probe.presence != "unknown":
             return probe.presence == "present"
-        if is_hal0_service_user():
+        if probe.reason != "not-service-user":
             log.warning(
                 "hal0.podman_ro.image_present_unanswered image=%s reason=%s — the rootful "
                 "seam did not answer; reporting image_status=unknown rather than consulting "
@@ -2650,13 +2660,23 @@ class ContainerProvider(Provider):
             return None
         try:
             result = subprocess.run(
-                [runtime, "image", "inspect", image],
+                [runtime, "image", "exists", image],
                 capture_output=True,
                 check=False,
             )
         except (OSError, subprocess.SubprocessError):
             return None
-        return result.returncode == 0
+        if result.returncode == 0:
+            return True
+        if result.returncode == 1:
+            return False
+        log.warning(
+            "hal0.podman_ro.image_present_unanswered image=%s reason=podman-failed rc=%s — "
+            "the local podman ran but failed operationally; reporting image_status=unknown",
+            image,
+            result.returncode,
+        )
+        return None
 
     def running_image(self, slot: Mapping[str, Any] | str) -> str | None:
         """Return the image ref of the running container for a slot (#663).
--- a/src/hal0/providers/container.py
+++ b/src/hal0/providers/container.py
@@ -2657,6 +2657,12 @@ class ContainerProvider(Provider):
         try:
             runtime = _container_runtime()
         except RuntimeError:
+            log.warning(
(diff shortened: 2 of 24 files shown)


thinmintdev:
Triaged the three P2s from the Codex review. All three were correct and all three are fixed in `925c4288`.

**1. "Preserve unknown in the rootless fallback" — accepted, and widened.**
Right on both counts. `podman image inspect` returns 125 for "not found" *and* for "podman broke", which is exactly why the wrapper uses `image exists` root-side — and the dev path was still using `inspect`, so it reproduced the original conflation on every non-service-account box. That path now runs `podman image exists` with the same rc 0 / rc 1 / else contract, and an operational rc gets a journal line plus `unknown`. (`_container_runtime()` resolves podman only — "docker is unsupported" per its docstring — so `image exists` is always available.)

Also took the other half of the comment: the fallback is now gated on the seam's *reason* rather than on `is_hal0_service_user()`. Previously a ref that our own mirror of the wrapper's validator rejected (`invalid-argument`, which short-circuits before the service-user gate) would fall through to the rootless store on a dev box and come back `False`. A ref this side rejects is one podman would reject too, so that could only ever launder a validation failure into an authoritative "missing". Only `not-service-user` falls back now.

**2. "Keep profileless timeout results as not-configured" — accepted; my test was pinning the wrong thing.**
Correct, and the sharper point is that `cfg` already answers this without any probe: no profile → no declared image → `not-configured` (#1226), and a timeout does not make that less true. The honest-`unknown` argument only applies to a slot that *does* declare an image and whose store we never got to read. Fallback is now `"unknown" if profile_name else "not-configured"`, and — as you spotted — the concurrency test I extended was hitting the profileless case, so it is back to asserting `not-configured`, with a new sibling test wedging a *profiled* slot for the `unknown` half.

**3. "Log a reason for non-seam unknown probe results" — accepted.**
This one mattered most, because the PR body, the UI tooltip and three kit briefs all send operators to the journal for the reason the payload deliberately omits — and this was the single `unknown` with nothing there. The branch now emits `slot_view.image_probe_failed slot=… image=… reason=probe-error error=<exc>`, with a test asserting the slot name and the exception text reach the log. `readonly/api.md`, `stateful/slots.md` and the `image-status-wrong-podman-store` expect clause now name both events, so the kit's rule reads as an invariant an agent can actually check: every `unknown` has exactly one of those two lines behind it, and an `unknown` with neither is itself a finding.

Re-verified after the fixes: `pytest tests/providers tests/slot_view tests/slots tests/api` → 3377 passed, 1 skipped; `ruff check` and `ruff format --check` clean; kit YAML/TOML re-parsed. UI is untouched by this commit (its gates passed on the parent and re-run in CI).

thinmintdev:
Second review round: both new P2s accepted and fixed in `6f5ef016`.

They are the same invariant as the previous round's third finding — and, awkwardly, one this branch itself wrote into the validation kit two commits ago: *"every `unknown` has exactly one of those two journal lines behind it; an `unknown` with neither is itself a finding."* Two paths still degraded silently, which made that sentence false as written. Now it holds.

**"Log pull-route probe failures before returning unknown"** — `inspect_image_state`, behind `/pull/status` and the no-job `/pull/stream`, swallowed a raising probe. It now emits `slot_view.image_probe_failed … reason=probe-error error=<exc>`, deliberately the same event name and reason token the aggregator uses, so an operator greps once and gets the slot list and both pull surfaces.

**"Log failures in the rootless fallback probe"** — `image_present`'s dev-checkout fallback returned `None` wordlessly both when no container runtime exists on the box and when podman fails to exec. Both now log using the seam's own vocabulary (`reason=podman-absent`, `reason=seam-error`), so a dev or CI box reads the same way a deployed one does instead of having a second, undocumented silence.

Each of the four log lines added across the two rounds is asserted by a test (event name, reason token, and the underlying error/rc where there is one) rather than left to inspection — otherwise they are exactly the kind of thing that rots into a stale format string.

`pytest tests/providers tests/slot_view tests/slots tests/api/test_slots_image_pull.py tests/api/test_slots_container_state.py` → 1612 passed; `ruff check` and `ruff format --check` clean. Full CI running on the push.

thinmintdev:
## Independent review — **approve**, no blocking findings

Reviewed `925c4288` (the post-Codex-triage head, not `cda26c52`) as a second pair of eyes; I did not write this. Traced all seven chain sites, hunted for consumers the planner's list missed, and re-ran every gate locally on a clean checkout of the branch. Four non-blocking notes at the bottom, one of which I'd like a follow-up issue for.

### 1. Chain completeness — all 7 sites mapped, no unmapped consumer

| # | Site | Verified |
|---|---|---|
| 1 | `podman_introspect.image_presence` | ✅ `_RC_REASON` covers 1/64/65/66; anything else → `seam-error`. `_run` swallows `OSError`/`SubprocessError` (and `TimeoutExpired` is a `SubprocessError`, so a timed-out seam is `seam-error`, not silence). rc 0 + an undefined word → `unknown/seam-error`, **not** `missing` — that's the drift case, and getting it right matters. |
| 2 | `ContainerProvider.image_present` | ✅ `bool \| None`; fallback gated on `reason == "not-service-user"` and probing with `image exists` (rc 0/1/else), not `inspect`. |
| 3a | `_image_present_cached` | ✅ tri-state cached verbatim; the deliberate absence of `bool()` is called out in the docstring and pinned by a test. |
| 3b | aggregator `except` | ✅ `unknown` + a `slot_view.image_probe_failed` WARNING. |
| 3c | probe-timeout fallback | ✅ `"unknown" if profile_name else "not-configured"` — the right split. |
| 4 | payload (`/api/slots`, `/pull/status`, `/pull/stream`) | ✅ all three route through the one `image_status_for` helper. |
| 5 | `useSlots.ts` union | ✅ `Slot.image_status` **and** `ImagePullState` **and** `IMAGE_PULL_TERMINAL`. That last one is the easy miss — omitting it hangs the EventSource forever on a state that will never advance. |
| 6 | `slots.jsx` / `inference-pane.jsx` | ✅ neutral dashed `image ?`, on the live `SlotScard` and the retired grid card, classifier shared via `slot-status.js`. Not red, not green. |
| 7 | e2e + mock data | ✅ 3 specs incl. a real-DOM assertion against the forced-mock seed, plus the negative (definitive neighbour → no chip). |

**Consumer sweep** — I grepped the whole tree for `image_status` / `image_present` / `image_exists` beyond the planner's list and found **no unmapped consumer**:
- `api/routes/runner_images.py` — false positive, it's a function named `pull_runner_image_status`.
- `installer/lib/preflight.sh`, `system/seam_check.py` — comments only.
- No CLI renderer, no `hal0 doctor` code path, no bench consumer reads the field.
- `ui/src/**` reads `image_status` in exactly three places and every one of them tests `=== "pulling"`, so a fifth member can't crash or mis-route them.
- No `openapi.*` in the repo and nothing under `docs/` names the field — the docstrings genuinely are the tracked schema surface here, and they were all updated.

### 2. Semantics discipline — clean in both directions

No site converts `unknown → missing` or `unknown → present`. The one place I went looking hardest for a leak was the dev-checkout fallback, since it's the only surviving path where a non-seam store gets consulted; gating it on `reason == "not-service-user"` (rather than `is_hal0_service_user()`) is the correct fix, and it closes the `invalid-argument` short-circuit that would otherwise have laundered a validation failure into an authoritative `False`.

The #1937 invariant survives and is arguably sharper than before: `missing` is now unreachable except from a real negative, which is exactly what makes "running slot + `missing`" a defect. `test_declared_image_still_reports_present_and_missing` is the right test to have written here — without it the tri-state could have quietly made `missing` unemittable and left the regression entry pinning a ghost.

Nor is `unknown` a dumping ground. The two places it would have been tempting to over-apply are both resisted: a profileless timeout stays `not-configured` (knowable from `cfg`, no probe needed), and `inspect_image_state`'s no-image case stays `missing` (a resolution failure, not a store read).

### 3. Tests — red-first confirmed by revert, not by assertion

I reverted each source file to `92d1a0e2` and re-ran, rather than taking the PR body's word for it:

```
revert src/hal0/slot_view/__init__.py:
  test_unanswerable_image_probe_reports_unknown_not_missing   FAILED (assert 'missing' == 'unknown')
  test_probe_exception_reports_unknown_not_missing            FAILED (assert 'missing' == 'unknown')
  test_a_wedged_profiled_slot_reports_unknown_not_not_configured  FAILED
  test_declared_image_still_reports_present_and_missing       passed  (correctly — it pins the unchanged half)

revert src/hal0/providers/container.py:
  17 failed, 37 passed — incl. all five [grant-denied|invalid-argument|podman-absent|
  podman-failed|seam-error] parametrisations, the cache round-trip, and the reason-logging test
```

Every collapse site the planner named has a test that genuinely goes red without the fix.

Gates on `925c4288`, my checkout:

```
pytest tests/providers tests/api tests/slots tests/slot_view tests/installer
                                4179 passed, 2 skipped in 12:49
ruff check src tests            All checks passed!
ruff format --check src tests   1162 files already formatted
ui: npx tsc --noEmit            clean
ui: npx vitest run              14 files, 135 tests passed
ui: npx eslint .                clean
kit YAML + TOML                 parse OK
```

CI on `925c4288`: `python (3.12)` pass (24m52s), `γ-suite (chromium)` pass (10m59s), `ui` pass, all three CodeQL analyses pass, `sunset` pass. The γ-suite pass is the one that mattered most to me, since `data.jsx` is the shared seed every γ spec renders from.

### 4. Kit + schema hygiene

`kit_version` 5 → 6 with a changelog entry, per the kit's own "bump it in the same PR as any change to lane briefs" rule. `readonly/api.md` 6, `stateful/slots.md` 1 and the `image-status-wrong-podman-store` expect clause all now distinguish the two failures and — importantly — tell the validating agent that `unknown` is *a finding of a different kind*, not a pass. That's the right call; the failure mode of a weaker edit would have been an agent that shrugs at a fleet-wide `unknown` and reports the lane green while the sudoers grant is missing on every box. No lane/tier/phase change, so the `LANES` table in the workflow script needs nothing.

No `CHANGELOG.md` hunk. ✅

### 5. Codex triage

All three P2s were correct, all three are addressed in `925c4288`, and the triage comment reads as genuine engagement rather than compliance — P1 in particular was widened beyond what Codex asked for (the `not-service-user` gating), and P2 was accepted *against* the author's own test, which is the right instinct.

---

## Non-blocking findings

**N1 — three `unknown` return paths emit no journal line, while three kit briefs state the invariant absolutely.** `readonly/api.md`, `stateful/slots.md` and the regressions entry now all say *"Every `unknown` has exactly one of those two lines behind it; an `unknown` with neither is itself a finding."* Three paths break that:

- `slots/image_pull.py:139` — `except Exception: return "unknown"`. Reachable on a provisioned box (anything `image_present` *raises*, as opposed to returning `None`), and this is the one that would produce a spurious kit finding.
- `providers/container.py:2662` — `except RuntimeError: return None` (no container runtime).
- `providers/container.py:2671` — `except (OSError, subprocess.SubprocessError): return None` (podman failed to exec).

The last two are dev-checkout-only so they can't be hit by the kit, but the first can. A one-line `log.warning` mirroring `slot_view.image_probe_failed` in `inspect_image_state` would make the invariant true as written — it's exactly the fix already applied to the aggregator for Codex's P3, just one surface over. Your call whether it rides here or as a follow-up.

**N2 — same conflation survives one layer over, out of this PR's scope: `capabilities/catalog.py::_flm_image_present()`.** It runs `podman image inspect` against hal0-api's own **rootless** store, maps every non-zero rc to `False`, and *caches* that `False` — so a broken/absent podman permanently reads as "the FLM toolbox image is not there", which drops the NPU backend out of `/api/capabilities`. That is #1889 + #1939 in miniature: wrong store, and inspect's rc collapsing "not found" into "podman broke". Different field, different consumer, genuinely not this PR's job — but it's the same footgun and it should get its own issue rather than a comment that scrolls away.

**N3 (trivial)** — `slot-card-container-v3.spec.ts`: *"The neighbouring card, whose image_status is definitive, carries none."* The `primary` seed slot has no `image_status` key at all, so it's *absent*, not definitive. The assertion is still correct and still valuable; only the comment overstates.

**N4 (trivial)** — `chip image-unknown` has no rule in `dashboard.css`; the neutral/dashed look is entirely inline styles on the JSX. That makes the `not.toMatch(/\b(err|warn|ok)\b/)` assertion vacuously true and leaves the class as a test hook only. Fine as-is, just worth knowing it isn't load-bearing CSS.

---

Approve. This is a careful piece of work — the tri-state is honest at every hop, the two definitive answers stay reachable, the kit was taught the difference instead of being left to guess, and the UI declines the two easy wrong answers (red, or nothing at all). N1 is the only one I'd act on before merge, and only the `image_pull.py` half of it.

PR #1971, opened 2026-08-21
docs(release-validation): kit v8 — Vulkan-restoration coverage for the rc.7 fold

thinmintdev:
## Summary

Kit v8 finalizes release-validation coverage for the rc.7 Vulkan-restoration fold (#1948) against the LANDED fold chain — this PR was originally drafted (as kit v6) in parallel with the lane re-enable PR; it has now been rebased onto current `main` and every provisional marker resolved against real merged code:

- **#1954** — kfd identity by runner uid + device gid (`e5e5925a`)
- **#1973** — `gpu-vulkan` LLM lane re-enable behind an explicit image allowlist (`3474ec03`)
- **#1959** — repin to `ghcr.io/hal0ai/hal0-combined:0822` + the B1 `image_pin` retag fix (`8f23afba`)

Also composed with kit v6/v7's `image_status` tri-state work (#1968, #1980) that landed on `main` in the meantime, rather than duplicated.

### What's covered

- `boxes.toml`: ct151 is the fleet's PRIMARY Vulkan LLM lane box again (render node, no /dev/kfd), with the real gate mechanics named — `VULKAN_CAPABLE_IMAGE_REFS`, an explicit per-image allowlist earned by the #1948 §3-C matrix, NOT a tag/version comparison (Vulkan correctness was not monotonic in build recency). `[boxes.ct152-cpu-fresh]` REMOVED — the box was destroyed 2026-08-21; there is now no CPU-only fresh box in the fleet at all, stated plainly. ct150 update-box note gains the `image_pin` retag + precedence verification and the perf-row reference (Vulkan now measures AHEAD of ROCm, +13.96% prefill / +20.45% decode — replacing the old ade07ba-era "-10%" expectation).
- `lanes/stateful/slots.md`: Vulkan check set 8b-8e, finalized against `require_kfd_for_gpu_slot`'s `gpu-vulkan` branch and `_require_vulkan_lane_prerequisites` (`providers/_gpu.py`) by name — 8b (the #1922 gate, shipped), 8c (positive load, gates named), 8d (negative case — stale image refused, citing #1888 + `VULKAN_FIXED_IMAGE`), 8e (the deliberately-garbage lane: stale image + `ENV_ALLOW_VULKAN_FALLBACK`, the #1922 gate must convict it — called out as the single most important check of the release).
- `lanes/readonly/api.md`: 6c finalized — the capabilities catalog OFFERS `gpu-vulkan` without gating on image version (confirmed by source: "the picker is an OFFER, not a promise"), refusal happens at slot load.
- `lanes/update/upgrade.md`: new checks 5b (image_pin retag + precedence, #1959 B1) and 5c (the inverted perf-row reference envelope).
- `regressions.yaml`: `brain-vulkan-backend-garbage-output` gets a finalized `rc7_note` naming all three landed PRs; two new entries — `updater-image-pin-retag-blind-spot` (issue #1959) and `vulkan-lane-stale-image-preflight-refusal` (issue #1973, filled in from the earlier `issue: null` placeholder).
- `README.md`: kit changelog rewritten as v8, composed alongside (not replacing) the intervening v6/v7 entries. `kit.toml` bumped to `kit_version = 8`.
- `.claude/workflows/rc-validate.js`: verify-prompt fleet note (unchanged from the draft, already correct); `BOX_ROLES` map drops the dead `ct152-cpu-fresh` entry to match `boxes.toml`.

All five original `FINALIZE-AFTER-PHASE-D` markers are resolved — none remain live in the kit files.

## Test plan

- [x] `HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/release -q` — 112 passed (workflow-contract + kit-contract tests, which pin `kit.toml [defaults]` against the script's model pins).
- [x] `node --check .claude/workflows/rc-validate.js`
- [x] `uv run ruff check tests/release-validation tests/release .claude/workflows` — all checks passed.
- [x] `regressions.yaml` / `boxes.toml` / `kit.toml` parse clean (47 regression entries, 4 boxes, `kit_version = 8`).

Refs #1948.

### DIFF
--- a/.claude/workflows/rc-validate.js
+++ b/.claude/workflows/rc-validate.js
@@ -33,9 +33,9 @@ const KIT = `${REPO}/tests/release-validation`
 const RUN = `/mnt/mintdev/artifacts/hal0-release-validation/${VERSION}`
 
 // Box id -> role. Mirrors tests/release-validation/boxes.toml.
+// ct152-cpu-fresh was destroyed 2026-08-21 (kit v8) — there is no CPU-only fresh box left.
 const BOX_ROLES = {
   'ct151-cpu-fresh': 'fresh',
-  'ct152-cpu-fresh': 'fresh',
   'ct163-cpu-fresh': 'fresh-alt',
   'ct150-update': 'update',
   'ct105-prod': 'prod',
@@ -479,13 +479,16 @@ relitigates it.
 
 Also judge \`other_hardware_applicability\`: would a GPU box, a privileged container, or an
 upgraded (rather than freshly installed) box hit this? No box in the fleet is BOTH a fresh
-install AND has /dev/kfd visible (the fresh boxes lack /dev/kfd, so post-#1923 their
-gpu-rocm LLM slots REFUSE to start — the Vulkan LLM lane is retired and there is no CPU
-fallback for a gpu-device slot; LLM coverage on those boxes means slots explicitly set to
-device=cpu. The boxes with /dev/kfd — update and prod — are not fresh installs), so a
-read-only cross-check
-against the production box is the best available evidence — use it when the finding touches
-backend selection, profile flags, hardware probing, or native tool calling.
+install AND has /dev/kfd visible (the fresh boxes lack /dev/kfd, so their gpu-rocm LLM slots
+REFUSE to start — no ROCm compute node, no CPU fallback for a gpu-device slot). As of the rc.7
+Vulkan-restoration fold (#1948), that is no longer the whole story for the Vulkan LLM lane: a
+fresh box WITHOUT /dev/kfd (ct151) now DOES run gpu-vulkan LLM slots, gated on the runner image
+being at or after the signed :0822 pin rather than refused outright — so a fresh-install
+Vulkan-lane finding CAN be reproduced fleet-side, on ct151, and does not automatically need the
+production cross-check. gpu-rocm LLM slots remain fresh-install-untestable (still no fresh box
+with /dev/kfd) — use the production box or the update box (both have /dev/kfd but are not
+fresh installs) for anything that is specifically about the ROCm lane, backend selection,
+profile flags, hardware probing, or native tool calling on a kfd-visible box.
 
 Non-destructive repro only. If you load a slot, restore what you found. Set key="${c.key}".`,
   { label: `verify:${c.key}`, phase: 'Verify', schema: VERDICT, model: 'opus', effort: 'high' })))).filter(Boolean)
--- a/tests/release-validation/README.md
+++ b/tests/release-validation/README.md
@@ -130,6 +130,77 @@ known-issues, or regressions, and add a line to the changelog below. A report re
 
 ### Kit changelog
 
+* **8** (2026-08-21) — Vulkan-restoration coverage for the rc.7 fold (#1948), FINALIZED against
+  the landed fold chain: #1954 (kfd identity by runner uid + device gid, `e5e5925a`), #1973 (the
+  `gpu-vulkan` lane re-enable behind an explicit image allowlist, `3474ec03`), #1959 (repin to
+  `ghcr.io/hal0ai/hal0-combined:0822`, `8f23afba`). Composed with v6/v7's `image_status`
+  tri-state work rather than duplicated — the api.md/slots.md tri-state checks stay exactly as
+  v6/v7 wrote them; this version only adds the Vulkan-specific checks and resolves the five
+  `FINALIZE-AFTER-PHASE-D` markers a v6 draft (written in parallel with the lane re-enable PR)
+  had left open, against the actual merged mechanics rather than a guess:
+    1. **The gate is an explicit allowlist, not a version comparison.**
+       `VULKAN_CAPABLE_IMAGE_REFS` (`config/schema.py`) is a `frozenset` earned per-image by the
+       #1948 §3-C matrix — deliberately NOT "the pin or later", because the tags have no total
+       order (git shortrefs, date stamps, names) and Vulkan correctness was empirically NOT
+       monotonic in build recency (the bisect found an OLDER tree correct and a NEWER one
+       broken). `image_serves_vulkan_lane()` checks membership; `effective_runner_image()` looks
+       THROUGH a pending retag so a migration decision and a live-load decision can never
+       disagree; `vulkan_lane_serves()` composes the two and is the one predicate the
+       derivation ladders, the bench harness, the installer's shell-mirrored preflight, and the
+       updater's migration all share.
+    2. **Preflight refusal is `require_kfd_for_gpu_slot`'s `gpu-vulkan` branch**
+       (`providers/_gpu.py`), which on the `llama` runtime lane on an AMD host now delegates to
+       `_require_vulkan_lane_prerequisites`: gate 1 is the image allowlist above (refuses by
+       name, cites #1888, names `VULKAN_FIXED_IMAGE` as the fix), gate 2 is
+       `render_node_present()` for the runner identity (uid 0, the rootful slot container — not
+       whoever calls the function). `/dev/kfd` is NOT consulted on this path at all — Vulkan
+       needs no compute node, which is the whole point of the kfd-less-box restoration.
+    3. **The capabilities catalog OFFERS the lane without gating on image** — confirmed
+       `capabilities/catalog.py`'s llama.cpp fan-out: `gpu-vulkan` is offered whenever the host
+       advertises that backend (a render node exists), independent of which image is installed.
+       "The picker is an OFFER, not a promise" is the source's own comment; a slot created on
+       that row still has to clear the load-time image gate and then #1922's readiness probe.
+       api.md 6c rewritten to state this as fact, not a maybe.
+    4. **`ENV_ALLOW_VULKAN_FALLBACK` survives unchanged** and now downgrades TWO correctness
+       refusals to a warning — missing `/dev/kfd` on the ROCm path, and an unvalidated Vulkan
+       image on the restored path — never the render-node check, which is a passthrough fact no
+       env var can change. `boxes.toml` wording finalized accordingly.
+    5. **Seeds stay `gpu-rocm` by operator ruling.** `install/profile_derive.derive_device`
+       still derives ROCm first whenever `/dev/kfd` is usable, even though the §3-C matrix shows
+       Vulkan is FASTER on the reference hardware (+13.96% prefill, +20.45% decode — the
+       corrected direction from the old ade07ba-era "Vulkan ~-10%" expectation kit v5 carried;
+       drop that stale expectation everywhere it appears). ROCm is offered first in the picker
+       because it has the longer validation history and is what MTP/speculative decode is tuned
+       on. The documented way to run Vulkan deliberately is `hal0 slot edit <slot> --hardware
+       vulkan` — a manual opt-in, not a new default. Kit checks must not flag "seeded slot is
+       still gpu-rocm on a box that could run Vulkan" as a defect.
+
+  `boxes.toml`: `[boxes.ct152-cpu-fresh]` REMOVED — the box was destroyed 2026-08-21. There is
+  now no CPU-only fresh box in the fleet at all; the fleet-coverage note says so plainly instead
+  of the old "do not claim no CPU-only box remains" hedge, since that claim is now simply true.
+  ct151's note and the fleet-coverage note both drop their `FINALIZE-AFTER-PHASE-D` HTML
+  comments and state the real gate: an explicit `VULKAN_CAPABLE_IMAGE_REFS` allowlist, not a
+  version/tag comparison.
+  `lanes/stateful/slots.md` 8c/8d finalized: 8c names `_require_vulkan_lane_prerequisites`'s two
+  gates explicitly instead of hedging on the wording; 8d's repro now greps
+  `_require_vulkan_lane_prerequisites` and `image_serves_vulkan_lane` by name and expects the
+  `GpuPreflightError` naming both #1888 and `VULKAN_FIXED_IMAGE`.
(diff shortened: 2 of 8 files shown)


thinmintdev:
**Finalize-pass note (GA session):** PR #1968 (image_status unknown) also bumps `kit_version` 5→6 with its own kit edits (api.md check 6, slots.md check 1, the `image-status-wrong-podman-store` expect) and is armed to land first. When this draft finalizes after Phase D: rebase onto main, resolve the kit.toml collision by bumping to **v7**, merge the api.md/regressions wording with #1968's (they teach the same tri-state — compose, don't duplicate), and add a v7 changelog entry. The five FINALIZE-AFTER-PHASE-D markers remain the main work.

thinmintdev:
**Version-ladder update (supersedes my earlier note):** PR #1980 (the #1968 follow-up pair) takes `kit_version` 7 with its own brief edits. When this draft finalizes after Phase D: rebase and bump to **v8** (not v7), composing with both #1968's and #1980's api.md/slots.md/regressions wording.

thinmintdev:
**Fleet change for the finalize pass:** ct152 was destroyed 2026-08-21 ~08:45 UTC (operator-confirmed; pre-destroy sweep found no unique data, rcchat slot unused). Remove the `[boxes.ct152-cpu-fresh]` section from boxes.toml in this PR's rebase and drop any references to it in the fleet-coverage note (its 'CPU-only fresh box exists' caveat is now false — note the coverage change honestly: no CPU-only fresh box remains in the fleet).

thinmintdev:
## Review — kit v8 (`c55cb7e2`), against the merged fold chain

**Verdict: 0 blocking. 2 non-blocking, one of them worth fixing before a blind run.**

I reviewed #1954/#1973/#1959 end-to-end, so I checked this brief against the code by **execution**, not by reading it. Base is `8f23afba`, i.e. the whole chain is on `main`: `e5e5925a` (#1954) → `3474ec03` (#1973) → `8f23afba` (#1959). Every commit ref the kit cites is correct.

---

## 1. Marker resolutions — all five accurate against the merged mechanics

I re-ran the preflight paths against the installed tree rather than trusting the prose:

```
8d  no env         : REFUSED | cites 1888=True | names VULKAN_FIXED_IMAGE=True
8e  env=1          : ADMITTED (no refusal)
8e  env=1, no dri  : REFUSED | cites 1888=False   <- render-node message, no image content
```

| Marker | Kit claim | Verified |
|---|---|---|
| 1 | allowlist not ordering; `effective_runner_image` looks THROUGH a pending retag; `vulkan_lane_serves` is the one shared predicate | ✅ `VULKAN_CAPABLE_IMAGE_REFS = frozenset({VULKAN_FIXED_IMAGE})`, and the shared-predicate claim is exactly the B5 fix — five consumers route through it |
| 2 | `require_kfd_for_gpu_slot` → `_require_vulkan_lane_prerequisites`; image gate **before** render gate; `/dev/kfd` never consulted | ✅ all three, by execution. Forcing `kfd_status → KFD_NOT_OPENABLE` on that path yields the render-node message with zero kfd content |
| 3 | catalog offers `gpu-vulkan` independent of image; offer-not-promise | ✅ no image dependency in `_backend_variants`; "OFFER, not a promise" is the source's own wording |
| 4 | `ENV_ALLOW_VULKAN_FALLBACK` downgrades **TWO** refusals, never the render node | ✅ `opted_in` branches at `_gpu.py:645` (Vulkan image gate) and `:822` (ROCm kfd path); the render gate has no such branch — and the third run above proves it |
| 5 | seeds stay `gpu-rocm`; `hal0 slot edit <slot> --hardware vulkan`; perf inverted | ✅ `--hardware` exists (`slot_commands.py:716`); `+13.96%` / `+20.45%` match #1948's §3-C row exactly |

Supporting facts spot-checked: `DEFAULT_ROCMFPX_IMAGE = ghcr.io/hal0ai/hal0-combined:0822`; `ade07ba` **is** in `STALE_ROCMFPX_IMAGE_REFS` and **not** in `VULKAN_CAPABLE_IMAGE_REFS` (so 8d's "concrete example" is live post-#1959, not stale advice); the quadlet really writes `Image=` (`container.py:739`), so 8c's `systemctl cat hal0-slot@<slot> | grep '^Image='` works verbatim; the journal event name `gpu_slot_vulkan_lane_unvalidated_image_allowed` in 8e is character-exact.

**One correction to the review brief I was given, in the kit's favour.** The brief says "kit calls [`HAL0_ALLOW_VULKAN_FALLBACK`]'s presence a finding — verify that framing matches". The kit does **not** do that, and it is right not to. Marker 4 says the env "survives unchanged", and 8e uses it as a *tool* to defeat every earlier refusal so the #1922 net can be tested in isolation. That is the correct framing: it is a documented, tested, loudly-logged escape hatch, not a defect. Treating its presence as a finding would have been the mis-teach, and the kit avoided it.

---

## 2. Decisive checks — executable blind, with one honesty gap

**8c** — every named symbol resolves; both gates named in the right order; the `Image=` grep is runnable; requiring a *second, different prompt through the slot's own port* rather than trusting the gate's internal probe is the right instinct (a canary that the gate itself issued proves the gate ran, not that the lane works).

**8d** — accurate and runnable. The "fires BEFORE the render-node check, so a box with no render node still gets the image-specific message" claim is true and I confirmed it: with a stale image and no render node the refusal is the image one. The grep line resolves against the installed tree.

**8e — constructible at the preflight seam (proved above), but it is missing the honest-failure clause that 8b has.** This is the one thing I would change.

8b ends with: *"If no bad case can be forced on this box, say so explicitly and record it as a coverage gap rather than a pass."* 8e has no equivalent. It inherits 8d's `(or simulate … say so either way)` only implicitly via "as in 8d", and closes on reproducibility (*"Record the exact env/command combination used"*), which is a different obligation.

That matters because of how 8e fails when it cannot be constructed. Forcing the garbage lane end-to-end needs the `ade07ba` image actually pullable on the box — plausible, but not guaranteed on a validation box that has already been upgraded and pruned. If the agent cannot get a broken image, the slot loads on `:0822` and goes **`ready`** — and 8e's stated expectation is *"terminal `error` … NEVER `ready`"*. A blind agent then either records a **false FAIL** (the box is fine; the fixture wasn't built) or, worse, reasons its way to a PASS. For the check the kit itself calls *"THE single most important check of this release"* and *"if this one fails, nothing else in this lane matters"*, that ambiguity is the wrong place to leave it.

One sentence fixes it, mirroring 8b: *if a broken-image fixture cannot be constructed on this box, record it as a coverage gap — not a pass and not a failure — and say what blocked it.*

**upgrade.md 5b** — accurate, and I checked the load-bearing claim rather than assuming #1959 landed as reviewed. The merged sweep binds:
```
2882   key = "image_pin"
2883   if isinstance(raw.get("image_pin"), str):            holder = raw
2885   elif raw["slot"].get("image_pin") …                  holder = raw["slot"]
2887   elif isinstance(raw.get("image"), str):              holder, key = raw, "image"
2889   elif raw["slot"].get("image") …                      holder, key = raw["slot"], "image"
```
`image_pin` first, matching `_resolve_image_ref` — so 5b's *"the retag sweep binds in the same order after the B1 review's precedence fix"* is **true as merged**. Had that fix not landed, 5b would have taught a validator to assert the opposite of the code on the exact both-keys slot it singles out. Worth stating that I checked it, because it was a live review finding on #1959 and the kit is asserting its resolution.

**upgrade.md 5c** — numbers match #1948's row exactly, direction correctly inverted, the `--hardware vulkan` recipe is the real flag, and *"do not flag 'Vulkan is faster than ROCm' as a regression"* is the right explicit instruction for an agent carrying the old kit-v5 envelope.

---

## 3. Composition with #1968/#1980 — merged, not duplicated

The changelog states the tri-state work is composed rather than restated, and the diff bears it out: no `image_status` check is rewritten, and the speculative `6b` from the v6 draft is **dropped** with the reason given (v6/v7's check 6 plus `image-status-wrong-podman-store`'s expect clause already cover it; a second check would duplicate). Dropping your own draft check because a neighbouring PR landed the coverage is the right call and it is recorded rather than silently deleted.

`kit_version` 7 → 8 ✓, and the v8 changelog entry is specific — it enumerates all five resolved markers with the mechanics, not "Vulkan support added". Historical entries left untouched, correctly.

---

## 4. Fleet honesty — ct152 fully removed from live surfaces, history preserved

- `.claude/workflows/rc-validate.js` — `BOX_ROLES` is now exactly `ct151-cpu-fresh / ct163-cpu-fresh / ct150-update / ct105-prod`, with a comment recording the destruction. No stale id.
- `boxes.toml` — the `[boxes.ct152-cpu-fresh]` stanza is gone; remaining mentions are prose explaining the removal.
- `regressions.yaml` / `lanes/` — no `ct152` reference at all.
- Contract tests: `tests/release/test_workflow_contract.py` + `test_rc_validate_kit_contract.py` → **24 passed**.

Two `README.md` lines still read `ct152-cpu-fresh still exists` and `ct152 has no snapshot` — I checked, and both are inside the **kit-v4 and kit-v1/2 changelog entries**, i.e. an append-only record of what those versions did at the time. Not touched by this PR, and correctly so: rewriting a v4 entry to match today's fleet would falsify the history, exactly as rewriting `reports/v1.0.0-rc.5.md` would. **Not a finding.**

The coverage change is stated plainly (*"There is now no CPU-only fresh box in the fleet at all"*), replacing the old hedge — which is the honest form.

---

## 5. `regressions.yaml`

47 entries parse, **no duplicate ids**. Exactly two new: `updater-image-pin-retag-blind-spot` (issue **1959**) and `vulkan-lane-stale-image-preflight-refusal` (issue **1973**) — both real issue numbers, both carrying `summary` / `severity` / `tier` / `lane` / `repro` / `expect` per house style. `promote_ready` is absent on the 1973 entry, which is within convention (23 of 47 carry it).

Both `expect` clauses are falsifiable rather than aspirational — the retag one names the two specific inversions that constitute the regression (`image_pin` stale-and-untouched, or `image` rewritten while a stale `image_pin` survives), which is the shape that actually catches it.

### The one thing to fix before a blind run

`regressions.yaml`, `vulkan-lane-stale-image-preflight-refusal` repro:
```sh
hal0 slot edit <throwaway> --device gpu-vulkan --model <small chat model>
```
**`--device` is not a flag on `hal0 slot edit`.** The real option set is `--ctx-size --hardware --model --port --profile --provider`, and the value form is `--hardware vulkan` (a `SlotHardware` enum), not `--hardware gpu-vulkan`. The kit is self-consistent everywhere else — `slots.md:31`, `upgrade.md` 5c and README marker 5 all use `--hardware` — so this is one isolated slip in the only concrete command in a `severity: ga-blocker` entry's repro.

Non-blocking because it fails **loudly** (`No such option: --device`) rather than silently doing something wrong, and the surrounding comment already tells the agent to force `image_pin` by hand. But it is the exact class of thing this review was asked to catch, and it is a one-word fix:
```sh
hal0 slot edit <throwaway> --hardware vulkan --model <small chat model>
```
(`hal0 slot load` on the next line **does** exist — `slot_commands.py:288` — so the rest of the block is runnable.)

---

## 6. Verification

```
node --check .claude/workflows/rc-validate.js   → OK
tests/release                                   → 112 passed      (matches the author's number)
tests/release contract tests, isolated          → 24 passed
regressions.yaml                                → 47 entries, no duplicate ids
```
Scope is clean — 8 files, all under `tests/release-validation/` plus the workflow script. **No `CHANGELOG.md` hunk**, correct for a kit revision.

---

## Standing

**0 blocking.** Two non-blocking:

1. **`--device` → `--hardware vulkan`** in the new regression repro (above). Worth taking — one word, and it is the only executable line in a ga-blocker entry.
2. **8e should carry 8b's honest-failure clause.** For the check the kit calls the most important of the release, "I could not construct the fixture" must be a recordable third outcome, not something the agent has to invent a verdict for.

Neither changes what the kit teaches; both are about what a validator does when reality does not cooperate.

Worth saying: the five marker resolutions are the part I checked hardest, because a brief written against a PR-in-flight and finalized after the merge is where mis-teaching normally creeps in. All five describe the code that actually shipped — including two things that were **review findings on #1973 and #1959 and could easily have been documented in their pre-fix form** (the shared `vulkan_lane_serves` predicate, and the retag's `image_pin`-first precedence). The kit describes the fixed behaviour in both cases. That is the check I most expected to fail, and it held.

thinmintdev:
## Delta confirmation — `c55cb7e2..05dcfde0`

Both non-blocking findings taken, exactly as prescribed, with no scope creep: **2 files, +5 / −2.**

1. `regressions.yaml` — `--device gpu-vulkan` → `--hardware vulkan`. Correct on both halves: the real flag *and* the real enum value (`SlotHardware`, bare `vulkan`, not `gpu-vulkan`). The repro block is now runnable end-to-end — `hal0 slot edit … --hardware vulkan` then `hal0 slot load`, both real commands.
2. `slots.md` 8e — the honest-failure clause is in, and it does the better thing of naming 8b explicitly ("Same honest-failure rule as 8b") rather than restating it, so the two cannot drift. It also enumerates the two realistic blockers (no stale image in the store, override refused for another reason) and closes with "do not invent a verdict for a check you could not run" — which is the sentence that actually prevents the false-FAIL I was worried about.

Re-verified on `05dcfde0`:
```
node --check .claude/workflows/rc-validate.js  → OK
tests/release                                  → 112 passed
regressions.yaml                               → 47 entries, no duplicate ids
```

**Verdict unchanged and now with nothing outstanding: 0 blocking, 0 non-blocking.** Kit v8 is validation-ready from my side.

CI on this head: CodeQL, sunset, ui, γ-suite and the Analyze jobs green; `python (3.12)` still running (it is consistently the ~30-minute job here) — the delta is two prose/YAML lines with no Python surface, and `tests/release` passes locally on this exact head.

thinmintdev:
**CI follow-up on `05dcfde0`:** `python (3.12)` polled to conclusion — **pass**. All 8 checks green.

Verdict stands: **0 blocking, 0 outstanding.** Kit v8 validation-ready.

PR #1972, opened 2026-08-21
feat(packaging): tracked build recipe for the ROCmFPX runner image

### LINKED ISSUE #1970: runner: the :0822 MiniCPM5 parser port exists only in the built image — not reproducible from any tracked source
Flagged by the Vulkan-restoration session itself (2026-08-21) while delivering `hal0-combined:0822`: the MiniCPM5 pre-tokenizer/parser restoration lives only in the built image. It is not upstreamed, and not a tracked patch in any repo — so `:0822` cannot be rebuilt from a clean checkout.

## Why this is a release problem, not a nit

This is precisely the failure mode that made #1888 expensive: `ade07ba` was also a hand-build whose provenance wasn't reproducible, and the #1948 bisect had to reconstruct its lineage from scratch to isolate the defect. Pinning a second irreproducible image re-arms that trap for the next defect.

## What done looks like (either)

1. The parser port committed as a tracked patch series in the runner-image build recipe (CT130 pipeline), so a build from the recipe reproduces `:0822` from known source + refs; or
2. The port upstreamed (charlie12345/ROCmFPX or llama.cpp mainline as appropriate) and the recipe pinned to the upstream ref containing it.

Either way: the recipe reference + exact base refs recorded in image labels and in `tests/release-validation` provenance notes, so a future bisect starts from a known tree.

## Gate relevance

Should block making `:0822` the SIGNED release pin (GA ledger row 10) — candidate tags are free, but a default pin the project cannot rebuild is technical debt at the worst address. Does not block continued validation of the candidate tag.

thinmintdev:
Closes #1970. **Stacked on `feat/pin-combined-runner`** (#1959) — base is that branch, not `main`, because the recipe is coupled to the pin it describes.

> **Stack note (revision `0416c107`).** The base branch moved to `6c6a84d7`; this branch has been rebased onto it and force-pushed with `--force-with-lease`. Stack shape is unchanged. The original author session was retired; this revision was written by a second agent against the review at [#issuecomment-5365405959](https://github.com/Hal0ai/hal0/pull/1972#issuecomment-5365405959) and the status confirmation at [#issuecomment-5366723946](https://github.com/Hal0ai/hal0/pull/1972#issuecomment-5366723946).

## Why

The MiniCPM5 restoration in `:0822` lived only in the built image. `:0822` could not be rebuilt from a clean checkout, which is the same shape that made #1888 expensive: `ade07ba` was also a hand-build, and isolating the defect meant reconstructing its lineage from scratch. Pinning a second irreproducible image re-arms that trap for the next defect.

## What lands

`packaging/runner/rocmfpx/`:

| File | Purpose |
|---|---|
| `manifest.toml` | base image **+ digest**, source repo + **full** pinned sha, cmake flags, patch series with a `why` per entry |
| `patches/` | the three deltas against `charlie12345/ROCmFPX@0a59add89` |
| `build.sh` | executes the manifest; `--check` verifies the series applies and builds nothing |
| `entrypoint.sh` | the #1936 zero-device preflight, previously untracked |

The three patches: the `minicpm5` pre-tokenizer mapping; the MiniCPM5 XML tool-call parser (written in charlie's own idiom, including the explicit `grammar_triggers` this tree's PEG API needs since it does not auto-derive them from `trigger_rule`); and the glslc serialisation carried from kyuz0.

The built image carries its own provenance as OCI labels — source, revision, base digest, recipe revision, patch-series hash — so a box can answer "where did this come from" without the repo.

---

# Revision `0416c107` — what changed since `736f76a4`

## 🔴 Blocking 1 — base image pinned by a mutable tag

`[base]` now carries a digest, and `build.sh` pulls `${BASE}@${DIGEST}` in **both** stages (the tag is kept alongside for readability only).

```toml
digest = "sha256:4f5418c1b1e39e5ad9bbadfd2e8381c6b8a70e9b03f667179a7233c6f681462a"
```

**How it was resolved, and when.** `skopeo` was not available on the resolving host, so the digest came from ghcr.io's Registry v2 API directly at **2026-08-21 08:46 UTC** — a manifest `GET` with an anonymous pull token, reading the `Docker-Content-Digest` header. Cross-checked by sha256-ing the returned manifest bytes, which reproduces the value exactly. Single-arch schema2 manifest, not a multi-arch index. All of this is recorded in the manifest comment rather than only here.

**It is not a guess at what the tag points to today — it is the base `:0822` was actually built from.** The base manifest's **9 layer digests are an exact ordered prefix of `:0822`'s 13**; the 4 extra layers are this recipe's own dnf / `COPY` / `chmod` steps. That also answers the review's open question: the tag has **not** been re-pushed since `:0822` was built, so blocking-1 was preventive rather than already-concrete — and this pin records real lineage instead of freezing a drifted tag after the fact.

**Base derivation, stated as far as it is known.** The base is a hal0-namespace republication of the Kyuz0 [`amd-strix-halo-toolboxes`](https://github.com/kyuz0/amd-strix-halo-toolboxes) family, whose `rocm-*-rocmfp4-*` images build on `fedora-minimal:43` — corroborated (not proven) by the inherited `vendor=Fedora Project` / `version=43` labels still on `:0822`. What is genuinely **not** knowable from here is recorded as unknown rather than guessed: which upstream Kyuz0 commit produced this image, and whether hal0's copy carries local modifications. Tracking that derivation is the real fix and stays a follow-up; the digest is what makes the gap survivable meanwhile.

**dnf, deferred in writing** (`dnf_packages_pinned = false`, with the reasoning in the manifest). Pinning every package is out of scope for the rc.7 cut. The position is deliberate: the digest freezes the layer that matters — the whole ROCm toolchain, the Fedora base, and every library the binaries link against all live *inside* the pinned base. What the dnf lines add is the builder toolchain (does not ship) and three Vulkan runtime packages. That residue is the accepted risk.

**The guarding test no longer passes on a floating tag.** `test_the_recipe_records_its_base_image`'s registry-prefix + version-substring check is kept, and two real ones added: `test_the_base_image_is_pinned_by_digest_not_only_by_tag` (asserts `sha256:` + 64 hex chars) and `test_build_sh_actually_pulls_the_base_by_digest` (parses every `FROM` line and requires all of them to be `FROM ${BASE_REF}` — recording a digest the build then ignores would be worse than none).

## 🔴 Blocking 2 — #1970's second recording site

New file: **`tests/release-validation/provenance/runner-images.md`**. Deliberately a new file — #1971 is rebasing this area later, so nothing existing is touched beyond one added row in the kit README's layout table (without which a validator still has no thread to pull, which was the finding).

It records: which recipe builds the default pin, the manifest path, the exact refs + base digest, the patch series with rationale, the runtime-less `--check` command, what `--check` does **not** prove, the known gaps, and the validation matrix the pin moved on.

`Closes #1970` is kept, because both recording sites named in the issue's done-bar are now genuinely delivered — labels (already there) and the kit provenance note. Left explicitly out of scope and worth separate issues: **tracking the base image's own derivation**, and **pinning dnf packages**.

## 🟡 The four Codex P2s

**`--check` required a container runtime it never uses.** `RUNTIME` resolution moved below the `--check` exit and guarded with `|| true`. Confirmed the failure mode was worse than a bad message — under `set -e` the assignment takes the failed substitution's status:

```
$ bash -c 'PATH=/tmp/emptybin; set -euo pipefail; RUNTIME="${HAL0_CONTAINER_RUNTIME:-$(command -v docker || command -v podman)}"; echo reached-next-line'
EXIT=1          # "reached-next-line" never printed; exit 65 never reached
```

After the fix, the same shape reaches its own error properly (`no docker/podman on PATH`, `EXIT=65`), and `--check` runs to completion with no runtime at all:

```
$ python3 -c '<run build.sh --check with PATH = only bash/sh/env/git/python3/coreutils>'
RETURNCODE: 0
==> tag    ghcr.io/hal0ai/hal0-combined:0822
==> base   ghcr.io/hal0ai/amd-strix-halo-toolboxes:rocm-7.2.4-rocmfp4-server
==> base@  sha256:4f5418c1b1e39e5ad9bbadfd2e8381c6b8a70e9b03f667179a7233c6f681462a
==> source https://github.com/charlie12345/ROCmFPX.git @ 0a59add89b8cba06fb6a0baf25a253a4e45faa78
==> patches 3
==> checked out 0a59add89b8cba06fb6a0baf25a253a4e45faa78
==> applied 0001-vocab-minicpm5-pretokenizer-mapping.patch
==> applied 0002-chat-minicpm5-tool-call-parser.patch
==> applied 0003-vulkan-shaders-serialize-glslc.patch
==> --check: patch series applies cleanly against 0a59add89…; nothing built
```

**CI hook — this is now a standing guard, not a one-host result.** `test_check_mode_does_not_require_a_container_runtime` shells `build.sh --check` under exactly that PATH sandbox. CI runs `uv run pytest tests/`, so it is already wired in with no workflow change. The test is non-vacuous by construction: it asserts the script reaches its banner, and under the old ordering stdout was **empty**. A network failure downstream of the clone is tolerated (asserted to be a fetch failure specifically), and a clone that exceeds the timeout skips rather than flakes.

**Labels named mutable filenames.** Added `dev.hal0.recipe.revision` (the recipe's git sha, suffixed `-dirty` when built from edited files rather than silently stamping a clean sha), `dev.hal0.runner.patches.sha256` (a hash over the series in manifest order — survives renames and same-name edits), and the standard `org.opencontainers.image.base.name` / `.base.digest` so generic OCI tooling finds the base. Asserted by `test_the_image_labels_name_immutable_things`.

**SELinux relabel.** `-v "${SRC}:/src:z"` — the same requirement `config/paths.py` and the `selinux="z"` `Mount` flags in `providers/` document. `:z` not `:Z` because `$SRC` is reused across builds; no NFS caveat here since `$WORK` is always local scratch.

**The inert symlink loop — removed, not revived.** As the review traced, impact was low (bench uses the absolute path, `LD_LIBRARY_PATH` covers the skipped `ldconfig`) and removing the ABI-mismatching stock binaries is arguably the correct behaviour. So rather than fix dead code nothing needs, the state it *actually* produced is now the stated intent, with the history of the bug written down next to it. Answering the review's question directly: yes, "deliberately no `llama-*` on PATH; use `/opt/rocmfpx/bin`" is the position.

## 🟡 Smaller review points, also taken

* **`HEAD == REF` asserted** after checkout (exit 66 with both shas), not merely echoed.
* **No submodules recorded** (`submodules = false`) with the evidence, so the manifest's silence reads as a finding rather than an omission.
* **`--check`'s limits stated in the script itself**, right where it prints its success line: dnf resolution, cmake, the four targets and both container builds are all unexercised. A green `--check` is not a green build.
* **The heredoc-vs-tracked-Containerfile tradeoff written into the header**, including that its cost has already been paid once (the symlink bug) and what would justify inverting it.

## 📋 Final ask — `podman inspect` of the published `:0822` vs. the recipe's labels

Posted in full as a separate comment. Short version, and it is the unflattering answer: **`:0822` carries none of this recipe's labels.** Its config blob has only inherited Fedora base labels — no `org.opencontainers.image.revision`, no `dev.hal0.runner.*`. It was pushed `2026-08-20T12:45:54Z`, before the recipe existed.

So for `:0822` this recipe is **reconstructed provenance, not a build record**, and the ledger should say so. The digest-pinned recipe governs the **next** build, from which the labels answer the question directly. Both that statement and its supporting evidence (the layer-prefix match, and `:0822`'s `Env`/`Entrypoint` matching the recipe's runtime stage exactly) are written into the provenance note, not just into this PR.

## Verification

```
bash -n build.sh                                    clean
build.sh --check (runtime-less PATH)                EXIT=0, series applies
pytest tests/packaging                              15 passed  (was 11)
pytest tests/release tests/packaging                127 passed
ruff check / ruff format --check                    clean
make check-sunset                                   192 <= baseline 192
CHANGELOG.md                                        no hunk
```

## Scope

This is option 1 from the issue (tracked patch series). Option 2 — upstreaming the parser to `charlie12345/ROCmFPX` — remains worth doing and would let the recipe drop patch 0002 and simply move the pinned ref, but it depends on an external maintainer and cannot gate a cut. Worth filing as a follow-up rather than blocking on it, alongside the two named above (base derivation, dnf pinning).


thinmintdev:
## Independent adversarial review (`736f76a4`) — verdict: **comment, two blocking findings**

Second pair of eyes; I did not write this code. Stacked on #1959, which is **operator-merged**, so this one inherits that gate — no automerge from me either way.

**Preamble: the central claim holds, and I verified it the same way you did — by running it.** Not on your build host, on mine:

```
$ ./build.sh --check
==> tag    ghcr.io/hal0ai/hal0-combined:0822
==> base   ghcr.io/hal0ai/amd-strix-halo-toolboxes:rocm-7.2.4-rocmfp4-server
==> source https://github.com/charlie12345/ROCmFPX.git @ 0a59add89b8cba06fb6a0baf25a253a4e45faa78
==> patches 3
==> checked out 0a59add
==> applied 0001-vocab-minicpm5-pretokenizer-mapping.patch
==> applied 0002-chat-minicpm5-tool-call-parser.patch
==> applied 0003-vulkan-shaders-serialize-glslc.patch
==> --check: patch series applies cleanly against 0a59add89b8cba06fb6a0baf25a253a4e45faa78; nothing built
EXIT=0
```

Fresh clone, pinned sha reachable and checked out, all three patches apply. The "verified by running it" argument in the body is legitimate and both defects you describe are the kind only execution finds. This is the right shape of PR.

Everything below is what running it **did not** catch.

---

### What I verified clean ✅

* **Source ref is a genuine full 40-char sha** — `0a59add89b8cba06fb6a0baf25a253a4e45faa78`, 40 chars, hex. `test_the_source_ref_is_pinned_not_a_branch` is a real check (`len(ref) == 40` + hex charset + branch blocklist), not a regex that would pass an abbreviated sha.
* **No submodules — I resolved this, since it was the one input the manifest is silent on.** In the tree `--check` left behind: `.gitmodules` is **empty** and `git submodule status --recursive` returns nothing. `ggml` is vendored in-tree at that sha. So the source side really is fully pinned by one ref, and the manifest's silence is correct rather than an omission. Worth a one-line note in the manifest saying so, because the next reader will ask.
* **Fetch mechanics are right.** Clone-then-checkout (`build.sh:81-83`), and the script documents *why* `fetch --depth 1 <sha>` can't work (no `uploadpack.allowReachableSHA1InWant`). `set -euo pipefail` at line 16. `git apply --check` precedes the real `git apply`, so a bad patch fails before mutating the tree.
* **All three patches are real, substantive, tracked diffs** — well-formed `diff --git` + hunk headers, nothing pointing at an image. `0002` in particular is not a token: it carries the full `common_chat_params_init_minicpm5`, the PEG construction, the template-detection dispatch, and the explicit `grammar_triggers` block the manifest's `why` claims this tree needs. It matches its stated rationale.
* **Tag coupling is real and tested.** `manifest.toml:14` == `schema.py:1013` == `ghcr.io/hal0ai/hal0-combined:0822`, asserted by `test_the_manifest_tag_matches_the_shipped_pin` via a live `from hal0.config.schema import DEFAULT_ROCMFPX_IMAGE` — not a duplicated string literal. That coupling justifies the stack.
* **`test_no_untracked_patch_files`** is a set-equality both directions. Genuinely non-vacuous; it's the test that stops the series silently drifting from the manifest.
* **No `CHANGELOG.md` hunk.** ✅

---

## 🔴 Blocking 1 — the base image is pinned by a *mutable tag*, and that re-arms the exact trap this PR exists to disarm

```toml
# manifest.toml:20
image = "ghcr.io/hal0ai/amd-strix-halo-toolboxes:rocm-7.2.4-rocmfp4-server"
```

No digest, anywhere. `build.sh:98` and `:126` both `FROM ${BASE}`. Nothing stops that tag being re-pushed, and when it is, every future run of this recipe produces a different image from identical tracked inputs — silently, with the same manifest, the same patches, the same `--check` output.

#1970 asks for "**the recipe reference + exact base refs** recorded in image labels and in `tests/release-validation` provenance notes, so a future bisect starts from a known tree." A floating tag is not an exact base ref. If the next `:0822`-class defect turns out to live in the base layer rather than in the three patches, the bisect lands in exactly the position #1948 was in — reconstructing lineage from scratch — which is the sentence the issue opens with.

The test that should catch this doesn't:

```python
# test_the_recipe_records_its_base_image
assert base.startswith("ghcr.io/") or base.startswith("docker.io/")
assert MANIFEST["base"]["rocm_version"] in base
```

Registry prefix + substring. Passes happily on a floating tag. **Ask:** record `base.digest = "sha256:..."` alongside the human-readable tag, `FROM ${BASE}@${BASE_DIGEST}`, and assert the `sha256:` shape in that test.

Related and in the same class: **the base image's own derivation is untracked anywhere.** The manifest describes it as a Kyuz0 `amd-strix-halo-toolboxes` derivative but it is published under hal0's namespace, and there is no recipe in this repo for that derivation — no Kyuz0 upstream ref, no build steps. Pinning its digest is the cheap fix that makes that gap survivable; tracking the derivation is the real one, and is fair as a follow-up rather than a blocker here.

Also unpinned, same theme, lower severity: every `dnf install` in both stages (`build.sh:99-105` builder toolchain — `rocm-llvm hipcc cmake ninja glslc glslang ...`; `build.sh:127` runtime — `mesa-vulkan-drivers vulkan-loader vulkan-tools`). Zero version constraints. Whatever `dnf` resolves on build day is compiled into the artifact. I'd accept "we pin the base by digest and accept dnf drift inside it" as a deliberate, documented position — but it should be written down as a decision rather than left as an omission, because the digest pin is what makes it *mostly* moot and the tag pin is what makes it *not*.

## 🔴 Blocking 2 — `Closes #1970` but half of #1970's done-bar is not delivered

#1970's "what done looks like" names two recording sites:

> the recipe reference + exact base refs recorded **in image labels** and **in `tests/release-validation` provenance notes**

The labels are there (`build.sh:140-144`) — `org.opencontainers.image.source` / `.revision` both interpolate the full 40-char sha, plus `dev.hal0.runner.recipe` / `.base` / `.patches`. Good.

`tests/release-validation` has **nothing**. I searched the tree: no file under `tests/release-validation/` and no file under `docs/` mentions `packaging/runner`, `rocmfpx/manifest`, or `build.sh --check`. The kit (`kit.toml`, `regressions.yaml`, `known-issues.yaml`, `lanes/`) is untouched.

That matters more than a docs nit because #1970 is explicitly a **GA gate on the signed pin** ("Should block making `:0822` the SIGNED release pin, GA ledger row 10"). The gate is satisfied by the process being able to *find* the provenance, and right now a validator running the kit has no thread to pull. Note #1971 (`docs/rc-validate-kit-v6-vulkan`) is in flight and is the natural place. **Ask:** either add the provenance note here, or drop `Closes #1970` to `Refs #1970` and let whichever PR adds the kit note close it. Closing an issue on half its stated bar is the thing that makes a ledger untrustworthy later.

---

### 🟡 Nit — `--check` demands a container runtime it never uses (Codex P2, confirmed)

`build.sh:61` resolves `RUNTIME` and hard-exits 65 if neither docker nor podman is present. The `--check` early-exit is at **line 91**. So the advertised "validates the patch series, builds nothing" mode is unusable on a git+python CI runner or a laptop without a container runtime — which is most of where you'd want to run it. Worse under `set -e`: `RUNTIME="${...:-$(command -v docker || command -v podman)}"` — the assignment takes the substitution's status, so on such a host the script dies at line 61 *before* reaching your own intended error at line 62. Move the `RUNTIME` resolution below the `CHECK_ONLY` exit. This is also what would let `--check` run in CI, which would turn this recipe from "verified once on one host" into a standing guard — worth doing for that reason alone.

### 🟡 Nit — the generated Containerfile's symlink loop is inert (Codex P2, confirmed by rendering it)

`build.sh:130-133` writes `\\\$b` into an **unquoted** heredoc. Bash collapses `\\`→`\` and `\$`→`$`, so what actually lands in the Containerfile is:

```dockerfile
for b in /opt/rocmfpx/bin/llama-*; do \
  [ -f "\$b" ] && [ -x "\$b" ] && ln -sf "\$b" "/usr/local/bin/\$(basename "\$b")"; \
done && ldconfig || true
```

`sh` reads `"\$b"` as the **literal string** `$b`, so the loop tests a file named `$b`, creates no symlinks, and short-circuits before `ldconfig` — which `|| true` then masks. Meanwhile the `rm -f /usr/local/bin/llama-*` on the preceding line **does** execute. Net: the image ships with no `llama-*` on `PATH` at all.

Impact is genuinely low, and I checked rather than assumed: `bench/harness.py:166-171` invokes `/opt/rocmfpx/bin/llama-bench` by **absolute path** — and explicitly documents that the stock toolbox's `/usr/local/bin/llama-bench` "ABI-mismatches and segfaults on that image", so removing it is arguably correct. `LD_LIBRARY_PATH=/opt/rocmfpx/bin` covers the skipped `ldconfig`. So nothing hal0 does breaks. But an operator who execs into the container loses `llama-cli`/`llama-bench`/`llama-quantize` — the three targets the build goes out of its way to compile — and the Containerfile claims to do something it does not. Two characters. Fix it, and consider whether the intent is now "deliberately no `llama-*` on PATH", which would be worth stating rather than achieving by accident.

### 🟡 Nit — labels can't identify *which* recipe built the image (Codex P2, valid)

`dev.hal0.runner.patches` records filenames only, and there is no hal0 commit or recipe digest. Since the patch files are mutable in-tree under stable names, an older image cannot be traced back to the recipe revision that produced it — which is the label block's stated purpose ("so a box can answer 'where did this come from' without the repo"). Adding `dev.hal0.recipe.revision=$(git -C "$HERE" rev-parse HEAD)` and a sha256 over the patch series closes it. This is the same class as blocking-1: provenance that names a mutable thing isn't provenance.

### 🟡 Nits, briefly

* **No post-checkout `HEAD == REF` assertion** (`build.sh:83` only echoes the short sha for a human). Low risk — `git checkout <full-sha>` can't silently resolve elsewhere and would fail under `set -e` — but it's one line and it's the assertion this whole file is about.
* **No `org.opencontainers.image.base.name`** — base provenance is only under the non-standard `dev.hal0.runner.base`. Generic OCI tooling won't find it.
* **SELinux bind mount** (Codex P2, `build.sh:109`): unlabelled `-v "${SRC}:/src"` fails on an enforcing host; `store.py:274-278` documents `:z`/`:Z` as required. Valid, cheap.
* **No tracked Containerfile.** Both are heredoc-generated. Defensible (single source of truth in the manifest), but it means reviewers review bash interpolation instead of a diffable image definition — and blocking-2's escaping bug is precisely what that costs. Worth stating as a deliberate tradeoff in the manifest header.
* **`test_every_patch_explains_itself` asserts `len(why) > 80`** — length only. Passes for 80 characters of anything. Soft, but harmless given a human reads the manifest.
* **`--check` proves patches apply, not that the build succeeds.** The body is honest about this, but since it's the PR's headline evidence: everything downstream of line 91 — dnf resolution, cmake against an unpinned toolchain, the four build targets, the two container builds — is unexercised. Say so explicitly next to the `--check` transcript.

### Codex triage

Four inline P2s on `736f76a4`, **all four valid**, all confirmed above by execution or line-order rather than by agreement: the symlink escaping, the `--check`/runtime ordering, the missing immutable recipe revision in the labels, and the SELinux relabel. None are blocking on their own; the labels one reinforces blocking-1.

### CI

`sunset` ✅, `ui` ✅, `γ-suite (chromium)` ✅. **`python (3.12)` still pending** — it's the only job running the new `tests/packaging/`, so it needs to be green. Base branch is #1959, so merge order is fixed by that PR's operator gate; `mergeStateStatus` is `UNSTABLE` accordingly.

---

### What I could not verify

* **That the recipe reproduces `:0822` bit-for-bit.** I ran `--check`, not a build — no gfx1151 build host and a full HIP+Vulkan compile is out of scope for a review pass. What I *can* say is that every tracked input is now pinned **except the base image and its dnf layer**, which is exactly why blocking-1 is blocking: with the base pinned by digest, "the recipe reproduces the image" becomes a checkable claim; without it, it isn't one, no matter how careful the patch series is.
* **Whether the published `:0822` was in fact built from this recipe**, as opposed to the recipe being a faithful reconstruction after the fact. The labels would answer it for future builds; for the current image, a `podman inspect` of the pushed `:0822` against these labels would settle it and is worth pasting into the PR — if the shipped image lacks these labels, that's worth knowing before the ledger row cites the recipe.
* Whether `ghcr.io/hal0ai/amd-strix-halo-toolboxes:rocm-7.2.4-rocmfp4-server` has already been re-pushed since `:0822` was built (would make blocking-1 concrete rather than preventive).

### Summary

Genuinely good work and the right instinct — executing the tool caught two defects inspection wouldn't have, and reproducing `--check` on an independent host is the strongest evidence a review pass can produce. Two things before this closes #1970: **pin the base by digest** (the manifest currently pins the interesting 5% and floats the other 95%), and **either add the `tests/release-validation` provenance note or downgrade `Closes` to `Refs`**. The nits are all cheap; the `--check`-without-a-runtime one is worth doing because it unlocks running this in CI.

thinmintdev:
## Delta re-review requested — **there is no delta; head is unchanged at `736f76a4`**

I was asked to re-review a pushed revision of this PR. Checking before reviewing: the branch has not moved.

```
$ git ls-remote github refs/heads/feat/rocmfpx-build-recipe
736f76a46f23efd0d40727cc804fa878a12ec096

$ gh api repos/Hal0ai/hal0/pulls/1972/commits
736f76a46f  2026-08-21T04:45:58Z  feat(packaging): tracked build recipe for the ROCmFPX runner image
```

One commit on the branch, authored **04:45:58Z** — which *predates* my review comment at **05:00:58Z**. No `head_ref_force_pushed` event in the timeline, and the only post-review events are `referenced` / `cross-referenced` (links from elsewhere), not code. `git diff 736f76a4 <remote head>` is empty. Base is still `feat/pin-combined-runner` at `d8008586` — #1959 has not moved either, so the stack is intact.

So this is not a re-review, just a status confirmation. **Both blockers stand, verified unchanged at this sha:**

1. **Base image still tag-pinned.** `manifest.toml:20` → `ghcr.io/hal0ai/amd-strix-halo-toolboxes:rocm-7.2.4-rocmfp4-server`, no `digest` key anywhere; `build.sh:98` and `:126` are both still bare `FROM ${BASE}`. No `sha256:` in either file. The guarding test is unchanged. dnf installs remain unpinned (not addressed, and not yet consciously deferred in writing either — a stated deferral would be a fine answer).
2. **#1970's second recording site still empty.** Nothing under `tests/release-validation/` or `docs/` references the recipe, and the PR body still opens `Closes #1970`. Either the note lands or that becomes `Refs`.

The four Codex P2s are also all still open as written — `RUNTIME` still resolves at `build.sh:61`, ahead of the `--check` exit at `:92`, so `--check` still can't run on a runtime-less CI box. And I don't see a `podman inspect` of the published `:0822` against these labels anywhere on the PR yet; that one matters for the ledger, since it's what distinguishes "this recipe built the shipped image" from "this recipe reconstructs it."

### One thing that *did* change: CI is now fully green ✅

```
python (3.12)      pass  24m22s     ← was pending at my review
sunset             pass
ui                 pass
γ-suite (chromium) pass
```

`mergeStateStatus` is `CLEAN`. So the new `tests/packaging/` suite is confirmed green in CI, which closes the one open item from my review that wasn't a request for a change. No `CHANGELOG.md` hunk (re-checked).

Happy to turn this around quickly once a revision actually lands — ping me with the new sha.

thinmintdev:
## Final ask: the published `:0822` inspected against the labels this recipe writes

Requested in the review ("if the shipped image lacks these labels, that's worth knowing before the ledger row cites the recipe"). Done, and the answer is the unflattering one.

`skopeo` was not available on the inspecting host, and pulling a multi-GB ROCm image to read one JSON blob is not worth it, so this went at ghcr.io's Registry v2 API directly — anonymous pull token, `GET` the manifest, then `GET` the config blob it names. Same bytes `podman inspect` would show, one hop earlier.

### The manifest

```
$ curl -sI -H "Authorization: Bearer $TOK" \
    -H 'Accept: application/vnd.oci.image.manifest.v1+json, …' \
    https://ghcr.io/v2/hal0ai/hal0-combined/manifests/0822

HTTP/2 200
content-type: application/vnd.oci.image.manifest.v1+json
docker-content-digest: sha256:9b411fc21afc19d6e1e6562b9480726997ff6087f79c7b485287478f87ebd71d
```

`:0822` is `sha256:9b411fc21afc19d6e1e6562b9480726997ff6087f79c7b485287478f87ebd71d`, 13 layers, config `sha256:59886c9fde01…`.

### The labels on the shipped image

```json
{
  "io.buildah.version": "1.43.1",
  "license": "MIT",
  "name": "fedora-minimal",
  "org.opencontainers.image.license": "MIT",
  "org.opencontainers.image.licenses": "MIT",
  "org.opencontainers.image.name": "fedora-minimal",
  "org.opencontainers.image.title": "fedora-minimal",
  "org.opencontainers.image.url": "https://fedoraproject.org/",
  "org.opencontainers.image.vendor": "Fedora Project",
  "org.opencontainers.image.version": "43",
  "vendor": "Fedora Project",
  "version": "43"
}
```

```
created: 2026-08-20T12:45:54.546908189Z
```

**Every one of those is inherited from the Fedora base.** Not present: `org.opencontainers.image.source`, `org.opencontainers.image.revision`, `dev.hal0.runner.recipe`, `dev.hal0.runner.base`, `dev.hal0.runner.patches` — i.e. none of the label block at `build.sh`, in either its original form or this revision's.

### Stated plainly, for the ledger

**`:0822` predates this recipe. It was not built by it.** It was pushed `2026-08-20T12:45:54Z`; the recipe's first commit is `2026-08-21T04:45:58Z`. For `:0822`, this PR is **reconstructed provenance, not a build record**, and a ledger row citing it should carry that qualifier. The digest-pinned recipe governs the **next** build; from that build on, the labels answer "where did this come from" directly and the qualifier can be dropped.

That sentence is in `tests/release-validation/provenance/runner-images.md`, not only in this comment, so the honesty survives the PR.

### What the inspection *does* corroborate

The reconstruction is not merely asserted. Two independent checks, both re-runnable:

**1. The base digest is the one `:0822` was actually built on.** The base manifest's 9 layer digests are an **exact ordered prefix** of `:0822`'s 13:

```
base layers: 9   0822 layers: 13
base is prefix of 0822: True
extra layers in 0822: 4
```

The 4 extra layers are exactly what the recipe's runtime stage adds: `dnf install`, `COPY bin/`, `COPY entrypoint`, `chmod`+`rm`. This also settles the review's third open question — **`ghcr.io/hal0ai/amd-strix-halo-toolboxes:rocm-7.2.4-rocmfp4-server` has not been re-pushed since `:0822` was built.** Blocking-1 was preventive, not already-concrete. And it means the digest now in the manifest is real recorded lineage rather than a snapshot of a tag that may already have drifted.

**2. `:0822`'s runtime config matches what the recipe's runtime stage writes**, byte for byte on the parts the recipe controls:

```json
"Env": ["PATH=/usr/local/bin:/usr/bin", "container=oci",
        "LD_LIBRARY_PATH=/opt/rocmfpx/bin:/opt/rocm/lib",
        "HSA_OVERRIDE_GFX_VERSION=11.5.1",
        "GGML_HIP_ENABLE_UNIFIED_MEMORY=1"],
"Entrypoint": ["/opt/rocmfpx/hal0-runner-entrypoint.sh"]
```

Those are the three `ENV` values and the `ENTRYPOINT` from the recipe's second stage.

There is a third, smaller corroboration worth noting because it is evidence *against* the recipe having been faithful in one respect and *for* the fix in this revision: `PATH` is `/usr/local/bin:/usr/bin` and the image ships no `llama-*` on it — which is precisely the state the inert symlink heredoc produced. The shipped image and the (broken) script agree, because the script's brokenness is what shipped. This revision makes that the stated intent rather than an accident.

### What this still does not prove

That the recipe reproduces `:0822` bit-for-bit. That needs a gfx1151 build host and a full HIP+Vulkan compile, and it is not something a review pass or CI can settle. What can now be said, which could not be said at `736f76a4`: **every tracked input is pinned by a digest or a full sha**, so "the recipe reproduces the image" is at last a *checkable* claim rather than an unfalsifiable one. Someone with the build host can settle it; before the base was digest-pinned, they could not have.

PR #1980, opened 2026-08-21
fix(slots): close the last two unknown-image gaps from the #1968 review

thinmintdev:
Two P2s landed on #1968's final commit and did not make it in before that PR was merged. Both are the same two invariants #1968 established, unmet in one place each. Neither is a regression from it — they are places the new contract simply had not reached.

## 1. The timeout path was an undiagnosable `unknown`

#1968 taught the probe-timeout fallback to report `image_status: "unknown"` for a slot that *does* declare an image (rather than the old `not-configured`, which was a claim about config from a probe that read none). But it emitted only the generic `slot_view.container_probe_timeout` line.

That made #1968's own release-validation text false: the kit tells operators every `unknown` carries a reason-bearing journal line, and that an `unknown` with none is itself a finding. A legitimate timeout therefore looked like a defect through the documented workflow. The timeout path now emits `slot_view.image_probe_failed … reason=probe-timeout` alongside the existing line.

The converse is pinned too: a **profileless** timeout still reports `not-configured` and deliberately stays silent. That is a definitive answer, and emitting the unknown-diagnostic there would send an operator hunting a seam that is fine. One test per half.

## 2. The indeterminate chip was missing from an entire tier

`SlotImageUnknownChip` was inserted into `SlotScard` only, but `InferencePane` routes embed / rerank / stt / tts slots through `utilRows` → `MiniCard`, a different component. So the seam failure was invisible for every utility slot — which run the same container images, out of the same store, and can be seam-blind in exactly the same way. The tiers differ in density, not in what they are allowed to hide.

The chip now renders in both. The forced-mock seed carries one seam-blind slot per tier (`legacy` headline, `tts` utility) so the γ-suite exercises both card shapes rather than assuming they agree — which is precisely the assumption that produced this gap.

## Kit

`kit_version` 6 → 7, per the kit's versioning policy (any change to lane briefs). `readonly/api.md` check 6 now names both `reason=` values on `slot_view.image_probe_failed`. Landing before v6's first run, so no report was produced under the incomplete wording.

## Verification

```
pytest tests/slot_view tests/providers    751 passed
pytest tests/release                      112 passed
ruff check src tests                      All checks passed!
ruff format --check src tests             1164 files already formatted
ui: npx eslint .                          clean
ui: npx tsc --noEmit                      clean
ui: npx vitest run                        14 files, 135 tests passed
ui: npm run build                         clean
ui: npx playwright test (full γ-suite)    649 passed, 16 skipped
```

Full γ-suite rather than the affected spec alone, because this touches the shared forced-mock seed (`src/dash/data.jsx`) every γ spec renders from.

Refs #1939, #1968.

### DIFF
--- a/src/hal0/slot_view/__init__.py
+++ b/src/hal0/slot_view/__init__.py
@@ -792,6 +792,22 @@ async def container_enrichment(
             except TimeoutError:
                 log.warning("slot_view.container_probe_timeout slot=%s", name)
                 profile_name = str(cfg.get("profile") or "")
+                if profile_name:
+                    # This is the third way to reach image_status="unknown",
+                    # and it needs the same reason-bearing line as the other
+                    # two: the release-validation kit tells operators that
+                    # every `unknown` has one behind it and that an `unknown`
+                    # with none is itself a finding. A timeout is a legitimate
+                    # unknown, so it must be diagnosable through the documented
+                    # route rather than only via the generic timeout line above.
+                    log.warning(
+                        "slot_view.image_probe_failed slot=%s profile=%s reason=probe-timeout — "
+                        "the slot probe exceeded %ss before the image store was read; "
+                        "reporting image_status=unknown",
+                        name,
+                        profile_name,
+                        _PROBE_TIMEOUT_S,
+                    )
                 return name, {
                     "container_status": "stopped",
                     "container_health": False,
--- a/tests/release-validation/README.md
+++ b/tests/release-validation/README.md
@@ -130,6 +130,13 @@ known-issues, or regressions, and add a line to the changelog below. A report re
 
 ### Kit changelog
 
+* **7** (2026-08-21) — follow-up to 6, landed before its first run. `readonly/api.md` check 6
+  now names both `reason=` values on `slot_view.image_probe_failed` (`probe-error` when the
+  probe raised, `probe-timeout` when the slot probe blew its deadline first). v6 promised that
+  every `unknown` has a reason-bearing journal line behind it while the timeout path emitted
+  only the generic `container_probe_timeout`; the product side of that gap is fixed in the
+  same PR, so the promise now holds as written. No check added or removed.
+
 * **6** (2026-08-20) — `image_status` gains an `unknown` member (#1939), so the three briefs
   that adjudicate that field learn to tell its two failures apart. `missing` on a running slot
   is still regression `image-status-wrong-podman-store` — a confident wrong answer. `unknown`
@@ -142,9 +149,9 @@ known-issues, or regressions, and add a line to the changelog below. A report re
   three point at the `reason=` field on the `podman_ro.image_present_unanswered` journal line
   (`grant-denied` / `podman-failed` / `podman-absent` / `invalid-argument` / `seam-error`),
   which is where the wrapper's exit-code contract now surfaces, or at
-  `slot_view.image_probe_failed` for an unknown that never reached the seam. Every `unknown`
-  has one of those two lines behind it; one with neither is a finding. No lane, tier, or phase
-  change.
+  `slot_view.image_probe_failed` (`reason=probe-error` / `probe-timeout`) for an unknown that
+  never reached the seam. Every `unknown` has one of those two lines behind it; one with
+  neither is a finding. No lane, tier, or phase change.
 
 * **5** (2026-08-20) — pre-rc.7 brief hardening from the GA-plan handoff, applied before the
   kit's first v-4-era full run. Four new checks close verification gaps for fixes that would
--- a/tests/release-validation/kit.toml
+++ b/tests/release-validation/kit.toml
@@ -4,7 +4,7 @@
 # scripts cannot read the filesystem, so when you add a lane here you must also add it to the
 # LANES table in the script. The brief itself is only ever read by the agent that runs it.
 
-kit_version = 6
+kit_version = 7
 
 [defaults]
 # Model tier per phase. Rationale in README "Phases".
--- a/tests/release-validation/lanes/readonly/api.md
+++ b/tests/release-validation/lanes/readonly/api.md
@@ -48,8 +48,9 @@ Probe the REST surface from the workstation with curl against `$API` (from `CONT
    finding**. Follow it up rather than passing it: grep the journal for
    `podman_ro.image_present_unanswered`, which carries a `reason=` naming which case fired
    (`grant-denied` / `podman-failed` / `podman-absent` / `invalid-argument` / `seam-error`),
-   or for `slot_view.image_probe_failed` (`reason=probe-error`, carrying the exception) when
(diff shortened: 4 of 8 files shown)


thinmintdev:
## Follow-up review — **approve**, no blocking findings

Same reviewer as #1968. Fast pass, scoped to the two deltas; I did not relitigate the merged work.

Reviewed `3f1d3105` against merge-base `f90d986a`. 8 files, +82/−8, all on-topic.

### 1. Both fixes are the established invariants, one place each

**Timeout reason line.** The condition guarding the new `log.warning` is `if profile_name:` — byte-for-byte the same predicate that already selects `"unknown"` over `"not-configured"` two lines below. So the log fires exactly when the status is `unknown` and never otherwise; there is no way for the two to disagree, and no new classification was introduced. No `unknown → missing` or `unknown → present` conversion anywhere in the diff.

The converse half is the part I'd have asked for if it weren't here: `test_a_wedged_profileless_slot_logs_no_image_probe_failure` pins that a profileless timeout stays `not-configured` **and** stays silent. That's the right guard — `not-configured` is a definitive answer, and emitting the unknown-diagnostic there would send an operator hunting a healthy seam.

**MiniCard chip.** It's the *same* `SlotImageUnknownChip` component, imported from the same module, so neutrality is structural rather than re-implemented — there is no second copy of the colour decision to drift. The premise checks out too: `isUtil` is `{embedding, reranking, tts, transcription}`, those rows go to `utilRows → MiniCards → MiniCard`, and that component genuinely had no chip. Real gap, correctly located.

I also checked the tier sweep is now complete rather than merely one-better. Three components render slot cards, and all three now carry the chip: `SlotScard` (:519), `MiniCard` (:624), and the retired grid `SlotCard` in `slots.jsx` (:376). The `img` slot is `type: "image"`, which `isUtil` rejects, so it lands in `headlineRows` → `SlotScard` — covered, not a forgotten fourth tier.

### 2. Test coverage — red-first confirmed, and the negative is non-vacuous

Reverted `slot_view/__init__.py` to the merged #1968 state (`d37fa48c`) and re-ran:

```
test_a_wedged_profiled_slot_reports_unknown_not_not_configured   FAILED
  (only `slot_view.container_probe_timeout slot=s1` in caplog — no reason= line)
1 failed, 5 passed
```

The profileless guard passes on both sides, correctly — it's a guard, not a red-first.

The e2e negative is real, not accidentally green: `infer-slot-embed` is `type: "embedding"`, so it genuinely renders as a `MiniCard`, and it genuinely has no `image_status` key. `toHaveCount(0)` is asserting something. The spec also pins `toHaveClass(/\bmcard\b/)` before asserting the chip, which stops the test silently passing if `tts` ever migrated tiers.

### 3. Journal line matches the kit's grammar

```
slot_view.image_probe_failed slot=%s profile=%s reason=probe-timeout — …
```

Event name and `reason=` token both match what `readonly/api.md` check 6 now tells the operator to grep for. The brief's two named values (`probe-error`, `probe-timeout`) are exactly the two the code emits.

### 4. Kit v7 — accurate

Every claim in the changelog entry checks out: check 6 does name both values, no check was added or removed, the product-side fix is in the same PR, and `kit_version` is 7. "Landed before v6's first run" holds — `reports/` contains rc.4/rc.5/rc.6 only, nothing produced under v6.

One neutral observation: the **v6** entry was edited in place to add `probe-timeout` rather than being left standing and superseded by v7. Normally I'd flag rewriting a shipped changelog entry, but v6 never produced a report, so no artifact refers to the old wording — and leaving v6 promising something v6 didn't deliver would be worse. Fine as done.

### 5. Scope

Tight. Two fixes, their tests, the kit text they invalidate. Nothing unrelated. **No `CHANGELOG.md` hunk.** ✅

### 6. My #1968 N1 — the accounting is now *closed*, and this PR closes the part I missed

Worth stating precisely, since it's changed since my last review. N1 named three unlogged `unknown` paths. All three were fixed inside #1968's merged commit `d37fa48c` (after the head I reviewed):

- `slots/image_pull.py` — now `except Exception as exc:` + `slot_view.image_probe_failed … reason=probe-error`
- `container.py` no-runtime — now logs `reason=podman-absent`
- `container.py` exec failure — now logs `reason=seam-error`

**#1980 closes a fourth that neither Codex nor I flagged**: the timeout path. I checked N1 by walking the aggregator's `except` and the provider's returns, and never walked the timeout fallback, because #1968 had *just* changed it and I read it as already-handled. It wasn't. Good catch.

I re-enumerated every path that can yield `unknown` on this head. All seven now emit a reason-bearing line:

| Path | reason= |
|---|---|
| `image_present` seam unanswered | `<seam reason>` |
| `image_present` dev fallback, no runtime | `podman-absent` |
| `image_present` dev fallback, exec failed | `seam-error` |
| `image_present` dev fallback, rc > 1 | `podman-failed` |
| aggregator `except` | `probe-error` |
| aggregator timeout (profiled) | `probe-timeout` ← this PR |
| `inspect_image_state` `except` | `probe-error` |

The kit's "every `unknown` has one of those two lines behind it; one with neither is a finding" is now true as written, which it was not before this PR.

### Gates (mine, on `3f1d3105`)

```
pytest tests/slot_view tests/providers   751 passed
ruff check src tests                     All checks passed!
ruff format --check src tests            1164 files already formatted
ui: npx tsc --noEmit                     clean
ui: npx vitest run                       14 files, 135 tests passed
ui: npx eslint .                         clean
kit.toml → kit_version 7, regressions.yaml parses
```

CI on `3f1d3105`: all 8 pass, including `γ-suite (chromium)` 11m49s and `python (3.12)` 24m40s. The γ pass is the one that matters, since `data.jsx` is the shared seed every γ spec renders from and this PR adds a second seam-blind slot to it.

---

## Non-blocking nits

**N1 (cosmetic, inherited from #1968, not introduced here)** — the three `slot_view.image_probe_failed` emitters don't agree on their key set. The aggregator's two carry `slot=`; `inspect_image_state`'s carries `image=` only, and it's emitted from `hal0.slots.image_pull` under a `slot_view.*` event name. Grepping by event name — which is what the brief instructs — works fine. An operator filtering by slot won't match that one. A `slot=` would make the three uniform; not worth its own PR, worth folding into the next touch of that file.

**N2 (judgement, probably correct as-is)** — the chip is now on all three *card* tiers but not on the two list-row surfaces (`slots.jsx:469 SlotListRow`, `slot-list.jsx:99 SlotRow`). Those are deliberately dense and drop far more than this, so I think omitting it is right — flagging only so the "which surfaces show it" decision is on the record rather than implicit.

---

Approve. The diff does exactly what the body says, both fixes are the existing contract applied one place further rather than new semantics, and the tests fail without them. Land it.

**One for the changelog agent:** PR #1965 still needs a #1939 entry (it had none when I checked during the #1968 review). It can now cover #1968 + #1980 in one line, and should mention the kit went 5 → 7.

PR #1986, opened 2026-08-21
feat(packaging): tracked, digest-pinned build recipe for the ROCmFPX runner image

### LINKED ISSUE #1970: runner: the :0822 MiniCPM5 parser port exists only in the built image — not reproducible from any tracked source

thinmintdev:
Successor to **#1972**, which GitHub auto-closed when its stack base (`feat/pin-combined-runner`) was deleted by #1959's squash-merge — same branch, rebased onto main (the two recipe commits only; #1959's content is already in main as `8f23afba`). All review history lives on #1972; its final revision (`0416c107`) is content-identical to this head modulo the rebase.

**Closes #1970.** The runner-image build recipe is tracked in-repo with pinned sources:

- Base image pinned by DIGEST in both FROM stages (`sha256:4f5418c1…`), resolved via the ghcr Registry v2 API and cross-checked by hashing the manifest bytes — and verified to be the base the shipped `:0822` was actually built from (its 9 layer digests are an exact ordered prefix of :0822's 13; the tag has not been re-pushed).
- Second #1970 recording site delivered: `tests/release-validation/provenance/runner-images.md` + a kit README row.
- All four Codex P2s from #1972's review fixed (`--check` now runs runtime-less and is exercised IN CI by a pytest that shells it under a container-runtime-free PATH; labels carry recipe revision + patch-series sha + base digest; SELinux `:z`; the inert symlink loop removed honestly).
- **Provenance honesty, on the record:** the published `:0822` carries none of the recipe's labels and predates the recipe's first commit — it is reconstructed provenance, not a build record; this digest-pinned recipe governs the NEXT build. Stated in the provenance note itself.

Verification (from #1972's final revision, re-validated post-rebase in CI): `bash -n` clean, `build.sh --check` EXIT=0 without a container runtime, tests/packaging 15 passed, tests/release+packaging 127 passed, ruff + format clean, sunset 192≤192, no CHANGELOG hunk.

Inherits the chain's convention: reviewed-then-landed, no automerge.


thinmintdev:
## Final verdict on the recipe revision (`4c853cc2`) — **1 blocking: the red CodeQL check**

Independent verification, carrying over the two blockers I raised on #1972. **Both are resolved, and the headline claim reproduces exactly.** The one blocking item is a required check that is currently red — not a defect in the recipe.

---

### 1. Base digest-pin — verified, including the strong claim ✅

I re-derived the digest from scratch against ghcr's Registry v2 API with an anonymous pull token, both ways the manifest says it was obtained:

```
Docker-Content-Digest: sha256:4f5418c1b1e39e5ad9bbadfd2e8381c6b8a70e9b03f667179a7233c6f681462a
sha256sum(manifest bytes): sha256:4f5418c1b1e39e5ad9bbadfd2e8381c6b8a70e9b03f667179a7233c6f681462a
```

Both match `manifest.toml` exactly. Single-arch schema2, not an index — as documented.

**The lineage proof reproduces.** This was the claim worth attacking, because "the digest the shipped image was actually built from" is much stronger than "whatever the tag resolves to today":

```
base layers: 9   :0822 layers: 13   PREFIX MATCH: True
 0-8  identical, in order
 9-12 recipe-added (dnf / COPY / chmod)
```

All nine base layer digests are an exact ordered prefix of `:0822`'s thirteen. So `:0822` really was built on this exact base and the tag has not been re-pushed since. That converts the pin from a plausible guess into recorded lineage, which is the difference between the pin being useful to a future bisect and being decorative.

`FROM ${BASE_REF}` in **both** stages (`build.sh:133`, `:186`), with `BASE_REF="${BASE}@${BASE_DIGEST}"`. `test_build_sh_actually_pulls_the_base_by_digest` parses **every** `FROM` line and asserts each equals `FROM ${BASE_REF}` — so a third stage added later cannot quietly float. `test_the_base_image_is_pinned_by_digest_not_only_by_tag` asserts the `sha256:` prefix, 64 chars, and hex. That is a real digest-shape check, and it closes the exact gap the old prefix-only test left open.

I also re-ran `--check` on the digest-pinned script end to end: clean, exit 0, now echoing `==> base@ sha256:4f5418c1…`.

**dnf deferral** is written down as `dnf_packages_pinned = false` with the reasoning — and the reasoning is correct: the ROCm toolchain, the Fedora base, and every library the binaries link against live *inside* the pinned base, so the dnf lines add build-only tooling plus three Vulkan runtime packages. That is a defensible accepted risk, and a stated position rather than an omission is what I asked for.

**Base derivation** records the Kyuz0 family and fedora-minimal:43, and states the upstream commit as `"unknown — see the note above; deliberately not guessed"`. Stating an unknown as unknown is the right call.

### 2. #1970's second site — delivered, and `Closes` is honest ✅

`tests/release-validation/provenance/runner-images.md` (131 lines) plus a `provenance/*.md` row in the kit README. Against #1970's bar — "recipe reference + exact base refs recorded in image labels **and** in `tests/release-validation` provenance notes" — both sites now exist.

One nuance worth naming: strictly, what is proven is that the recipe reproduces `:0822`'s **inputs** (same base layer set, same source sha, patches apply cleanly), not a bit-identical rebuild. The note says exactly that rather than implying more. Since #1970's stated purpose is "so a future bisect starts from a known tree", **`Closes #1970` is honest.**

### 3. The four Codex P2s — all taken ✅

* **`--check` runtime-less:** `RUNTIME` moved to line 128, *after* the `--check` exit at 111, and now ends `|| true` so `set -e` cannot kill the script on a runtime-less host. And it is genuinely exercised: `test_check_mode_does_not_require_a_container_runtime` builds an inclusion-only PATH sandbox, asserts `docker`/`podman` are unreachable, and actually runs `build.sh --check`. I checked the skip paths specifically, because a test that silently no-ops in CI is worse than no test — the skips cover only missing coreutils and a clone timeout, and the assertions that matter (`"no docker/podman on PATH" not in output`, `returncode != 65`, banner reached, `==> base@ sha256:` printed) run **unconditionally**, before the network branch. It cannot pass vacuously on an offline runner. `tests/packaging` → **15 passed, 0 skipped** locally, so it ran for real here.
* **Labels:** now carry `org.opencontainers.image.base.name` + `.base.digest` (the standard keys I flagged as missing), plus `dev.hal0.recipe.revision` and `dev.hal0.runner.patches.sha256`. That closes "provenance that names a mutable thing isn't provenance".
* **SELinux:** `-v "${SRC}:/src:z"`.
* **Symlink loop:** removed, with the history and rationale kept as a comment — and the `rm -f` retained *deliberately*, because the stock `llama-*` ABI-mismatches these binaries. Turning an accidental behaviour into a documented intentional one is the right resolution.

### 4. Published-image audit — on the record in the note, and independently true ✅

The note states `:0822` is **reconstructed provenance, not a build record**, that it predates the recipe, and that the recipe governs the next build. That statement is in the note itself, not only in a comment — which is what the ledger reads.

I verified every factual claim it makes by pulling `:0822`'s config blob:

```
labels: fedora-minimal / vendor=Fedora Project / version=43 …
recipe-era dev.hal0 labels: NONE
created: 2026-08-20T12:45:54Z
```

No `dev.hal0.*` and no `org.opencontainers.image.revision` — confirming it was not built by this recipe. The inherited Fedora labels are present, which is exactly the corroboration the manifest cites for the fedora-minimal:43 lineage, correctly hedged there as "corroboration, not proof". The timestamp matches the note to the second. The `Env`/`Entrypoint` corroboration also checks out. **Every claim in the note is independently true**, including the ones that make the recipe look worse. That is the standard this document needed to meet.

### 5. Scope ✅

Two commits, recipe files only, no `CHANGELOG.md` hunk, `ruff check` clean, `ruff format --check` clean (1171 files), `check-sunset` 192 ≤ 192.

---

## 🔴 Blocking — CodeQL is red: 2 high-severity alerts

```
py/incomplete-url-substring-sanitization  [high]  ×2
tests/packaging/test_rocmfpx_recipe.py:87
  "The string ghcr.io/ may be at an arbitrary position in the sanitized URL."
  "The string docker.io/ may be at an arbitrary position in the sanitized URL."
```

Line 87 is `assert base.startswith("ghcr.io/") or base.startswith("docker.io/"), base`.

**On security substance this is not exploitable.** `MANIFEST["base"]["image"]` is read from a tracked file in this repo — not untrusted input, not network-derived — and this is a test asserting a repo invariant, not a runtime sanitiser. There is no attack surface here, and I want that stated plainly so the disposition is made on the real risk rather than on the severity badge.

**But it is still a merge blocker**, for two reasons: CodeQL is a check on this PR and it is failing, and a red required check needs a deliberate disposition rather than being carried into a release. Flagging rather than deciding, since dismissing a high-severity security alert is an operator call, not a reviewer's.

The cheap resolution also happens to *improve* the assertion — and note this is the same test whose prefix-only check failed to catch the tag-vs-digest gap in the first place. A real host check satisfies CodeQL legitimately instead of suppressing it:

```python
assert base.split("/")[0] in {"ghcr.io", "docker.io"}, base
```

I verified this keeps the invariant and is strictly stronger — it rejects everything the current form rejects (`https://evil.com/ghcr.io/x`, `ghcr.io.evil.com/x`, `quay.io/x`, `docker.io.evil/x`) while passing the real value. Now that `test_the_base_image_is_pinned_by_digest_not_only_by_tag` owns the digest check, this assertion's only remaining job is the registry host, so tightening it costs nothing.

Alternative if you'd rather not touch it: dismiss both alerts as "used in tests" with that rationale recorded. Either way the check needs to go green or be consciously dismissed before this lands.

### CI

`sunset` ✅ `ui` ✅ `γ-suite` ✅ (10m33s) `Analyze` ×3 ✅ · **`CodeQL` ❌** · `python (3.12)` still running at the time of writing (it takes ~25m on this repo). Locally on this head: `tests/packaging` 15 passed / 0 skipped, lint and sunset clean.

---

### Verdict

The substance is done, and done well. Both of my blockers are properly closed rather than papered over: the digest pin is backed by a lineage proof I re-derived independently, and the provenance note is honest about `:0822` in a way I was able to confirm against the registry line by line — including the parts that don't flatter the recipe. `Closes #1970` is earned.

**One thing stands between this and green: the CodeQL check.** Fix the assertion or dismiss the alerts deliberately, and I have no remaining objection.

PR #1987, opened 2026-08-21
feat(ui): memory v2 — redesigned memory surface with ranked units, curate, and history

thinmintdev:
## What

Full Memory v2 redesign from the approved design handoff (memory_v2): the Memory section becomes two sub-tabs — **Overview** (engine panel, growth chart with consolidation ticks, bank table) and **Bank** (unified workspace: ranked fact list, filter card with tag chips/when-brush/type toggles/state chip, sources panel, inspector with curation flows, ego focus view depth 1–10, animated web graph with lenses and a 120-node salience cap). Legacy Graph/Tools views and their five e2e specs are removed.

Backend: curate `PATCH` passthrough, ego depth clamp raised to 10, and a new composed `GET /api/memory/banks/{bank}/units` endpoint (upstream-forwarded q/type/state/document_id, hal0-side tags/time-window/multi-type OR/salience ranking over the cached graph slab, truthful `truncated`/`total_matched`, `salience: null` for out-of-slab rows).

## Why

Ships the operator-approved memory_v2 direction (design handoff 2026-08-21); replaces the three-tab v1 memory surface whose graph views and Tools tab it supersedes.

## How verified

- vitest 196/196 · Playwright e2e 665 passed / 0 failed (13 pre-existing non-memory skips) · pytest tests/api 1833 passed · `make lint` + `ruff format --check` clean — all at the pushed tip.
- Every task passed an independent spec+quality review; a final whole-branch review (opus) verified the rebase seams, contract alignment, deleted-global sweep, and spec completeness — verdict ready-to-merge after its two Important findings and N1 were fixed (`final-review.md` on the team board).
- Hindsight semantics (single-value `type` params, archive-on-invalidate, link-type set, history/curation constraints) verified against the hindsight-api 0.8.4 wheel by a dedicated reference agent.

## Outstanding before merge

1. **Manual visual pass** on the app-wide token-cascade change (`--dur-fast`/`--ease`/`--shadow-drawer` now take dashboard.css's canonical values — previously silently overridden by memory-overhaul.css import order). Un-automatable.
2. Live-backend smoke on CT105 once it's back up (box currently down; includes the `metadata.fact_id` aliasing check and the prod Hindsight version probe).

## Follow-ups filed (out of scope)

#1989 client multi-type comma-join footgun · #1993 ops-retry affordance · #1994 approval toast · #1995 mock graph wire-faithfulness · #1996 web-view tag-dim design call · #1997 truncated-notice e2e coverage. Deferred by design: server-side graph cap, entity mode, co-occurrence matrix.


thinmintdev:
## Independent adversarial review — release-gate rigor

Reviewed at `17e02378` (9 commits, 19 files, diff vs `main` @ `8f23afba`). I did not write this code. Comment-only; no approve/merge.

**Verdict: changes requested.** CI is red on the surface this PR touches, and there are three server-contract defects I reproduced locally. All fixes are small. The architecture is sound — the composed-endpoint design, the cache-sharing story, the auth posture and the depth-cap analysis all check out (details at the bottom).

---

### BLOCKING

**B1 — γ-suite is red because this PR grows a shared mock fixture that an existing spec pins.**
`ui/tests/e2e/specs/memory-graph-explorer-v3.spec.ts:66` asserts `expect(factCount).toBe(22)`. `mockFixtures.ts` appends `f23`–`f26` to the shared `MEM_FACTS` seed (comment: "so a unit lands inside the timeseries window"), so the same graph now renders 26 nodes.

```
2 failed
  memory-graph-explorer-v3.spec.ts:41 › direction A renders fact nodes and edges as SVG
  memory-graph-explorer-v3.spec.ts:61 › source toggle redraws as the entity co-occurrence graph
    Expected: 22
    Received: 26
646 passed (8.5m)
```

Deterministic — it failed on retry too. Fix: derive the expected count from the fixture (`MEM_FACTS.length`) rather than hardcoding it, or seed the new unit fixtures from their own fact list instead of mutating the shared one. Please don't just bump 22→26 without making the coupling explicit; the next fixture addition will do this again.

**B2 — `bank_units` silently truncates at the 2000-row slab, with no signal.**
`src/hal0/api/routes/memory_admin.py:377` pins `src_params["limit"] = "2000"`, then computes `total_matched`/`next_offset` over that slab. Reproduced with a 2000-row stub:

```
GET /api/memory/banks/shared/units?limit=5&offset=1995
→ 200 {"total_matched": 2000, "next_offset": null, keys=['items','next_offset','total_matched']}
```

On the production shared bank (thousands of units) the UI will render "2000 units" as though it were the total and stop paging, with no way to tell a full result from a clipped one. The sibling `bank_subgraph` gets this right — it emits `truncated` (line 313). The same cap applies to the graph slab, so under `sort=salience` a large bank is ranked against a partial graph with the unseen units silently pinned at `salience: 0.0`.

Fix: `"truncated": len(rows) >= 2000` (and the same for the graph fetch), surfaced in the UI. Two lines server-side.

**B3 — `limit=0` returns `next_offset == offset`; a client following `next_offset` loops forever.**

```
GET /api/memory/banks/shared/units?limit=0&offset=0
→ 200 {"items": [], "total_matched": 4, "next_offset": 0}
```

`limit=-1` is also accepted and silently drops the last row (`kept[0:-1]`). Fix: `limit = max(1, min(int(qp.get("limit", 20)), 200))`.

---

### MINOR

**M4 — `?limit=abc` is an unhandled `ValueError` → 500, not a 422.**
Verified: `ValueError: invalid literal for int() with base 10: 'abc'` escapes to `ServerErrorMiddleware`. The `int()` pattern is pre-existing in `bank_subgraph`, but this PR adds two more parse sites on a brand-new endpoint whose other invalid input (`sort=bogus`) correctly 422s with `memory.invalid_query`. A small `_int_param()` helper would make the endpoint's error contract consistent.

**M5 — Cache coherence: `_GRAPH_CACHE` has a 45s TTL and no invalidation path at all.**
Grepped: the cache is only ever `peek`/`put` — nothing anywhere clears it. That was tolerable while the surface was read-only; this PR adds the curate PATCH write in the same diff, so a **curate → units GET within 45s serves a stale `salience` / `link_counts_by_type`**. It is bounded and rank-only — text and `state` come from the uncached listing, so no stale content or stale archive state is served — but nothing in the response says so. Either clear `_GRAPH_CACHE` for the bank on the curate PATCH (needs lifting it out of `_FORWARDS`, or a targeted post-forward hook), or return a `salience_as_of` and state the bound in the docstring. Right now the only record is an inline `# (45s TTL)` comment.

**M6 — Under `state=invalidated`, every row reports `salience: 0.0` / `link_counts_by_type: {}`.**
The graph fetch carries only `type`/`q`, never `state`, and the default graph excludes invalidated units. The docstring calls out `state=invalidated` as exactly what the inspector's revert flow needs — so that flow will show a uniformly-zeroed salience column. Hide the column in that mode or document it.

**M7 — The curate PATCH is unaudited.**
`_DELETE_ACTIONS` puts a `record_action` row behind every destructive forward. `PATCH .../memories/{id}` with `state="invalidated"` is a soft-delete of a memory unit and lands with no audit row. The inline comment justifies this as reversible, which is fair — but the audit trail is what makes "reversible" actionable (who invalidated it, when). Suggest a `memory.memory.curate` action. Correctly *not* in `DESTRUCTIVE_MEMORY_ROUTES` / `CONFIRM_GUARDED_MEMORY_ROUTES` — no argument there.

**M8 — `memory-v2.css` leaks three global rules into the whole dashboard.**
It's imported unconditionally at `main.tsx:56`, after `dashboard.css:48`, so equal-specificity bare element selectors win app-wide:

- `:278` `html,body{height:auto;overflow:visible}` overrides `dashboard.css:125` `html,body{height:100%}` — **for every view**, not just Memory v2.
- `:264` `a{color:var(--accent)}a:hover{color:#FFC533}` recolors every anchor in the app; `#FFC533` is a hardcoded literal for the existing `--hal0-accent-hover` token.
- `:267-270` `.sb-sublist`/`.sb-subrow` and `:284` `.mem-tabs` redeclare classes that already exist in `memory-overhaul.css:17-20` and `dashboard.css:3761` and are consumed by the existing `dash/memory.jsx` — a fork of the shared sidebar/tab styles rather than an additive layer.

Scope these under `.mv-page`. The file's own header comment ("already `mv-*`/`mvi-*` namespaced… nothing needed stripping") is not true of these five rules and should be corrected with the fix. Everything else in the stylesheet does consume the existing tokens correctly — no token fork.

**M9 — Docs leak an operator-local path.**
`docs/superpowers/plans/2026-08-21-memory-v2.md:11`:

> `**Design source (read-only reference):** /mnt/mintdev/resources/hal0/memory_v2_handoff/`

This repo is public; that's a local workspace path and an unresolvable reference for every other reader. Please drop it. The rest of the file is clean — no IPs, hostnames, credentials, session ids, or transcript content, and it reads as a real implementation plan rather than session scratch.

Separately (not this PR's to fix): `docs/README.md` and `.gitignore:92` designate `docs/.devdocs/` — gitignored, local-only — as the home for "plans/specs from agent sessions (`superpowers/`)". This file lands in the tracked public tree instead. Three earlier plan docs are already there, so this is an established practice contradicting the stated policy. Worth an issue against the docs policy owner rather than a change here.

**M10 — Mock/server contract drift on `sort`.** `mockFixtures.ts` `buildBankUnits` accepts `sort=oldest`, which the server 422s. `BankUnitsParams.sort` is typed as bare `string`, so nothing catches it at compile time. Tighten to `'recency' | 'salience'` and drop `oldest` from the mock.

---

### NITS

- **N1** Salience tie-break is inverted relative to recency. `key=(-salience, ts)` with `reverse=False` breaks salience ties **oldest-first**, while `sort=recency` is newest-first; undated (`""`) rows also float to the *top* of a salience tie but correctly sink to the bottom under recency.
- **N2** `from`/`to` compare a mixed key space (`mentioned_at` ISO datetimes vs `date` day strings) lexicographically, so `from=2026-08-21T10:00` drops every `date`-only row for that day. `to` prefix-truncates (`_unit_ts(r)[:len(hi)]`) but `from` does not — the asymmetry is deliberate and correct for whole-day `to`, but undocumented.
- **N3** No test pins the default `sort=recency` ordering (the existing specs only assert counts and id sets under the default), nor the slab-sharing between `bank_units` and `bank_subgraph` that the PR description claims. I verified by reading that the key formats do match — `f"{bank}:{kind}:{type}:{q}"` with `kind` hardcoded to `"memories"`, same upstream path, same `limit=2000` — but nothing stops that drifting.

---

### Verified good (checked by reading, not taken on trust)

- **`degree_by_node` math genuinely matches `rank_by_degree`.** Both compute `sum(type_weight(lt) * w)` over the same undirected adjacency; `rank_by_degree`'s secondary key is exactly `degree_by_node(...)[1]`. Worth knowing, though: the units endpoint ranks on that weighted salience *alone*, while `rank_by_degree` ranks primarily on raw degree — so the graph explorer's "top by degree" and the list's "sort by salience" will legitimately disagree about which units matter most. Not a bug; a UX note.
- **No route shadowing.** `GET /banks/{bank_id}/units` is registered before the `_FORWARDS` loop, and every table entry at that path depth is a literal segment — nothing can capture `units`.
- **Auth posture is correct with no `exposure.py` change needed.** `exposure.py:260` classifies the entire `/api/memory` prefix ADMIN, so both new routes inherit it.
- **Depth 10 is safe — no cap needed.** `ego_bfs` early-returns the moment `len(reached) >= limit` (default 240, hard cap 500) and runs over the ≤2000-node cached slab, so BFS cost is bounded by the node cap, not by depth; there is no pathological-request OOM/hang path. `test_subgraph_ego_depth_honoured_to_ten` pins it. The old cap of 2 looks like a UI-era default rather than a cost guard.
- **Route tests pin real contracts**, not just happy path: tag filter is asserted *not* forwarded upstream, `document_id`/`state` *are*, the time-window undated-row drop is pinned both ways, salience order and the exact per-type link counts are pinned, `sort=bogus` → 422 with its error code, transport failure → 503. Gaps are N3 plus the B2/B3 edges above.
- **Scope is clean.** 19 files, all within the memory surface plus the plan doc. No `CHANGELOG.md` hunk. The rebase carried nothing extraneous.

### Local verification

```
tests/api/test_memory_units_route.py + test_memory_subgraph.py + test_memory_admin_routes.py
  → 56 passed
ruff check + ruff format --check on both touched Python files → clean
ui: npx tsc --noEmit → clean;  npx vitest run → 18 files / 172 tests passed;  npx eslint . → clean
```

CI at this head: `ui`, `sunset`, CodeQL ×3 pass; **γ-suite (chromium) fail** (B1); `python (3.12)` still running at time of writing.

The blockers are B1 (red CI, caused by this PR), B2 and B3 — all small. Happy to re-review on a push.

thinmintdev:
## Revision — addresses review comment #5368232818

Pushed `65987f4a` on top of `17e02378`. All three blockers and every minor taken are fixed and covered by new tests; the affected γ specs pass locally.

### Blockers

**B1 (γ-suite red)** — `mockFixtures.ts` now `export`s `MEM_FACTS`; `memory-graph-explorer-v3.spec.ts` imports it and asserts against `MEM_FACTS.length` instead of a hardcoded `22`. This kills the coupling class outright — the next fixture addition can't silently desync the spec again.
```
npx playwright test memory-graph-explorer-v3 memory-
→ 44 passed, 3 skipped (unrelated ADR-0023 extraction-panel specs)
```

**B2 (`bank_units` silent truncation)** — added `"truncated": len(rows) >= 2000` to the response (mirrors `bank_subgraph`'s existing field), plus a docstring caveat that `sort=salience` ranks against a partial graph slab when truncated. Covered by `test_units_truncated_false_under_the_2000_row_slab` / `test_units_truncated_true_at_the_2000_row_slab`.

**B3 (pagination clamp)** — `limit = max(1, min(_int_param(qp, "limit", 20), 200))`. Covered by `test_units_limit_zero_does_not_loop_forever`, `test_units_negative_limit_does_not_drop_last_row`, `test_units_limit_clamped_to_200_ceiling`.

### Minors taken

- **M4** — added `_int_param()`; applied to both `bank_units` (`limit`, `offset`) and `bank_subgraph` (`limit`, `depth`, `top_k`), so `?limit=abc` 422s with `memory.invalid_query` everywhere instead of escaping to a 500 on the new endpoint. Tests: `test_units_limit_abc_422_not_500`, `test_units_offset_abc_422_not_500`, `test_subgraph_limit_abc_422_not_500`.
- **M5** — added `GraphCache.clear_bank(bank_id)`; the curate PATCH handler calls it post-forward so a curate → units/subgraph GET inside the 45s TTL never serves the pre-mutation slab. Tests: `test_graph_cache_clear_bank_drops_only_that_banks_keys`, `test_curate_patch_clears_bank_graph_cache`.
- **M7** — curate PATCH now lands a `memory.memory.curate` `record_action` row (actor/target/outcome), same as every DELETE forward. Stays out of `DESTRUCTIVE_MEMORY_ROUTES`/`CONFIRM_GUARDED_MEMORY_ROUTES` — still reversible, no confirm gate. Test: `test_curate_patch_lands_audit_row`.
- **M8** — namespaced `memory-v2.css`'s three global leaks under `.mv-page`: `.sb-sublist`/`.sb-subrow` (unused so far, future-proofed) and `.mem-tabs` (was forking `dash/memory.jsx`'s existing class) are now `.mv-page`-scoped; the bare `a{}` rule is scoped and its hardcoded `#FFC533` swapped for the existing `--hal0-accent-hover` token. The `html,body{height:auto;overflow:visible}` rule is dropped outright — it's an ancestor selector of `.mv-page`, not a descendant, so it can't be namespaced, and `dashboard.css:125` already owns `html,body{height:100%}` app-wide. Corrected the file's stale "already namespaced" header comment to say what's actually still global (`body{background}`, restates `--bg`, harmless no-op) and why. (Caught and fixed a self-inflicted CSS-comment bug from this same edit along the way — an unescaped run of apostrophes broke the Tailwind/lightningcss parser and 500'd the dev server; γ specs are what caught it.)
- **M9** — dropped the `/mnt/mintdev/resources/hal0/memory_v2_handoff/` operator-local path from `docs/superpowers/plans/2026-08-21-memory-v2.md:11`.
- **M10** — `mockFixtures.ts`'s `buildBankUnits` no longer accepts `sort=oldest` (the server 422s on it); `BankUnitsParams.sort` is now typed `'recency' | 'salience'`. Tests: `mockFixtures.bankUnits.test.ts` additions.

### Not in scope (per review, noted here rather than fixed)

- **M6** (`state=invalidated` → uniformly-zeroed `salience`) — known, left for the inspector's revert-flow UI work.
- The `docs/.devdocs/` policy contradiction — tracked separately as an issue per the review's own suggestion, not this PR's to fix.

### Verification run

```
env -u FORCE_COLOR HAL0_HOME=$(mktemp -d) uv run --extra dev pytest tests/api -q
→ 1827 passed, 1 skipped

npx tsc --noEmit        → clean
npx vitest run          → 18 files / 174 tests passed (172 baseline + 2 new)
npx eslint .             → clean

ruff check + ruff format --check (src/hal0, tests/api) → clean

npx playwright test memory-graph-explorer-v3 memory-
→ 44 passed, 3 skipped (pre-existing ADR-0023 skips, unrelated)
```

No `CHANGELOG.md` hunk. 11 files touched, all within the memory surface + the plan doc + the two composed-endpoint test files.

thinmintdev:
**Coordination notice (memory-v2 team session, 2026-08-21):** an active multi-agent build is running on this branch — board at `/mnt/mintdev/artifacts/hal0-memory-v2-board/`. Commit 65987f4a unknowingly **reverted** the reviewed multi-type filtering fix (e2b7af5a, local at the time); a reconciling rebase that preserves 65987f4a's improvements (`_int_param`, `truncated`, limit clamp, `clear_bank`, audit row) AND restores the multi-type semantics is landing next, followed by ~7 reviewed UI commits (C2–C6: overview, bank bar, workspace, web graph, routing). **Please hold further pushes to this branch** until the board shows the merge complete — coordinate via the board's MESSAGES.md. This PR should stay open; do not merge yet (e2e migration C7/C8 + final whole-branch review pending).

thinmintdev:
**GA-session coordination — branch freeze for review.** Two workstreams pushed to this branch in the last hour: the assigned review-revision (`65987f4a` — all 3 blockers + 6 minors from comment 5368232818) and two further commits (`c6de524c`, `6eeb92fe` — truthful multi-type filtering + out-of-slab scoring, addressing the review's slab-ranking concerns beyond the assigned scope). To whoever is driving the second stream (memory-UI session, most likely): good additions by their titles — **please HOLD further pushes** until the delta re-review of the full head completes; a moving head burns review rounds. If you have more in flight, comment here instead and it gets folded in one pass. The re-review is being dispatched now; this PR lands in the rc.7 cut immediately after it clears.

thinmintdev:
## Delta re-review — head `6eeb92fe`

Re-reviewed the three commits since my pass at `17e02378`: the revision (`65987f4a`) plus the two concurrent commits (`c6de524c`, `6eeb92fe`), which had had no review. Comment-only, no approve/merge.

**Verdict: no blocking findings.** All three blockers are genuinely closed, every minor was addressed, and the two foreign commits are real improvements rather than drift — they fix a class of bug I only half-caught. Two minors and two nits remain, none of them gating.

---

### Blockers — all closed, verified on the current head

**B1 — γ-suite.** Fixed the right way: `FACT_COUNT = MEM_FACTS.length` imported from the fixture, with `MEM_FACTS` exported and the spec's NOTE reworded. Not a 22→26 bump, so the coupling can't silently desync again. **γ-suite (chromium) passes on this head (12m17s)** — confirmed against CI, not just locally.

**B2 — silent truncation.** `truncated = len(rows) >= 2000` returned in the envelope, mirroring `bank_subgraph`, plumbed through `BankUnitsPage.truncated` in TS and the mock. Both slab boundaries are now documented in the docstring. The graph-slab half is handled better than I asked for — see the foreign-commit section.

**B3 — pagination loop.** `limit = max(1, min(_int_param(qp, "limit", 20), 200))`, with `test_units_limit_zero_does_not_loop_forever`, `test_units_negative_limit_does_not_drop_last_row`, and `test_units_limit_clamped_to_200_ceiling` pinning all three edges I reproduced.

### Minors — all addressed

- **M4** `_int_param` 422s on garbage, applied to *both* routes (`bank_units` limit/offset, `bank_subgraph` limit/depth/top_k) with tests on each. Good — I only asked for the new endpoint.
- **M5** `GraphCache.clear_bank()` called post-forward on curate success, driven by a `_CACHE_CLEARING_FORWARDS` set rather than an inline special-case. Correctly placed after the forward, so a failed PATCH doesn't clear. Stale-slab test present.
- **M7** `record_action` with `memory.memory.curate`, wired through `_make_handler` without disturbing the `_DELETE_ACTIONS` path, and correctly still out of `DESTRUCTIVE_MEMORY_ROUTES`/`CONFIRM_GUARDED_MEMORY_ROUTES`.
- **M8** The `html,body{height:auto;overflow:visible}` override is **gone** (not scoped — correctly, since html/body can't be descendants of `.mv-page`), `a{}` is now `.mv-page a` using `var(--hal0-accent-hover)` instead of the `#FFC533` literal, and `.sb-sublist`/`.sb-subrow`/`.mem-tabs` are all `.mv-page`-prefixed. The false header comment is rewritten into an accurate one that explains the import-order hazard. Keeping unscoped `body{background:var(--bg)}` is fine — it restates what `dashboard.css:125` already sets.
- **M9** The `/mnt/mintdev/...` path is gone, replaced with a neutral reference. Doc is public-clean.
- **M10** `sort?: 'recency' | 'salience'`, and `sort=oldest` dropped from the mock.

---

### The two unreviewed commits — adversarial pass

**`c6de524c` "truthful multi-type filtering".** The real bug: upstream `/memories/list`'s `type` is single-value exact-equality, so `type=world,experience` matched nothing and returned an empty page with no error. Now a single type still forwards upstream verbatim; 2+ types deliberately forward *nothing*, pull the unfiltered slab, and OR hal0-side via `r.get("fact_type") in types`. Correct, and `truncated` is computed off the unfiltered slab so it stays honest in that path. Pinned by `test_units_single_type_still_forwarded_upstream` and `test_units_multi_type_filters_hal0_side_not_forwarded`. **Does not break the contract my earlier tests pinned** — single-type forwarding, tags-stay-local, and `state`/`document_id` forwarding all still hold.

**`6eeb92fe` "unfiltered salience slab + truthful out-of-slab scoring".** This addresses my B2 partial-graph concern, and the reasoning goes further than mine did. The graph fetch is now unconditionally unfiltered, for a second reason I had missed and which is the more important one: upstream only returns an edge when *both* endpoints are in the filtered node set, so passing even a single valid `type` would drop every cross-type edge and understate every node's degree. `test_units_salience_sees_cross_type_edges_despite_type_filter` pins exactly that.

Checked the three things that could have gone wrong here:

- **The slab kept its bound.** `params={"limit": "2000"}` — still capped, just no longer filtered.
- **The shared-slab property survives the cache-key change.** The key is now constant per bank (`f"{bank_id}:memories::"`), which is byte-identical to `bank_subgraph`'s key when it fetches unfiltered (`f"{bank}:{kind}:{type}:{q}"` with kind `memories` and empty type/q) — and both now fetch with the same `limit=2000` params, so there's no risk of one poisoning the other with a differently-shaped graph. Verified empirically, not just by reading: a `/units` GET followed by a `/graph/subgraph` GET issues **one** upstream `/graph` fetch, not two.
- **"Truthful" is actually truthful.** A row outside the 2000-node graph slab reads back `salience: null` + `link_counts_by_type: {}`, distinct from a row that *is* in the slab with no edges (genuine `salience: 0.0`). Under `sort=salience` the unscored rows sort after every scored row rather than being interleaved at a false zero. That's the right call — `null` for unknown, `0.0` for known-zero.

The +127 test lines pin real contracts, not smoke: invalid type → 422, the graph fetch is never type/q-filtered, cross-type edges survive a type filter, out-of-slab rows get null and sort last, truncated true *and* false at the slab boundary, and 422-not-500 on both int params.

**Composition (your Q4).** The three layers compose cleanly. The revision's `truncated`/clamp logic sits before the filter; the foreign rework replaces the scoring loop and the sort block below it. The one place they genuinely interact — `truncated` describing the listing slab while the *graph* slab is now described by `salience: null` instead — is handled deliberately and spelled out in the docstring, rather than one flag being overloaded to mean both. No lost hunks, no double-applied logic.

---

### Remaining — non-blocking

**M11 (minor, new, from `c6de524c`) — the hardcoded `_VALID_FACT_TYPES` allowlist is a drift trap, and it isn't needed for correctness.**
`?type=episodic` now returns **422 `memory.invalid_query`** (reproduced), where before it forwarded verbatim and let Hindsight decide. The multi-type OR works on whatever string values come back — `r.get("fact_type") in types` needs no allowlist at all. So the allowlist buys typo-catching and costs a hard failure the day Hindsight adds a fourth fact type. This PR also contradicts itself on the point: `useHindsight.ts:620` types the field `'world' | 'experience' | 'observation' | string` — deliberately open — while the server now closes it. `hindsight_provider.py:272`'s `_DEFAULT_RECALL_TYPES` uses the same three, but that's a default *selection*, not a proven exhaustive enum. Suggest dropping the allowlist (keep the multi-type OR), or at minimum an issue tracking it against the Hindsight version. Not gating — the UI only ever sends those three.

**M12 (minor) — the docstring's first paragraph is now stale.** It still says ``q``/``type``/``state``/``document_id`` "forward upstream verbatim", which the second paragraph and the code both contradict for `type` (validated always, dropped entirely at 2+ values). Worth a one-line fix on a commit whose theme is truthfulness.

**N4 (nit) — `clear_bank` over-clears sibling banks whose id is a prefix.** `_BANK_RE` permits colons, so `clear_bank("global")` also drops `global:hal0`'s slab (verified: both come back `None`). Harmless in both directions — it only ever over-clears, never under-clears, so the worst case is a cache miss and a refetch. Fine to leave; a `split(":", 1)`-aware key or a stored bank tag would tighten it.

**N1 (nit, carried) — the salience tie-break is still inverted, and now inconsistently so within one sort.** `scored.sort(key=lambda r: (-r["salience"], _unit_ts(r)))` breaks ties **oldest-first**, while the `unscored` list right beside it sorts `reverse=True` (newest-first), as does `sort=recency`. Two adjacent blocks in the same code path disagree about what a tie means.

---

### Verification on `6eeb92fe`

```
tests/api + tests/security  →  2008 passed, 1 skipped (12m44s)
ruff check src/hal0/api/routes/ → clean;  ruff format --check → clean
ui: npx tsc --noEmit → clean;  npx vitest run → 18 files / 174 passed;  npx eslint . → clean
```

CI: **γ-suite (chromium) pass**, ui pass, sunset pass, CodeQL + all three Analyze jobs pass. `python (3.12)` was still in progress when I posted — my full local `tests/api`+`tests/security` run above is the same surface and is green.

Scope re-checked: 11 files in the delta, all within the memory surface plus the plan doc and the one e2e spec that had to change. **No `CHANGELOG.md` hunk.** No Codex bot comments on the PR.

Nothing here blocks. M11 is the one I'd most like to see picked up, as a follow-up issue if not in this PR.

thinmintdev:
Post-live-smoke wave landed (tip `bbb98d70`): reflect body `query` fix (v1 was silently broken against hindsight-api 0.8.4 too), `caused_by` link type weighted/colored as causal on both backend salience and UI (live wire emits `caused_by`, not `causal`), MvError copy branched by status (404 deploy-skew vs 422 vs real unreachable), tag chip row capped at 24 with expander, count bar scoped to list view. Gates: vitest 200/200 · e2e 666/0 · pytest 1835 · lint clean. Note for reviewers: the `/units` 404 seen on CT105 is deploy skew (box runs rc.5 backend with the new UI rsynced) — resolved by a full deploy after merge.

thinmintdev:
Visual-polish wave from operator screenshot feedback (tip `2f35c054`): bare-UUID tags filtered from chip row and tag map (prefixed tags kept; display-only — queries unaffected), atlas bubbles only labeled above a legibility threshold (hover keeps tag+count), bank-row type legend unwrapped, and engine-panel graph stats now render exactly when extraction is off (the state where the ADR-0023 panel hides its copy — no duplication on, no blindness off). Gates: vitest 204/204 · e2e 666/0 · build/lint clean.

PR #2039, opened 2026-08-24
fix(slots): fail-fast doomed model loads via entrypoint exit 64

### LINKED ISSUE #2037: slots: llama-server exits 1 for everything, so RestartPreventExitStatus= cannot fail-fast a doomed load
Follow-up from #2012 / #2015.

## Problem

With the restart backoff merged in #2012, a slot whose model is genuinely doomed (corrupt GGUF, invalid CLI flag) still burns ~10 restart attempts over ~20 minutes before systemd gives up. The natural fail-fast directive exists — `RestartPreventExitStatus=` — but it is unusable today:

```
corrupt GGUF   → llama-server exit 1
bogus CLI flag → llama-server exit 1
```

Tested directly on the live reference box: llama-server returns exit code 1 for every failure class, so systemd cannot discriminate "doomed, never retry" from "transient, retry with backoff".

## Proposal

1. Teach the container entrypoint to translate load-phase failures into a distinct exit code (e.g. `exit 64` when llama-server dies before its health endpoint ever answered, or when the GGUF fails to parse). The entrypoint owns the child and can distinguish "died during model load" from "died while serving".
2. Emit `RestartPreventExitStatus=64` in the Quadlet template (`providers/container.py`), alongside the existing backoff keys.
3. The `hal0-systemctl` wrapper's `[Service]` allowlist needs the new key added (uint charset, same as `RestartSec=`).

Result: a corrupt model fails immediately and loudly (unit lands in `failed` with a distinct status visible in `systemctl status` and the #1791 failure-reason probe), while transient causes keep the full backoff runway.

## Notes

- Requires #2012's `LogDriver=passthrough` (merged) — before it, the entrypoint's own diagnostics were invisible.
- The crash-loop breaker (manager-side) is unaffected; this only short-circuits the systemd retry layer.

thinmintdev:
Closes #2037. Follow-up to #2012/#2015.

## Why

llama-server exits 1 for every failure class — corrupt GGUF and bogus CLI flag look identical to systemd — so `RestartPreventExitStatus=` was unusable and a genuinely doomed model burned the whole #2012 backoff ramp (~10 restarts over ~20 minutes) before parking.

## What

Three layers, one repo (the entrypoint source lives in `packaging/runner/rocmfpx/`, not the image repo):

1. **Entrypoint** (`packaging/runner/rocmfpx/entrypoint.sh`): now supervises llama-server instead of `exec`-ing it, polling `/health` on the slot's `--port`. "Loaded" = `/health` answered 200 at least once (during load llama-server either isn't listening or answers 503, so a bare TCP accept is not enough). A **normal (non-signal) death before that point** is translated to **exit 64**; signal deaths (OOM kill, GPU reset — possibly transient) and post-load crashes propagate untouched and keep the backoff runway. The #1936 GPU preflight (exit 78) is unchanged. Test seams: `HAL0_RUNNER_SERVER`, `HAL0_RUNNER_DEV_ROOT` (unset in the shipped image).
2. **Template** (`src/hal0/providers/container.py`): `[Service]` gains `RestartPreventExitStatus=64` next to the backoff keys.
3. **Wrapper** (`installer/wrappers/hal0-systemctl`): the `[Service]` allow-list accepts the new key, pinned to a single uint (`_QRE_UINT`) — signal names / lists / ranges systemd would accept are refused; both comment inventories updated.

## Version skew

Safe in both directions: an old image never exits 64, so the directive is inert; a new image under an old template just keeps today's backoff behaviour. The template change reaches live units on the next slot load/restart re-render.

## Tests (written red-first)

- `tests/installer/test_runner_entrypoint.py` (new): runs the real script with a fake server — death-before-health → 64 (with diagnostic), clean exit 0 not translated, SIGKILL during load → 137 untouched, crash after health → real code (7), preflight still 78.
- `tests/providers/test_container.py`: template emits `RestartPreventExitStatus=64` in `[Service]`.
- `tests/installer/test_systemctl_quadlet_validation.py`: renderer output accepted byte-for-byte (existing parametrised suite now covers the new key); `RestartPreventExitStatus=SIGKILL` and `=64 78` refused.

## Verification

- Full suite: 11081 passed, 18 skipped, 1 xfailed.
- `ruff check` / `ruff format --check` clean.
- mypy: zero new errors vs main (only a pre-existing sentry_sdk env diff and a line-number shift).

### DIFF
--- a/installer/wrappers/hal0-systemctl
+++ b/installer/wrappers/hal0-systemctl
@@ -277,7 +277,8 @@ validate_dropin_body() {  # arg: kind (gateway|hindsight); body on stdin
 #               HealthCmd= HealthStartPeriod= HealthInterval= HealthRetries=
 #               HealthTimeout= Exec=
 #   [Service]   Restart= RestartSec= RestartSteps= RestartMaxDelaySec=
-#               SyslogIdentifier= StandardOutput= StandardError=
+#               RestartPreventExitStatus= SyslogIdentifier= StandardOutput=
+#               StandardError=
 #   [Install]   WantedBy=hal0.target        (omitted when autoload = false)
 #
 # Sections may appear at most once and only in that order; [Container] is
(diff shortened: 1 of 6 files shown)


thinmintdev:
Review pass applied (independent reviewer; Codex bot rate-limited): added `--port=NNNN` single-token parsing (operator-edited profile flags may use it; the template always renders the two-token form) with a regression test. Reviewer's stale-listener note assessed: a stale 200 on the port sets ready early, which *skips* the 64 translation — the failure direction is "falls back to today's backoff", never a false fail-fast. `sleep 0.5` is GNU coreutils, which the fedora-minimal base ships.

Issue #2015, opened 2026-08-22
Slot units emit zero logs, and a crash-looped slot stays down until a human intervenes


thinmintdev:
Two defects in the container slot lifecycle. They are separable, but (1) should land first — while it holds, (2) is undebuggable.

Found while diagnosing a real outage on hal0 (2026-08-22): `hal0-slot@1` crash-looped from ~04:15 and `hal0-slot@15` was taken down as collateral. Recovering required `systemctl reset-failed` by hand, and the root cause was only obtainable by re-running the unit's `ExecStart` manually.

---

## 1. Slot containers produce no logs at all

**Severity:** high — every inference slot is unobservable.

### Evidence

24h of journal, all four running slots:

| unit | total journal lines | actual container output lines |
|---|---|---|
| `hal0-slot@1` | 337 | **0** |
| `hal0-slot@2` | 65 | **0** |
| `hal0-slot@15` | 98 | **0** |
| `hal0-slot@stt` | 8 | **0** |

The only non-systemd/non-podman lines are the 64-hex container IDs that `podman run -d` echoes. Directly:

```
$ podman logs hal0-slot-2
Error: this container is using the 'none' log driver, cannot read logs: this container is not logging output
```

`hal0-slot@2` had been healthy for 20 hours and logged nothing. This is not specific to failure — it is the steady state.

### Cause

`providers/container.py:744` emits Quadlet `LogDriver=none`. The comment at :741 reasons:

> `LogDriver=none` keeps conmon→journal the single sink so `journalctl -u` isn't double-fed (podman's own journald driver would tag a second copy — the old B3 note).

The double-logging it fixes (#721) was real. But the premise — that the unit already routes the container's stdout to journald via `StandardOutput=journal` — does not hold under Quadlet. The generated service runs podman **detached** (`-d`, `--sdnotify=conmon`); the podman client prints the container ID and exits, while conmon holds the real streams. `StandardOutput=journal` therefore captures only the client's trivial output. With podman's own sink disabled, the container's stdout/stderr has **no sink at all**.

The wrong copy was removed.

### Fix

```diff
-        "LogDriver=none",
+        "LogDriver=passthrough",
```

`passthrough` hands the container's standard streams to the unit, which systemd journals exactly once — one copy, and output that actually exists. Confirmed supported: podman 4.9.3 on hal0 (`file, journald, none and passthrough`).

Please verify no double-logging returns on the 4.x generator path before closing.

---

[...]

PR #2040, opened 2026-08-24
feat(ui): render the crash-loop breaker on slot cards and the slot drawer

### LINKED ISSUE #2038: ui: render the crash-loop breaker state (metadata.breaker) on slot cards
Follow-up from #2012.

## Problem

Since #2012, slot snapshots carry a live crash-loop breaker view in metadata:

```json
"breaker": {"state": "backoff" | "parked" | "half-open", "failures": 6, "retry_after_s": 512}
```

The API serves it (`/api/slots`, single-slot status), but nothing renders it. An operator looking at the dashboard still sees only a red ERROR chip with no hint that the slot is deliberately refusing loads, how many consecutive failures it has accumulated, or when the next automatic trial will run. Invisibility of exactly this state is how the 2026-08-22 outage went unnoticed for four hours (#2015).

## Proposal

Render `metadata.breaker` on the slot card (and the slot detail drawer):

- `backoff` — amber hint: "retrying in Ns (N failures)".
- `parked` — red hint: "parked after N failures — next automatic trial in Ns", with the existing manual load/restart actions as the escape hatch.
- `half-open` — neutral hint: "trial pending — next request attempts a load".

`retry_after_s` is computed at snapshot time, so a simple client-side countdown from fetch time is accurate enough between polls.

## Notes

- No backend work needed; the field is already in every snapshot where the breaker is non-closed, and absent otherwise.
- The `parked` extra (`metadata.parked`) predates this and is stamped only on ERROR transitions; `metadata.breaker` is the live truth and should be preferred.

thinmintdev:
Closes #2038. Follow-up to #2012.

## Why

The API has served `metadata.breaker = {state, failures, retry_after_s}` since #2012, but nothing rendered it. An operator looking at the dashboard saw only a red ERROR chip — no hint the slot was deliberately refusing loads, how many consecutive failures it had, or when the next automatic trial runs. Invisibility of exactly this state is how the 2026-08-22 outage went unnoticed for four hours (#2015).

## What

- `ui/src/dash/slot-status.js` — `breakerChip(slot, elapsedS)`: pure classifier (state → tone/label/tooltip), beside its siblings (`imageStatusChip`, …) so surfaces can't drift. `retry_after_s` is computed at snapshot time, so the caller passes seconds-since-fetch and the countdown clamps at 0. Unknown breaker state renders nothing rather than guessing a colour. Prefers `metadata.breaker` over the older, stale-prone `metadata.parked` extra per the issue.
- `ui/src/dash/breaker-chip.jsx` — `SlotBreakerChip`: owns only the 1s countdown tick (clock re-keyed when a fresh poll delivers a new breaker view) and the colour tokens (`--warn*` / `--err*` / dashed neutral, matching `BackendMismatch` and `SlotImageUnknownChip`).
- Rendered in two places: the inference slot card foot (first chip — most actionable thing on the card) and beside the lifecycle state chip in the slot edit drawer, so "error" + "parked · 7 failures" read as one story.
- `ui/src/api/hooks/useSlots.ts` — typed `SlotBreaker` + `Slot.metadata.breaker`.

States per the proposal: `backoff` → amber "retry in Ns" (failure count in tooltip); `parked` → red "parked · N failures" (trial countdown + manual Start/Restart escape hatch in tooltip); `half-open` → neutral dashed "trial pending".

No backend work — the field is present whenever the breaker is non-closed and absent otherwise, so presence is the render signal.

## Tests

`ui/src/dash/__tests__/breaker-chip.test.ts` (written red-first): null without a breaker view, all three states' tone/label/tooltip, countdown from elapsed seconds + clamp at zero, singularised "1 failure", unknown state → null. Full UI unit suite 213 passed; `npm run typecheck` clean; γ-suite covers the panes in CI.

### DIFF
--- a/ui/src/api/hooks/useSlots.ts
+++ b/ui/src/api/hooks/useSlots.ts
@@ -46,6 +46,13 @@ export interface SlotMetrics {
   res?: string
 }
 
+/** #2012 crash-loop breaker snapshot view — see Slot.metadata.breaker. */
+export interface SlotBreaker {
+  state: 'backoff' | 'parked' | 'half-open'
+  failures: number
+  retry_after_s: number
+}
+
 export interface Slot {
   name: string
   type: string
@@ -101,6 +108,16 @@ export interface Slot {
    *  used / max". Absent when the slot configures no context window —
    *  the UI shows an em-dash, never a fabricated number. */
   ctx_max?: number
+  /** Crash-loop breaker view (#2012), stamped into slot metadata by the
+   *  manager whenever the breaker is non-closed and ABSENT otherwise —
+   *  presence of `metadata.breaker` is the render signal. `retry_after_s`
+   *  is computed at snapshot time; the UI counts down from fetch. Prefer
+   *  this over the older `metadata.parked` extra, which is stamped only on
+   *  ERROR transitions and goes stale. */
+  metadata?: {
+    breaker?: SlotBreaker
+    [key: string]: unknown
+  }
   /** Wall-clock epoch (seconds) of the most recent request served by
    *  this slot. ``null``/undefined means hal0-api has not seen a request
    *  for this slot since startup. Used by the slots view to render the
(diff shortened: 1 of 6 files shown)


thinmintdev:
Review pass applied (independent reviewer): countdown clock reset moved from an in-render ref mutation to a `useEffect` keyed on the breaker view (one-frame staleness, invisible at 1s granularity, render stays pure); `normalizeSlot` now names `metadata` explicitly so the pass-through is a documented contract. CSS tokens (`--warn*`/`--err*`/`--fg-3`) verified present in `dashboard.css`.

Issue #2052, opened 2026-08-24
Fresh install ships an updater that cannot update: cosign is not installed, so  stage fails on a pristine rc.7 box


thinmintdev:
Found on the rc.8 update lane (ct151, pristine rc.7 fresh install, Ubuntu 26.04 unprivileged LXC).

`hal0 update -y --target 1.0.0-rc.8` fails at stage:

    Error: prepare failed: privileged update seam failed: stage (rc=1)

Journal (privileged child):

    updater.privileged_failed error='cosign is not installed; install from https://docs.sigstore.dev/cosign/installation/' verb=stage

The keyless-signature verification the updater rightly insists on depends on a cosign binary that neither install.sh nor the rc.7 tree provisions, so every fresh install's FIRST update hard-fails until the operator hand-installs cosign. The error itself is honest and typed (good), but the dependency belongs to the platform: either install.sh should install/bundle cosign (pinned version, checksum-verified), or the updater should carry a vendored verifier.

Also cosmetic: the failure surfaces as the generic "privileged update seam failed: stage (rc=1)" at CLI level — the underlying cosign message only lives in the journal; the CLI should surface the child's typed error directly.

Found by the rc.8 update-lane run, 2026-08-24. Lane proceeded after `cosign` was hand-installed; that workaround is recorded in the lane report.

PR #2069, opened 2026-08-26
fix(installer): remove bootstrap work dir after a successful install

### LINKED ISSUE #2065: Bootstrap leaks its work dir (~150 MB) after every successful install
## Summary

bootstrap.sh sets an EXIT cleanup trap, but the trap dies at the `exec` into install.sh — exec replaces the process, so the trap never runs and nothing ever deletes the work dir. Confirmed live on ct151: `/tmp/hal0-install-IkMrMN`, 152 MB (tarball + unpacked tree), left behind after a successful install. This leaks on every install, and repeated installs accumulate.

## Fix options

- install.sh removes the handed-off tree at the end of a successful run (it knows the path it was exec'd from), or
- bootstrap re-arms cleanup differently — e.g. run install.sh as a child instead of exec so the trap survives, or hand the work-dir path to install.sh explicitly for deletion.

Note `HAL0_BOOTSTRAP_KEEP_TMP` (debug keep-tmp) must keep working either way.

## Acceptance

- Successful bootstrap install leaves no `/tmp/hal0-install-*` residue.
- Failed install still leaves the tree when `HAL0_BOOTSTRAP_KEEP_TMP` is set (debuggability preserved).

thinmintdev:
## Summary

Every successful one-line install leaked bootstrap's entire `/tmp/hal0-install-*` work directory (release tarball + unpacked tree, ~152 MB measured on ct151). This PR makes install.sh remove the handed-off tree at the end of a successful run, while keeping every debugging path intact.

## Root cause

`installer/bootstrap.sh` arms an `EXIT` trap (`cleanup_workdir`) to delete its work dir, but it hands control to the installer with `exec bash .../install.sh` — `exec` replaces the process, so the trap never fires on the success path. Nothing else ever deleted the directory, so it leaked on every install and repeated installs accumulated ~150 MB each.

## Fix

Mirrors the #2052 cosign hand-off pattern:

- `bootstrap.sh` exports `HAL0_BOOTSTRAP_WORK` pointing at the work dir just before the `exec`. It is deliberately left unset when `HAL0_BOOTSTRAP_KEEP_TMP=1`, so the debug knob keeps the tree exactly as before.
- A new `cleanup_bootstrap_workdir` function in `installer/lib/preflight.sh` (already sourced by install.sh) removes that tree. It re-checks `HAL0_BOOTSTRAP_KEEP_TMP` as defense in depth, and a strict `hal0-install-*` basename check keeps a stray or mangled value from ever aiming `rm -rf` anywhere else. Never fatal.
- `install.sh` calls it as its very last step (a single call plus comment — the install.sh touch is kept minimal on purpose). Dead-last ordering satisfies the #2058 constraint: `persist_bootstrap_cosign` has long since copied bootstrap's cosign out of the tree during pre-flight, and a failed install exits before ever reaching the cleanup, so the tree stays behind for debugging.

Deleting the tree install.sh itself runs from is safe: bash holds an open fd on the script file, which Linux keeps readable after the unlink.

## Acceptance mapping

- Successful bootstrap install leaves no `/tmp/hal0-install-*` residue — cleanup runs at end of a successful run.
- `HAL0_BOOTSTRAP_KEEP_TMP=1` keeps the tree — bootstrap withholds the hand-off env var, and the cleanup function honors the knob again independently.
- Failed install leaves the tree — the cleanup call is the last line of the script; any failure exits through the ERR trap before reaching it.

## Test evidence

New `tests/installer/test_bootstrap_workdir_cleanup.py` (written first, all 9 red before the fix, green after):

- drives `cleanup_bootstrap_workdir` directly under real bash: removes a handed-off tree; no-ops without the env var; keeps the tree under `HAL0_BOOTSTRAP_KEEP_TMP=1`; refuses non-`hal0-install-*` paths untouched; tolerates an already-gone dir.
- drives `bootstrap.sh` end-to-end through the `test_bootstrap_contract.py` fixture harness: the stub install.sh observes `HAL0_BOOTSTRAP_WORK` pointing at a live `hal0-install-*` dir, and observes it unset under `HAL0_BOOTSTRAP_KEEP_TMP=1`.
- static wiring assertions: the export exists in bootstrap code (non-comment), and install.sh's call site sits after `persist_bootstrap_cosign` and after the final summary box.

Ran locally (macOS dev box): `pytest tests/installer/test_bootstrap_workdir_cleanup.py tests/installer/test_install_cosign_persist.py` → 15 passed. `bash -n` clean on all three touched scripts (shellcheck not installed on this box). The full `tests/installer` directory shows 220 pre-existing failures on macOS (Darwin-only environment gaps, e.g. the contract harness's real `uname`); the count is identical with and without this change — zero regressions. CI on Linux is the authoritative run.

Fixes #2065

### DIFF
--- a/installer/bootstrap.sh
+++ b/installer/bootstrap.sh
@@ -505,6 +505,19 @@ main() {
     # HAL0_INSTALL_SKIP_VERIFY opt-out.
     export HAL0_BOOTSTRAP_VERIFIED=1
 
+    # The EXIT trap armed above dies at the exec below (exec replaces this
+    # process), so nothing bootstrap-side can delete the work dir after a
+    # successful hand-off — it used to leak ~150 MB per install (#2065).
+    # Hand the path to install.sh, which removes it as the very last step
+    # of a successful run — strictly after persist_bootstrap_cosign has
+    # copied the fetched cosign out of the tree (#2052). Deliberately left
+    # unset under HAL0_BOOTSTRAP_KEEP_TMP=1 so the debug knob keeps the
+    # tree, and a failed install exits before install.sh's cleanup — the
+    # tree stays for debugging either way.
+    if [[ "${HAL0_BOOTSTRAP_KEEP_TMP:-0}" != "1" ]]; then
+        export HAL0_BOOTSTRAP_WORK="${work}"
+    fi
+
     # Pass through stdin so install.sh's interactive prompts work when
     # the user invoked us as `sudo bash install.sh`. When invoked via
     # curl|bash, stdin is closed and install.sh falls back to defaults.
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -3461,5 +3461,14 @@ ui_box "hal0 is ready" "${SUMMARY_LINES[@]}"
 # install is picking models per slot, which is a dashboard job — see the
 # "Next steps" box above. Do NOT reintroduce a post-install CLI wizard here.
 
+# Bootstrap hand-off cleanup (#2065): remove the /tmp/hal0-install-* work
+# dir this tree was unpacked into — bootstrap.sh's own EXIT trap died at
+# its exec into us, so the removal is ours. No-op for git-checkout / --dev
+# / tarball-direct installs (HAL0_BOOTSTRAP_WORK unset) and under
+# HAL0_BOOTSTRAP_KEEP_TMP=1. Dead-last on purpose: a failed install never
+# gets here (the tree stays for debugging), and persist_bootstrap_cosign
+# already salvaged bootstrap's cosign out of the tree in pre-flight.
+cleanup_bootstrap_workdir
+
 # Restore the caller's umask (see the save near the top of the file).
 umask "${_HAL0_ORIG_UMASK}"
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1650,3 +1650,35 @@ persist_bootstrap_cosign() {
     warn "cosign is not installed and this run has no bootstrap-verified binary to persist — the first 'hal0 update' will fail its signature check until cosign is installed (https://docs.sigstore.dev/cosign/installation/)"
     return 0
 }
+
+# ── bootstrap work-dir cleanup (#2065) ─────────────────────────────────────
+# bootstrap.sh arms an EXIT trap to delete its /tmp/hal0-install-* work dir
+# (release tarball + unpacked tree, ~150 MB), but that trap dies at the
+# `exec` into install.sh — so every successful install used to leak the
+# whole tree. Bootstrap hands the path over via HAL0_BOOTSTRAP_WORK (left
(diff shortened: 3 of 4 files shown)


thinmintdev (commented):
Independent review (install-fix wave). Verified against issue #2065 including the #2058 cosign-ordering comment constraint. LGTM — findings below, none blocking.

Correctness of the fix:
- installer/bootstrap.sh:505-518 — export of HAL0_BOOTSTRAP_WORK sits immediately before the exec and is guarded by the exact same KEEP_TMP condition that gates the EXIT-trap arming (bootstrap.sh:436), so the debug knob and the hand-off can never disagree. Bootstrap-phase failures still clean up via the surviving trap; the export only matters once exec succeeds.
- installer/install.sh:3472 — cleanup call is after persist_bootstrap_cosign (install.sh:303, in pre-flight), satisfying the binding comment constraint on #2065, and after the final summary box, so any failure exits through the ERR trap (install.sh:248, 'exit 1') before reaching it. Confirmed the file runs set -euo pipefail (:17) and cleanup_bootstrap_workdir cannot return non-zero (every branch '|| return 0' or explicit return 0), so the new last-line call introduces no set -e hazard.
- installer/lib/preflight.sh:1669-1684 — basename glob check plus 'rm -rf --' correctly bounds the blast radius to hal0-install-* paths; rm does not follow a symlink argument, and mktemp -d -t hal0-install-XXXXXX (bootstrap.sh:434) always yields a matching absolute path. Deleting the tree install.sh runs from is safe on Linux (unlinked-but-open script fd); only the umask builtin runs after.
- Env-var survival across install.sh's non-root re-exec ('exec sudo -E', install.sh:~275) rides the same mechanism HAL0_BOOTSTRAP_VERIFIED and HAL0_BOOTSTRAP_COSIGN already depend on — no new failure mode.

Test claims check out:
- 9 tests in the new file matches the class contents; the direct-function tests' info/warn stubs actually take effect because preflight.sh only sources ui.sh when 'info' is undefined (preflight.sh:115), so the INFO:/WARN: assertions are sound. The wiring test pins the #2058 ordering and the after-summary-box placement. No scope creep — the diff is additive and touches only the three scripts plus the new test file.

Nits (non-blocking):
- installer/install.sh:125 — 'curl|bash -s -- --help' through bootstrap downloads and verifies the full ~150 MB release, then exits 0 at the --help branch before cleanup ever runs, leaking the tree on a nominally successful exit. Pathological invocation, pre-existing shape, fine to leave; noting for completeness.
- installer/lib/preflight.sh:1676 — a trailing-slash value (basename expands to empty) fails the glob and warns/leaks rather than cleaning; unreachable from bootstrap's mktemp, and fail-soft is the right bias here.
- PR text says the cleanup call is install.sh's 'very last step'; strictly it is followed by the umask restore. Harmless (builtin, no fs access) — the code comment could say 'last filesystem-touching step' if anyone cares.

Comment-only per wave protocol; not formally approving or merging.

PR #2070, opened 2026-08-26
fix(hermes): root hermes runs resolve the hal0 service home, and say so (#2063)

### LINKED ISSUE #2063: hermes mcp list/test as root reads ~root/.hermes and falsely reports "No MCP servers configured"
## Summary

Running `hermes mcp list` / `hermes mcp test` as root resolves the config under `~root/.hermes` instead of the service account's home, and reports "No MCP servers configured" on a healthy install. This bit a live debugging session: an operator diagnosing MCP would reasonably conclude the install is broken when it's fine — the CLI just read the wrong home.

## Fix options (in order of preference)

1. hermes CLI resolves the service home (the hal0 service user's HOME / HERMES_HOME) when run as root, matching what the running agent actually uses.
2. hal0 ships a wrapper that sets `HOME=/var/lib/hal0` (or HERMES_HOME) for hermes CLI invocations.
3. At minimum: print which config path was read, so the empty result is diagnosable ("No MCP servers configured (read /root/.hermes/...)").

Option 3 should probably happen regardless — silent path resolution is what turned a wrong-home read into a false diagnosis.

## Acceptance

- `hermes mcp list` as root on a configured box shows the service account's servers (or, minimally, names the config path it read).
- Behavior as the service user unchanged.

thinmintdev:
## Summary

`hermes mcp list` / `hermes mcp test` run as root resolved the config under `~root/.hermes` and falsely reported "No MCP servers configured" on a healthy install, sending an operator down a "broken install" path when the CLI had simply read the wrong home. This PR hardens hal0's `hermes` wrapper so root invocations resolve the hal0 service home, and makes every root-context path print which `~/.hermes` tree it resolved.

## Root cause

Upstream `hermes` resolves its state dir from `$HOME` (`~/.hermes`). The hermes CLI source is **not** in this repo — it is pinned from upstream `NousResearch/hermes-agent` (`installer/agents/hermes/requirements.txt`) — so the CLI-side fix (resolve the service home when euid == 0) is an upstream matter. hal0's seam is the wrapper at `installer/wrappers/hermes`.

That wrapper already drops root to the `hal0` service user via the run-as-hal0 guard (`/usr/lib/hal0/guards/run-as-hal0.sh`, #843), which covers a fully installed box. But two root paths remained silent and wrong:

- guard lib missing/unreadable (partial install, upgrade drift, dev box) or the `hal0` user absent → the wrapper proceeded as root unguarded, and hermes read `/root/.hermes`;
- `HAL0_ALLOW_ROOT=1` (deliberate root debug) → root's own tree was read with no indication of which config path was in play.

Either way the empty result was undiagnosable — exactly what turned a wrong-home read into a false diagnosis in the live session behind #2063.

## Fix

In `installer/wrappers/hermes`, after the guard block, if the process is still root:

- **Guard couldn't drop** → point `HOME` at the hal0 service home (`getent passwd hal0`, `HAL0_SERVICE_HOME` override for tests/nonstandard accounts), strip any inherited `HERMES_HOME` (matching the guard's `env -u` posture), and print the resolved `~/.hermes` path on stderr.
- **No `hal0` user resolvable** (true dev shell) → behavior unchanged, but the resolved path is printed so the empty result is diagnosable.
- **`HAL0_ALLOW_ROOT` set** → environment left alone, but the wrapper states it is reading root's tree, not the service home.

Non-root invocations (the service user, the systemd unit) are untouched and stay silent. The `--hal0-ready` installer sentinel stays silent too (the block sits after it). Docs: a short "Running `hermes` as root" section added to `docs/guides/run-agents.mdx`, noting that the euid-aware CLI fix belongs upstream.

## Test evidence

Test-first: the four new behavior tests in `tests/agents/test_hermes_wrapper.py` fail against the pre-fix wrapper —

```
FAILED test_wrapper_root_without_guard_points_home_at_service_home
FAILED test_wrapper_root_without_guard_strips_inherited_hermes_home
FAILED test_wrapper_root_without_service_user_warns_with_resolved_path
FAILED test_wrapper_allow_root_keeps_root_env_but_names_the_path
4 failed, 1 passed
```

With the fix: `pytest tests/agents/test_hermes_wrapper.py` → **36 passed** (all pre-existing wrapper tests included; notably `test_hermes_wrapper_does_not_pin_hermes_home` still holds — the fallback only redirects `HOME`, it never reintroduces a `HERMES_HOME` pin). `sh -n installer/wrappers/hermes` passes; shellcheck is not installed on this dev Mac — relying on CI for that. `tests/agents/test_run_as_hal0_guard.py::test_reexec_actually_runs_with_clean_env` fails on macOS before and after this change (the guard's `getent` lookup doesn't exist on Darwin) — pre-existing, untouched here.

Fixes #2063

### DIFF
--- a/docs/guides/run-agents.mdx
+++ b/docs/guides/run-agents.mdx
@@ -185,6 +185,22 @@ Default uninstall tears down the agent's private memory namespace too;
 </TabItem>
 </Tabs>
 
+## Running `hermes` as root
+
+Upstream `hermes` resolves its state from `~/.hermes`, so a bare root
+invocation would read `/root/.hermes` — an empty tree the agent never
+uses — and commands like `hermes mcp list` would falsely report
+"No MCP servers configured" on a healthy install. hal0's `hermes`
+wrapper prevents that: run as root, it re-executes as the `hal0`
+service user (via the run-as-hal0 guard), so you see the same config
+the running agent uses. If the guard can't drop privileges, the
+wrapper points `HOME` at the service home instead and prints which
+`~/.hermes` tree it resolved on stderr. `HAL0_ALLOW_ROOT=1` opts out
+for deliberate root debugging — the wrapper then leaves root's
+environment alone but still names the tree it is reading. The
+euid-aware fix inside the hermes CLI itself is an upstream
+(NousResearch/hermes-agent) matter; the wrapper is hal0's seam.
+
 ## Personas
 
 A persona is a TOML file pairing a system prompt with tool gating:
--- a/installer/wrappers/hermes
+++ b/installer/wrappers/hermes
@@ -67,6 +67,44 @@ case "${1:-}" in
     --hal0-ready) exit 0 ;;
 esac
 
+# root-home fallback (#2063): reaching this point as root means the guard
+# above could NOT drop privileges — the guard lib is missing/unreadable
+# (partial or dev install) or the hal0 user is absent — or the operator
+# forced root via HAL0_ALLOW_ROOT. Upstream hermes resolves its state dir
+# from $HOME (`~/.hermes`; the CLI-side euid-aware fix belongs upstream),
+# so an unguarded root run silently reads /root/.hermes — a tree the hal0
+# service never writes — and `hermes mcp list` falsely reports "No MCP
+# servers configured" on a healthy install. Point HOME at the hal0 service
+# home instead (and strip any inherited HERMES_HOME, matching the guard's
+# `env -u` posture), and ALWAYS name the tree this invocation resolves so
+# an empty result is diagnosable rather than a false "broken install".
+# HAL0_SERVICE_HOME bypasses the getent lookup (tests / nonstandard
+# service accounts; set-but-empty means "no service user resolvable").
+# HAL0_RUNAS_TEST_UID is the guard's documented test-only euid seam.
+_hal0_euid="${HAL0_RUNAS_TEST_UID:-$(id -u 2>/dev/null || echo 0)}"
+if [ "$_hal0_euid" = "0" ]; then
+    case "${HAL0_ALLOW_ROOT:-}" in
+        1 | true | yes | TRUE | YES)
+            # Deliberate root debug: env left alone, resolved path named.
+            printf 'hermes: HAL0_ALLOW_ROOT set — reading %s/.hermes (root debug; NOT the hal0 service home)\n' \
+                "${HOME:-/root}" >&2
+            ;;
+        *)
+            _hal0_svc_home="${HAL0_SERVICE_HOME-$(getent passwd hal0 2>/dev/null | cut -d: -f6)}"
+            if [ -n "$_hal0_svc_home" ]; then
+                HOME="$_hal0_svc_home"
+                export HOME
+                unset HERMES_HOME
+                printf 'hermes: running as root without the run-as-hal0 guard — reading %s/.hermes (hal0 service home)\n' \
+                    "$_hal0_svc_home" >&2
+            else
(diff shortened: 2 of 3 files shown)


thinmintdev (commented):
Independent review (install-fix wave). Verified against issue #2063 (no issue comments carrying extra constraints), the run-as-hal0 guard (`installer/lib/run-as-hal0.sh`), and by running the tests myself against the PR head (6477285).

**Verification done**
- Reproduced the test-first claim exactly: the five new tests run standalone against the pre-fix wrapper give 4 failed / 1 passed; against the PR wrapper all 5 pass. `sh -n` passes on the PR wrapper.
- Confirmed the PR wrapper is the current base plus only the new block — no unrelated drift, no scope creep (wrapper + tests + one doc section).
- set -e/quoting checked: `${HAL0_SERVICE_HOME-$(getent ...| cut ...)}` is safe when `getent` is absent (pipeline ends in `cut`, exit 0 → empty → warn branch; probed under sh/bash/dash/zsh), and the `id -u ... || echo 0` guard mirrors the guard lib's line. Wrapper is stateless, so re-runs are idempotent; non-root and `--hal0-ready` paths verified untouched (block sits after the sentinel, and `HAL0_RUNAS_TEST_UID=1000` test proves silence).
- Guard interplay checked: all three still-root-after-guard cases (lib unreadable; `HAL0_ALLOW_ROOT`; hal0 user absent → guard returns 0) are covered, and the `HAL0_ALLOW_ROOT` value list (1|true|yes|TRUE|YES) matches the guard's exactly. The guard-present-but-no-dropper case still hard-refuses before the fallback, which preserves the "never proceed silently as root" posture. Legacy `hal0-hermes` pins HERMES_HOME so it never had the wrong-home bug — correctly out of scope.
- Issue acceptance: option 2 (wrapper points HOME at the service home) plus option 3 (always name the resolved tree) are both delivered; service-user behavior unchanged. Matches the issue's stated preference order given the CLI fix is upstream.

**Non-blocking findings**
1. installer/wrappers/hermes:95 — the fallback re-opens a sliver of the #843 root-clobber surface: on a guard-less partial install, a state-writing root invocation (e.g. `hermes reset`) now writes root:root files into the service tree instead of harmlessly split-braining into /root/.hermes. This is exactly what issue option 2 asked for and the stderr line makes it diagnosable, but the block comment (or the docs section) should name the trade-off so a future reader doesn't "fix" it back.
2. installer/wrappers/hermes:89,101 — the ALLOW_ROOT and no-service-user branches print `$HOME/.hermes` as the resolved tree, but in those branches an inherited `HERMES_HOME` is deliberately left alone and would actually win inside upstream hermes — so the printed path can be wrong in the one sub-case (root debug with HERMES_HOME exported) where precision matters most. Cheap fix: prefer `${HERMES_HOME:-$HOME/.hermes}` in those two messages.
3. docs/guides/run-agents.mdx:193 — "If the guard can't drop privileges, the wrapper points HOME at the service home instead" slightly overstates: when the guard lib is present but no runuser/setpriv/sudo exists, the guard refuses and the wrapper exits 1 — the HOME fallback only applies when the guard is absent or skipped (lib missing / hal0 user absent). Worth one clarifying clause.
4. tests/agents/test_hermes_wrapper.py:676 — every root-path test sets `HAL0_SERVICE_HOME`, so the real `getent passwd hal0` production lookup is never exercised (understandable — CI has no hal0 user — but worth knowing the seam-free path is uncovered).

Could not reproduce the full "36 passed" suite number locally (this machine's checkout predates the PR base and the repo venv is incomplete); the five new tests and the wrapper-only pre-existing shell tests were verified directly, the rest is CI's to confirm. Overall: correct, well-scoped, honest test evidence — LGTM with the minor notes above.

PR #2071, opened 2026-08-26
fix(installer): grant the service user write access to /etc/avahi/services

### LINKED ISSUE #2059: mDNS advertise/withdraw always fails EACCES: installer never grants the service user write access to /etc/avahi/services
## Summary

`POST /api/services/mdns` (dashboard Services page advertise/withdraw) fails with:

```
comfyui: [Errno 13] Permission denied: '/etc/avahi/services/hal0-addon-comfyui.service'
openwebui: [Errno 13] Permission denied: '/etc/avahi/services/hal0-addon-openwebui.service'
```

`services/mdns.py` writes/removes `hal0-addon-*.service` files directly from the API process, which runs as the unprivileged `hal0` service user. `/etc/avahi/services` is `root:root 0755`, and nothing in the installer (`install/perms.py`, `installer/install.sh`) grants the service user write access. There is also no privileged helper path for this write.

This is a regression from the root-to-unprivileged switch: the feature worked when the API ran as root (existing addon files on affected boxes are owned `root:root` from that era), and has been dead on every unprivileged install since. Stale root-owned addon files also cannot be withdrawn or refreshed.

## Observed on

Live box 10.0.1.142 (hal0, CT105): avahi active, addon files dated from the root era, every advertise/withdraw attempt returns EACCES.

## Suggested fix

When avahi is present, the installer should make `/etc/avahi/services` group-writable by the service user, e.g.:

```
chgrp hal0 /etc/avahi/services
chmod g+w /etc/avahi/services
chgrp hal0 /etc/avahi/services/hal0-addon-*.service 2>/dev/null || true
```

avahi-daemon inotify-watches the directory, so no daemon interaction is needed. Alternatively route the write through a privileged helper, but the group-write grant is the minimal fix and matches the existing perms model in `install/perms.py`. The manual chgrp/g+w above was applied by hand on 10.0.1.142 as a workaround and confirmed to restore advertise/withdraw.

## Acceptance

- Fresh unprivileged install with avahi present: dashboard advertise then withdraw both succeed, files appear/disappear under `/etc/avahi/services`.
- Upgrade path fixes ownership of pre-existing root-owned `hal0-addon-*.service` files.

thinmintdev:
## Summary

Dashboard mDNS advertise/withdraw (`POST /api/services/mdns`) has been dead on every unprivileged install: `services/mdns.py` runs as the `hal0` service user and writes `hal0-addon-*.service` files into `/etc/avahi/services`, which ships `root:root 0755` — every write dies EACCES. This adds the missing permission grant to the installer.

## Root cause

A regression from the root-to-unprivileged switch: the feature worked while the API ran as root, and nothing in `installer/install.sh` or the ownership table (`src/hal0/install/perms.py`) was ever taught to make the avahi services dir writable by the service user. `mdns.py` writes via tmp + rename, so *directory* write is the one capability it needs.

## Fix

Two small, mutually reinforcing pieces, both idempotent and guarded on avahi actually being present:

- **`src/hal0/install/perms.py`** — a new dir-only `PermRow` for the avahi services dir (resolved via `mdns.services_dir()`, the writer's own path authority): `root:hal0 0775` under the hardened flip — owner stays root because the dir belongs to avahi; group-write is the minimal grant, matching the manual workaround already verified on 10.0.1.142. `optional` is the avahi-presence guard (absent dir plans clean, never drifts), and the root-era table keeps the distro default `root:root 0755`. The installer's existing `doctor perms --fix --force` backstop applies it on fresh installs *and* upgrades, and `hal0 doctor perms` can now see/heal this drift forever. Deliberately no glob into the addon files: rewrites and prunes only need dir write, and a fresh advertise births files hal0-owned while root-era leftovers are root-owned — a single declared owner would fight one era or the other.
- **`installer/install.sh`** — the same `chgrp hal0` + `chmod g+w` at the filesystem-layout step so the dir is correct before the daemon's first touch, plus the upgrade path: pre-existing root-era `hal0-addon-*.service` files get `chgrp hal0`. avahi-daemon inotify-watches the dir, so no daemon interaction is needed.
- **`tests/conftest.py`** — `tmp_hal0_home` now also points `HAL0_AVAHI_SERVICES_DIR` under the isolated tree so the perms converge tests never observe (or drift against) the host's real avahi install.

`services/mdns.py` itself is unchanged.

## Test evidence

Test-first: the two new tests fail against `main`'s `perms.py` (`KeyError` — no row for the avahi dir) and pass with the fix:

- `test_avahi_services_dir_is_group_writable_under_flip` — row is `root:hal0 0775` dir-only under the flip, `root:root 0755` in the root-era table, `optional` (the presence guard).
- `test_avahi_services_dir_absent_is_never_drift` — a box without avahi plans clean.

Ran locally (macOS, uv): `tests/install/` + `tests/security/test_api_env_mode.py` → **306 passed, 1 skipped** (pre-existing sandbox skip); `tests/api/test_services_page.py` + `tests/cli/test_doctor_verify.py` → 34 passed. `ruff check` + `ruff format --check` clean on touched files; `bash -n installer/install.sh` clean (shellcheck not installed on this Mac — relying on CI for that).

Fixes #2059

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -756,6 +756,25 @@ if [[ "${DEV_MODE}" -eq 0 ]] && getent passwd hal0 >/dev/null 2>&1; then
     chmod 2775 "${VAR_DIR}/slots" "${VAR_DIR}/registry" "${VAR_DIR}/models" 2>/dev/null || true
 fi
 
+# mDNS addon advertisements (#2059): services/mdns.py (hal0-api, User=hal0)
+# writes /etc/avahi/services/hal0-addon-<id>.service via tmp+rename, which
+# needs *directory* write — the distro ships the dir root:root 0755, so every
+# dashboard advertise/withdraw died EACCES since the root-to-unprivileged
+# switch. Group-write grant (root:hal0, g+w) mirrors the OwnershipStore row in
+# src/hal0/install/perms.py (the `doctor perms --fix` backstop heals the same
+# values); done here too so the dir is correct before the daemon's first
+# touch. The chgrp of pre-existing hal0-addon-*.service files is the upgrade
+# path: root-era installs wrote them as root, and handing them to the hal0
+# group keeps the daemon's own advertisement surface consistently
+# group-readable. avahi-daemon inotify-watches the dir — no reload needed.
+# Guarded on avahi actually being present; idempotent on re-run.
+if [[ "${DEV_MODE}" -eq 0 && -d /etc/avahi/services ]] \
+    && getent group hal0 >/dev/null 2>&1; then
+    chgrp hal0 /etc/avahi/services 2>/dev/null || true
+    chmod g+w /etc/avahi/services 2>/dev/null || true
+    chgrp hal0 /etc/avahi/services/hal0-addon-*.service 2>/dev/null || true
+fi
+
 # Production (FHS, #495) ships the source tree into the versioned dir
 # ${PREFIX} (=${FHS_ROOT}/hal0-<version>) and points `current` at it, so
 # `hal0 update` can atomically swap `current` to a new versioned tree.
--- a/src/hal0/install/perms.py
+++ b/src/hal0/install/perms.py
@@ -191,6 +191,20 @@ def _shared_dir_roots() -> tuple[Path, ...]:
 # ── the declarative table ─────────────────────────────────────────────────────
 
 
+def _avahi_services_dir() -> Path:
+    """The avahi services dir the mDNS row declares (#2059).
+
+    Delegates to :func:`hal0.services.mdns.services_dir` — the writer's own
+    path authority, honoring its ``HAL0_AVAHI_SERVICES_DIR`` override — so the
+    table and the writer can never disagree about which dir the grant covers.
+    Imported lazily to keep this module free of the services layer at import
+    time.
+    """
+    from hal0.services.mdns import services_dir
+
+    return services_dir()
+
+
 @dataclass(frozen=True)
 class PermRow:
     """One path's declared ownership + mode.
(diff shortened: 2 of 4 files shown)


thinmintdev (commented):
Independent review (install-fix wave). Verified against issue #2059 (no issue comments, so no extra binding constraints) and the PR-head tree at cee563b.

What I checked and confirmed:
- installer/install.sh:771 — the grant matches the issue's prescribed fix byte-for-byte (chgrp hal0 + chmod g+w on the dir, chgrp of root-era hal0-addon-*.service leftovers). Guards are correct: DEV_MODE, dir presence, getent group hal0 — and the hal0 group is created earlier at install.sh:700, so a fresh install takes this path rather than relying on the doctor backstop. All three commands are '2>/dev/null || true' fail-soft, so set -euo pipefail is safe and the unmatched-glob case (no addon files yet) is harmless. Idempotent on re-run.
- src/hal0/install/perms.py:875 — row shape is right: dir-only (no glob/recursion), owner stays root, group-write is the minimal grant for the tmp+rename/unlink writer in services/mdns.py (dir write is genuinely all it needs — rename-over and unlink of root-owned files work without file ownership since the dir has no sticky bit). Root-era table keeps distro-default root:root 0755. Resolving the target via mdns.services_dir() keeps the table and the writer agreeing, including the HAL0_AVAHI_SERVICES_DIR override.
- Absent-dir behavior: plan()/PermDiff.changed never flags an absent path and audit_rows reports it as 'absent — not present' (doctor treats absent as not-a-problem), so non-avahi boxes stay clean. The new absent-plan test confirms it.
- Upgrade coverage: installer re-runs and 'hal0 update' (which runs the new tarball's installer) both apply the grant; drifted boxes are additionally healable forever via the doctor perms row. Acceptance criteria of #2059 are met.
- tests/conftest.py isolation is necessary and correct — without it, test_ownership_table_builds_under_hal0_home would fail on the new row and perms tests could observe/heal the host's real avahi dir. Test claims are honest: both new tests KeyError against main's table.

Minor, non-blocking observations:
- perms.py:869/test_perms.py:300 — 'optional' is documented as the presence guard, but nothing in the machinery actually consumes PermRow.optional (absent paths are skipped for every row, optional or not, and True is the field default). The behavior is fine; the comment/assert slightly overstate the field's mechanical role.
- install.sh:774 — additive 'chmod g+w' vs the row's exact 0775: on a hardened box shipping the dir at e.g. 0750, install.sh leaves 0770 and the doctor backstop then widens to exactly 0775 (adds o+rx). Same end state on default distros; just noting the exact-mode convergence broadens other-bits on hardened hosts.
- Security tradeoff (accepted per the issue): dir group-write lets the hal0 service user delete/replace any file in /etc/avahi/services — including the installer-owned hal0.service — and advertise arbitrary mDNS records. That is inherent to the issue's chosen minimal fix; a privileged-helper route was explicitly deferred there.
- Dev-mode caveat: services_dir() ignores HAL0_HOME, so on a dev box that happens to have avahi installed, the row points at the real /etc/avahi/services and doctor perms will report drift there (tests are insulated via the conftest env var). Consistent with the writer, which also targets the real dir in dev — noting only.

LGTM — fixes the issue as specified, idempotent, fail-soft, no scope creep (mdns.py untouched).

PR #2072, opened 2026-08-26
fix(providers): honest kfd-less refusal + documented device contract for the non-llama gpu-rocm seeds (#2067)

### LINKED ISSUE #2067: Non-llama seeds (comfyui img, qwen3tts) skip device derivation — gpu-rocm seeded on kfd-less box
## Summary

#2054 fixed device derivation for llama slots only. The non-llama seeds still skip it: on a kfd-less box, the `img` (comfyui) and `qwen3tts` seeds landed with `gpu-rocm`, creating the same dead-slot class #2054 addressed — slots that will fail on first use because the device doesn't exist.

## Proposed

Either:

- Extend the derivation ladder from #2054 to cover the non-llama seed paths (img/comfyui, qwen3tts, and any other non-llama seeds), so they honor the detected device the same way llama slots now do, or
- If these providers legitimately manage their own device selection, verify their load refusal is honest — a clear "no ROCm device" error at load, not a hang or a misleading failure — and document that as the contract.

## Acceptance

- Fresh install on a kfd-less box: img and qwen3tts slots either seed with a device that exists, or their first-use failure states plainly that the seeded device is absent.
- No regression on ROCm boxes.

thinmintdev:
## Summary

#2067 asked an either/or: extend #2054's device-derivation ladder to the non-llama seeds (img/ComfyUI, qwen3tts), or — if those providers legitimately own device selection — make their kfd-less load refusal honest and document the contract. This PR takes the second branch, because the first is impossible by construction: both runner images are ROCm-only builds (`RUNNER_IMAGES["comfyui"].supported_backends == ("rocm",)`, same for `qwen3tts`), so a kfd-less box has no CPU or Vulkan lane a derivation ladder could pick. Extending the ladder would only relabel the slot into a device/profile pair the runtime cannot serve.

## Root cause

The seeded `device = "gpu-rocm"` on a kfd-less box is deliberate and mostly safe: the seeds ship no `[model].default` pin, so nothing autoloads — the slot lands as an inactive tile, and first use hits `require_kfd_for_gpu_slot`, which already refuses `gpu-rocm` in every lane and (since #1952) explains the rocm lane without blaming llama.cpp's #1888 fallback. The one dishonest piece was the missing-node remedy's closing clause, shared across lanes: "…or move this slot to device='cpu'". For ComfyUI/Qwen3-TTS that is advice the runtime cannot honour — following it launches a HIP image with no GPU lane at all, trading a clear refusal for a confusing failure.

## Fix

- `src/hal0/providers/_gpu.py`: the missing-kfd remedy is now lane-scoped. The `rocm` lane states the image is ROCm-only and keeps only the forward-the-node remedy; the `llama` and `no-rocm` lanes keep the `device='cpu'` suggestion (llama-server genuinely runs on CPU; for `no-rocm` the device label is the actual defect). Docstring updated.
- `src/hal0/install/static_seeds.py` + `installer/etc-hal0/slots/{img,qwen3tts}.toml`: the contract is documented where the next reader will look — verbatim `gpu-rocm` on these two seeds is deliberate (#2067), why no derivation applies, and what the first-use failure looks like on a kfd-less box.
- Tests pin the contract end-to-end:
  - `tests/install/test_seed_device_derivation.py`: the shipped non-llama `gpu-rocm` seeds are exactly `img`/`qwen3tts`; each resolves (via `_spec_provider_for` on the actual seed TOML) to a rocm-lane provider over a ROCm-only runner; and the kfd-less first-use refusal names the slot, the device, the missing `/dev/kfd` and the forward remedy while containing neither "#1888" nor `device='cpu'`.
  - `tests/providers/test_gpu_kfd_preflight.py` (`TestGuardRemedyText`): lane-scoped remedy pins — rocm lane never offers a CPU lane, llama lane still does.

No regression on ROCm boxes: kfd-present hosts never reach the refusal path, and the #2054 llama-seed derivation is untouched (its ladder still picks `gpu-rocm` when kfd exists — pinned by the existing `test_kfd_present_host_still_derives_gpu_rocm`).

## Test evidence

Test-first, verified red on the pre-fix code: `TestGuardRemedyText::test_the_rocm_lane_missing_remedy_does_not_offer_a_cpu_lane` and `test_kfd_less_first_use_refusal_is_plain_and_honourable` both fail before the `_gpu.py` change and pass after.

Ran on the Mac dev checkout: `pytest tests/install/ tests/providers/test_gpu_kfd_preflight.py tests/providers/test_gpu_vulkan_lane_gate.py tests/providers/test_vulkan_lane_perimeter.py tests/providers/test_comfyui.py tests/providers/test_qwen3tts_container_spec.py tests/installer/test_seed_order.py` — all green except pre-existing environment failures unrelated to this diff (`tests/installer/*` bash-contract tests refuse Darwin, and `tests/install/test_perms_converge.py` trips a UnicodeDecodeError on NFS AppleDouble files); relying on CI for those. `ruff check` and `ruff format --check` clean on all touched files.

Fixes #2067

### DIFF
--- a/installer/etc-hal0/slots/img.toml
+++ b/installer/etc-hal0/slots/img.toml
@@ -15,6 +15,11 @@
 name = "img"
 type = "image"
 provider = "comfyui"
+# gpu-rocm is a contract, not a preference (#2067): the pinned ComfyUI image
+# is a ROCm-only build (RUNNER_IMAGES["comfyui"].supported_backends ==
+# ("rocm",)), so seeding never derives another device — there is no CPU or
+# Vulkan lane to derive. On a kfd-less box this seeds as an inactive tile and
+# first use refuses loudly, naming the absent /dev/kfd and how to forward it.
 device = "gpu-rocm"
 runtime = "container"
 profile = "comfyui"
--- a/installer/etc-hal0/slots/qwen3tts.toml
+++ b/installer/etc-hal0/slots/qwen3tts.toml
@@ -10,6 +10,12 @@
 # triggers a surprise download or a crash-loop.
 name = "qwen3tts"
 type = "tts"
+# gpu-rocm is a contract, not a preference (#2067): the Qwen3-TTS image is a
+# ROCm-only build (RUNNER_IMAGES["qwen3tts"].supported_backends == ("rocm",)),
+# so seeding never derives another device — there is no CPU or Vulkan lane to
+# derive; the cpu `tts`/kokoro sibling above IS the CPU speech lane. On a
+# kfd-less box this seeds as an inactive tile and first use refuses loudly,
+# naming the absent /dev/kfd and how to forward it.
 device = "gpu-rocm"
 runtime = "container"
 profile = "qwen3-tts"
--- a/src/hal0/install/static_seeds.py
+++ b/src/hal0/install/static_seeds.py
@@ -64,7 +64,14 @@ STATIC_SEED_SLOTS: tuple[str, ...] = (
 #:
 #: Deliberately absent: ``flm`` (npu), ``tts`` (kokoro/cpu), ``img``
 #: (ComfyUI) and ``qwen3tts`` — non-llama runtimes with their own device
-#: logic. Their seed device ships verbatim, whatever the host.
+#: logic. Their seed device ships verbatim, whatever the host. For the two
+#: that ship ``gpu-rocm`` (``img``, ``qwen3tts``) verbatim is a CONTRACT, not
+#: an omission (#2067): both runner images are ROCm-only builds
+#: (``RUNNER_IMAGES[...].supported_backends == ("rocm",)``), so a kfd-less box
+#: has no CPU or Vulkan lane to derive instead. The seed lands as an inactive
+#: tile (no ``[model].default`` pin, #1369 — nothing autoloads), and first use
+#: refuses loudly via ``_gpu.require_kfd_for_gpu_slot``'s ``rocm`` lane,
+#: naming the absent ``/dev/kfd`` and how to forward it.
 LLAMA_SEED_CAPABILITIES: dict[str, str] = {
     "agent": "chat",
     "brain": "chat",
--- a/src/hal0/providers/_gpu.py
+++ b/src/hal0/providers/_gpu.py
@@ -717,7 +717,9 @@ def require_kfd_for_gpu_slot(
       ROCmFPX runner. Gated, with #1888's silent-garbage explanation.
     * ``"rocm"`` — a non-llama runtime whose image resolves ROCm/HIP.
       Gated, but the refusal does not blame llama.cpp's fallback: quoting
-      #1888 at a ComfyUI operator sends them chasing the wrong defect.
+      #1888 at a ComfyUI operator sends them chasing the wrong defect. Its
+      missing-node remedy also never suggests ``device='cpu'`` (#2067):
+      these images are ROCm-only, so forwarding the node is the only fix.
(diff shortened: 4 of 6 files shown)


thinmintdev (commented):
Independent review (issue #2067; the issue has zero comments, so the issue body is the full spec — no extra comment constraints apply).

Verdict: LGTM. The PR takes the issue's sanctioned second branch (honest kfd-less refusal + documented contract), and its justification for rejecting the first branch is factually correct — I verified against the head tree, not just the diff.

Verified:
- src/hal0/runners/__init__.py:205,216 — RUNNER_IMAGES qwen3tts and comfyui both declare supported_backends == ("rocm",), so there genuinely is no CPU/Vulkan lane a #2054-style derivation ladder could pick; extending derivation would only relabel the dead slot. The PR's central premise holds.
- src/hal0/providers/comfyui.py:122,131 and qwen3tts.py:109,117 — provider names match the RUNNER_IMAGES keys and both declare gpu_runtime_needs_rocm = True, so the new registry/lane pins in tests/install/test_seed_device_derivation.py assert real production wiring.
- src/hal0/providers/container.py:2494-2500 — the real first-use path (load_sync) passes runtime_lane=runtime_lane_for_provider(_spec_provider_for(...)), so the tests' direct guard invocation mirrors production, and the seed TOMLs resolve to the rocm lane exactly as the tests pin (img via family/provider/type, qwen3tts via the qwen3-tts profile family).
- src/hal0/providers/_gpu.py:854-876 — remedy scoping is correct in every lane: KFD_NOT_OPENABLE remedy is untouched (never offered device='cpu'); llama and no-rocm lanes keep the cpu suggestion (honest for both — llama-server runs on CPU, and for no-rocm the device label is the defect); only the rocm lane loses it. Sentence composition/punctuation is right in both branches. Note the gpu-vulkan + rocm lane (a picker-labeled Qwen3-TTS slot) also falls through to this gate and now gets the ROCm-only wording — correct for these images too.
- Acceptance pins: refusal message contains slot name, device=gpu-rocm, the kfd path, and the dev1: forward remedy; contains neither '1888' nor device='cpu'. ROCm-box no-regression holds — KFD_OK returns at _gpu.py:797 before any changed text, and the #2054 llama ladder is untouched.
- New tests are hermetic: kfd_status short-circuits to KFD_MISSING on a nonexistent tmp path before uid resolution, SLOT_RUNNER_UID is the int 0 (no sudo/podman round-trip), env={} blocks a HAL0_ALLOW_VULKAN_FALLBACK leak from the host environment. Existing rocm-lane tests only assert 'pct stop/start', which survives, so no collateral breakage.
- Seed TOML comment claims are true against the files: neither img.toml nor qwen3tts.toml ships a [model].default pin, so the inactive-tile story is accurate. No scope creep; no shell diffs, so no quoting/set -e or KEEP_TMP concerns.

Two non-blocking nits:
- _gpu.py:873 — 'no other device can serve it' is true of the shipped pinned-profile seeds (hand-editing device='cpu' with profile=qwen3-tts pinned would indeed launch a HIP image with no GPU), but the tts CAPABILITY does have a CPU lane: the picker's device switch remaps tts to the kokoro profile, and the cpu 'tts' sibling seed exists. The refusal for qwen3tts could point the operator at the kokoro sibling as the speech-on-CPU escape hatch; the seed TOML comment mentions it, the runtime error does not.
- installer/etc-hal0/slots/qwen3tts.toml:16 — 'the cpu tts/kokoro sibling above' reads as if the sibling were earlier in this file; it lives in tts.toml (the file's own header mentions it). Cosmetic.

Issue #2020, opened 2026-08-22
/v1 injects chat_template_kwargs.enable_thinking=false into every chat body — with response_format: json_object the seeded model hard-400s "Failed to initialize samplers", taking all memory retention and every JSON-mode client down on a fresh install


thinmintdev:
## GA blocker — found by rc-validate v1.0.0-rc.7 (kit v8)

**Severity:** ga-blocker · **Lane:** memory · **Register key:** `hindsight-response-format-400-via-thinking-policy` · **First seen:** v1.0.0-rc.7

### Summary

`hal0-api`'s `/v1/chat/completions` unconditionally injects `chat_template_kwargs: {"enable_thinking": false}` into every chat body that doesn't already carry the kwarg (`src/hal0/normalize/thinking.py::apply_thinking_policy`, called from `api/routes/v1.py`). Combined with `response_format: {"type":"json_object"}` on a Qwen3-family-templated model (the installer-seeded `hal0-brain-sft-*` is the ONLY model a fresh install ships), llama-server throws `Failed to initialize samplers: std::exception` and returns HTTP 400 — deterministically, A/B'd on byte-identical bodies (through hal0-api /v1: 400; direct to the same slot's own port: 200; with `chat_template_kwargs. enable_thinking: true` instead of false, direct to the slot: 200). Hindsight's own fact- extraction request never sends `chat_template_kwargs` — hal0-api adds it — so EVERY `retain_extract_facts` call against the seeded model 400s, and hindsight's poller retries forever. This is NOT hindsight-specific: any OpenAI-compatible client sending `response_format: json_object` through hal0's `/v1` against a Qwen3-templated model hits the identical 400, so the blast radius extends to any structured-output client, not just memory. Root cause is `hal0-brain-sft.jinja`'s think-block emission (`{%- if enable_thinking is false %}{{- '<think>\n\n</think>\n\n' }}`) colliding with the json_object grammar in llama.cpp b109 sampler init. Does NOT reproduce on production (ct105), because its bound extraction model has no such template branch — this is model- template-scoped, not backend/hardware-scoped, and every fresh install is maximally exposed because the brain-sft template is the only one staged.

### Repro

```sh
# A/B on an identical body, direct-to-slot vs through hal0-api (use the port a chat/utility
# llm slot serving a hal0-brain-sft-* model is bound to; find via `hal0 slot list`):
curl -s -X POST http://127.0.0.1:8080/v1/chat/completions -d \
[...]

PR #2074, opened 2026-08-26
feat(installer): post-install smoke — structured-output probe + hal0 update --check (#2066)

### LINKED ISSUE #2066: Post-install smoke: add structured-output probe and hal0 update --check probe
## Summary

The installer's verify step has two blind spots this release cycle proved expensive:

1. **Structured output.** The smoke probes coherence ("Paris") but never a `json_object` / structured-output request — which is how extraction has been silently dead on the fresh-install surface since the :0820 lineage. A working chat probe passes while every structured-output consumer is broken.
2. **Updater path.** It never dry-runs `hal0 update --check`, which is how #2052 shipped through three RCs — the install completes green with a broken update path, and the breakage is only discovered at the next release.

## Proposed

Add to the installer's post-install verify step:

- A structured-output probe: minimal `json_object` (or equivalent response-format) request against the fresh surface; assert parseable JSON comes back.
- An update-check probe: `hal0 update --check` dry-run; assert it exits cleanly (manifest fetch + cosign path exercised, nothing applied).

Both fail-soft consistent with the rest of the verify step (report loudly, don't necessarily abort the install) — but visible in the summary so an rc-validate run catches them.

## Acceptance

- Fresh install on a box with the extraction-dead condition reports the structured-output failure in the install summary.
- A broken updater path (e.g. #2052-class cosign gap) is surfaced at install time, not at next release.

thinmintdev:
## Summary

Adds two fail-soft probes to the installer's post-install verify step (end of the Service start branch in `installer/install.sh`), closing the two blind spots the v1.0 RC cycle proved expensive: a green install with structured output silently dead, and a green install with a broken updater (#2052 shipped through three RCs that way).

## Root cause

The verify step only checked that services came up (`wait_active`, `/health` polls). It never exercised a `json_object` request — so a coherent plain-chat reply masked a dead extraction path — and never dry-ran the updater, so update-path breakage stayed invisible until the next release.

## Fix

- **`smoke_structured_output`** — POSTs a chat completion **via the gateway** (`http://127.0.0.1:${HAL0_PORT}/v1/chat/completions`), never direct-to-slot (direct-to-slot 200s mask gateway-path failures, rc.7's #2020). The body is **kwarg-absent** — exactly `model` (`hal0/utility`, the canonical virtual alias with anchor fallback) + `messages` + `response_format: {"type": "json_object"}` — because extra kwargs mask template-branch differences between slot backends (probe shape per the issue comments, live-verified on ct151 during rc.9 validation). Asserts the reply *content* parses as JSON; an HTTP 200 carrying prose fails the probe. Reuses the same `/etc/hal0/api.env` key discovery as the hindsight block for KB-1 auth boxes.
- **`smoke_update_check`** — runs `hal0 update --check` (thin client over `/api/updates/check`; manifest fetch + cosign path exercised daemon-side, nothing applied) and asserts a clean exit.

Both are fail-soft consistent with the rest of the verify step: loud `warn` lines at probe time, and the failures are repeated **inside the "hal0 is ready" summary box** as a `Verify FAILED:` section (driven by `SMOKE_FAILED`), so an rc-validate run can't scroll past them. Neither probe can abort a finished install. No new `ui_step` (step total stays 16).

## Test evidence

New `tests/installer/test_install_smoke_probes.py` (written failing-first, 9 tests): contract tests that the probe targets the gateway port with a kwarg-absent `json_object` body, plus behavioral tests driving the extracted functions through fake `curl`/`hal0` shims (the `test_api_restart_on_upgrade.py` technique) — parseable-JSON content passes, a coherent non-JSON reply ("Paris") fails, gateway-down fails, `update --check` is invoked and its exit code respected, and both probes failing leaves the driver at exit 0 with both failures recorded and warned.

Ran on the dev Mac:

- `uv run --no-project --with pytest pytest tests/installer/test_install_smoke_probes.py` — 9 passed
- `pytest tests/installer/test_api_restart_on_upgrade.py test_install_single_entry_point.py test_install_cosign_persist.py test_hal0_update_wrapper.py` — 83 passed
- `bash -n installer/install.sh` — clean (shellcheck not installed locally)

Note: the full `tests/installer/` run has pre-existing failures on this macOS host (bash 3.2 lacks `mapfile`, Linux-only wrapper tests) unrelated to this diff — CI on Linux is the real gate there.

Fixes #2066

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -2952,6 +2952,77 @@ if [[ "${DEV_MODE}" -eq 0 ]]; then
     fi
 fi
 
+# ── Post-install smoke probes (#2066) ──────────────────────────────────────
+# Two blind spots the v1.0 RC cycle proved expensive:
+#   1. Structured output — a plain chat probe passes while every json_object
+#      consumer is broken (extraction silently dead since the :0820 lineage).
+#   2. The updater — #2052 shipped through three RCs because nothing ever
+#      dry-ran `hal0 update --check`; the breakage only surfaced at the NEXT
+#      release.
+# Both probes are fail-soft like the rest of the verify step: warn loudly and
+# repeat the failure inside the summary box (via SMOKE_FAILED), never abort a
+# finished install. Defined at top level so tests/installer/
+# test_install_smoke_probes.py can extract and drive them through fakes.
+SMOKE_FAILED=()
+
+smoke_structured_output() {
+    # Probe shape is load-bearing (live-verified on ct151, rc.9 validation):
+    #   * via the GATEWAY (:${HAL0_PORT}/v1), never direct-to-slot — a
+    #     direct-to-slot 200 masks gateway-path failures (rc.7's #2020);
+    #   * kwarg-ABSENT body — model + messages + response_format ONLY.
+    #     Extra kwargs (temperature, max_tokens, …) mask template-branch
+    #     differences between slot backends.
+    # hal0/utility is the canonical virtual alias (normalize/resolver.py):
+    # it routes to the cheap helper slot and falls back to the anchor, so
+    # the probe works on any box with at least one loaded model.
+    local body resp
+    body='{"model":"hal0/utility","messages":[{"role":"user","content":"Reply with only a JSON object of the form {\"ok\": true}."}],"response_format":{"type":"json_object"}}'
+    # KB-1 auth: same key discovery as the hindsight EnvironmentFile above —
+    # a fresh open-LAN box has no api.env and needs no header.
+    local -a auth=()
+    local key
+    if [[ -f /etc/hal0/api.env ]]; then
+        key="$(grep -oP '(?<=^HAL0_CLIENT_KEY=).*' /etc/hal0/api.env 2>/dev/null | head -1 || true)"
+        [[ -n "${key}" ]] && auth=(-H "Authorization: Bearer ${key}")
+    fi
+    resp="$(curl -fsS -m 120 -X POST "http://127.0.0.1:${HAL0_PORT}/v1/chat/completions" \
+        -H 'Content-Type: application/json' ${auth[@]+"${auth[@]}"} \
+        -d "${body}" 2>/dev/null)" || return 1
+    # The assertion: the CONTENT of the reply parses as JSON — an HTTP 200
+    # carrying prose ("Paris") is exactly the failure this probe exists for.
+    printf '%s' "${resp}" | "${PY}" -c 'import json, sys
+try:
+    reply = json.load(sys.stdin)
+    json.loads(reply["choices"][0]["message"]["content"])
+except Exception:
+    sys.exit(1)' 2>/dev/null
+}
+
+smoke_update_check() {
+    # Dry-run the updater end to end (manifest fetch + cosign path via the
+    # daemon's /api/updates/check); --check applies nothing. A clean exit is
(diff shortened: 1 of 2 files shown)


thinmintdev (commented):
Independent review (install-fix wave). Verified against the issue #2066 comment constraints and by re-running the PR's tests from the head SHA on a clean checkout.

**Constraint compliance (issue comments) — all met:**
- Probe goes via the gateway (`http://127.0.0.1:${HAL0_PORT}/v1/chat/completions`), not direct-to-slot: installer/install.sh:2989 (head). Matches the ct151/rc.9 constraint and the #2020 lesson.
- Body is kwarg-absent: exactly model + messages + response_format json_object (installer/install.sh:2979), and the contract test test_install_smoke_probes.py::test_structured_output_body_is_kwarg_absent_json_object json-parses the literal and asserts the key set — this is enforced, not just claimed.
- Asserts the reply *content* parses as JSON, so an HTTP 200 carrying prose fails (installer/install.sh:2994-3000).
- `hal0 update --check` exists with those exact semantics (src/hal0/cli/update_commands.py:245-248, check-only exits after /api/updates/check), and `hal0/utility` is a real virtual alias with anchor fallback (src/hal0/normalize/resolver.py:29). Neither claim is invented.

**Fail-soft / set -e:** install.sh runs under `set -euo pipefail`; both probes are only invoked as `if` conditions inside run_post_install_smoke, so errexit is suppressed and a failing probe cannot abort the install — confirmed by the behavioral test (both probes fail, driver exits 0, WARN on stderr, both names in SMOKE_FAILED). Summary-box block is guarded by `${#SMOKE_FAILED[@]}` so the empty-array "${SMOKE_FAILED[@]}" expansion is never reached on old bash. The `${auth[@]+"${auth[@]}"}` idiom is the correct set -u-safe form. DEV_MODE/--no-start skip the probes (call site is inside the service-start else branch, installer/install.sh:3413), and re-running the installer is idempotent — both probes are read-only.

**Test evidence honest:** I fetched the head SHA files and ran `uv run --no-project --with pytest pytest tests/installer/test_install_smoke_probes.py` — 9 passed — and `bash -n` on head install.sh — clean.

**Nits (none blocking):**
- installer/install.sh:2985 — key discovery reads only HAL0_CLIENT_KEY, not HAL0_ADMIN_KEY. Identical to the pre-existing hindsight discovery at line 3119, so consistent by design, but an admin-key-only api.env would send no auth header and report a structured-output failure that is really a 401. The warn text at least points at journalctl.
- installer/install.sh:2987 — `[[ -n "${key}" ]] && auth=(...)` is a latent errexit hazard if smoke_structured_output is ever called outside an if-condition with an empty key; harmless at every current call site (probe failure semantics want nonzero anyway), just worth knowing before reusing the function elsewhere.
- curl stderr is discarded (2>/dev/null), so a 401 vs connection-refused vs timeout all collapse into one failure line; the warn tells the operator to re-run by hand, which is acceptable for a smoke probe.
- test_smoke_failures_surface_in_the_summary_box is a textual grep of install.sh rather than a behavioral test of the summary rendering — weak, but the behavioral gap is small and the other 8 tests carry the real weight.

No scope creep (diff is exactly the two probes + summary wiring + tests; step total unchanged, no ui_step added). KEEP_TMP/debug paths untouched. LGTM as comment-only review — nits are optional follow-ups.

Issue #1584, opened 2026-08-01
installer: hermes toolchain apt-get races unattended-upgrades dpkg lock on fresh boots


thinmintdev:
Hit during the v1.0.0-rc.1 fresh-install test on CT151 (Ubuntu 26.04, pristine box, installer run ~10 min after boot).

## What happened

Step 16's hermes provisioning failed:

```
[hermes-prereqs] apt-get install -y python3 python3-venv python3-pip pipx
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 6593 (unattended-upgr)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
[hermes-prereqs] ERROR: toolchain install failed.
```

`unattended-upgrades` runs on every fresh Ubuntu boot, so a fresh install shortly after provisioning a box will commonly lose this race. The installer's own earlier apt phase won the lock; the hermes prereq script ran later and lost it.

## Impact

Degrades gracefully — install exits 0, hermes left unprovisioned with a clear remediation line (`hal0 agent install hermes`). Not release-blocking, but it makes the default first-boot experience flaky.

## Fix

Add `-o DPkg::Lock::Timeout=120` (apt >= 1.9.11) to the hermes-prereqs apt-get invocations — and to any other post-phase apt calls — so they wait for the lock instead of dying on it.

Found in: v1.0.0-rc.1 fresh-install validation (main @ b230af4f).

PR #2075, opened 2026-08-26
fix(installer): batch bootstrap dep preflight and auto-install via package manager

### LINKED ISSUE #2062: Bootstrap missing-dep preflight is one-at-a-time and never offers to install
## Summary

On minimal Ubuntu 26.04 (lacks both curl and jq), bootstrap died on curl; after hand-installing curl it died again on jq. Each missing dependency costs a full retry cycle, and the operator has to install by hand.

install.sh already has the install-or-fail pattern for exactly this (`preflight_venv`, `preflight_container_runtime`). Bootstrap should:

1. Batch-report ALL missing deps in one pass instead of failing on the first.
2. Ideally auto-install them via the detected package manager (the installer already detects distro/package manager for its messages), falling back to a single copy-pasteable `apt install curl jq ...` line when it can't.

## Acceptance

- Minimal box missing curl+jq: one bootstrap run either installs both and proceeds, or fails once listing both with the exact install command.
- No behavior change on boxes with deps present.

thinmintdev:
## Summary

`installer/bootstrap.sh`'s dependency preflight now collects **all** missing dependencies in one pass and then either auto-installs the whole batch via the host's detected package manager, or fails **once** with a single copy-pasteable install command listing everything. Boxes with all deps present see no behavior change.

## Root cause

`preflight()` called `need <tool>`, and `need` `die`d on the first absent command. On a minimal Ubuntu box missing both curl and jq, bootstrap died on curl; after hand-installing curl it died again on jq — one full retry cycle per missing tool, with no install offer, even though the repo already has the install-or-fail pattern (`preflight_venv`, `preflight_container_runtime` in `installer/lib/preflight.sh`).

## Fix

- `need` now appends to `_MISSING_DEPS` instead of dying; `preflight()` checks curl/tar/sha256sum/jq/python3 in one pass (the `need curl` / `need tar` / … lines are kept verbatim so the parity contract in `test_bootstrap_prereq_parity.py::TestMirrorsBootstrapSh` still holds).
- New `install_missing_deps()` mirrors the `preflight_venv` install-or-fail pattern: one package-manager transaction for the whole batch (non-interactive apt with the #1584 DPkg lock timeout; dnf/yum/zypper/pacman/apk supported), then a `command -v` re-check as the source of truth.
- Anything still missing (install failed, ran unprivileged, unrecognized manager) fails once, listing every dep (`missing dependency: X` per line, preserving the message contract of `test_bootstrap_contract.py`) plus one sudo-prefixed install command built from per-manager package names (`sha256sum` → `coreutils`, Arch `python3` → `python`).
- Package-manager detection is inline (not sourced from `installer/lib/distro.sh`) because bootstrap runs standalone via `curl|bash` before any repo file exists on disk; it mirrors `distro.sh`'s `pkg_mgr()` preference order.

## Test evidence

Test-first: `tests/installer/test_bootstrap_dep_preflight.py` was written before the fix and failed 5/7 against the old one-at-a-time behavior; all 7 pass with the fix (hermetic PATH, faked `uname`, stub `apt-get`). It covers: batch reporting of curl+jq in one run, single copy-pasteable command on failed install, exactly one package-manager transaction, auto-install success path, and the no-op path when deps are present.

Neighbors: `test_bootstrap_contract.py` / `test_bootstrap_prereq_parity.py` / `test_bootstrap_cosign_fetch.py` show **identical** failure sets before and after the change on this macOS dev host (63 pre-existing environment failures — the suites use the real `uname`, which reports Darwin, so they only pass on Linux; verified `test_bootstrap_missing_jq_fails_preflight_before_network` fails at the Linux gate on baseline too). The static contract tests (`TestMirrorsBootstrapSh`, `test_bootstrap_requires_jq_and_artifact_bundle_without_detached_fallback`) pass. `bash -n installer/bootstrap.sh` is clean; shellcheck is not installed on this host — relying on CI for the Linux run of the full installer suite.

Fixes #2062

### DIFF
--- a/installer/bootstrap.sh
+++ b/installer/bootstrap.sh
@@ -171,7 +171,8 @@ dep_package() {
 
 # The one-liner an operator runs to install PKG… with the detected manager.
 # Mirrors distro.sh's pkg_install_cmd, including the apt DPkg lock timeout
-# (#1584: unattended-upgrades commonly holds the lock on fresh Ubuntu boots).
+# that #1733 added there for #1584 (unattended-upgrades commonly holds the
+# lock on fresh Ubuntu boots) — keep the two copies in step if either changes.
 pkg_install_hint() {
     local IFS=' '   # global IFS is \n\t; the hint must stay a one-liner
     local pm="$1"; shift
--- a/installer/bootstrap.sh
+++ b/installer/bootstrap.sh
@@ -130,17 +130,122 @@ banner() {
 }
 
 # ── preflight ──────────────────────────────────────────────────────────────
+# Missing deps are collected, not fatal one-by-one (#2062): a minimal host
+# lacking curl AND jq should learn about both in a single run instead of
+# paying one retry cycle per tool. preflight() then either auto-installs
+# the whole batch via the host's package manager or fails once with a
+# single copy-pasteable install command covering everything.
+_MISSING_DEPS=()
 need() {
-    command -v "$1" >/dev/null 2>&1 || die "missing dependency: $1 — install it and re-run"
+    command -v "$1" >/dev/null 2>&1 || _MISSING_DEPS+=("$1")
+}
+
+# Package-manager detection. Inline — NOT sourced from installer/lib/distro.sh
+# — because this script runs standalone via curl|bash before any repo file
+# exists on disk. Mirrors distro.sh's pkg_mgr(): command presence beats
+# os-release IDs for derivatives (CachyOS ships pacman, PikaOS ships apt),
+# first match in preference order wins.
+detect_pkg_mgr() {
+    local m
+    for m in apt-get dnf yum zypper pacman apk; do
+        if command -v "${m}" >/dev/null 2>&1; then
+            printf '%s\n' "${m}"
+            return 0
+        fi
+    done
+    return 1
+}
+
+# Command name → package name for the detected manager. sha256sum ships in
+# coreutils everywhere; Arch names its Python package "python".
+dep_package() {
+    local pm="$1" dep="$2"
+    case "${dep}" in
+        sha256sum) printf 'coreutils\n' ;;
+        python3)
+            if [[ "${pm}" == "pacman" ]]; then printf 'python\n'; else printf 'python3\n'; fi
+            ;;
+        *) printf '%s\n' "${dep}" ;;
+    esac
+}
+
+# The one-liner an operator runs to install PKG… with the detected manager.
+# Mirrors distro.sh's pkg_install_cmd, including the apt DPkg lock timeout
+# (#1584: unattended-upgrades commonly holds the lock on fresh Ubuntu boots).
+pkg_install_hint() {
+    local IFS=' '   # global IFS is \n\t; the hint must stay a one-liner
+    local pm="$1"; shift
+    case "${pm}" in
+        apt-get) printf 'sudo apt-get -o DPkg::Lock::Timeout=120 install -y %s\n' "$*" ;;
+        dnf)     printf 'sudo dnf install -y %s\n' "$*" ;;
+        yum)     printf 'sudo yum install -y %s\n' "$*" ;;
+        zypper)  printf 'sudo zypper install -y %s\n' "$*" ;;
+        pacman)  printf 'sudo pacman -S --noconfirm %s\n' "$*" ;;
+        apk)     printf 'sudo apk add %s\n' "$*" ;;
+    esac
+}
+
+# Batch-resolve everything need() collected — the install-or-fail pattern of
+# install.sh's preflight_venv / preflight_container_runtime. One package-
+# manager transaction for the whole batch, then a re-check; anything still
+# missing (install failed, ran unprivileged, no recognised manager) fails
+# ONCE, listing every dep plus the exact command that installs them all.
+install_missing_deps() {
+    local IFS=' '   # global IFS is \n\t; keep "${arr[*]}" messages one-line
+    local pm="" dep pkgs=()
+    pm="$(detect_pkg_mgr)" || pm=""
+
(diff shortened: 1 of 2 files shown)


thinmintdev (commented):
Independent review of the #2062 fix. The core change is sound: need() batches into _MISSING_DEPS, preflight() keeps the five need lines verbatim (parity regex in test_bootstrap_prereq_parity.py::TestMirrorsBootstrapSh still matches — verified the non-greedy preflight() body regex against the new function), install_missing_deps mirrors the preflight_venv install-or-fail shape with the command -v re-check as source of truth, and both issue acceptance criteria are met (batch report + single copy-pasteable command, or auto-install-and-proceed; verified no-op path leaves boxes with deps present untouched, with a tripwire test proving no package-manager call). set -e/IFS handling is careful: || true on pm transactions, || pm="" on detection, local IFS=' ' scoped to the functions that join arrays. Idempotent re-run and KEEP_TMP paths unaffected (preflight runs before workdir/trap setup). Test claims check out against the actual test file (6 tests + bash -n = 7).

Findings, blocking first:

1. installer/bootstrap.sh:3-6 — the file header mandates that this script stay byte-identical with Hal0ai/hal0-web:public/install.sh and be synced "in the same PR". This PR only edits the canonical copy and neither links a companion hal0-web PR nor mentions the mirror. Without the sync, the actually-served https://hal0.dev/install.sh keeps the old one-at-a-time behavior and the fix never reaches the users the issue describes. Please open/link the hal0-web companion PR.

2. CI: the python (3.12) job fails on the Format check — "Would reformat: tests/installer/test_bootstrap_dep_preflight.py" (run 32929160451). Run the repo formatter over the new test file and repush; everything else in that job never ran, so the new suite has not actually executed on Linux CI yet. Given the PR's own test evidence leans on "relying on CI for the Linux run", a green Linux CI run is needed before merge.

3. installer/bootstrap.sh (pkg_install_hint comment) — minor: the comment says the hint "mirrors distro.sh's pkg_install_cmd, including the apt DPkg lock timeout", but installer/lib/distro.sh:93-104 pkg_install_cmd has no DPkg::Lock::Timeout (grepped the repo: this PR introduces the only occurrence). The timeout itself is right per #1584; just reword the comment (or add the timeout to pkg_install_cmd in a follow-up) so a future sync doesn't "fix" the divergence in the wrong direction.

4. Non-blocking observation: when run unprivileged with a detected manager, the script now attempts apt-get update/install as non-root and lets them fail before the clean single-shot failure — functional (re-check decides, sudo hint is printed), but the apt permission noise lands on the operator's terminal. An EUID check to skip the doomed transaction would be quieter; fine to leave as-is.

5. Non-blocking: behavior-change note for the missing-dep+package-manager-present case — preflight now performs network I/O (apt-get update) before any manifest fetch. That is the intent of the issue and the hermetic contract test (missing jq fails before network) still holds because it has no package manager on PATH; flagging only so the trust-boundary comment at the top of the script stays honest about what runs pre-verification.

Verdict: needs the mirror sync (1) and a green Linux CI run after the format fix (2); otherwise this is a good, well-tested fix.

thinmintdev:
Addressed the review — thanks for the thorough pass.

**1. Mirror sync (blocker) — no companion hal0-web PR is needed; the sync contract is post-merge automation, and a hand-sync PR is explicitly forbidden.**

The header quoted in the review ("sync the other in the same PR") is the *old* wording — it survives only in the served copy at hal0-web:`public/install.sh` (last hand-synced in hal0-web#44) and in the prose of `scripts/check-bootstrap-parity.sh`. The current canonical header on this branch (`installer/bootstrap.sh:3-8`) says:

> `.github/workflows/mirror-bootstrap.yml` publishes this file to the web copy on every main push, gated on the stable channel manifest being signed. **Never hand-edit the web copy; it is overwritten.**

Evidence in `.github/workflows/mirror-bootstrap.yml`:
- lines 30-35: triggers on `push` to `main` with path `installer/bootstrap.sh` — i.e. the moment this PR merges;
- lines 64-85: publish gate — it curls `https://releases.hal0.dev/stable.json.bundle` and only publishes on HTTP 200;
- lines 95-121: copies `bootstrap.sh` → `hal0-web:public/install.sh` (`bash -n` sanity check) and pushes to `master` as `hal0-docs-mirror[bot]`.

Current gate status: `stable.json.bundle` → **404**, so the mirror will record a skip rather than publish. That is the workflow's designed behavior (lines 9-19): the canonical bootstrap is fail-closed and requires a Sigstore bundle next to the channel manifest; publishing it while `stable` still points at v0.9.8 (unsigned) would break every one-line install. The gate opens automatically when the signed stable v1.0.0 is published (`release: published` + the `workflow_call` from release.yml re-run it at that instant), and the daily `bootstrap-parity.yml` reports drift in the interim. Opening a manual hal0-web PR now would both violate the "never hand-edit" rule and ship exactly the breakage the gate exists to prevent.

**2. ruff format (blocker)** — ran the repo formatter over `tests/installer/test_bootstrap_dep_preflight.py` (`uvx ruff format` + `ruff check` clean) and re-ran the suite: 7/7 green locally. Pushed for a fresh Linux CI run.

**3. DPkg-timeout comment (minor)** — checked before editing: `installer/lib/distro.sh` `pkg_install_cmd` **does** carry `-o DPkg::Lock::Timeout=120` on its apt-get branch (distro.sh:105, added by #1733 for #1584; it's one line below the 93-104 range cited, which may be why the grep missed it — `installer/install.sh` and `installer/agents/hermes-prereqs.sh` carry it too). So the comment was accurate and I did not invert it; I did reword it to cite #1733 explicitly so a future sync can't "fix" the parity in the wrong direction.

Findings 4 and 5 acknowledged as non-blocking; leaving as-is per the review.

Issue #2056, opened 2026-08-26
hal0-combined:0824 + seeded brain model: every response_format request 400s at sampler init — fresh-install memory extraction 100% dead


thinmintdev:
## Summary

On the `ghcr.io/hal0ai/hal0-combined:0824` runner image (the rc.9 default), any structured-output request (`response_format: json_object` or `json_schema`) against the installer-seeded `hal0-brain-sft-q8-rocmfpx` model fails at sampler initialization:

```
HTTP 400 {"error":{"code":400,"message":"Failed to initialize samplers: std::exception","type":"invalid_request_error"}}
```

This is deterministic (temp 0), direct-to-slot (the gateway is not the mechanism), and independent of `chat_template_kwargs.enable_thinking` (true, false, and absent all 400). Plain chat completions on the same slot return 200 and coherent text (coherence canary passes).

Because `HINDSIGHT_API_LLM_BASE_URL` points at the hal0 gateway and fact extraction uses JSON mode, **every `retain_extract_facts` call on a fresh rc.9 install fails** — the journal on a fresh ct151 shows `Fact extraction failed: 1/1 chunks failed ... 400 Failed to initialize samplers` and operations marked failed. Blast radius is the same as rc.7's GA-blocker `hindsight-extraction-samplers-400` (#2020): all memory retention plus any OpenAI-compatible client requesting JSON mode against the only model a fresh install ships.

## Not a #2020 regression — different mechanism

#2020/#2036 was gateway kwarg injection: direct-to-slot succeeded, and `enable_thinking=true` succeeded. Here the A/B does not flip on either axis — direct-to-slot 400s and all kwarg variants 400. The mechanism is in the `:0824` image's llama-server build: grammar/sampler init throws for this model's template (or quant) whenever a grammar is requested.

## Evidence matrix (2026-08-26, all direct to slot port, temp 0)

| Box | Slot | Image | Backend | Model | json_object |
|---|---|---|---|---|---|
| ct151 (fresh rc.9) | brain :8089 | combined:0824 | Vulkan (no /dev/kfd in ctr) | hal0-brain-sft-q8-rocmfpx | **400** |
| ct150 (update rc.9) | agent :8081 | combined:0824 | ROCm (/dev/kfd) | hal0-brain-sft-q8-rocmfpx | **400** |
| ct150 | brain :8089 | rocmfpx:ade07ba (drift residue) | ROCm | hal0-brain-sft-q8-rocmfpx | 200 |
[...]

PR #2076, opened 2026-08-26
feat(install): LFM2.5-2.6B Q8_0 as the default seeded brain model

thinmintdev:
## DO NOT MERGE before #2073

The runner pin to `hal0-combined:0826` (PR #2073) is a hard prerequisite: LFM2.5's native tool-call parsing and think-extraction need that image. Draft until #2073 merges; validate the combination in the rc.10 fresh lane. Trees are independent (installer/ + registry vs packaging/runner pins), so no merge conflict is expected.

## What changes

- **Curated entry `lfm2.5-2.6b`** — `LiquidAI/LFM2.5-2.6B-GGUF` / `LFM2.5-2.6B-Q8_0.gguf`, 2.87 GB, context 131072, `tool_calling=True`, `architecture=lfm2`. The pull layer verifies the HF LFS sha256 automatically.
- **`install/brain_model.py`** — new `BRAIN_MODEL_DEFAULT = "lfm2.5-2.6b"`, selected on every hardware class: plain Q8_0 GGUF loads on the FPX runner and stock llama.cpp alike, so the ROCm-vs-CPU fork collapses to override-only. The `hal0-brain-sft` ids remain valid `HAL0_BRAIN_MODEL` overrides with their existing runner constraints. The install step now discloses the download size (2.87 GB, up from the 1.1 GB sft Q8 — per operator decision: brain stays default-install, size disclosed).
- **`profile.brain`** — drops `--reasoning off --reasoning-format none`. LFM2.5 emits `<think>` from weights and `enable_thinking=false` does not suppress it, so that pair leaked raw think prose into `message.content`; the runner-default reasoning-format extracts it cleanly into `reasoning_content`. Cost note documented in the profile: extraction/JSON calls pay ~100-300 think tokens per call before the grammar-constrained output starts.
- **`[brain_chat] tool_model` routing unchanged** (default `hal0/agent`). Flipping steward tool turns to the brain natively is the rc.10 steward re-eval's call, not this PR's.
- brain.toml / install.sh comments updated; the model-less-seed + stamp-after-pull design (#1108/#1369) is untouched.

## Provenance (verified tonight)

- HF LFS sha256 `1e22128dfa128bdfb684da167e74e072d0a056baa7d06d9f280291e2839b0fc9`, 2,874,779,648 bytes — exact match to the ct150 reference copy (sha256sum run on the box).
- GGUF header read from the actual file: `general.architecture=lfm2`, `lfm2.context_length=131072`, embedded chat template present and carries the `List of tools:` parser-detection string.
- Validation matrix on `:0826` (ct150, rc.9 lane): temp-0 canary, `json_object`, `json_schema`, degenerate contentless prompt, native `tool_calls` — all 200.

## Tests

- `tests/install/test_brain_model.py` rewritten contract-first: selection matrix asserts every hardware class gets the default; pullability parametrized per repo (LiquidAI for the default, the public sft repo for overrides); provisioning reuse/untrustworthy-file tests moved to the default id.
- `tests/config/test_schema_seeds_c5.py` brain-seed test updated the same way (seed still ships model-less).
- Batch result: 143 passed (install/registry/slots/config target files); wider sweep `tests/install tests/registry tests/slots tests/config tests/brain`: 2579 passed, 10 skipped. Ran on macOS via a local-disk uv env; CI is the authoritative Linux run.

Depends-on: #2073. Related: #2056 (parser fix lineage), #2054 (seed device derivation — untouched here).

### DIFF
--- a/installer/etc-hal0/slots/brain.toml
+++ b/installer/etc-hal0/slots/brain.toml
@@ -18,20 +18,23 @@
 # `[model].default` only once the bytes have landed. That is why no
 # `[model].default` is hardcoded below:
 #
-#   * The right variant is hardware-dependent. All three live in the public
-#     repo Hal0ai/hal0-brain-sft-ROCmFPX-GGUF, but the Q8/Q4 builds carry
-#     CUSTOM GGML tensor type ids (103 / 100) that STOCK llama.cpp REJECTS —
-#     they need the ROCmFPX runner. A GPU box gets the small, fast
-#     `hal0-brain-sft-q8-rocmfpx` (agent/tool-use preset); a box with no
-#     ROCm/Vulkan device gets the portable `hal0-brain-sft-f16`. Pinning
-#     either one here would crash-loop the other kind of box.
+#   * The default is `lfm2.5-2.6b` (LFM2.5-2.6B Q8_0, 2.87 GB — plain GGUF,
+#     loads on every runner class, so it is no longer hardware-forked). The
+#     hal0-brain-sft variants remain as HAL0_BRAIN_MODEL overrides, and THEY
+#     are hardware-constrained: the sft Q8/Q4 builds carry CUSTOM GGML
+#     tensor type ids (103 / 100) that STOCK llama.cpp REJECTS (ROCmFPX
+#     runner only); only the sft F16 is portable. Pinning any id here
+#     instead of letting the installer stamp it would still race the pull.
 #   * Model-presence IS the activation signal (#1369), so an id written
 #     before the file exists on disk is precisely the start-before-model
 #     race (#1108). If the pull cannot happen (no network, no space) the
 #     slot stays model-less and the install still succeeds — a grey tile,
 #     not a crash loop.
 #
-# WHAT REPLACED WHAT — the previous seed pinned `MiniCPM5-1B-Agentic-Tooluse`.
+# WHAT REPLACED WHAT — lineage: MiniCPM5-1B pin → hal0-brain-sft variants
+# (hardware-forked) → `lfm2.5-2.6b` default (rc.10, #2073 runner pin is the
+# prerequisite). The sft story below is kept for the override path.
+# The previous seed pinned `MiniCPM5-1B-Agentic-Tooluse`.
 # That is a REAL, PULLABLE model: the public GGUF repack of the upstream
 # tool-use base at ewinregirgojr/MiniCPM5-1B-Agentic-Tooluse-GGUF
 # (`...Nemotron-DPO.F16.gguf`, 2,166,552,576 B, anonymous read). Moving to the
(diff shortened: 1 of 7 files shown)


thinmintdev:
Gate update: #2073 merged (2026-08-26T04:18Z) — the :0826 pin is on main. Remaining gate is the rc.10 fresh-lane validation; per the release session's sequencing this PR flips to ready and merges before the rc.10 version-bump PR so the fresh lane exercises the LFM seed + :0826 together.

Renderer nit from review checked and confirmed clean: no Python in the tree references `--reasoning`/`reasoning_format` — slot argv is assembled purely from the profile flag strings (`seed_profiles.toml`) plus model `defaults.extra_args`, and `normalize_argv` only collapses duplicates, never injects defaults. `normalize/thinking.py` is request-side (`chat_template_kwargs`), not argv. So with the pair dropped from `profile.brain`, the slot launches with llama-server's own default reasoning-format — the exact configuration verified on :0826 to extract think text into `reasoning_content`.

thinmintdev:
Gate status: v1.0.0-rc.10 is cut and published WITHOUT this PR (tag on e598eec1, operator ruling — the fresh lane validates the known brain-sft + :0826 combination first, so a lane failure means release, not new-model ambiguity). This PR flips to ready and lands for rc.11 once the rc.10 ct151 fresh lane passes. Full cut + lane handoff: /mnt/mintdev/artifacts/hal0-rc10-handoff-2026-08-26.md.

thinmintdev:
Independent review from the rc.11 sequencing session (posted as a comment — GitHub refuses formal reviews on same-account PRs, per the #2082 precedent). **Approving on substance**; merging once CI is green.

What I verified rather than took on trust:

1. **No semantic collision with #2088.** Both PRs touch something called "the brain profile", which is the kind of overlap that survives a clean textual merge and breaks at runtime. They are different objects: #2088 fixed `memory_headers` in `src/hal0/agents/hermes_provision.py` (the `X-hal0-Private` int→string coercion), while this PR rewrites `[profile.brain]` in `src/hal0/config/data/seed_profiles.toml` (llama-server launch flags). No file overlap, and the `"1"` string fix is intact on the merged tree.
2. **Branch updated onto current main** (`963c412e`, which carries #2087 + #2088) and the merged result exercised locally: `tests/install tests/config tests/installer tests/agents tests/registry tests/release` → **3300 passed**, 11 skipped, 1 xfailed. The only failure was `test_avahi_hostname.py::test_no_avahi_on_the_box_is_a_silent_noop`, which fails identically on clean main on that workstation (it runs avahi) — environment-shaped, CI is authoritative.
3. **The pull target is real and anonymously reachable** — this was the open provenance question when the work was handed over. `LiquidAI/LFM2.5-2.6B-GGUF` → `gated: false`, `LFM2.5-2.6B-Q8_0.gguf` present in the sibling list, HTTP 302 to the LFS CDN on an unauthenticated HEAD. Combined with the sha256 you verified against the ct150 copy, the fresh-lane pull should not 404.
4. **Fixed the CI format gate** (`ruff format` on `src/hal0/install/brain_model.py` and `tests/config/test_schema_seeds_c5.py` — pushed as a separate commit; `ruff check` and `format --check` now clean across `src tests`, and `tests/install tests/config` re-run green at 1067 passed).

The one behavioural change worth naming for the lane: dropping `--reasoning off --reasoning-format none` from `profile.brain` is what routes think text into `reasoning_content` instead of `message.content`. That is the correct call for this model and matches what was verified on `hal0-combined:0826`, but it means the rc.11 fresh lane's steward probes should read `reasoning_content` explicitly rather than assuming content-only. Noted in the lane plan, not a change request here.

PR #2077, opened 2026-08-26
fix(cli): run hermes gateway install as root on the root path (#2061)

### LINKED ISSUE #2061: Hermes gateway bridge sub-install falsely demands sudo when the installer already runs as root
## Summary

During an rc.9 install running entirely as root, the bridge sub-install refused to proceed:

```
System gateway install requires root. Re-run with sudo.
Telegram/Discord bridge unavailable; continuing
```

(rc.9 log line 2317.) The same message appeared in the rc.7 install log, so this is persistent, not a fluke. The core hermes-gateway unit installs and works fine — only the bridge sub-install's root detection is wrong, so Telegram/Discord bridges are silently skipped on every install.

## Expected

A root installer run should pass the bridge sub-install's privilege check. Whatever the check inspects (EUID vs UID, SUDO_USER presence, a re-exec boundary that drops the detection context) needs to align with how install.sh actually invokes it.

## Acceptance

- Fresh root install: bridge sub-install proceeds without the sudo demand; Telegram/Discord bridge components present.
- The "unavailable; continuing" fallback still fires when genuinely unprivileged.

thinmintdev:
## Summary

During rc.7 and rc.9 installs running entirely as root, the hermes gateway bridge sub-install refused with `System gateway install requires root. Re-run with sudo.` (rc.9 log line 2317), so the Telegram/Discord bridge was silently skipped on every install. This PR makes the bridge sub-install run as root when the caller is root, restoring the behavior PR #1337 already established for `installer/install.sh`'s inline gateway block.

## Root cause

`hermes gateway install --system` checks `os.geteuid() == 0` internally and refuses a non-root caller — that is where the "Re-run with sudo" message comes from. `installer/install.sh` learned this in PR #1337 and runs the inline gateway block as root. But the CLI path `_install_hermes_gateway()` in `src/hal0/cli/agent_commands.py` — which `hal0 agent install hermes` invokes, and which install.sh shells out to during provisioning — later reintroduced a privilege drop: when `euid == 0` it routed the hermes invocation through `_run_as_hal0()` (an "m1" change aimed at the #843 stray `/root/.hermes` concern). Dropping to the unprivileged `hal0` user made hermes' internal root check refuse the `--system` install, so the sub-install failed on exactly the runs that had the privileges it demands.

## Fix

- `_install_hermes_gateway()` now runs `hermes gateway install --system --run-as-user hal0` as root directly when `euid == 0`, mirroring install.sh's inline block: `HERMES_HOME` is stripped from the env and `HOME` is pinned to the hal0 service user's home (`/var/lib/hal0` fallback), which satisfies the #843 stray-`/root/.hermes` safeguard without a privilege drop. `--run-as-user hal0` continues to name the runtime user baked into the systemd unit.
- The genuinely-unprivileged path is unchanged: a non-root run still invokes hermes as the current user, hermes refuses, and the existing "Telegram/Discord bridge unavailable; continuing" fallback fires.
- The retry hint printed when the unit is missing no longer suggests `sudo -u hal0 ...` (the exact invocation hermes refuses); it now suggests the root invocation with the sanitized env.
- Factored the hal0-home lookup into `_agent_home()`, shared by `_run_as_hal0()` and the gateway install.

## Test evidence

Test-first: replaced `test_install_hermes_gateway_drops_to_hal0_when_root` (which enshrined the regression) with `test_install_hermes_gateway_runs_as_root_directly`, asserting that as root the gateway install runs via a bare `subprocess.run` with `HOME=/var/lib/hal0`, no `HERMES_HOME`, `stdin=DEVNULL`, and never routes through `_run_as_hal0`.

- `uv run pytest tests/cli/test_agent_install_hermes.py` — 36 passed (the new test fails against the pre-fix source, passes with the fix).
- `uv run pytest tests/cli/test_agent_shim.py` — 53 passed, 4 skipped (Linux-only /proc scans; run on macOS).
- `ruff check` and `ruff format --check` clean on both touched files.

Note: run on a macOS dev box, so the Linux-only shim tests skipped; CI covers those.

Fixes #2061

### DIFF
--- a/src/hal0/cli/agent_commands.py
+++ b/src/hal0/cli/agent_commands.py
@@ -547,6 +547,16 @@ def _hermes_root_prelude(terminal_env: dict[str, str] | None = None) -> None:
             console.print(f"[dim]hermes prelude: could not pre-own {d}: {exc}[/dim]")
 
 
+def _agent_home() -> str:
+    """The hal0 service user's home dir (``/var/lib/hal0`` fallback)."""
+    import pwd as _pwd
+
+    try:
+        return _pwd.getpwnam(_AGENT_RUNTIME_USER).pw_dir or "/var/lib/hal0"
+    except KeyError:
+        return "/var/lib/hal0"
+
+
 def _run_as_hal0(
     argv: list[str], *, stdin: int | None = None, extra_env: dict[str, str] | None = None
 ) -> int:
@@ -568,14 +578,10 @@ def _run_as_hal0(
     three tools — the ``sudo`` fallback resets the environment, and relying on
     inheritance would silently lose it there (see HAL0_HERMES_TERMINAL).
     """
-    import pwd as _pwd
     import shutil as _shutil
     import subprocess as _subprocess
 
-    try:
-        home = _pwd.getpwnam(_AGENT_RUNTIME_USER).pw_dir or "/var/lib/hal0"
-    except KeyError:
-        home = "/var/lib/hal0"
+    home = _agent_home()
     assignments = [f"{k}={v}" for k, v in sorted((extra_env or {}).items())]
     env_argv = ["env", "-u", "HERMES_HOME", f"HOME={home}", *assignments, *argv]
     if _shutil.which("runuser"):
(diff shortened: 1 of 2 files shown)


thinmintdev (commented):
Independent review (install-fix wave), comment-only.

**Verdict: LGTM.** The fix matches the issue and the established install.sh behavior; test claims reproduce independently.

Verified:
- Root cause confirmed against the repo: installer/install.sh:2264-2274 (the PR #1337 inline block) documents hermes' internal `os.geteuid() == 0` check and runs `env -u HERMES_HOME HOME=/var/lib/hal0 hermes gateway install --system --run-as-user hal0` as root. The pre-fix CLI path routed the same invocation through `_run_as_hal0`, which hermes then refuses — exactly the rc.7/rc.9 "Re-run with sudo" symptom in #2061. The PR makes src/hal0/cli/agent_commands.py:811-820 mirror install.sh exactly (HERMES_HOME popped, HOME pinned to hal0's home when euid==0, bare subprocess.run).
- Acceptance criteria: root path now passes hermes' privilege check; the genuinely-unprivileged path is unchanged (hermes refuses, rc != 0, and the existing "unavailable; continuing" fail-soft at agent_commands.py:821-825 plus unit-missing warning still fire). Best-effort/warn-and-continue posture preserved throughout; drop-in pre-write ordering and foreign-gateway detection untouched.
- Test evidence reproduced on a fresh clone of the PR head: tests/cli/test_agent_install_hermes.py 36 passed; tests/cli/test_agent_shim.py 53 passed 4 skipped; ruff clean on both touched files. The new test also fails against the pre-fix source as claimed.
- No other call sites drop to hal0 for `gateway install` (grepped src + installer); no scope creep — the `_agent_home()` extraction is a minimal shared helper; no shell/set -e hazards (Python-only diff).

Notes, non-blocking:
- src/hal0/cli/agent_commands.py:802-806 — "the #843 safeguard is satisfied by env alone" is slightly overstated: HOME pinning prevents the stray /root/.hermes row, but check_hermes_ownership (src/hal0/cli/doctor_commands.py:532) also flags root-owned files inside /var/lib/hal0/.hermes as drift. If `hermes gateway install` ever writes into ~/.hermes as root, those files land root-owned in hal0's tree. install.sh has accepted this same trade-off since #1337 (provisioning creates ~/.hermes as hal0 first), so this is consistency, not a new risk — just worth remembering if doctor perms ever reports drift under .hermes post-install.
- tests/cli/test_agent_install_hermes.py:389 — against the pre-fix source the new test fails via AttributeError on the `_agent_home` monkeypatch rather than the `_run_as_hal0` boom assertion. Regression proof still holds (pre-fix code routes through `_run_as_hal0`), but the failure mode is incidental rather than behavioral.
- src/hal0/cli/agent_commands.py:831 — the retry hint interpolates `_agent_home()` unquoted into a shell one-liner; fine for /var/lib/hal0, cosmetic only.

PR #2078, opened 2026-08-26
fix(installer): pin avahi host-name to HAL0_HOSTNAME so a custom <name>.local resolves

### LINKED ISSUE #2060: Installer: honor a custom hostname choice all the way to avahi (custom <name>.local instead of machine hostname)
## Summary

The plumbing for a custom hostname already exists end-to-end in config: `HAL0_HOSTNAME` env is honored by the installer derivation (`install/network.py:resolve_hostname`), the mDNS module (`services/mdns.py:mdns_hostname`), and the WS-origin allowlist, and the answer file supports `network.hostname` (`install/answers.py`). The base avahi service file uses the `%h` wildcard, so it follows the announced hostname automatically.

What is missing is the last hop: avahi announces the **machine** hostname from `/etc/hostname`, not `HAL0_HOSTNAME`. An operator who installs with `HAL0_HOSTNAME=jarvis` gets `jarvis.local` URLs in the dashboard, but the name never resolves on the LAN unless the machine hostname happens to match. It works today on boxes literally named `hal0` and silently breaks everywhere else.

## Proposed change

When the resolved hostname differs from the machine hostname, the installer should make avahi answer to it — either:

- write `host-name=<name>` under `[server]` in `/etc/avahi/avahi-daemon.conf` and reload avahi-daemon (scoped to mDNS only, does not rename the box), or
- run `hostnamectl set-hostname <name>` (renames the whole box — probably too invasive as a default).

The avahi conf edit is the safer default. Roughly 20–30 lines in `installer/install.sh`, guarded on avahi being present, idempotent on re-run.

UX stays non-interactive per the install.sh contract: `HAL0_HOSTNAME=jarvis sudo -E bash install.sh`, or `network.hostname` in `hal0-setup.yaml` — same pattern as HF_TOKEN.

## Wrinkles

1. `MDNS_ALIAS = "hal0.local"` is hardcoded in `install/network.py` and docs promise every box answers to `hal0.local`. Setting a custom avahi host-name drops that alias unless it is separately published (avahi CNAME publishing is messy). Suggest: renamed boxes simply stop answering to `hal0.local`; keep the alias in the origin allowlist (harmless) and note the behavior change in docs.
2. Upgrade path: re-running the installer with a different `HAL0_HOSTNAME` should update both `api.env` and the avahi conf, not leave them diverged.

## Acceptance

- Fresh install with `HAL0_HOSTNAME=jarvis`: `http://jarvis.local:8080` resolves and loads the dashboard from another LAN host; WS chat-proxy upgrade passes the origin gate.
- Install with no override: behavior unchanged (machine hostname announced).
- Re-run with a new name updates the announcement; avahi conf edit is idempotent.

thinmintdev:
## Summary

Closes the last hop of the custom-hostname story: `HAL0_HOSTNAME=jarvis sudo -E bash install.sh` now makes `http://jarvis.local:8080` actually resolve on the LAN, by pinning avahi's announced name to the resolved hostname.

Scope note (per review): this covers the **env path only**. `network.hostname` in the answer file is consumed by the separate `hal0 setup --answers` process (`setup_command.py` / `answers.py`) — `install.sh` never reads `hal0-setup.yaml`, so the answer-file path does not reach the avahi sync. That's the same gap that already exists for the api.env seeding; wiring the answer-file flow into the sync is left as a follow-up.

## Root cause

The custom-hostname plumbing already reached `api.env`, the mDNS URL building (`services/mdns.py`), and the WS-origin allowlist (`install/network.py`) — but avahi announces the **machine** hostname from `/etc/hostname`. The dashboard advertised `jarvis.local` URLs that only resolved on boxes literally named `jarvis`; everywhere else the name silently never resolved.

## Fix

New `sync_avahi_hostname()` in `installer/install.sh`, called right after the network-env derivation (prod installs only — `--dev` never elevates and must not touch `/etc/avahi`):

- When the resolved hostname differs from the machine hostname and avahi is present (conf file or `avahi-daemon` binary), it writes a marker-tagged `host-name=<name>` under `[server]` in `/etc/avahi/avahi-daemon.conf` — creating the file/section if needed, preserving every other line (including commented `#host-name=` examples) — then `systemctl try-restart avahi-daemon.service` (a restart, because SIGHUP/reload only re-reads the services dir, not the conf; try-restart is a no-op when the daemon isn't running).
- mDNS-scoped only: no `hostnamectl`, the box is never renamed.
- Re-run with a new `HAL0_HOSTNAME`: the marker-managed line is replaced. Re-run with the same name: byte-identical conf, no daemon bounce. Re-run with no override: the managed line is withdrawn; on that path an operator's own hand-written `host-name=` is never touched.
- Caveat (per review): while an override **is** in effect, pinning must replace an operator's pre-existing active `host-name=` under `[server]` (two `host-name` lines would be invalid conf), and a later no-override run withdraws only the managed pair — it does not restore the operator's earlier value. The installer now warns loudly when this clobber happens.
- Dotted overrides (reverse-proxy FQDNs) and `.local`-suffixed choices are normalized the same way `services/mdns.py` does (suffix stripped; dotted names skipped — not mDNS-announceable).
- Fail-soft throughout: no avahi, unwritable conf, a failed mktemp, or a failed final `mv` (e.g. an immutable conf) degrades to a warned no-op rather than aborting the install.

Documented wrinkle (in `docs/reference/env-vars.mdx`): a renamed box stops answering to the `hal0.local` alias — avahi has no clean CNAME publishing — while the alias stays in `HAL0_ALLOWED_ORIGINS`, which is harmless.

## Test evidence

Test-first: `tests/installer/test_avahi_hostname.py` (now 11 tests) was written before the fix and failed on the unfixed installer; after the fix all pass (11/11 locally). It extracts the function body and drives it against a tmp conf (`HAL0_AVAHI_CONF`) with fake `hostname`/`systemctl`/`avahi-daemon` shims on PATH — the same technique as `test_api_restart_on_upgrade.py`. Covers: the core pin, surgical edit (other sections/keys/comments survive), re-run update, idempotent no-bounce, withdrawal on no-override, operator-owned `host-name=` untouched on the no-override path, warn-on-clobber when pinning over an operator's `host-name=`, machine-name/dotted-name no-op, no-avahi no-op, `.local` suffix stripping, and that the call site sits after the network derivation.

Also run locally: `uvx ruff format --check` + `uvx ruff check` on the test file (clean — this fixes the red python-3.12 CI format gate), `bash -n installer/install.sh` (clean), and `tests/install/test_network.py` (11 passed). shellcheck is not installed on this machine. Note: the wider `tests/installer` suite has pre-existing failures in this macOS dev environment (bash 3.2 lacks `mapfile`, Linux-only tooling) that are unrelated to this change — relying on CI for those.

Fixes #2060


thinmintdev (commented):
Independent review (issue #2060 — no binding comment constraints on the issue; body only). Verification done: ran the 10 new tests against the PR head locally (10/10 pass), probed extra edge cases by driving the extracted function, and read network.py/mdns.py/answers.py at the head SHA.

The fix itself is sound: sync_avahi_hostname correctly pins host-name= under [server], the marker-managed pair updates/withdraws on re-run, an unchanged conf skips the daemon bounce (verified byte-identical + no systemctl call), the sed extraction matches network.py main()'s unquoted KEY=value output, the .local strip matches mdns.py, DEV_MODE guard and placement after NETWORK_ENV_LINES are correct, and try-restart-not-reload is the right call for a conf (not services-dir) change.

Findings, most severe first:

1. CI is red because of this PR: the python (3.12) job fails at ruff format --check — tests/installer/test_avahi_hostname.py:133 (the write_text line in test_operator_own_host_name_is_left_alone_without_override) exceeds the line length; ruff wants it wrapped. Because the format step fails, the pytest step never ran in CI, so the new tests are unverified there. Fix: run ruff format on the file and re-push.

2. Docs/summary overstate the answer-file path: env-vars.mdx (the HAL0_HOSTNAME row) and the PR summary say network.hostname in the answer file also pins avahi. It does not — the answer file is consumed by 'hal0 setup --answers' (setup_command.py), a separate process; install.sh never reads hal0-setup.yaml, and load_answers only exports into that process's os.environ (answers.py:_apply_network). Only the HAL0_HOSTNAME env install path reaches sync_avahi_hostname. The same gap already exists for the api.env seeding so behavior stays coherent, but the new doc sentence promises something the code cannot do — tighten it to the env path, or wire the sync into the answer-file flow in a follow-up.

3. installer/install.sh (pin branch of sync_avahi_hostname): with an override in effect, an operator's pre-existing hand-written host-name= under [server] is silently deleted (verified: [server]/host-name=operator-choice + HAL0_HOSTNAME=jarvis -> operator line gone), and a later no-override run withdraws only the managed pair — the conf reverts to the machine hostname, not the operator's earlier choice, which is lost permanently. Replacing while pinned is necessary (two host-name lines would be invalid), but the PR summary's 'an operator's own hand-written host-name= is never touched' only holds on the no-override path. Suggest at least a warn when clobbering a non-managed active host-name, and correcting the claim.

4. installer/install.sh (~line 1320, end of sync_avahi_hostname): chmod 0644 and mv -f "${tmp}" "${conf}" are unguarded under set -euo pipefail — a failing mv (e.g. chattr +i on the conf; mktemp in the same dir would still succeed) aborts the entire install, contradicting the stated fail-soft posture. '|| { rm -f "${tmp}"; return 0; }' on the mv would close it.

Nit (no action needed): the awk section tracking compares $0 == "[server]" exactly, so a header with stray whitespace would be missed and a duplicate [server] section appended at EOF — unlikely in real confs.

Verdict: needs-fixes — #1 is objective and blocking (CI red from this PR's own file); #2 is a doc correction; #3/#4 are hardening/accuracy. The core mechanism is correct and well-tested.

thinmintdev:
Addressed all four review findings in d19fcc7:

1. **CI red (ruff format)** — ran `ruff format` on `tests/installer/test_avahi_hostname.py`; `ruff format --check` and `ruff check` are now clean locally, which unblocks the pytest step in the python-3.12 job.

2. **Docs/summary overstated the answer-file path** — `docs/reference/env-vars.mdx` and the PR body now say only the `HAL0_HOSTNAME` env install path reaches `sync_avahi_hostname`; `network.hostname` in the answer file is consumed by the separate `hal0 setup --answers` process and `install.sh` never reads it. Wiring the answer-file flow into the sync is noted as a follow-up rather than implemented here (it would mean teaching install.sh to parse `hal0-setup.yaml`, which is beyond a few lines).

3. **Silent clobber of an operator's own `host-name=`** — the pin branch now detects a pre-existing non-managed active `host-name=` under `[server]` and warns that it is being replaced and will not be restored by a later no-override run (replacement itself is unavoidable — two `host-name` lines are invalid conf). The function's comment block and the PR body no longer claim "never touched" unconditionally; a new test covers the warn (suite is now 11 tests).

4. **Unguarded `chmod`/`mv` under `set -euo pipefail`** — both are now fail-soft: `chmod ... || true`, and a failed `mv` cleans up the tmp file, warns, and returns 0 instead of aborting the install.

The awk `$0 == "[server]"` whitespace nit is left as-is per the review (no action needed).

Verification: 11/11 tests pass locally via `uv run --extra dev pytest tests/installer/test_avahi_hostname.py`; `bash -n installer/install.sh` clean.

PR #2082, opened 2026-08-26
fix(installer): persist_bootstrap_cosign called undefined `ok` — fresh cosign-less installs die at pre-flight (#2081)

### LINKED ISSUE #2081: persist_bootstrap_cosign success path calls undefined `ok` — every fresh rc.10 bootstrap install without system cosign dies at pre-flight 1/16
## Summary

Every fresh bootstrap install of **v1.0.0-rc.10** on a host without a system cosign dies at pre-flight step 1/16:

```
/tmp/hal0-install-nkez5q/unpacked/installer/lib/preflight.sh: line 1644: ok: command not found
```

Found on the rc.10 validation fresh lane (ct151, pristine Ubuntu 26.04, `HAL0_CHANNEL=preview`, bootstrap.sh from the `v1.0.0-rc.10` tag). This is a release blocker for the fresh-install path: rc.10's headline fix #2058 (persist bootstrap-verified cosign, closes #2052) crashes the installer on its own success path.

## Root cause

`persist_bootstrap_cosign()` (installer/lib/preflight.sh:1644, added by #2058) reports success with an `ok` helper:

```sh
if install -m 0755 "${HAL0_BOOTSTRAP_COSIGN}" "${dest_dir}/cosign"; then
    ok "cosign: persisted bootstrap's pinned build to ${dest_dir}/cosign (hal0 update requires it)"
```

No `ok()` exists anywhere in the installer — `installer/lib/ui.sh` defines `info`/`warn`/`err`/`die` only (info prints the green check glyph). Under `set -euo pipefail` the unknown command (rc 127) kills install.sh at the `persist_bootstrap_cosign` call site (install.sh:303), immediately after the "bootstrap prereqs" line.

## Why every gate missed it

- Hosts with a system cosign return early (`command -v cosign` guard) and never reach line 1644 — CI runners and dev boxes fall in this class.
- The failure needs the exact success path: no system cosign AND `HAL0_BOOTSTRAP_COSIGN` set + executable — i.e. precisely a real end-user fresh install via the one-liner.
- The `warn` fallback paths of the same function use real helpers and are fine.

## Sequence on the box

1. cosign IS persisted to `/usr/local/bin/cosign` (the `install` succeeds) — the #2052 fix itself works.
2. The success message then crashes the installer at step 1/16; nothing under `/usr/lib/hal0` or `/var/lib/hal0` is created.
3. Bootstrap work dir `/tmp/hal0-install-*` stays behind (by design on failure — #2065 cleanup only runs on success).
4. Perverse follow-on: a straight re-run now passes, because step 1 left a system cosign and the early return skips the broken line. Retry-hides-the-bug.

## Fix

One line: `ok` → `info` at preflight.sh:1644. Worth pairing with a guard test that greps installer/lib/*.sh for helper invocations not defined in ui.sh (this exact class — a fix's success branch calling a nonexistent reporter — is invisible to every current test).

## Evidence

- ct151 `/root/rc10-install.log` (bootstrap verify chain all green; dies at step 1/16)
- `/usr/local/bin/cosign` present on ct151 post-crash, `/usr/lib/hal0` absent
- rc.10 validation lane, 2026-08-26

thinmintdev:
Fixes #2081.

## What

- `installer/lib/preflight.sh:1644` — `ok` → `info`. `persist_bootstrap_cosign()`'s success branch (added by #2058) reported via an `ok` helper that no installer source defines; ui.sh's reporters are `info`/`warn`/`err`/`die`. Under `set -euo pipefail` the 127 killed install.sh at pre-flight step 1/16 — right after the cosign was successfully persisted to /usr/local/bin.
- New `tests/installer/test_ui_helper_contract.py` — collects every function defined across `installer/**/*.sh` and flags any command-position reporter-shaped call (`ok`/`okay`/`success`/`fail`/`note`/known reporters) with no definition, so this class (a fix's success branch calling a nonexistent reporter — invisible to every existing test) cannot land again. Also pins ui.sh's reporter contract itself.

## Why every gate missed it

Hosts with a system cosign early-return before the broken line — that covers CI and every dev box. The failure needs the exact end-user shape: fresh box, no cosign, bootstrap-verified binary handed over via `HAL0_BOOTSTRAP_COSIGN`. Reproduced on the rc.10 validation fresh lane (ct151, pristine Ubuntu 26.04, preview channel); log in #2081.

## Impact

Every fresh one-liner install of v1.0.0-rc.10 on a cosign-less host dies. Re-runs mask it (run 1 leaves /usr/local/bin/cosign, run 2 early-returns past the bug).

Do not self-merge — needs independent review per kit process.

### DIFF
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1641,7 +1641,7 @@ persist_bootstrap_cosign() {
     fi
     if [[ -n "${HAL0_BOOTSTRAP_COSIGN:-}" && -x "${HAL0_BOOTSTRAP_COSIGN}" ]]; then
         if install -m 0755 "${HAL0_BOOTSTRAP_COSIGN}" "${dest_dir}/cosign"; then
-            ok "cosign: persisted bootstrap's pinned build to ${dest_dir}/cosign (hal0 update requires it)"
+            info "cosign: persisted bootstrap's pinned build to ${dest_dir}/cosign (hal0 update requires it)"
         else
             warn "could not persist cosign to ${dest_dir}/cosign — the first 'hal0 update' will fail its signature check until cosign is installed (https://docs.sigstore.dev/cosign/installation/)"
         fi
--- a/tests/installer/test_install_cosign_persist.py
+++ b/tests/installer/test_install_cosign_persist.py
@@ -37,12 +37,16 @@ _BOOTSTRAP = _REPO_ROOT / "installer" / "bootstrap.sh"
 _INSTALL_SH = _REPO_ROOT / "installer" / "install.sh"
 
 # ui.sh helpers the function calls; stubbed so output is capturable and the
-# restricted environment needs no extra sourcing.
+# restricted environment needs no extra sourcing. Stub ONLY the reporters
+# ui.sh actually defines (info/warn/err/die) — #2081 shipped because an
+# earlier version of this stub set also defined an `ok()` that exists
+# nowhere in the installer, masking the undefined-helper 127 that killed
+# every real cosign-less fresh install at pre-flight.
 _STUBS = """
 info() { printf 'INFO:%s\\n' "$*"; }
-ok()   { printf 'OK:%s\\n' "$*"; }
 warn() { printf 'WARN:%s\\n' "$*" >&2; }
 err()  { printf 'ERR:%s\\n' "$*" >&2; }
+die()  { err "$*"; exit 1; }
 """
 
 
@@ -97,7 +101,7 @@ class TestPersistBootstrapCosign:
         assert installed.read_bytes() == src.read_bytes()
         mode = stat.S_IMODE(installed.stat().st_mode)
         assert mode & stat.S_IXUSR, f"not executable: {oct(mode)}"
-        assert "OK:" in proc.stdout
+        assert "INFO:" in proc.stdout
 
     def test_never_overwrites_a_system_cosign(self, tmp_path: Path) -> None:
         src = _make_bootstrap_binary(tmp_path)
--- a/tests/installer/test_ui_helper_contract.py
+++ b/tests/installer/test_ui_helper_contract.py
@@ -62,6 +62,5 @@ def test_no_reporter_shaped_call_without_a_definition():
                 offenders.append(f"{f.relative_to(REPO)}:{n}: {line.strip()}")
     assert not offenders, (
         "reporter-shaped call to a helper no installer source defines "
-        "(#2081 class — this dies with 127 under set -e at runtime):\n"
-        + "\n".join(offenders)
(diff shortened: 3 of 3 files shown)


thinmintdev:
CI note on the previous run: `test_install_cosign_persist.py::test_persists_bootstrap_binary_when_host_has_none` failed BECAUSE of this fix — the test stubbed its own `ok()` (a helper that exists nowhere in the installer) and asserted `OK:` on stdout. That stub is precisely how #2081 shipped invisibly: the only coverage of the success path supplied the missing function itself. Pushed a0f59d06: stub set now mirrors ui.sh exactly (info/warn/err/die only), assertion follows the fix to `INFO:`.

`tests/slots/test_fail_watcher_warming.py::test_warming_slot_recovers_when_stale` (SlotState STARTING vs READY) is untouched by this diff — installer-only change; same workflow is green on main tip e598eec1. Looks like a timing flake on the shared runner.

thinmintdev:
The `ok` → `info` fix and the stub correction are both right and verified (the stub now mirrors ui.sh's real contract, which is exactly what let #2081 through my #2058 — good catch).

**Blocking finding on the new guard test: it cannot catch the #2081 class it exists to pin.** `test_no_reporter_shaped_call_without_a_definition` collects function definitions globally across `installer/**/*.sh` — and `installer/bootstrap.sh` defines its own `ok()` (line ~115, its private reporter set). So `"ok"` lands in `defined`, and any stray `ok "..."` call in preflight.sh/install.sh passes the guard.

Verified adversarially: on this PR's branch with the preflight.sh fix reverted (`info` → `ok` reinstated), `pytest tests/installer/test_ui_helper_contract.py` → **2 passed**. The guard is green over the exact bug that killed the fresh lane.

Fix shape: scope definitions per calling file — a call in file F is satisfied only by functions defined in F itself or in the files F actually sources (for `installer/lib/*.sh` and install.sh that's the lib set incl. ui.sh; bootstrap.sh is standalone — piped from curl, sources nothing — so its `ok()` must not satisfy anyone else's calls, and its own calls resolve only against its own definitions). After scoping, re-run the reverted-fix check above: the guard must go red on `preflight.sh:1644` before this merges.

Happy to re-review immediately — everything else is merge-ready.

thinmintdev:
Resolution given the author session is idle: **merging now** — the substantive fix (ok→info + honest stubs) is verified correct and every cosign-less fresh install is dead without it; holding a fresh-install-killer fix hostage to test-hardening inverts priorities. The guard-scoping finding above moves to a follow-up PR (per-file/source-scoped resolution, proven red on the reverted fix) so it gets its own independent review instead of me both writing and approving it inside this PR.

thinmintdev:
Follow-up landed as promised: #2086 scopes the guard per sourcing topology (bootstrap standalone vs shared lib scope) with a self-test pinning the scoping, and the reverted-fix acceptance check now goes RED on preflight.sh:1644. CI running.

Issue #1530, opened 2026-07-30
release: the stable channel pointer is still 0.9.8 — tagging v1.0.0 does not deliver it to a single existing user


thinmintdev:
Verified live from an installed box and from the published endpoints (2026-07-30):

```
halo150 (10.0.1.150, hal0 0.9.8, channel=stable):
  $ hal0 update --check
  hal0 0.9.8  →  0.9.8  (stable)  up to date
  released_at  2026-07-13T07:40:37+00:00

published pointers:
  https://releases.hal0.dev/stable.json   → 200  version 0.9.8   (2026-07-13)
  https://releases.hal0.dev/preview.json  → 404  (see #TBD)
  https://releases.hal0.dev/nightly.json  → 200  version 1.0.0-nightly.20260727191504
```

So **every existing stable-channel install is pinned at 0.9.8 and is told it is up to date.** 1.0.0-rc.1 and the nightlies shipped without the stable pointer ever advancing.

This matters for the GA cut because `release.yml` is explicit that it **never advances an external channel pointer** — its `authorize-pointer` job is a read-only re-verification (workflow header, lines 13-15: *"This workflow only authorizes the immutable remote publication; it never advances an external channel pointer"*). The pointer lives in hal0-web.

**Consequence: pushing the `v1.0.0` tag is not the release.** Without a matching hal0-web pointer advance, GA is invisible to every user who isn't on nightly, and `hal0 update` will keep answering "up to date" at 0.9.8.

## Wanted

- The GA runbook states the pointer advance as an explicit, owned step, not an implicit one — including who performs it and how it is verified (`curl https://releases.hal0.dev/stable.json` returning the new version is the acceptance test).
- Consider a post-release CI check that fails loudly if a `v*` GA tag exists with no matching stable pointer after N hours — the current failure mode is silent and indefinite.
- Decide and document the 0.9.8 → 1.0.0 upgrade path, given #1464: 0.9.8's own updater predates the self-update posture fix, so the very clients that need to reach 1.0 are running the updater that could not self-update. If 0.9.8 → 1.0.0 cannot work unattended, that needs saying out loud in the release notes with the manual path.

Found by the v1.0 polish/fix program (release umbrella #1443).

thinmintdev:
[...]

PR #2087, opened 2026-08-28
fix(installer): persist bootstrap HAL0_CHANNEL so fresh preview installs can self-update (#2083)

### LINKED ISSUE #2083: bootstrap's HAL0_CHANNEL is never persisted — fresh preview installs default the updater to stable, fail the new install-time update-check probe until GA, and cannot self-update to the next rc
## Summary

On the rc.10 validation fresh lane (ct151, pristine Ubuntu 26.04, `HAL0_CHANNEL=preview` bootstrap), the new #2066/#2074 install-time update-check probe failed and the install summary printed:

```
!!  '/usr/lib/hal0/venv/bin/hal0 update --check' FAILED — the update path is broken and the next release will not install
...
Verify FAILED:
  update-check probe failed — see the warnings above
```

The update path is NOT broken. The installed updater simply defaulted to the **stable** channel, and `https://releases.hal0.dev/stable.json` is 404 until the first signed stable (v1.0.0 — the accepted #2057/#1530 posture):

```
GET http://127.0.0.1:8080/api/updates/check → HTTP 500: could not fetch
release manifest: release manifest fetch returned HTTP 404 from
https://releases.hal0.dev/stable.json
```

After `PUT /api/updates/channel {"channel":"preview"}`, the same box immediately reports `1.0.0rc10 → 1.0.0-rc.10 (preview) up to date` — manifest fetch, admission and cosign path all healthy.

## Root cause

bootstrap.sh admits and verifies the release against `HAL0_CHANNEL`, but nothing hands that channel to the installed system. The updater then starts life on its stable default. Two consequences:

1. **Every preview/nightly fresh install fails the new update-check probe until GA**, and the summary's "the update path is broken and the next release will not install" is a false alarm — which trains operators to ignore the one probe that exists to catch #2052-class breakage.
2. **Functional gap, not just cosmetic**: an rc box installed from the preview channel can never see the next rc via `hal0 update` until someone manually flips the channel. The rc.10 upgrade lane only worked on ct150 because its `HAL0_RELEASES_URL` was hand-pinned.

## Suggested fix

Persist the bootstrap channel into the installed config (env var handoff bootstrap → install.sh → api.env or the updates config, mirroring the `HAL0_BOOTSTRAP_COSIGN` pattern from #2058), so the first `update --check` runs against the channel the install was actually admitted from. The probe itself is fine once that's true.

## Evidence

- ct151 `/root/rc10-install-2.log` (summary block), rc.10 validation lane 2026-08-26.
- Same run, other probe: structured-output probe PASSED (gateway /v1, kwarg-absent json_object) — first install-time confirmation of the #2056/#2073 fix on a fresh box.

thinmintdev:
## Why

Fixes #2083 — the last code gap in the "first self-update works" story (#2052 → #2058 → #2082).

A preview-channel fresh install (`HAL0_CHANNEL=preview` at bootstrap) verifies and installs correctly, but nothing writes the channel anywhere durable. The installed updater defaults to `stable`, whose manifest pointer is 404 until GA (#1530), so:

1. the #2066 install-time update-check probe fails with a false "the update path is broken" alarm on every preview/nightly fresh install (rc.10 lane, ct151 `/root/rc10-install-2.log`), and
2. an rc box can never see the next rc via `hal0 update` until someone manually flips the channel.

## What

Mirrors the `HAL0_BOOTSTRAP_COSIGN` hand-off pattern from #2058:

- `installer/bootstrap.sh`: `export HAL0_CHANNEL` in `validate_channel()` so the admitted channel survives the `exec` into install.sh.
- `installer/lib/preflight.sh`: new `persist_bootstrap_channel <toml>` — writes `telemetry.channel` into `hal0.toml`, the exact key `PUT /api/updates/channel` persists and the updater's `_current_channel()` reads. Locked read-modify-write (system python3, `hal0.toml.lock` fcntl + O_NOFOLLOW — the `[models].store` seeding discipline) because a live daemon may be serving config writes. Acts only when `HAL0_CHANNEL` came across the exec; validates the value; never overwrites an explicit differing value (warns and points at `hal0 update --channel`); never dies.
- `installer/install.sh`: call site in the Configuration step, right after `hal0.toml` exists and before the hal0-api (re)start (the daemon caches config at startup) and before the update-check smoke probe.

Persistence only — verification/probe logic untouched, per the rc.10 lane finding that the update path is healthy end-to-end once the channel is set.

## Tests

`tests/installer/test_install_channel_persist.py` (real-bash subprocess, ui.sh reporter contract stubs only — info/warn/err/die, #2081):

- writes channel into existing `[telemetry]` (rest of config intact, valid TOML via tomllib)
- appends `[telemetry]` when absent
- no-op when `HAL0_CHANNEL` unset / value already matches
- never overwrites a differing explicit channel — warns with the `hal0 update --channel` pointer
- rejects invalid channel without writing; survives a missing toml with a warning
- wiring: bootstrap exports the channel; install.sh call sits after `ui_step "Configuration"` and before `smoke_update_check`

**Proven red on the reverted fix** (house rule from #2081/#2082/#2086): with only `installer/` stashed, all 9 fail (`persist_bootstrap_channel: command not found` + both wiring asserts); restored, 9 pass.

Full `tests/installer/`: 874 passed, 1 skipped, 1 failed — the pre-existing local-only `test_no_avahi_on_the_box_is_a_silent_noop` (thinmint runs avahi; CI is authoritative).

https://claude.ai/code/session_01QtnrnWxxv8XEBiGrVRXqsh

### DIFF
--- a/installer/bootstrap.sh
+++ b/installer/bootstrap.sh
@@ -254,6 +254,11 @@ validate_channel() {
         stable|preview|nightly) ;;
         *) die "HAL0_CHANNEL must be one of: stable, preview, nightly (got ${HAL0_CHANNEL})" ;;
     esac
+    # The admitted channel must survive the exec into install.sh so the
+    # installed updater starts on the channel this install was actually
+    # verified against (persist_bootstrap_channel -> telemetry.channel,
+    # #2083) - same hand-off shape as HAL0_BOOTSTRAP_COSIGN (#2058).
+    export HAL0_CHANNEL
 }
 
 # ── cosign acquisition ─────────────────────────────────────────────────────
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1154,6 +1154,13 @@ fi
 chmod 0755 "${ETC_DIR}" 2>/dev/null || true
 chmod 0644 "${HAL0_TOML}" 2>/dev/null || true
 
+# Persist the channel this install was admitted from (bootstrap hand-off,
+# #2083) so the first `hal0 update --check` - including the update-check
+# smoke probe below - runs against it instead of the stable default (404
+# until GA, #1530). Must run before the hal0-api (re)start: the daemon
+# caches hal0.toml at startup.
+persist_bootstrap_channel "${HAL0_TOML}"
+
 # Pin the dashboard's built assets. Prod installs hal0 NON-editable, so the
 # package's __file__ lives in the venv site-packages and the walk-up that
 # finds ui/dist in a checkout no longer reaches it — point HAL0_UI_DIST at
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1651,6 +1651,140 @@ persist_bootstrap_cosign() {
     return 0
 }
 
+# ── bootstrap channel persistence (#2083) ──────────────────────────────────
+# bootstrap.sh admits and cosign-verifies the release against HAL0_CHANNEL,
+# then execs into install.sh — but nothing used to write that channel
+# anywhere durable, so the installed updater started life on its `stable`
+# default, whose manifest pointer is 404 until GA (#1530). Every preview/
+# nightly fresh install then failed the #2066 install-time update-check
+# probe with a false "the update path is broken" alarm, and an rc box could
+# never see the next rc via `hal0 update` without a manual channel flip.
+# Same hand-off shape as persist_bootstrap_cosign above: act only when the
+# env var came across the exec, never overwrite an explicit differing value
+# (warn instead), never die — the install itself is fine either way.
+# Writes telemetry.channel in hal0.toml: the exact key PUT
+# /api/updates/channel persists and the updater's _current_channel() reads.
+# Locked read-modify-write via system python3 with the same hal0.toml.lock
+# fcntl + O_NOFOLLOW discipline as install.sh's [models].store seeding —
+# this can run against a live daemon serving config writes, and /etc/hal0
+# is service-writable while we run as root.
+persist_bootstrap_channel() {
+    local toml="$1"
+    local channel="${HAL0_CHANNEL:-}"
+    [[ -n "${channel}" ]] || return 0
+    case "${channel}" in
+        stable|preview|nightly) ;;
+        *)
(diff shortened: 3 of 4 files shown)


thinmintdev:
**Independent review pass** (same-account constraint per the #2082 precedent: findings as a comment, not a formal review).

Verdict was **"with fixes"** — both Important findings valid, both addressed in `4fde63fe`:

**Important 1 — regex blind spots could corrupt the config.** The existing-key regex missed valid TOML (`channel = "preview"  # comment`, single-quoted values, `[telemetry]  # comment` headers), and a miss appended a *duplicate key* — a hard `tomllib` parse error that takes the daemon down with the config, violating the function's own never-break contract. **Fixed:** detection now parses the file with `tomllib` (any valid spelling is seen); the patched text is re-parsed and value-checked before the atomic replace, so every residual blind spot degrades to warn-and-leave-alone. The commented-header case now also inserts correctly.

**Important 2 — stable-default poisoning.** Bootstrap defaults `HAL0_CHANNEL=stable`, so every default install would persist an explicit `channel = "stable"` — which the never-overwrite rule would then defend against a later deliberate `HAL0_CHANNEL=preview` re-bootstrap (the exact recovery #2083 is about). **Fixed:** the schema default is not persisted when no key exists (behavior-identical for the updater); an explicit on-disk value is still respected with a warn.

**Minor:** single-quote misreport — fixed by the tomllib detection. Test gaps — 9 new tests cover inline comments (same + differing), single quotes, a list-valued following table (splice guard), commented header, unparsable config left alone, stable-noop, default-then-preview persists, explicit-stable respected. Dir-fsync and lock-timeout nits left as-is (consistent with `hal0.config.locking` precedent). Stale "stable | nightly" docstring in `routes/updater.py` noted as out-of-scope follow-up.

Evidence: review-round tests proven red against the pre-review function (7 fail on `f6ab6593`'s preflight.sh, 18 pass on `4fde63fe`); full `tests/installer/`: 883 passed, 1 skipped, plus the pre-existing local-only avahi failure (thinmint runs avahi; CI authoritative).

Issue #2021, opened 2026-08-22
hermes venv is built from hermes-agent[web] without the mcp extra — hermes has no MCP client, both hal0-provisioned MCP servers are dead, and tools list / mcp list / provision.json mcp_wire all report success over it


thinmintdev:
## GA blocker — found by rc-validate v1.0.0-rc.7 (kit v8)

**Severity:** ga-blocker · **Lane:** hermes · **Register key:** `hermes-venv-missing-mcp-extra` · **First seen:** v1.0.0-rc.7

### Summary

`installer/agents/hermes/requirements.txt` installs `hermes-agent[web]`, but upstream gates the MCP client behind a separate `[mcp]` extra (`Requires-Dist: mcp==1.26.0; extra == "mcp"`) that hal0 never requests. The hermes venv therefore has NO `mcp` package at all. Both MCP servers hal0 itself provisions into hermes' config.yaml (hal0-admin, hal0-memory) are dead: `hermes mcp test hal0-admin` fails immediately ("requires HTTP transport but mcp.client.streamable_http is not available"), `hermes mcp test hal0-memory` times out at 40s. `hermes tools list` / `hermes mcp list` both claim "all tools enabled" / "✓ enabled", and the installer's own `mcp_wire` provisioning phase reports `status: ok, mcp_server_count: 2` — every self-check surface is green over a wholly dead client. Net effect: hermes' system prelude tells the model it has hal0-admin/hal0-memory MCP tools, the model receives a tools array with ZERO hal0-admin tools, and every live-state question is answered from static prelude/STATE.md text with a confident fabrication (wording varies — "no slots loaded", or hitting the tool-iteration ceiling — the invariant is "wrong, tool-free answer", not a specific string). Reproduces on every rc.7 box checked: two fresh installs (ct151, ct163) AND the upgrade box (ct150) — an agent-upgrade rebuild loses MCP too. Production (ct105) still has `mcp` importable only because of a pre-rc.7 venv left over from before this requirements.txt regressed; that is stale residue masking the defect on the one box people actually chat through, not evidence it's fixed there.

### Repro

```sh
/var/lib/hal0/venvs/hermes/bin/python -c "import mcp"   # expect ModuleNotFoundError
sudo -u hal0 /usr/local/bin/hermes mcp test hal0-admin   # expect a transport-unavailable failure
sudo -u hal0 /usr/local/bin/hermes mcp test hal0-memory  # expect a ~40s timeout
[...]

Issue #245, opened 2026-05-23
model_automap + voice_wire phases


thinmintdev:
## What to build

Implement `model_automap` and `voice_wire` phases.

`model_automap`: walks `slot_list` + `model_list`, writes `model_aliases` entries to config.yaml. One alias per chat slot. `/model <alias>` inside Hermes switches slots. Embed/rerank slots stay UNWIRED (per grilling Q6 resolution; deferred to v0.4 per ADR-0011 §3).

`voice_wire`: conditional on STT/TTS slot presence + reachability. Emits:

```yaml
stt: { provider: "openai" }
tts: { provider: "openai" }
```

…plus `STT_OPENAI_BASE_URL` to `/var/lib/hal0/secrets/agents/hermes.env`. If slots absent: phase status = `skip` with reason.

## Acceptance criteria

- [ ] Every chat slot in `capabilities.toml` becomes a `model_aliases` entry.
- [ ] `/model <alias>` switches inside Hermes; chat continues against the new endpoint.
- [ ] STT/TTS config emitted only when both slots are `ready` in `slot_list`.
- [ ] Phase logs which slots were mapped + which were skipped (with reason).
- [ ] Idempotent: hash-equal output skips re-write.
- [ ] Unit tests cover slot enumeration + voice-conditional emission.

## Blocked by

- #241

## References

- `docs/internal/hermes-bootstrap-plan-2026-05-23.md` §12, §13

thinmintdev:
**Sharpening from audit 2026-05-23.** Pinning MCP contract + config write mechanism + secrets file.

## MCP contract

- [ ] Phase calls \`hal0-admin:slot_list()\` to enumerate currently-running capability slots.
- [ ] Phase calls \`hal0-admin:model_list()\` for live model metadata (\`model_id\`, \`backend_url\`, \`context_length\`).
- [ ] On unavailable/unreachable MCP, phase gracefully degrades: log warning, write empty aliases, mark phase \`status = "degraded"\` (NOT fail).

## Config write mechanism

- [ ] \`model_aliases\` written directly to \`\$HERMES_HOME/config.yaml\` — rewrite the whole \`[model_aliases]\` section atomically (same envelope as \`config_write\` phase). **Do NOT call \`hermes config set\`** (per plan §8 gotcha — fragments the single-write invariant).

## STT/TTS env vars

[...]

PR #2088, opened 2026-08-28
fix(agents): X-hal0-Private must be a YAML string — hermes memory MCP dead on fresh installs (#2085)

### LINKED ISSUE #2085: hermes cannot connect to the hal0-memory MCP server on fresh rc.10 (CancelledError after initialize) while hal0-admin works — hermes runs with zero memory tools; server healthy under every direct probe
## Summary

rc.10 fresh install (ct151, rc.10 validation lane): hermes cannot connect to the **hal0-memory** MCP server while **hal0-admin works perfectly** in the same process. Every hermes-side surface fails:

```
$ sudo -u hal0 hermes mcp test hal0-memory
  Transport: HTTP → http://127.0.0.1:8080/mcp/memory/mcp
  ✗ Connection failed (40775ms): MCP call timed out after 40.8s (configured timeout: 40.0s)
```

and the live gateway at boot:

```
WARNING tools.mcp_tool: Failed to connect to MCP server 'hal0-memory': CancelledError
INFO  tools.mcp_tool: MCP: registered 184 tool(s) from 1 server(s) (1 failed)
```

So hermes runs with **zero hal0-memory tools** (memory_recall/memory_add/... absent) on a fresh rc.10 box. This re-fails the `hermes-venv-missing-mcp-extra` (#2021) regression entry's expect clause — "`hermes mcp test hal0-admin` AND `hermes mcp test hal0-memory` both connect and pass" — though the mechanism is different (the mcp client package is present and admin works).

## What it is NOT (discrimination done on the box)

The server side is healthy under every probe shape tried:

- initialize on `/mcp/memory/mcp`: HTTP 200 in **5ms** (plain), 17ms with hermes' exact headers (`X-hal0-Agent: hermes`, `X-hal0-Private: 1`)
- full manual handshake (initialize → notifications/initialized → tools/list) succeeds in milliseconds under BOTH `protocolVersion: 2025-03-26` and `2025-11-25` (the version hermes negotiates), returning the full memory tool list
- identical handshake against `/mcp/admin/mcp` behaves identically — no server-side difference between the routes
- hermes' own log shows initialize DOES complete ("Received session ID", "Negotiated protocol version: 2025-11-25") — the hang is after that, inside the client, ending in `CancelledError` at the 40s ceiling
- retention itself works: installer-era retains landed (`shared`=7 facts), `hal0 memory status` ON/landing — the dead surface is specifically hermes' MCP tool client for this one server

## Config delta between the working and failing server (hermes config.yaml)

```yaml
hal0-admin:  {type: http, url: .../mcp/admin/mcp,  headers: {X-hal0-Agent: hermes}, timeout: 14491}
hal0-memory: {type: http, url: .../mcp/memory/mcp, headers: {X-hal0-Agent: hermes, X-hal0-Private: 1}, timeout: 900}
```

Suspects worth chasing in hermes' mcp_tool client: the per-server `timeout: 900` unit interpretation, the post-initialize GET/SSE listener flow for this server, or a collision with hermes' `memory.provider: hal0-memory` plugin holding its own connection to the same server.

## Impact

Every live-state memory question through hermes falls back to model guesswork (compare #2021's "confident fabrication over a green surface" mode — hermes one-shots on this box produced fabricated tool-call tags as literal text). Memory WRITES land via the non-MCP path, so the defect is invisible to `hal0 memory status` and every green self-check.

## Environment

ct151 fresh rc.10 (hal0 1.0.0rc10, hermes-agent venv per installer), 2026-08-26. Reproduced 4/4 (root + hal0 user, with/without --dev).

thinmintdev:
## Why

Fixes #2085 — hermes runs with **zero hal0-memory tools** on every fresh rc.10 install (`CancelledError` after initialize, 184 admin tools registered, memory server failed) while `hal0-admin` works in the same process and the server answers every direct probe in milliseconds.

## Root cause (found by bisection against a live gateway)

The issue's config delta contained the whole answer:

```yaml
hal0-memory: {headers: {X-hal0-Agent: hermes, X-hal0-Private: 1}, timeout: 900}
```

That `1` is an **unquoted YAML int**. The provisioner passes `("mcp_servers.<name>.headers.X-hal0-Private", "1")` through `hermes config set`, and hermes coerces bare integers on the way in (the documented behavior `_fmt_config_value` already relies on for timeouts) — so the string `"1"` lands on disk as the int `1`. Hermes' own MCP client then wedges post-initialize on the non-string header value (`_discover_tools`/`tools/list` never completes) and dies with `CancelledError` at its outer ceiling. `hal0-admin` carries no private header — hence the clean admin-works/memory-hangs split. Manual probes never reproduced it because every hand-built probe sends the header as a proper string.

Reproduction (thinmint, pinned hermes build `9de9c25f` from `installer/agents/hermes/requirements.txt`, mcp 1.26.0, against ct150's live rc.10 gateway over an SSH tunnel):

- `X-hal0-Private: 1` (int, verbatim fresh-install rendering) → `✗ Connection failed (41230ms): MCP call timed out after 41.2s` — the exact #2085 signature
- header removed → `✓ Connected, 26 tools`
- `X-hal0-Private: "1"` (string) → `✓ Connected, 26 tools`

This also disproves the "fresh-install-shaped server" framing: ct150's server triggers it fine — the discriminator was always the client-side config rendering, which only fresh installs (and re-provisions) receive.

## What

`src/hal0/agents/hermes_provision.py`:

- `_build_config_overlay`: the private header no longer goes through `hermes config set` (which cannot express a numeric-looking string — quoting the argv stores literal quote characters, silently disabling private mode instead).
- `_config_list_keys` gains `mcp_servers`: private servers get `headers: {X-hal0-Private: "1"}` in the PyYAML deep-merge layer, where `safe_dump` renders `'1'` quoted. The nested-dict merge preserves the sibling headers config set wrote (`X-hal0-Agent`, `Authorization`), and because the merge runs **after** the config-set phase with overlay-wins semantics, a re-provision (`hal0 agent bootstrap hermes --repair`) also repairs an already-poisoned config.yaml.
- `_build_brain_profile_mcp_servers`: same bug as a literal int `1`; now `"1"`.

Two existing test asserts locked the int in (`== 1`); flipped to the string contract.

## Tests

`tests/agents/test_hermes_private_header_type.py`:

- overlay emits no `X-hal0-Private` config-set pair
- `_config_list_keys` carries the header as `str` `"1"` for private servers only; no `mcp_servers` table when none are private
- `_merge_config_yaml_layers` repairs an int-poisoned config (value becomes `str`, sibling headers survive, on-disk YAML carries `'1'`)
- brain-profile header is a `str`

**Proven red on the reverted fix**: all 5 fail with `src/` stashed, pass restored.

**End-to-end**: the fixed `_config_list_keys` + `_merge_config_yaml_layers` (real provision code, not a hand-written config) repaired an int-poisoned config.yaml, after which `hermes mcp test hal0-memory` → `✓ Connected (191ms), ✓ Tools discovered: 26` against ct150's live gateway.

Full `tests/agents/`: 802 passed, 1 skipped, 1 xfailed (the pre-existing terminal-backend drift watch).

#2021's regression entry ("both `hermes mcp test` connect") becomes honest again with this fix — no expectation change needed.

https://claude.ai/code/session_01QtnrnWxxv8XEBiGrVRXqsh

### DIFF
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -2080,7 +2080,9 @@ def _merge_config_yaml_layers(
             if isinstance(_hdrs, dict) and "X-hal0-Private" in _hdrs:
                 _val = _hdrs["X-hal0-Private"]
                 if not isinstance(_val, str):
-                    _hdrs["X-hal0-Private"] = str(_val).lower() if isinstance(_val, bool) else str(_val)
+                    _hdrs["X-hal0-Private"] = (
+                        str(_val).lower() if isinstance(_val, bool) else str(_val)
+                    )
     out = yaml.safe_dump(merged, sort_keys=False, default_flow_style=False)
     if config_path.exists() and config_path.read_text(encoding="utf-8") == out:
         return False
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -1756,7 +1756,10 @@ def terminal_tool_enabled(
 
 
 def _config_list_keys(
-    *, terminal_enabled: bool, existing_disabled: list[str] | None = None
+    *,
+    terminal_enabled: bool,
+    existing_disabled: list[str] | None = None,
+    mcp_servers: list[dict[str, Any]] | None = None,
 ) -> dict[str, Any]:
     """The list-valued overlay, with the terminal posture folded in.
 
@@ -1773,6 +1776,19 @@ def _config_list_keys(
     if not terminal_enabled:
         disabled += TERMINAL_OFF_TOOLSETS
     keys["agent"] = {"disabled_toolsets": disabled}
+    # X-hal0-Private must reach config.yaml as the STRING "1" (#2085): the
+    # config-set path can't express a numeric-looking string (hermes coerces
+    # it to the YAML int 1, which wedges hermes' MCP client post-initialize),
+    # so private servers get their header here, where safe_dump quotes it.
+    # The nested-dict merge preserves the sibling headers config set wrote
+    # (X-hal0-Agent, Authorization) and overlay-wins repairs poisoned boxes.
+    private_headers = {
+        srv["name"]: {"headers": {"X-hal0-Private": "1"}}
+        for srv in (mcp_servers or [])
+        if srv.get("private")
+    }
+    if private_headers:
+        keys["mcp_servers"] = private_headers
     return keys
 
 
@@ -1918,8 +1934,15 @@ def _build_config_overlay(
         ]
         if bearer:
             pairs.append((f"mcp_servers.{name}.headers.Authorization", f"Bearer {bearer}"))
-        if srv.get("private"):
-            pairs.append((f"mcp_servers.{name}.headers.X-hal0-Private", "1"))
+        # X-hal0-Private deliberately does NOT go through `hermes config set`:
+        # hermes coerces bare integers on the way in (see _fmt_config_value),
+        # so "1" lands in config.yaml as the YAML int 1 — and hermes' own MCP
+        # client wedges post-initialize on a non-string header value and dies
+        # with CancelledError at its outer ceiling (#2085: zero memory tools
+        # on every fresh install while hal0-admin connects fine). The header
+        # is layered as a real string by _config_list_keys via the PyYAML
+        # deep-merge, which runs AFTER these config-set writes — so a
+        # re-provision also repairs an already-poisoned config.yaml.
 
     pairs.append(("skills.creation_nudge_interval", 15))
     # terminal.cwd is the working directory Hermes' built-in `local` terminal
@@ -2327,6 +2350,7 @@ def _phase_config_write(ctx: _StepCtx) -> PhaseResult:
         list_keys=_config_list_keys(
             terminal_enabled=terminal_enabled,
             existing_disabled=_disabled_toolsets(_parse_yaml_config(pre_merge)),
+            mcp_servers=mcp_servers,
         ),
         overrides_path=OVERRIDES_PATH,
     )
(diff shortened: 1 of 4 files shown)


thinmintdev:
**Independent review pass** (same-account constraint per the #2082 precedent: findings as a comment, not a formal review).

Verdict was **"with fixes"**. Reviewer confirmed the mechanism end-to-end from the code (config-set coercion → int on disk; merge layer runs strictly after config set; `_deep_merge` overlay-wins repairs poisoned boxes and preserves sibling headers; both server-side resolvers accept the string, so private scoping is preserved; idempotency #245 converges — the existing phase idempotency test exercises the new merge key). Both Important findings valid, both addressed in `4c0a40cf`:

**Important 1 — no phase-level wiring guard.** The single line `mcp_servers=mcp_servers` into `_config_list_keys` was the only connection between fix and production; dropping it in a future `_phase_config_write` refactor would keep every unit test green while fresh installs ship **no** private header at all — hermes connects fine but private memory writes silently land in the `shared` bank, invisible to every smoke test. **Fixed:** `test_config_write_phase_writes_yaml_idempotently` now asserts the phase-written config.yaml contains `X-hal0-Private: '1'`.

**Important 2 — overrides.yaml re-poisoning.** `overrides.yaml` merges LAST by design, and the natural thing to copy from a pre-fix config.yaml is the unquoted int `1` — permanently reinstating the 40s wedge with zero diagnostics. **Fixed:** `_merge_config_yaml_layers` normalizes the known-poisonous `mcp_servers.*.headers.X-hal0-Private` leaf to a string after the final merge, whichever layer supplied it (also mitigates the reviewer's residual concern about hermes' own re-serialization). Regression test proven red on the pre-fix code.

**Minor:** loose `"'1'"` assert tightened to the full header form. Headers-only-stub and PyYAML-absent paths acknowledged, left as-is (pre-broken box / dead-code guard). The latent same-class risk for a numeric `X-hal0-Agent` is theoretical (hal0 controls the id) — noted here as the reviewer suggested rather than coded around.

**Residual debt, stated explicitly per the operator ruling:** upstream hermes itself still (a) coerces bare ints in `config set` argv and (b) hangs rather than errors on a non-string header value — a hand-run `hermes config set …X-hal0-Private 1` re-triggers the wedge. hal0-side normalization now self-heals it on the next provision/`--repair`, but an upstream hermes-agent issue is worth filing.

Full `tests/agents/`: 803 passed, 1 skipped, 1 xfailed.

Issue #2126, opened 2026-08-30
CPU-only fresh install: brain slot crash-loops with SIGILL — the 'cpu' runner's image is FALLBACK_VULKAN_IMAGE, and no CPU toolbox image is published


thinmintdev:
Found on a v1.0.0 fresh install of ct163 (Ubuntu 24.04 privileged LXC, no GPU devices), installed with the public one-liner plus `HAL0_ALLOW_CPU_ONLY=1`.

The install completes and reports itself ready, but no model ever serves.

## Symptom

The brain slot never leaves `warming`:

```
│ brain    │ warming │ lfm2.5-2.6b │ cpu     │ 8089 │ local │
```

and the install's own verification catches it:

```
!!  structured-output probe FAILED — a json_object request via the gateway (:8080/v1) did not return parseable JSON
  Verify FAILED:
    structured-output probe failed — see the warnings above
```

Any request routed to it fails:

```json
{"error":{"code":"dispatch.no_route","message":"model 'hal0' not found in registry, no upstream advertised it, and legacy slot resolution failed"}}
```

## It is a crash loop, not a slow load

`journalctl -u hal0-slot@brain`:

```
hal0-slot-brain[21042]: 0.00.025.303 I srv    load_model: loading model '/var/lib/hal0/models/chat/lfm2.5-2.6b/model.gguf'
hal0-slot-brain[21042]: AMX is not ready to be used!
hal0-slot-brain[21042]: 0.00.729.231 I cmn          init: llama threadpool init, n_threads = 16
podman[21117]: container died ... (image=ghcr.io/hal0ai/amd-strix-halo-toolboxes:vulkan-radv-server, name=hal0-slot-brain)
systemd[1]: hal0-slot@brain.service: Main process exited, code=exited, status=132/n/a
systemd[1]: hal0-slot@brain.service: Failed with result 'exit-code'.
```

`status=132` is 128+4 — **SIGILL**. It dies ~0.7 s into model load, restarts, and repeats (`NRestarts=2` and climbing, unit `activating`).

## Root cause

The slot is configured correctly. `/etc/hal0/slots/brain.toml`:

```toml
device = "cpu"
```

`runner_for_backend` also routes correctly — `device_class == "cpu"` or `backend == "cpu"` selects the `cpu` runner. The problem is what that runner points at.

`src/hal0/runners/__init__.py:153-163`:

```python
"cpu": Runner(
    "cpu",
[...]

Issue #2084, opened 2026-08-26
post-install seam verify: transient false-negative on hal0-podman-ro right after grant write — same invocation passes moments later; warning also misquotes the command it ran


thinmintdev:
## Summary

rc.10 fresh install (ct151, pristine Ubuntu 26.04): the post-install seam verification warned

```
!!  seam hal0-podman-ro: 'sudo -n hal0-podman-ro check-slot-token hal0probe' failed as the hal0 user — the /etc/sudoers.d/hal0-podman-ro grant does not apply, or the wrapper is stale
```

immediately (same log second) after the installer's own

```
OK  wrote /usr/lib/hal0/bin/hal0-podman-ro
OK  wrote /etc/sudoers.d/hal0-podman-ro
OK  systemctl daemon-reload
```

Minutes later, on the untouched box, the exact invocation the check runs (`_preflight_seam` uses the full `${bin}` path) succeeds:

```
# sudo -n -u hal0 sudo -n /usr/lib/hal0/bin/hal0-podman-ro check-slot-token hal0probe
hal0-slot-hal0probe   (rc=0)
```

Wrapper root:root 0755, grant root 0440, `visudo`-clean. So the install-time warning was a false negative — some transient in the write→verify window (grant/wrapper freshly written, hal0 user freshly created), not a broken grant.

## Why it matters

- #1889 made podman-ro the only source of truth for a running slot's `image_status`/`actual_image`; kit v6 explicitly treats a warned-but-actually-healthy seam as its own finding class ("`unknown` on a box whose seam is demonstrably healthy is a defect of its own").
- An operator following the warning re-runs install.sh for nothing; a validation agent files a false regression.

## Two smaller nits in the same block

1. The warning text prints the bare name (`sudo -n hal0-podman-ro …`) while the check actually ran the full path — copy-pasting the printed command fails for an unrelated reason (not in sudo's `secure_path`), compounding the misdiagnosis.
2. A transient failure gets no retry: one `sleep 1` + re-probe (or moving the optional-seam probes after the service-start phase) would separate "grant genuinely broken" from "verified too fast".

## Evidence

ct151 `/root/rc10-install-2.log` (seam block), live re-probe 2026-08-26 ~19:30 UTC, rc.10 validation lane. hal0-systemctl and hal0-update probes passed in the same run; only the optional podman-ro probe tripped, so the install correctly continued.

thinmintdev:
[...]

PR #2148, opened 2026-08-30
fix(preflight): add DPkg::Lock::Timeout to every apt-get call in preflight.sh

### LINKED ISSUE #2125: preflight: git install races the dpkg lock and skips hermes — the call omits the DPkg::Lock::Timeout its own error hint recommends
Found on a v1.0.0 fresh install of ct163 (Ubuntu 24.04 privileged LXC, CPU-only), installed with the public one-liner.

## What happens

```
OK  installing git (required for Hermes agent provisioning)
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 12462 (apt-get)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
XX  git install failed — see output above
!!  hermes provisioning skipped — git is required (pip installs hermes-agent from a git URL) and could not be installed
!!    install git manually (sudo apt-get -o DPkg::Lock::Timeout=120 install -y git), then run '/usr/lib/hal0/venv/bin/hal0 agent install hermes'
```

The install completes and the box is otherwise healthy; the agent is simply absent.

## The call omits the flag its own hint recommends

`installer/lib/preflight.sh:758`:

```sh
apt-get) DEBIAN_FRONTEND=noninteractive apt-get install -y -q git || ok=0 ;;
```

No `-o DPkg::Lock::Timeout`. Two lines of context make this stand out:

1. The failure hint printed by this same function tells the operator to run `sudo apt-get -o DPkg::Lock::Timeout=120 install -y git` — so the remedy is already known here, it just is not applied to the call itself.
2. `installer/install.sh` uses `DPkg::Lock::Timeout` in **6** places. `installer/lib/preflight.sh` uses it in **0**. The convention exists and this file is the outlier.

With no timeout, `apt-get` fails immediately rather than waiting, so any concurrent dpkg holder loses hermes for the whole install.

## Why it will hit real installs

The lock holder does not have to be anything exotic. `unattended-upgrades` and `apt-daily` fire on a freshly booted Ubuntu box, which is exactly the state a one-liner install runs in. This is a race, so it will be intermittent and will look like flakiness rather than a bug.

## Suggested fix

Add the timeout to the `apt-get` branch, matching what `install.sh` already does and what the hint advises:

```sh
apt-get) DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 install -y -q git || ok=0 ;;
```

Worth a sweep of `installer/lib/preflight.sh` for other `apt-get` calls with the same gap, since the count above suggests the file was never brought in line with `install.sh`.

## Related

On the same v1.0.0 fresh-install pass, hermes also failed to provision on ct151 (Ubuntu 26.04) for an unrelated reason — see #2124. Between the two, hermes did not come up on either fresh box, though each cause is independent and each has a documented manual recovery that works.

thinmintdev:
## What changed

`preflight_git`'s apt-get branch (installer/lib/preflight.sh:758) omitted the
`-o DPkg::Lock::Timeout` flag that its own failure hint recommends
(`sudo apt-get -o DPkg::Lock::Timeout=120 install -y git`). On a freshly
booted Ubuntu box, `unattended-upgrades` / `apt-daily` routinely hold the
dpkg lock; without the timeout, `apt-get` fails immediately instead of
waiting it out, and Hermes provisioning is skipped for what is really just a
race — reported on a v1.0.0 fresh install of ct163.

## Root cause, and why the fix is broader than the reported line

The reported symptom (git install fails, hermes skipped) traces to a single
missing flag on one apt-get call, but `installer/install.sh` uses
`-o DPkg::Lock::Timeout=120` on *every* apt-get invocation it makes, while
`installer/lib/preflight.sh` used it on *none* of its six apt-get call
sites. That's a convention gap in the whole file, not a one-off typo, so
every other preflight apt-get install in this file races the same lock:

- `_main_py_autoinstall` (python3.\<floor\> autoinstall) — update + install
- `_hindsight_py_autoinstall` (Hindsight-compatible python autoinstall) — update + install
- `preflight_venv` (python3-venv/python3-pip autoinstall)
- `preflight_container_runtime` (podman autoinstall)
- `preflight_git` (git autoinstall — the reported call site)
- `_node_autoinstall` (nodejs install after the NodeSource setup script)

All six now carry `-o DPkg::Lock::Timeout=120`, matching install.sh's
established convention exactly.

## How it was verified

- Added a regression test to `tests/installer/test_preflight_git.py`
  (`test_apt_get_invocation_carries_lock_timeout`) asserting the apt-get
  invocation `preflight_git` shells out to (mocked, `HAL0_GIT_REQUIRED=1`
  path) contains `-o DPkg::Lock::Timeout=`.
- Confirmed **red**: sourced the pre-fix `preflight.sh` (git HEAD) directly
  and drove `preflight_git` through the same mocked-apt-get harness — the
  recorded invocation was `install -y -q git`, missing the flag, reproducing
  the exact defect from #2125.
- Confirmed **green**: `tests/installer/test_preflight_git.py` — 12/12
  passed after the fix.
- `bash -n` syntax check on `installer/lib/preflight.sh` and
  `installer/install.sh`.
- `ruff check` and `ruff format --check` on `src tests` (repo-wide) — both
  clean.

Also ran the full `tests/installer/` suite. 215 pre-existing failures were
observed there, all unrelated to this change: macOS `._*` AppleDouble
resource-fork files already present (untracked, gitignored) in this worktree
confuse shell-file-scanning tests (`test_ui_helper_contract.py`,
`test_systemctl_dropin_validation.py`, `test_systemctl_quadlet_validation.py`),
and `test_bootstrap_prereq_parity.py::TestAllPresentOnLinux` fails on this
macOS dev host's "hal0 only supports Linux" gate. Neither touches
`preflight_git`/apt-get and neither is affected by this diff.

## Scope

`installer/lib/preflight.sh` and its regression test only, per the fix
wave's isolation rule (this file is also touched by #2084 and #2126 — not
run in the same batch). CHANGELOG.md deliberately left untouched (#1545).

Closes #2125

### DIFF
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -180,8 +180,8 @@ _main_py_autoinstall() {
     info "no Python >=3.${MAIN_PY_MIN_MINOR} found — attempting to install ${cand} (${fam})"
     case "${fam}" in
         debian)
-            DEBIAN_FRONTEND=noninteractive apt-get update -qq >/dev/null 2>&1 || true
-            DEBIAN_FRONTEND=noninteractive apt-get install -y -q "${cand}" "${cand}-venv" >/dev/null 2>&1 || return 1 ;;
+            DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 update -qq >/dev/null 2>&1 || true
+            DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 install -y -q "${cand}" "${cand}-venv" >/dev/null 2>&1 || return 1 ;;
         fedora) "$(pkg_mgr)" install -y "${cand}" >/dev/null 2>&1 || return 1 ;;
         # Arch/openSUSE/Alpine ship a single rolling python; a pinned older
         # minor isn't reliably in the base repos (mirrors
@@ -266,8 +266,8 @@ _hindsight_py_autoinstall() {
         cand="python3.${v}"
         case "${fam}" in
             debian)
-                DEBIAN_FRONTEND=noninteractive apt-get update -qq >/dev/null 2>&1 || true
-                DEBIAN_FRONTEND=noninteractive apt-get install -y -q "${cand}" "${cand}-venv" >/dev/null 2>&1 || continue ;;
+                DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 update -qq >/dev/null 2>&1 || true
+                DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 install -y -q "${cand}" "${cand}-venv" >/dev/null 2>&1 || continue ;;
             fedora) "$(pkg_mgr)" install -y "${cand}" >/dev/null 2>&1 || continue ;;
             # Arch/openSUSE/Alpine ship a single rolling python; a pinned older
             # minor isn't reliably in the base repos. Skip — the metadata-gate
@@ -397,8 +397,8 @@ preflight_venv() {
         local ok=1
         case "$(distro_family)" in
             debian)
-                DEBIAN_FRONTEND=noninteractive apt-get update -qq || true
-                DEBIAN_FRONTEND=noninteractive apt-get install -y -q python3-venv python3-pip || ok=0
+                DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 update -qq || true
+                DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 install -y -q python3-venv python3-pip || ok=0
                 ;;
             fedora) "${pm}" install -y python3 python3-pip || ok=0 ;;
             suse) zypper install -y python3 python3-pip || ok=0 ;;
@@ -678,7 +678,7 @@ preflight_container_runtime() {
         info "installing podman (required for hal0 inference slots)"
         local ok=1
         case "${pm}" in
-            apt-get) DEBIAN_FRONTEND=noninteractive apt-get install -y -q podman || ok=0 ;;
+            apt-get) DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 install -y -q podman || ok=0 ;;
             dnf) dnf install -y podman || ok=0 ;;
             yum) yum install -y podman || ok=0 ;;
             zypper) zypper install -y podman || ok=0 ;;
@@ -755,7 +755,7 @@ preflight_git() {
         info "installing git (required for Hermes agent provisioning)"
         local ok=1
         case "${pm}" in
-            apt-get) DEBIAN_FRONTEND=noninteractive apt-get install -y -q git || ok=0 ;;
+            apt-get) DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 install -y -q git || ok=0 ;;
             dnf) dnf install -y git || ok=0 ;;
             yum) yum install -y git || ok=0 ;;
             zypper) zypper install -y git || ok=0 ;;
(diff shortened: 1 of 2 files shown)


thinmintdev:
## Review — approve

Verified against #2125's repro rather than the PR description. Findings below.

### 1. Does it fix the reported defect? Yes — exact path

The issue names `installer/lib/preflight.sh:758`, the `apt-get` branch of
`preflight_git`, reached with `HAL0_GIT_REQUIRED=1`. That is `installer/install.sh:3609`
(`elif ! HAL0_GIT_REQUIRED=1 preflight_git; then`), the only caller that flips the
required mode, and it is precisely the line the diff changes. Not an adjacent path.

### 2. Root cause, not symptom

The PR's broader claim checks out. On `main`, `installer/lib/preflight.sh` had **zero**
occurrences of `DPkg::Lock::Timeout` across nine `apt-get` lines; every other installer
surface already carries it — `installer/install.sh` (2531, 2560, 2562, 2574, 2637),
`installer/bootstrap.sh` (206-207), `installer/lib/distro.sh:105`,
`installer/agents/hermes-prereqs.sh` (129, 131). preflight.sh was the outlier, exactly as
the issue's "worth a sweep" note suspected. After the diff all nine lines carry the flag
and no `apt-get` remains unguarded in that file. Patching only :758 would have left five
live instances of the same race, so the wider fix is the right call, not scope creep.

### 3. Does the new test fail without the fix? Verified, not trusted

Ran it both ways in a clean worktree of `a5dbb09d`:

- green: `tests/installer/test_preflight_git.py` → 12 passed
- red: `git checkout github/main -- installer/lib/preflight.sh`, same test file →
  `1 failed, 11 passed`, with
  `AssertionError: apt-get invocation missing DPkg::Lock::Timeout: 'install -y -q git\n'`

The recorded invocation in the red run is byte-for-byte the command from the issue.
The test is a genuine regression guard, not a tautology.

### 4. Regressions

The only behaviour change is that these `apt-get` calls now wait up to 120s for the lock
instead of failing instantly. Every changed call site sits behind a `*_REQUIRED=1` gate
(`HAL0_VENV_REQUIRED`, `HAL0_CONTAINER_REQUIRED`, `HAL0_GIT_REQUIRED`) or an autoinstall
helper that only install.sh reaches, so the read-only `hal0 doctor` path
(`src/hal0/cli/doctor_commands.py`, which sources this same file) never reaches an
`apt-get` and cannot newly block. No subprocess timeout wraps preflight.sh from the Python
side — doctor's `timeout=10.0` values are HTTP probes, not this script. `-o` before the
subcommand is the same placement install.sh uses, and an unknown `-o` key is inert on
older apt, so no distro regression. No test anywhere asserts the bare command strings
(checked `tests/`, `docs/`).

Merge order with #2149 is clean: its hunks start at line 1497, above the highest line this
PR touches (1429).

### 5. Scope

Two files, both in scope. No CHANGELOG.md hunk (correct — forbidden this wave). DCO
sign-off present. CI green on all 8 checks.

### 6. Codex bot

No review, no review comments, no issue comments on this PR. Recording that as bot
silence — possibly a usage limit — not as an approval. Nothing to triage.

---

### Nits (non-blocking)

1. **`tests/installer/test_preflight_git.py:168-196` duplicates `test_missing_git_triggers_auto_install_via_apt_get` almost verbatim** — same `d`/`marker` setup, same heredoc apt-get stub, same `_run` call; only the final assertion differs. CONTRIBUTING rule 11 is explicit that a parallel which later drifts is a scar. Lifting the stub into a fixture, or just adding the one assertion to the existing test, would carry the same regression value without the copy.

2. **Only one of the six fixed call sites is guarded.** The PR's own argument is that this is a file-wide convention gap, but the test pins `preflight_git` alone — `_main_py_autoinstall`, `_hindsight_py_autoinstall`, `preflight_venv`, `preflight_container_runtime` and `_node_autoinstall` can silently drift back. A file-level static assertion (every `apt-get` line in preflight.sh carries `-o DPkg::Lock::Timeout=`) would ratchet the convention the change is actually about, and is cheaper than five more mocked-shell tests.

3. **No rationale comment at the new sites.** `installer/install.sh:2524` and `installer/lib/distro.sh:97` both explain why the flag is there (unattended-upgrades on a fresh box). The nine new occurrences carry none, so the next reader has to find the explanation in another file.

4. **`120` is now a bare literal in five files.** distro.sh — already sourced by preflight.sh — owns the flag's formatting in `pkg_install_cmd`; a shared `APT_LOCK_OPT` or a thin `apt_get()` wrapper there would be the one-owner form (rule 1). Pre-existing, and correctly not fixed here.

5. **Two `apt-get` calls outside this file still lack the flag** — `installer/uninstall.sh:661` (`apt-get remove -y fastflowlm`) and `scripts/proxmox-ve/hal0.sh:217-218`. The Proxmox one is the interesting one: it runs `apt-get update` + `install` under `set -e` inside a container that just booted, then `die "apt install failed"` — the same boot-race window #2125 describes, with a harder failure. Out of scope for this PR under the wave's isolation rule; worth a follow-up issue.

PR #2149, opened 2026-08-30
fix(installer): stop the post-install seam probe from calling a transient a broken grant (#2084)

### LINKED ISSUE #2084: post-install seam verify: transient false-negative on hal0-podman-ro right after grant write — same invocation passes moments later; warning also misquotes the command it ran

thinmintdev:
## What

`installer/lib/preflight.sh` — the post-install privileged-seam verification's grant probe now **retries, bounds each attempt, and reports what it observed** instead of taking one silent attempt and naming a cause it never saw.

## Why

rc.10 / ct151 (#2084): the installer warned

```
!!  seam hal0-podman-ro: 'sudo -n hal0-podman-ro check-slot-token hal0probe' failed as the hal0 user — the /etc/sudoers.d/hal0-podman-ro grant does not apply, or the wrapper is stale
```

in the **same log second** as its own `OK wrote /etc/sudoers.d/hal0-podman-ro`. Minutes later, on the untouched box, the identical invocation succeeded — wrapper `root:root 0755`, grant `root 0440`, `visudo`-clean.

The transient is real, but it is not the root cause. The root cause is that `_preflight_seam` let a **single, unretried, evidence-discarding** attempt be the verdict:

1. **The check threw away its own evidence and then asserted a diagnosis.** `sudo … >/dev/null 2>&1` discarded both the rc and the probe's stderr, and the message went on to name *"the grant does not apply, or the wrapper is stale"* — a cause nothing in the check had observed. `sudo: a password is required` (grant genuinely broken) and `hal0-podman-ro: bad slot token` (wrapper genuinely stale) printed **identically**. `src/hal0/system/seam_check.py` has carried the rc + stderr tail since #1465 (`seam_check.py:246-256`); the shell copy had quietly drifted out of the lock-step the file's own header asks for.
2. **One attempt**, so any transient in the write→verify window became a permanent verdict. Every probe verb in `_hal0_seam_probe` is side-effect-free by construction (`help` prints usage; `check` / `check-slot-token` / `check-image-ref` validate and print, never touching podman) — which is exactly what makes re-running one safe.
3. **The message misquoted the command it ran** (nit 1 on the issue): it printed the bare wrapper name while the check ran the full `${bin}` path through two sudo hops. An operator pasting `sudo -n hal0-podman-ro …` gets a *different* failure — the wrapper is not on sudo's `secure_path` — and mis-diagnoses a second time.

Cost of the old behaviour: an operator re-runs `install.sh` for nothing, and a validation agent files a regression that does not exist. #1889 made podman-ro the only source of truth for a running slot's `image_status`/`actual_image`, and kit v6 treats a warned-but-actually-healthy seam as its own finding class, so this warning is acted on.

## How

New in `preflight.sh`, all inside the seam block:

- `_hal0_seam_probe_run` — one attempt, as the hal0 user, **keeping stderr** (stdout discarded) and returning the probe's rc. Split out so the retry/report logic is exercisable in tests with no root, no sudo and no provisioned box.
- `_hal0_seam_grant` — up to `HAL0_SEAM_PROBE_ATTEMPTS` (3) attempts, `HAL0_SEAM_PROBE_DELAY` (1s) apart. A grant that only came up on a later attempt logs `grant verified on attempt 2/3 — the first probe ran before the grant was live, not a fault` rather than passing silently. A grant that never comes up reports **the command that actually ran** (`sudo -n -u hal0 sudo -n /usr/lib/hal0/bin/hal0-podman-ro check-slot-token hal0probe`), the exit code, the attempt count, and the probe's own last stderr line.
- `HAL0_SEAM_PROBE_TIMEOUT` (20s) bounds each attempt — load-bearing now that there are up to three of them, and it mirrors the `timeout=20` `seam_check.py` already passes to `subprocess.run`. `install.sh` is `set -euo pipefail`; the loop is written to be safe under it.

Behaviour for a genuinely broken grant is unchanged apart from the message: required seams still `err` + `die` the install, optional seams still `warn` and continue. Worst-case added wall time on a healthy box is zero (first attempt passes); on a genuinely broken one, 2s of `sleep` plus two extra instant `sudo -n` rejections.

## Verification

`tests/installer/test_seam_verification.py` gains 9 tests, all of which **fail on `main` and pass here** (verified by reverting only `preflight.sh` under them: `9 failed, 13 passed` → `22 passed`). They drive the real `_preflight_seam` past its `stat` checks with two shell-level stubs installed after sourcing (`stat`, because tmp files belong to the test user; `_hal0_seam_probe_run`, because CI has no root/sudo/hal0 user):

- the rc.10/ct151 shape — one failed probe on a healthy seam → `rc=0`, two attempts, no warning;
- the retry is announced, and is *not* announced when the first attempt passed;
- a genuinely broken grant still fails, after exactly `HAL0_SEAM_PROBE_ATTEMPTS` attempts;
- the failure quotes the full two-hop command and **not** the old bare-name form;
- the failure carries the probe's rc and its stderr tail;
- the loop does not abort a caller running `set -euo pipefail` (install.sh's mode);
- required seams get the same retry — the transient is not podman-ro's;
- one attempt is bounded by a timeout, in lock-step with `seam_check.py`.

Checks run:

- `pytest tests/installer/test_seam_verification.py tests/installer/test_podman_ro_validation.py` → **388 passed**
- `pytest tests/installer/` → same 215 pre-existing macOS-environment failures as `main` (AppleDouble `._*.sh` files under `installer/comfyui/`, `stat -c` etc.), `962 passed` on main → `971 passed` here, i.e. +9 and nothing newly red
- `ruff check` / `ruff format --check` on the touched file → clean (repo-wide `ruff check` shows only the 17 pre-existing `packaging/toolbox/**` findings)
- `bash -n installer/lib/preflight.sh` → clean
- the #2081 reporter-scope guard (`test_ui_helper_contract.py`) re-run manually with the AppleDouble files skipped → no offenders; the new code only calls `info` and the existing `"${report}"` indirection

## Deliberately out of scope

- **Reordering the optional-seam probes relative to the service-start phase** (the issue's second suggestion). Bigger blast radius in `install.sh`, and it would only move the window rather than close it — the retry already separates "verified too fast" from "genuinely broken", and the new message tells you which one you got.
- **The Python mirror** (`src/hal0/system/seam_check.py`). It runs long after install (`hal0 doctor`), has no write→verify window, and already reports rc + stderr tail. Read-only context here.
- **`CHANGELOG.md`** — per #1545, one consolidated CHANGELOG PR lands at the end of the wave.

Closes #2084

### DIFF
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1575,7 +1575,13 @@ _hal0_seam_grant() {
     # included. The old text printed the bare wrapper name, so an operator who
     # pasted it hit "command not found" (the wrapper is not on sudo's
     # secure_path) and mis-diagnosed a second time.
-    "${report}" "seam ${name}: 'sudo -n -u hal0 sudo -n ${bin} $*' exited ${probe_rc} after ${attempt} attempt(s) — the ${grant} grant does not apply, or the wrapper is stale${last:+ (${last})}"
+    #
+    # Facts first, then a HEDGED reading — the whole point of #2084 is not to
+    # name a cause we did not observe. Three different bugs land here (grant
+    # broken, wrapper stale, transient that outlasted the retry window) and
+    # this line cannot tell them apart. Keep the wording in lock-step with
+    # src/hal0/system/seam_check.py.
+    "${report}" "seam ${name}: 'sudo -n -u hal0 sudo -n ${bin} $*' exited ${probe_rc} after ${attempt} attempt(s)${last:+ (${last})} — usually the ${grant} grant not applying or a stale wrapper, though a transient outlasting the retry window reports identically"
     return 1
 }
 
--- a/installer/lib/preflight.sh
+++ b/installer/lib/preflight.sh
@@ -1497,6 +1497,88 @@ _hal0_seam_probe() {
     esac
 }
 
+# ── grant probe (#2084) ─────────────────────────────────────────────────────
+#
+# install.sh calls preflight_seams IMMEDIATELY after writing the wrappers and
+# the sudoers drop-ins, on a box whose `hal0` user was itself created minutes
+# earlier. A single failed probe in that window is therefore not evidence that
+# the grant is broken — rc.10/ct151 warned on hal0-podman-ro in the same log
+# second as "wrote /etc/sudoers.d/hal0-podman-ro", and the identical
+# invocation succeeded on the untouched box minutes later. An operator who
+# believes that warning re-runs install.sh for nothing; a validation agent
+# files a regression that does not exist.
+#
+# So: retry, and — the part that actually matters — REPORT WHAT HAPPENED
+# instead of asserting a cause we never observed. The old message threw the
+# probe's stderr away and then named a diagnosis ("the grant does not apply,
+# or the wrapper is stale") it had no evidence for. src/hal0/system/seam_check.py
+# has kept the rc + stderr tail since #1465; this is the shell copy catching
+# up, and the two are meant to stay in lock-step.
+#
+# Every probe verb in _hal0_seam_probe is side-effect-free by construction
+# (`help` prints usage; `check`/`check-slot-token`/`check-image-ref` validate
+# and print), which is exactly what makes re-running one safe.
+HAL0_SEAM_PROBE_ATTEMPTS="${HAL0_SEAM_PROBE_ATTEMPTS:-3}"
+HAL0_SEAM_PROBE_DELAY="${HAL0_SEAM_PROBE_DELAY:-1}"
+# Bounds ONE attempt. Load-bearing now that there are up to three of them: an
+# unbounded sudo that hangs would wedge the installer for good. Mirrors the
+# timeout=20 seam_check.py already passes to subprocess.run.
+HAL0_SEAM_PROBE_TIMEOUT="${HAL0_SEAM_PROBE_TIMEOUT:-20}"
+
+# Run the grant probe ONCE, as the hal0 user. Prints whatever the probe wrote
+# to STDERR on our stdout (its stdout is discarded) and returns the probe's
+# rc. Split out so the retry/report logic below is exercisable in tests
+# without root, without sudo and without a provisioned box.
+_hal0_seam_probe_run() {
+    local bin="$1"; shift
+    # The grant is written for the hal0 user, so the only honest test runs AS
+    # that user. -n keeps both hops non-interactive: a missing grant fails
+    # immediately instead of prompting.
+    local -a cmd=(sudo -n -u hal0 sudo -n "${bin}" "$@")
+    if command -v timeout >/dev/null 2>&1; then
+        cmd=(timeout "${HAL0_SEAM_PROBE_TIMEOUT}" "${cmd[@]}")
+    fi
+    # Order matters: 2>&1 first duplicates the capture pipe onto fd 2, THEN
+    # fd 1 is discarded — so we keep stderr and drop stdout.
+    "${cmd[@]}" 2>&1 >/dev/null
+}
+
+# Prove ONE seam's grant end-to-end, retrying a transient failure.
+# Args: <name> <bin> <grant> <report-fn> <probe word>...
+# Returns 0 when the grant works, 1 once the last attempt has failed.
+_hal0_seam_grant() {
+    local name="$1" bin="$2" grant="$3" report="$4"; shift 4
+    local attempt=0 probe_rc=0 probe_err='' last=''
+
+    while :; do
+        attempt=$(( attempt + 1 ))
+        probe_rc=0
+        probe_err="$(_hal0_seam_probe_run "${bin}" "$@")" || probe_rc=$?
+        (( probe_rc == 0 )) && break
+        (( attempt >= HAL0_SEAM_PROBE_ATTEMPTS )) && break
+        sleep "${HAL0_SEAM_PROBE_DELAY}"
+    done
+
+    if (( probe_rc == 0 )); then
+        if (( attempt > 1 )); then
+            info "seam ${name}: grant verified on attempt ${attempt}/${HAL0_SEAM_PROBE_ATTEMPTS} — the first probe ran before the grant was live, not a fault"
+        fi
+        return 0
+    fi
+
+    # Last non-empty stderr line, capped — the evidence the old message
+    # discarded. `sudo: a password is required` and `hal0-podman-ro: bad slot
+    # token` are entirely different bugs and used to print identically.
+    last="$(printf '%s\n' "${probe_err}" | grep -v '^[[:space:]]*$' | tail -n 1 || true)"
+    last="${last:0:160}"
+    # Quote the command we ACTUALLY ran, full path and both sudo hops
+    # included. The old text printed the bare wrapper name, so an operator who
+    # pasted it hit "command not found" (the wrapper is not on sudo's
+    # secure_path) and mis-diagnosed a second time.
+    "${report}" "seam ${name}: 'sudo -n -u hal0 sudo -n ${bin} $*' exited ${probe_rc} after ${attempt} attempt(s) — the ${grant} grant does not apply, or the wrapper is stale${last:+ (${last})}"
+    return 1
+}
+
 # Verify ONE seam: wrapper present + root-owned + 0755, sudoers drop-in
 # present + root-owned + 0440, and — the fact that actually matters — the
 # grant works when exercised AS the hal0 user. Returns non-zero on any
@@ -1538,13 +1620,7 @@ _preflight_seam() {
         # current one).
         local -a probe_argv=()
         read -r -a probe_argv <<< "${probe}"
-        # The grant is written for the hal0 user, so the only honest test runs
-        # AS that user. -n keeps it non-interactive: a missing grant fails
-        # immediately instead of prompting.
-        if ! sudo -n -u hal0 sudo -n "${bin}" "${probe_argv[@]}" >/dev/null 2>&1; then
-            "${report}" "seam ${name}: 'sudo -n ${name} ${probe}' failed as the hal0 user — the ${grant} grant does not apply, or the wrapper is stale"
-            rc=1
-        fi
+        _hal0_seam_grant "${name}" "${bin}" "${grant}" "${report}" "${probe_argv[@]}" || rc=1
     fi
     return "${rc}"
 }
--- a/src/hal0/system/seam_check.py
+++ b/src/hal0/system/seam_check.py
@@ -42,6 +42,7 @@ than silently implied.
 from __future__ import annotations
 
 import os
+import shlex
 import stat as stat_mod
 import subprocess
 from collections.abc import Callable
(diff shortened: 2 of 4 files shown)


thinmintdev:
Reviewed against #2084's repro, not the PR description. Verified locally on `71007e6e` in a clean worktree.

**What holds up.** The diff changes the exact code the issue quotes (`_preflight_seam`, the `sudo … >/dev/null 2>&1` line), not an adjacent one. The only caller is `installer/install.sh:2103`, so the blast radius is the post-install assertion and nothing else — `preflight_seams` is still out of `preflight_all`, and `hal0 doctor`'s seam surface goes through the Python path, untouched. I confirmed the red/green claim by reverting only `preflight.sh` under the new tests: `9 failed, 13 passed` → `22 passed`. `bash -n` clean, `tests/installer/test_seam_verification.py tests/installer/test_podman_ro_validation.py` → 388 passed. Two files, no `CHANGELOG.md` hunk — clean on the wave rule.

I also checked the two things most likely to be quietly wrong and they are right. The redirect ordering really does keep stderr and drop stdout, exercised against the real function with a fake `sudo` on `PATH`:

```
rc=7
captured=[sudo: a password is required]
```

And the loop is genuinely `set -euo pipefail`-safe with the *real* `_hal0_seam_probe_run` (not just the stub) — `attempt=$(( attempt + 1 ))` rather than `(( attempt++ ))`, and the `… && break` lists don't trip `-e`. I ran a failing probe end-to-end under `set -euo pipefail` and got the full message plus `SURVIVED set -e; grant rc=1`.

---

## Blocking

### 1. The PR's headline behaviour has zero test coverage — mutation-verified

Every one of the nine tests stubs `_hal0_seam_probe_run`, and the stub writes its payload to **stdout**:

```python
if (( n <= {fail_attempts} )); then
    printf '%s\\n' "{stderr_line}"   # stdout, not stderr
```

So `test_the_failure_reports_the_probe_rc_and_its_stderr` never exercises the redirect that makes "report the evidence" work. I flipped the one line that matters:

```diff
-    "${cmd[@]}" 2>&1 >/dev/null
+    "${cmd[@]}" >/dev/null 2>&1
```

**Result: `22 passed`.** The stderr tail silently vanishes from the message, `${last:+ (${last})}` expands to nothing, and the fix degrades back to the pre-fix behaviour with the whole suite green.

This isn't a hypothetical mutation. `>/dev/null 2>&1` is the *exact* idiom on the line being deleted, it's what anyone reading quickly will "correct" it to, and the comment above it exists precisely because the ordering is non-obvious. A fix whose thesis is "stop discarding the evidence" needs one test that proves the evidence isn't discarded.

One test closes it — stub `sudo` on `PATH` instead of stubbing `_hal0_seam_probe_run`, and assert the real function returns the probe's rc, captures its stderr, and drops its stdout. That shape works; it's what I used above.

### 2. Issue nit 1 is only half fixed, on the copy operators actually read

`src/hal0/system/seam_check.py:252-253`:

```python
f"{spec.name}: `sudo -n {spec.name} {' '.join(spec.probe)}` exited "
f"{proc.returncode} as {SERVICE_USER} — the {grant} grant does not apply…"
```

`grant_probe_argv` (`seam_check.py:191-193`) builds `["sudo","-n","-u","hal0","sudo","-n", "/usr/lib/hal0/bin/hal0-podman-ro", …]` under euid 0 — full path, two hops — and the message prints `sudo -n hal0-podman-ro check-slot-token hal0probe`. Bare name, one hop. That is character-for-character the misquote #2084 filed, and `grant_detail` feeds `problems` (`seam_check.py:157`) which lands straight in `hal0 doctor`'s seam row (`doctor_all.py:640-651`).

So the operator flow is: installer warns → operator runs `sudo hal0 doctor` (the remediation the warning points at) → gets the same wrong command to paste → hits `secure_path` → mis-diagnoses. The PR fixes the copy printed once and leaves the copy printed every time.

"Read-only context here" covers the rc/stderr half — the Python side genuinely has had those since #1465. It doesn't cover the misquote, and the PR leans on lock-step with `seam_check.py` as its own justification while the two messages now diverge in wording as well (shell says "does not apply, or the wrapper is stale"; Python says "does not apply"). This is a two-line change plus a test, or an explicit follow-up issue linked here before `Closes #2084` lands — but not silence.

---

## Non-blocking

**The retry budget isn't derived from anything.** 3 attempts × 1s = 2s of separation. The only timing evidence in #2084 is "minutes later, on the untouched box" — there is no observation that this transient clears inside 2s. If it's an nss/sssd negative-cache miss on the freshly-created `hal0` user, the default negative TTL is 15-20s and all three attempts land inside it. I think the reframe of the root cause is defensible and I'd rather have this than main, but the PR presents the retry as *separating* "verified too fast" from "genuinely broken" when it can only separate transients shorter than 2s. Two options, either is fine: derive the budget from a measurement on ct151, or drop the certainty from the final message — it currently reads `— the ${grant} grant does not apply, or the wrapper is stale` with no hint that a >2s transient produces the identical line. Which is, in miniature, the defect being fixed.

**The final message still asserts the cause.** The PR body says the old code "named a diagnosis it had no evidence for" and that the fix "reports what happened instead". The disjunction is retained verbatim; evidence is appended after it. Strictly better, but the body over-claims relative to the diff. Worth noting because the appended `(hal0-podman-ro: bad slot token)` case now reads oddly — the stderr says the wrapper is stale, the prose still offers the grant as an equal possibility.

**`test_one_probe_attempt_is_bounded_by_a_timeout` pins spelling, not behaviour.** It greps the function body for the substring `timeout` and asserts a literal `HAL0_SEAM_PROBE_TIMEOUT="${HAL0_SEAM_PROBE_TIMEOUT:-20}"`. `timeout 0 …` passes it; so does reformatting the default. As a lock-step pin against `seam_check.py` it's reasonable, but it's not the timeout test its name promises.

**`HAL0_SEAM_PROBE_DELAY` is never exercised at a non-zero value** (every test sets it to 0), so a `sleep 30` typo ships green. Minor — the value is a literal two lines from the loop.

---

Verdict: changes requested on (1) and (2). The direction is right and the shell implementation is careful — I went looking for fd-ordering and `set -e` bugs and found neither. The problem is that the suite doesn't know that, and the mirror still ships half the reported defect.

Note on the bot: Codex has posted nothing on this PR — no reviews, no issue comments, no review comments (`gh api …/pulls/2149/comments` → `0`). Nothing to triage, and nothing to read as an approval either.

thinmintdev:
Review findings addressed in 353138a1.

**Blocking 1 — the kept stderr had no coverage.** Correct, and the mutation reproduced exactly as described: flipping `2>&1 >/dev/null` back to `>/dev/null 2>&1` left all nine retry tests green, because every one of them stubs `_hal0_seam_probe_run` itself and the redirection lives inside it.

New test `test_the_probe_keeps_stderr_drops_stdout_and_returns_the_rc` stubs the layer below instead — `sudo` as a real executable on PATH, not a shell function, since the probe may exec through `timeout` — and drives the real `_hal0_seam_probe_run` with a stub that writes a marker to stdout, a marker to stderr, and exits 7. It asserts the rc is the probe's own (7), that stderr is what gets captured, that the stdout marker does *not* appear, and that the argv the stub saw is the two-hop `-n -u hal0 sudo -n /usr/lib/hal0/bin/hal0-podman-ro check-slot-token hal0probe`.

Mutation proof, on this branch:
- redirect flipped to `>/dev/null 2>&1` → `1 failed, 23 passed`, the failure being only the new test (`captured=` empty).
- restored → `24 passed`.

**Blocking 2 — the Python mirror misquoted the command.** Fixed. `seam_check.py` now renders `shlex.join(argv)` — the same argv `grant_probe_argv` built and `subprocess.run` actually executed — so `hal0 doctor`'s seam row quotes the full two-hop command through the wrapper's absolute path instead of the bare `sudo -n <name> <probe>`. Two tests in `tests/system/test_seam_check.py`: one pins the quoted command against `run.calls[0]` and asserts the bare-name form is gone, one pins the wording against `installer/lib/preflight.sh` so the two copies stay in lock-step.

**Nit — overclaimed certainty.** Taken, in both copies. A broken grant, a stale wrapper and a transient outlasting the 3x1s window all print the same line, so neither copy claims to have separated them any more:

```
seam hal0-podman-ro: 'sudo -n -u hal0 sudo -n /usr/lib/hal0/bin/hal0-podman-ro check-slot-token hal0probe' exited 1 after 3 attempt(s) (sudo: a password is required) — usually the /etc/sudoers.d/hal0-podman-ro grant not applying or a stale wrapper, though a transient outlasting the retry window reports identically
```

Python side is the same phrase minus the retry clause (it has no retry loop). Evidence moved ahead of the interpretation in both.

**Nit — exercise `HAL0_SEAM_PROBE_DELAY` non-zero.** Taken: the helper now stubs `sleep` to log its argument and pass through to the real one, and `test_the_configured_delay_is_what_the_loop_actually_sleeps` runs the loop at `0.25` and asserts that value reached `sleep` exactly once. Mutation-checked too — deleting the `sleep` line turns it red. A pure wall-clock assertion would have been vacuous here (each `bash`-source run costs seconds, dwarfing any tolerable delay).

Verification: `tests/installer/test_seam_verification.py tests/system/` → `203 passed`, against a `199 passed` baseline on this branch before the change (+4 new tests). Both runs carry the same one pre-existing local failure, `test_inventory_covers_every_wrapper_install_sh_ships`, from AppleDouble `._*` files in the checkout — not present in CI. `ruff check` and `ruff format --check` clean on the four touched files; `bash -n installer/lib/preflight.sh` clean.

PR #2151, opened 2026-08-30
fix(installer): verify hermes-prereqs' uv install by path, not PATH

### LINKED ISSUE #2124: hermes-prereqs: ensure_uv installs uv to /usr/local/bin then verifies via PATH, so provisioning dies on any caller whose PATH lacks it
Found on a v1.0.0 fresh install of ct151 (Ubuntu 26.04 unprivileged LXC, Python 3.14 only), installed with the public one-liner.

## What happens

`install.sh` runs the hermes toolchain prereqs and dies:

```
[hermes-prereqs] installing uv (managed-Python fallback for hermes-agent on 3.14+ boxes) via pipx → /usr/local/bin
[hermes-prereqs] ERROR: this box has only Python 3.14+ (hermes-agent wheels require <3.14) and
uv could not be installed for the managed-interpreter fallback. Install uv
(https://docs.astral.sh/uv/) or a python3.12, then re-run `hal0 agent install hermes`.
Error: toolchain prerequisites failed — see the output above.
!!  hermes provisioning failed
```

The install otherwise completes and the box comes up healthy — only the agent is missing.

## uv was installed successfully

The error is false. Immediately after the failure:

```
$ ls -l /usr/local/bin/uv
lrwxrwxrwx 1 root root 25 Aug 30 00:41 /usr/local/bin/uv -> /opt/pipx/venvs/uv/bin/uv
$ /opt/pipx/venvs/uv/bin/uv --version
uv 0.12.7 (x86_64-unknown-linux-gnu)
```

The timestamp is from during the install run. `pipx install uv` worked; the verification is what failed.

## Cause

`installer/agents/hermes-prereqs.sh:68-74`:

```sh
ensure_uv() {
    have_uv && { info "uv already present ($(command -v uv))"; return 0; }
    have_pipx || { warn "pipx unavailable — cannot install uv for the managed-Python fallback"; return 1; }
    info "installing uv (managed-Python fallback for hermes-agent on 3.14+ boxes) via pipx → /usr/local/bin"
    PIPX_HOME="${PIPX_HOME:-/opt/pipx}" PIPX_BIN_DIR=/usr/local/bin pipx install uv >/dev/null 2>&1 || true
    have_uv
}
```

It writes to a hardcoded `PIPX_BIN_DIR=/usr/local/bin` but verifies with `have_uv()` → `command -v uv`, which searches `$PATH`. When `/usr/local/bin` is not on `PATH`, the check fails even though the file is there, and `ensure_interpreter_path` calls `die`.

On this box:

```
$ echo $PATH
/sbin:/bin:/usr/sbin:/usr/bin
$ command -v uv        # NOT FOUND
$ cat /etc/environment
PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
```

`/etc/environment` does list `/usr/local/bin` — it is applied by PAM, which `pct exec` (and other non-login execution contexts) bypasses.

## Confirmation

Re-running the exact same command with a login-shell PATH provisions cleanly:

```
$ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
    /usr/lib/hal0/venv/bin/hal0 agent install hermes
...
✓ System service started
hermes-gateway is running (Telegram/Discord)
Installed hermes  (managed venv: /var/lib/hal0/venvs/hermes)
```

Same box, same uv, same everything — only PATH differs.

## Impact

An operator installing over SSH gets `/usr/local/bin` from `/etc/environment` and will not hit this. It bites non-PAM execution contexts: `pct exec`, `lxc-attach`, `docker exec`, cloud-init/`runcmd`, CI provisioning, and anything driving the installer from a service manager. In those environments hermes silently does not get provisioned on any Python-3.14-only box, and the emitted error message tells the operator to do something they have already done.

## Suggested fix

Verify the path that was written rather than asking `PATH` about it — e.g. after the `pipx install`, check `[ -x /usr/local/bin/uv ]` (or capture `PIPX_BIN_DIR` into a variable and test `"$PIPX_BIN_DIR/uv"`), and export that directory onto `PATH` for the remainder of the script so the provisioner that follows can also find it. `have_uv` is fine as the fast path for a box that already has uv on `PATH`; it is only wrong as the post-install verification.

Worth checking whether any sibling helper in that file has the same install-here-verify-via-PATH shape.

thinmintdev:
## What changed

`installer/agents/hermes-prereqs.sh`'s `ensure_uv` installs uv via pipx to a
fixed bin dir (`/usr/local/bin`) but verified the install with `have_uv`
(== `command -v uv`), which asks `$PATH` rather than the location that was
actually written. `ensure_uv` now verifies `[ -x "$UV_PIPX_BIN_DIR/uv" ]` —
the exact path pipx just wrote to — instead of asking PATH. `UV_PIPX_BIN_DIR`
defaults to the historical `/usr/local/bin` and honours an already-exported
`PIPX_BIN_DIR`, so tests can redirect it without touching a real system
directory. On success the bin dir is folded onto `PATH` for the rest of the
script's own run. `have_uv` is untouched and stays correct as the
pre-install fast path (a box that already has uv on PATH needs no reinstall).

**Root cause turned out broader than the issue described** — the exact same
install-here/verify-via-PATH shape lives one call site deeper:
`hermes_provision._uv_available` (consumed by both `_phase_preflight` and
`_provision_python_via_uv`) only asked `shutil.which("uv")`, i.e. *this
Python process's* PATH. The CLI (`hal0.cli.agent_commands._install_hermes`)
runs `hermes-prereqs.sh` as a bare `subprocess.run` and never propagates its
env back, and a root install additionally re-execs the bootstrap through
`runuser`/`setpriv`/`sudo` (whose `secure_path` can strip added PATH entries
regardless of what the bash script exports). So fixing only the bash script
would let the exact same box reproduce the bug one step later inside
`hal0 agent bootstrap hermes`, now surfaced as a more confusing
`RuntimeError: Python 3.12 is unavailable; install python3.12 or uv and
retry`. `_uv_available` now falls back to checking the fixed pipx install
location (`/usr/local/bin/uv`) directly when the PATH probe misses,
independent of whatever PATH actually made it through the CLI → subprocess →
privilege-drop chain.

## Why

#2124: on a non-login exec context — `pct exec`, `lxc-attach`, `docker
exec`, cloud-init/`runcmd`, a service manager — PATH commonly lacks
`/usr/local/bin` because PAM's `/etc/environment` PATH is never applied
there. `pipx install uv` (via `PIPX_BIN_DIR=/usr/local/bin`) succeeds, but
both the bash post-install check and the Python-side `shutil.which("uv")`
ask PATH instead of checking where the file was actually written, so
provisioning falsely dies (or, with only the bash script fixed, fails one
step later with a different, more confusing error) even though uv is
present and working. The reporting agent's repro was on a fresh Ubuntu
26.04 (Python-3.14-only) LXC provisioned non-interactively — exactly this
shape.

## How verified

- `tests/installer/test_hermes_prereqs_uv.py` (new) drives the real script
  end-to-end via `subprocess` with a PATH that excludes the pipx install
  dir. Confirmed red/green: fails against the pre-fix script with the exact
  reported error ("uv could not be installed…") even though a fake pipx
  wrote a working `uv`; passes after the fix. A second test guards that the
  die path stays intact when the install genuinely doesn't produce a usable
  uv.
- `tests/agents/test_hermes_provision.py::TestUvAvailableFallsBackToFixedPipxPath`
  (new) covers the Python-side fallback the same way — confirmed red before
  the fix, green after.
- `tests/installer/test_hermes_prereqs_git.py` (pre-existing) still green —
  no regression to the git-toolchain coverage sharing this script.
- `tests/agents/` + `tests/cli/` (1615 passed) — no regressions from the
  `_uv_available` signature-compatible change.
- `ruff check` / `ruff format --check` on all touched Python — clean.
- `bash -n` on the script — clean.
- CI has no hermes install available, so nothing here shells out to a real
  pipx/uv/hermes-agent; both new test files use fakes on a scoped `PATH`.

## Left out of scope

- Did not touch `installer/install.sh`'s own `preflight_git` gate or
  `agent_commands.py`'s subprocess call — the fix is fully contained to the
  two verify-by-PATH call sites (bash + Python).
- Did not add a PATH-propagation mechanism across the root→hal0 privilege
  drop (`cli/agent_commands._run_as_hal0`) — the fixed-location fallback in
  `_uv_available` makes that unnecessary, since it no longer depends on PATH
  surviving that hop at all.
- Did not touch CHANGELOG.md (per the fix-wave's consolidated-CHANGELOG
  policy, #1545).

Closes #2124.

### DIFF
--- a/installer/agents/hermes-prereqs.sh
+++ b/installer/agents/hermes-prereqs.sh
@@ -62,9 +62,19 @@ have_system_hermes_py() {
 }
 
 # Where pipx drops the uv shim. Hardcoded to /usr/local/bin in production;
-# overridable via an already-exported PIPX_BIN_DIR so tests can redirect it
-# without writing into a real system directory.
-UV_PIPX_BIN_DIR="${PIPX_BIN_DIR:-/usr/local/bin}"
+# overridable via an already-exported HAL0_UV_PIPX_BIN_DIR so tests can
+# redirect it without writing into a real system directory. Deliberately
+# namespaced instead of reading the ambient PIPX_BIN_DIR: that name is
+# pipx's own general-purpose contract (an operator may already export it to
+# steer every pipx install on the box), and reading it here would let it
+# silently move this specific install target while the Python-side fallback
+# (hermes_provision._UV_PIPX_FALLBACK_BIN) stayed hardcoded to
+# /usr/local/bin — reproducing #2124 one phase later with the two halves
+# desynced (#2151 review). ${UV_PIPX_BIN_DIR} is still passed to pipx as its
+# own PIPX_BIN_DIR explicitly on the invocation below, so the install target
+# is controlled by exactly this variable regardless of what else is in the
+# ambient environment.
+UV_PIPX_BIN_DIR="${HAL0_UV_PIPX_BIN_DIR:-/usr/local/bin}"
 
 # Install uv to a fixed location (pipx → ${UV_PIPX_BIN_DIR}) so the
 # provisioner's managed-interpreter fallback
@@ -86,7 +96,7 @@ ensure_uv() {
     # success, also fold ${UV_PIPX_BIN_DIR} onto PATH for the rest of this
     # script's own run, so subsequent `command -v uv` calls here (e.g. the log
     # line in ensure_interpreter_path) see it consistently.
-    if [ -x "${UV_PIPX_BIN_DIR}/uv" ]; then
+    if [ -f "${UV_PIPX_BIN_DIR}/uv" ] && [ -x "${UV_PIPX_BIN_DIR}/uv" ]; then
         case ":${PATH}:" in
             *":${UV_PIPX_BIN_DIR}:"*) ;;
             *) PATH="${UV_PIPX_BIN_DIR}:${PATH}"; export PATH ;;
@@ -112,8 +122,15 @@ uv could not be installed for the managed-interpreter fallback. Install uv
 }
 
 # Toolchain is "complete" only when the interpreter fallback is also satisfiable:
-# venv+pip+pipx+git AND (a system Python in range OR uv for the managed fallback).
-if have_venv && have_pip && have_pipx && have_git && { have_system_hermes_py || have_uv; }; then
+# venv+pip+pipx+git AND (a system Python in range OR uv reachable — either on
+# PATH or already installed at the fixed pipx location, since `have_uv` alone
+# is PATH-based and would false-negative on a non-login exec context where a
+# prior run already installed uv to ${UV_PIPX_BIN_DIR} but PATH doesn't carry
+# it — the same blind spot #2124 fixed for the post-install check in
+# ensure_uv, extended here so the "nothing to do" fast path doesn't redo that
+# work every invocation).
+if have_venv && have_pip && have_pipx && have_git \
+    && { have_system_hermes_py || have_uv || { [ -f "${UV_PIPX_BIN_DIR}/uv" ] && [ -x "${UV_PIPX_BIN_DIR}/uv" ]; }; }; then
     info "toolchain already present (python venv + pip + pipx + git + interpreter path) — nothing to do"
     exit 0
 fi
--- a/installer/agents/hermes-prereqs.sh
+++ b/installer/agents/hermes-prereqs.sh
@@ -61,15 +61,38 @@ have_system_hermes_py() {
     command -v python3.12 >/dev/null 2>&1
 }
 
-# Install uv to a PATH location (pipx → /usr/local/bin) so the provisioner's
-# managed-interpreter fallback (hermes_provision._provision_python_via_uv) can
-# find it via `command -v uv`. Idempotent; requires pipx. Returns non-zero if uv
-# still isn't on PATH afterwards.
+# Where pipx drops the uv shim. Hardcoded to /usr/local/bin in production;
+# overridable via an already-exported PIPX_BIN_DIR so tests can redirect it
+# without writing into a real system directory.
+UV_PIPX_BIN_DIR="${PIPX_BIN_DIR:-/usr/local/bin}"
+
+# Install uv to a fixed location (pipx → ${UV_PIPX_BIN_DIR}) so the
+# provisioner's managed-interpreter fallback
+# (hermes_provision._provision_python_via_uv) can find it. Idempotent;
+# requires pipx. Returns non-zero if uv still isn't usable afterwards.
 ensure_uv() {
     have_uv && { info "uv already present ($(command -v uv))"; return 0; }
     have_pipx || { warn "pipx unavailable — cannot install uv for the managed-Python fallback"; return 1; }
-    info "installing uv (managed-Python fallback for hermes-agent on 3.14+ boxes) via pipx → /usr/local/bin"
-    PIPX_HOME="${PIPX_HOME:-/opt/pipx}" PIPX_BIN_DIR=/usr/local/bin pipx install uv >/dev/null 2>&1 || true
+    info "installing uv (managed-Python fallback for hermes-agent on 3.14+ boxes) via pipx → ${UV_PIPX_BIN_DIR}"
+    PIPX_HOME="${PIPX_HOME:-/opt/pipx}" PIPX_BIN_DIR="${UV_PIPX_BIN_DIR}" pipx install uv >/dev/null 2>&1 || true
+    # Verify the exact path pipx just wrote to — NOT via `command -v`/PATH
+    # (#2124). A non-login exec context (pct exec, lxc-attach, docker exec,
+    # cloud-init/runcmd, a service manager — anything that bypasses PAM's
+    # /etc/environment) can have a PATH that lacks ${UV_PIPX_BIN_DIR} even
+    # though the install landed there just fine, and `have_uv` (== `command -v
+    # uv`) would then falsely report failure on a successful install. `have_uv`
+    # above remains the correct fast path (a box that already has uv on PATH
+    # needs no reinstall) — it is only wrong as the POST-INSTALL check. On
+    # success, also fold ${UV_PIPX_BIN_DIR} onto PATH for the rest of this
+    # script's own run, so subsequent `command -v uv` calls here (e.g. the log
+    # line in ensure_interpreter_path) see it consistently.
+    if [ -x "${UV_PIPX_BIN_DIR}/uv" ]; then
+        case ":${PATH}:" in
+            *":${UV_PIPX_BIN_DIR}:"*) ;;
+            *) PATH="${UV_PIPX_BIN_DIR}:${PATH}"; export PATH ;;
+        esac
+        return 0
+    fi
     have_uv
 }
 
--- a/src/hal0/agents/hermes_provision.py
+++ b/src/hal0/agents/hermes_provision.py
@@ -542,8 +542,30 @@ def _hal0_subprocess_env(**overrides: str) -> dict[str, str]:
     return env
 
 
+# hermes-prereqs.sh installs uv to this fixed location via pipx (see
+# UV_PIPX_BIN_DIR there) specifically so a caller can find it even when its
+# own PATH lacks /usr/local/bin — a real gap on non-login exec contexts (pct
+# exec, lxc-attach, docker exec, cloud-init/runcmd) that bypass PAM's
+# /etc/environment (#2124). This process inherits whatever PATH the CLI
+# invocation started with, and the root→hal0 privilege drop
+# (cli/agent_commands._run_as_hal0, via runuser/setpriv/sudo) can carry that
+# same deficient PATH straight through — `sudo`'s secure_path can even
+# override it. `shutil.which("uv")` alone reproduces, one call site later,
+# the exact "verify an install via inherited PATH instead of the location
+# that was actually written" bug hermes-prereqs.sh had to fix in itself.
+# Falling back to the fixed path here closes the same gap for this
+# Python-side consumer, independent of what PATH actually made it through.
+_UV_PIPX_FALLBACK_BIN = Path("/usr/local/bin/uv")
+
+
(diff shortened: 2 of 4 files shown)


thinmintdev:
## Review — changes requested

The fix is at the right layer and it does fix #2124. Two things must change before merge, both in the "fixed location" contract this PR introduces.

### Does it fix the reported defect? Yes

#2124 cites `hermes-prereqs.sh:68-74` — `ensure_uv` writes to a hardcoded `PIPX_BIN_DIR=/usr/local/bin` and verifies with `have_uv` (`command -v uv`). This diff replaces the post-install verification with `[ -x "${UV_PIPX_BIN_DIR}/uv" ]` on exactly that path, and leaves `have_uv` alone as the pre-install fast path — the reporter's own suggested fix, at the cited lines, not an adjacent one. Root cause, not symptom, and the extension to `hermes_provision._uv_available` is the right call: the CLI runs the script via a bare `subprocess.run` and never propagates the script's `PATH` edit back, so the bash-only fix would have re-surfaced the same bug as `RuntimeError: Python 3.12 is unavailable` one phase later.

### Do the new tests fail without the fix? Verified, not taken on faith

Reverted both source files to `465085b7~1` with the new tests in place:

```
FAILED tests/installer/test_hermes_prereqs_uv.py::TestUvInstallVerifiedByPathNotPATH::test_uv_install_succeeds_even_when_bin_dir_not_on_path
FAILED tests/agents/test_hermes_provision.py::TestUvAvailableFallsBackToFixedPipxPath::...
```

and `test_uv_install_succeeds_even_when_bin_dir_not_on_path` fails with the issue's verbatim symptom (`uv could not be installed for the managed-interpreter fallback`) while the fake `uv` sits on disk. Green on the PR head. No regressions: `tests/agents tests/installer` gives 217 failed / 1769 passed at the parent vs 214 / 1775 at the head on this (macOS, Linux-only-installer-tests) host — the delta is exactly the three new red→green tests, nothing else moves.

---

### Blocking 1 — the hardcoded fallback breaks isolation of an existing test (Codex P2, confirmed)

`_UV_PIPX_FALLBACK_BIN` is a module constant that `_uv_available` consults unconditionally, so an injected `prober` can no longer simulate a uv-less host. The pre-existing `tests/agents/test_hermes_provision.py::test_ensure_python_returns_none_without_uv` passes `prober=lambda _name: None` **with the default `runner=subprocess`** — on any host that has an executable `/usr/local/bin/uv`, it now falls through to the real binary. Reproduced by pointing the constant at a logging shim:

```
[warning] uv-install-dir-perms-failed error="[Errno 13] Permission denied: '/var/lib/hal0'" path=/var/lib/hal0/python
REAL-UV-INVOKED: python install 3.12
REAL-UV-INVOKED: python find 3.12
AssertionError: isolation broken: got '/opt/managed/python3.12'
```

That is a unit test shelling out to a real `uv python install 3.12` (a ~30 MB interpreter download) and trying to `mkdir /var/lib/hal0/python` — and then failing. The hosts where it triggers are not exotic: every box `hermes-prereqs.sh` has ever run on has a `/usr/local/bin/uv` by construction, so the fleet boxes are precisely where the suite breaks.

Fix: make the fallback injectable next to the prober, e.g.

```python
def _uv_available(
    prober: Callable[[str], str | None] = shutil.which,
    *,
    fallback: Path = _UV_PIPX_FALLBACK_BIN,
) -> str | None:
```

and have `_provision_python_via_uv` / the absence tests thread it through (or, minimally, `monkeypatch.setattr(hp, "_UV_PIPX_FALLBACK_BIN", tmp_path / "absent")` in `test_ensure_python_returns_none_without_uv`). Injectable is better — it's the same reason `prober` exists.

### Blocking 2 — `PIPX_BIN_DIR` as a test hook re-opens #2124 in production (Codex P2, confirmed)

```sh
UV_PIPX_BIN_DIR="${PIPX_BIN_DIR:-/usr/local/bin}"
```

Pre-fix, `PIPX_BIN_DIR=/usr/local/bin` was set *on the pipx invocation*, so an operator's ambient `PIPX_BIN_DIR` (a documented pipx variable; `~/.local/bin` is a common export) could not move the target. Post-fix it can — and only the bash half follows it. `_UV_PIPX_FALLBACK_BIN` stays hardcoded `Path("/usr/local/bin/uv")`, so with `PIPX_BIN_DIR` exported the script installs uv, verifies it, prints `toolchain ready`, and then `_uv_available` misses on both PATH and `/usr/local/bin` — the exact false "install python3.12 or uv and retry" this PR exists to kill, one phase later. Nothing in `installer/`, `src/`, `scripts/` or `packaging/` sets `PIPX_BIN_DIR`, so this override exists purely for test hermeticity yet is live in production.

Fix: namespace it so no ambient pipx config can steer it — `UV_PIPX_BIN_DIR="${HAL0_UV_PIPX_BIN_DIR:-/usr/local/bin}"` — and/or have `_uv_available` read the same variable so the two halves cannot desynchronize. The new installer test only needs its env dict updated; the fake pipx still reads the `PIPX_BIN_DIR` the script passes it.

---

### Nits

- **`[ -x ]` accepts a directory where `command -v` did not** (Codex P2 — real, but low-likelihood; nit). Verified: with a directory named `uv` on PATH, `command -v uv` skips it, while `[ -x .../uv ]` and `os.access(..., os.X_OK)` are both true. So the new check is strictly weaker than the one it replaces on that edge, and `pipx install`'s failure is already swallowed by `|| true`. This file's own `have_git` comment (#1727 review) argues for functional probes over resolution checks; either `[ -f "${UV_PIPX_BIN_DIR}/uv" ] && [ -x ... ]` plus `fallback.is_file() and os.access(...)`, or a real `"${UV_PIPX_BIN_DIR}/uv" --version >/dev/null 2>&1`.
- **The completeness gate at `hermes-prereqs.sh:116` still asks PATH.** `{ have_system_hermes_py || have_uv; }` means the boxes from #2124 never hit the "nothing to do" fast path: every later `hal0 agent install hermes` re-runs `apt-get update` and the full package install before reaching the now-correct `ensure_uv`. Correct, just needlessly slow on exactly the affected hosts — `|| [ -x "${UV_PIPX_BIN_DIR}/uv" ]` there would close it.
- **`TestStaticWiring` asserts three verbatim source lines**, including `have_uv() { command -v uv >/dev/null 2>&1; }`. These break on any harmless reformat while the behavioral tests above them already prove the property. Worth dropping at least the `have_uv` text assertion.

### Scope

Clean — four files, no drive-by refactors, no `CHANGELOG.md` hunk (correct for this wave, #1545).

### Codex triage

All three bot comments were verified against the code rather than taken at face value; all three are real. Two are blocking above, one is the `[ -x ]`-on-a-directory nit. None dismissed.

thinmintdev:
Addressed the review findings from this pass:

**Blocking finding 1** (`src/hal0/agents/hermes_provision.py`) — `_UV_PIPX_FALLBACK_BIN` was a non-injectable module constant, so `_uv_available` with an uv-less prober could still fall through to whatever `/usr/local/bin/uv` actually exists on the host running the test. Fixed: `_uv_available` now takes a keyword-only `fallback_bin: Path | None`, threaded through `_provision_python_via_uv` → `_ensure_supported_python` / `resolve_hermes_python`. Updated `test_ensure_python_returns_none_without_uv` to pin `fallback_bin` to a guaranteed-absent tmp path so the assertion holds on a fleet box with a real uv install, not just in CI. Added coverage for the file-vs-directory nit and the new `fallback_bin`/`HAL0_UV_PIPX_BIN_DIR` paths.

**Blocking finding 2** (`installer/agents/hermes-prereqs.sh:67`) — the script read the ambient `PIPX_BIN_DIR` (pipx's own general-purpose env var) to pick its uv install target, so an operator's ambient setting could silently desync the bash install location from the Python-side hardcoded `/usr/local/bin` fallback, reproducing #2124 one phase later. Renamed the override to `HAL0_UV_PIPX_BIN_DIR`; pipx's own `PIPX_BIN_DIR` is still set explicitly on the `pipx install uv` invocation itself, so the actual install target is controlled by exactly this variable regardless of what else is ambient. The Python-side default fallback (`_uv_pipx_fallback_bin`) now honors the same `HAL0_UV_PIPX_BIN_DIR` for symmetry. Added `test_ambient_pipx_bin_dir_does_not_redirect_the_uv_install_target` as the regression guard.

**Nits** — all taken:
- `[ -f ] && [ -x ]` (file-not-directory) checks on both the bash post-install verification and the Python-side `_uv_available` fallback; added a directory-as-fallback test proving the old `os.access` check alone would have wrongly accepted it.
- Extended the "toolchain already complete" fast-path gate to also accept uv already installed at the fixed pipx location (not just PATH), matching the same PATH-blind-spot fix `ensure_uv`'s post-install check already had. Added `TestToolchainCompleteFastPathHonorsFixedUvLocation`.
- Dropped the verbatim `have_uv` source-text assertion from `TestStaticWiring`.

**Verification**
- `.venv/bin/python -m pytest tests/installer/test_hermes_prereqs_uv.py tests/installer/test_hermes_prereqs_git.py tests/agents/test_hermes_provision.py -q` → 195 passed, 1 failed. The 1 failure (`test_hal0_memory_provider_plugin_file_present`, a `UnicodeDecodeError` reading a macOS AppleDouble resource-fork file) reproduces identically on the unmodified branch tip (verified via `git stash`) — pre-existing, unrelated to this change.
- `ruff check` / `ruff format --check` on all touched `.py` files: clean.
- `bash -n installer/agents/hermes-prereqs.sh`: clean.
- CI on this push: all required checks pass (CodeQL, both Analyze jobs, detect, `python (3.12)`, sunset, ui, γ-suite chromium); `engine-gate` shows `skipping` (not applicable to this change).

Not merging — leaving for maintainer review/merge.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01AcGmVXzvn28MJhtZwGd5f3

Issue #1953, opened 2026-08-20
gpu: kfd preflight checks /dev/kfd as the API user, refusing every GPU slot on boxes where the runner (root) can use it fine


thinmintdev:
## Summary

`require_kfd_for_gpu_slot` refuses **every AMD GPU llama.cpp slot** on a box whose GPU is fully functional, because `kfd_present()` evaluates `/dev/kfd` accessibility **as the `hal0-api` service user**, while the slot container actually runs **as root** with the device explicitly passed through.

Found on CT105 (production, Strix Halo) immediately after deploying post-rc.6 code.

## Mechanism

`src/hal0/providers/_gpu.py`:

```python
def kfd_present(kfd_path: str = KFD_DEVICE_PATH) -> bool:
    return os.path.exists(kfd_path) and os.access(kfd_path, os.R_OK | os.W_OK)
```

On CT105:

| Fact | Value |
|---|---|
| `/dev/kfd` | `root:root 660` |
| `/dev/dri/renderD128` | `root:render 660` |
| `hal0-api` unit | `User=hal0` (uid 996, groups: `hal0`, `video(44)`, `render(993)`) |
| `os.access('/dev/kfd', R_OK\|W_OK)` as `hal0` | **False** |
| `os.access('/dev/kfd', R_OK\|W_OK)` as root | **True** |
| `hal0-slot@.service` | no `User=` → runs as **root** (rootful podman) |
| `hal0-slot@1` ExecStart | `--device=/dev/kfd --device=/dev/dri/amdgpu --device=/dev/dri/renderD128` |

So the container has full ROCm access and HIP initialises correctly — ROCm inference on this box has been working all along. Only the *preflight*, running as a different and less-privileged identity than the runner, concludes otherwise.

Because the guard gates `gpu-rocm` **always** and `gpu-vulkan` on any AMD host, the practical effect is that **no AMD GPU llama.cpp slot can load**:

```
GpuPreflightError: slot 'brain' (device=gpu-vulkan) needs the ROCm compute node
/dev/kfd, which is not visible here. ...
```

The remediation the error text suggests (forward `/dev/kfd` from the host via `/etc/pve/lxc/<CTID>.conf` + `pct stop/start`) is a no-op here — the device is already forwarded. It sends the operator to reboot production for a problem that is a group-ownership mismatch.

## Why this is likely to be widespread

[...]

PR #2213, opened 2026-09-02
feat(proxmox): privileged GPU-passthrough LXC by default in the pve script

thinmintdev:
## Why

`scripts/proxmox-ve/hal0.sh` created a vanilla **unprivileged Debian 13** container with no accelerator access, and pointed operators at a hand-written recipe for the shape hal0 actually runs in production. Every step of that recipe is host-side work — `devN` entries, cgroup rules, memlock, apparmor — which an in-guest installer can detect but never perform. So it stayed manual, and the "quick path" and the "real path" had nothing in common.

## What changed

**The script builds the production shape itself.**

- Probes the host for `/dev/dri/renderD*`, `/dev/dri/amdgpu`, `/dev/kfd` and `/dev/accel/*`; forwards each as a `devN` entry carrying the node's gid.
- Writes the matching `lxc.cgroup2.devices.allow` rules (majors derived with `stat -c %t`, not hardcoded), `lxc.prlimit.memlock: unlimited`, and — on a privileged container — `lxc.apparmor.profile: unconfined`. The raw-config writer is idempotent, so a re-run never duplicates a line.
- Devices go on `pct create` itself rather than being appended after a first boot: a container that boots once without `/dev/kfd` gets CPU slots seeded, and that is exactly the failure this is meant to prevent.
- After start, each forwarded node is verified **inside** the guest before anything is installed. A broken passthrough now fails loudly instead of producing a successful-looking CPU-only install.

**Default changes (deliberate, called out in the script header, README and docs):**

| | before | after |
|---|---|---|
| privilege | unprivileged | **privileged** (`UNPRIVILEGED=1` restores) |
| OS | debian 13 | **ubuntu 26.04** |
| features | `nesting=1` | `nesting=1,keyctl=1,fuse=1,mknod=1` |
| GPU/NPU | none | `GPU_PASSTHROUGH=auto` |

A privileged container's root is effectively host root. That is the trade this default makes for direct hardware access, and it is stated plainly everywhere the default appears. `UNPRIVILEGED=1 GPU_PASSTHROUGH=0 OS_TYPE=debian OS_VERSION=13` reproduces the previous behaviour exactly.

**Prerequisites are no longer pre-staged by the script.** `bootstrap.sh` already installs curl/tar/jq/python3 and fetches a pinned, sha256-verified cosign; `install.sh` already auto-installs podman (`preflight_container_runtime`) and the venv stdlib (`preflight_venv`). The script's own `apt-get install` + cosign download duplicated all of it. The single package still staged is `curl` — the tool that downloads the bootstrap cannot be downloaded by the bootstrap.

**In-guest platform classification.** New `hal0_lxc_kind` in `installer/lib/preflight.sh` returns `none` / `lxc-privileged` / `lxc-unprivileged`, read from `/proc/self/uid_map`. The `/dev/kfd` gid-repair fallback (#1953) and `_container_runtime_gate` used to narrate the *unprivileged* remedy unconditionally — telling an operator of a privileged container to change a host `devN:` entry that was never the cause. Both now branch on the real classification.

**New env knobs** for headless/static-IP provisioning: `IP_CIDR`, `GATEWAY`, `NAMESERVER`, `SEARCHDOMAIN`, `TIMEZONE`, `TAGS`, `ONBOOT`, `STARTUP`, `MOUNTS`, `FEATURES`, `GPU_PASSTHROUGH`, `RUN_BOOTSTRAP`.

Docs: `docs/getting-started/proxmox.mdx` §2 now describes what the script does; §3 becomes "what it automates, by hand" and keeps the container-gid caution and the keyring/network/AppArmor notes.

## Testing

- `tests/installer/test_preflight_lxc_kind.py` (new, 6 tests) — the override seam, the `hal0_in_lxc` predicate, an in-vocabulary answer from the real probe, and a guard that the old inline `grep container=lxc` checks were not left behind to drift.
- `shellcheck -x scripts/proxmox-ve/hal0.sh` clean; `bash -n` clean on all three shell files.
- `tests/installer/test_preflight_gpu_gate.py` — 2 failures under a root test runner (`chgrp 0` expectation) reproduce identically on unmodified `main`; pre-existing, not from this change.
- End-to-end on a Strix Halo PVE 9.2 host (privileged Ubuntu 26.04 CT, all four device nodes) is in progress; the run log will be posted as a comment before merge.

https://claude.ai/code/session_01FQujuL1hQGETQAnrgUnyXf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,26 @@ applying. Add those subsections to a version's section to surface them; see
 
 ### Added
 
+- The Proxmox VE quick-start script (`scripts/proxmox-ve/hal0.sh`) now builds
+  the production hal0 container shape instead of a vanilla CPU-only one: a
+  **privileged Ubuntu 26.04** LXC with `nesting/keyctl/fuse/mknod`, every
+  GPU/NPU device node the host exposes forwarded as `devN`, the matching
+  `lxc.cgroup2.devices.allow` rules, `lxc.prlimit.memlock: unlimited`, and
+  `lxc.apparmor.profile: unconfined`. It verifies the forwarded nodes are
+  actually visible inside the container before installing anything, so a
+  broken passthrough fails loudly instead of quietly seeding CPU slots. New
+  env knobs: `IP_CIDR`/`GATEWAY`, `NAMESERVER`, `SEARCHDOMAIN`, `TIMEZONE`,
+  `TAGS`, `ONBOOT`, `STARTUP`, `MOUNTS`, `FEATURES`, `GPU_PASSTHROUGH`,
+  `RUN_BOOTSTRAP`. `UNPRIVILEGED=1 GPU_PASSTHROUGH=0 OS_TYPE=debian
+  OS_VERSION=13` reproduces the previous behaviour. Guest prerequisites are
+  no longer pre-staged by the script — bootstrap and install.sh install
+  their own (curl excepted: it is what fetches the bootstrap).
+- `hal0_lxc_kind` in `installer/lib/preflight.sh` classifies the platform as
+  `none` / `lxc-privileged` / `lxc-unprivileged` from `/proc/self/uid_map`,
+  and the installer's `/dev/kfd` gid-repair fallback and container-runtime
+  gate now narrate the remedy that is actually possible on the container at
+  hand instead of always describing the unprivileged one.
+
 - The dashboard now surfaces the decisive crash line when a slot container
   dies during model load. On a load failure the slot manager tails the
   unit's journal, extracts the one line that names the fault (e.g.
--- a/docs/getting-started/proxmox.mdx
+++ b/docs/getting-started/proxmox.mdx
@@ -1,6 +1,6 @@
 ---
 title: Install on Proxmox
-description: Run hal0 on Proxmox VE — the unprivileged one-liner for CPU-only, and the full GPU/NPU passthrough recipe for AMD Strix Halo LXCs.
+description: Run hal0 on Proxmox VE — the one-liner that creates a privileged Ubuntu LXC with GPU/NPU passthrough already wired, and the manual recipe behind it.
 sidebar:
   order: 13
 ---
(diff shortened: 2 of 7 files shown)


thinmintdev:
End-to-end run on a real box, as promised in the description.

**Target:** PVE 9.2.5 / kernel 7.0.14-8-pve, AMD Strix Halo (Ryzen AI Max+ 395, Radeon 8060S), new CT 220 — privileged Ubuntu 26.04, 28 cores / 96 GB / 500 G, static IP, six bind mounts, stable channel.

**What the script did unattended:**

```
✓  host devices: /dev/dri/renderD128,gid=993 …
✓  template: local:vztmpl/ubuntu-26.04-standard_26.04-1_amd64.tar.zst
✓  LXC 220 created
✓  raw lxc config applied
✓  LXC 220 started
✓  LXC network up
✓  devices visible in LXC
✓  curl available in the LXC
✓  hal0 installed
```

Guest prerequisites resolved themselves, with no pre-staging by the script:

```
» installing missing dependencies (jq) via apt-get
OK  installing the python venv stdlib (required to build hal0's venv)
OK  installing podman (required for hal0 inference slots)
✓ pinned cosign v3.1.2 sha256 OK
```

**The first run failed, and the failure is what the second commit fixes.** `preflight_gpu` hit `BROKEN_GID`:

```
!!    Fix on the Proxmox host: dev0: /dev/dri/renderD128,gid=991
XX  GPU passthrough is broken: the render device is visible but its gid does not map to the render group in this container
```

Host render gid is 993; Ubuntu 26.04's `render` group is 991. On a privileged container there is no idmap to shift it, so the host's number arrived verbatim and hal0 refused — correctly. `align_device_gids` in dbae013 now boots once, reads `getent group render` from the guest, re-points every `devN`, and restarts; `verify_devices` checks the gid the guest reports rather than mere presence.

The same run also produced a junk `lxc.cgroup2.devices.allow: c 0:* rwm` rule, because `/dev/dri/amdgpu` is a **symlink** to the render node on this host and `stat -c %t` on a symlink reports major 0. Symlinks are skipped now. Real majors here: 226 (dri), 509 (kfd), 261 (accel) — worth noting the docs' "234 = kfd" is host-specific, which is why the code derives them instead of hardcoding.

**After the fix, install completes and the box serves on ROCm:**

```
OK  gpu: /dev/dri/renderD128 → group render (gid 991)
OK  flm validate ok — NPU runtime reachable
OK  hal0-api is running
OK  hindsight-api is running (memory engine on 127.0.0.1:9177)
OK  hal0-openwebui is running (chat at :3001)
- ROCm0   : AMD Radeon 8060S Graphics (118784 MiB, 14432 MiB free)
```

`hal0 status` shows 11 slots, brain serving `lfm2.5-2.6b`, and a chat completion round-trips through the gateway.

**One finding filed separately:** #2216 — on this container the slots seeded `gpu-vulkan` despite `/dev/kfd` being present and healthy, because `derive_device`'s ROCm clause needs `platform == "strix-halo"` (unreachable in an LXC) or `rocm-smi` on PATH. Not caused by this PR and not fixed here.

Lint/tests: `shellcheck -x scripts/proxmox-ve/hal0.sh` clean, `bash -n` clean, `tests/installer/test_preflight_lxc_kind.py` 6 passed. `test_preflight_gpu_gate.py`'s two `chgrp 0` failures reproduce identically on unmodified `main` under a root test runner.

PR #2251, opened 2026-09-06
feat(openwebui): wire RAG and ComfyUI image generation from live capability state

thinmintdev:
Open WebUI is now fully pre-wired, not just chat + voice. Document uploads
route through RAG the moment an embed-capable slot is bound
(RAG_EMBEDDING_ENGINE / RAG_OPENAI_API_BASE_URL / RAG_EMBEDDING_MODEL point
at hal0's own /v1), and the image button generates through ComfyUI the
moment the img slot is bound (ENABLE_IMAGE_GENERATION, COMFYUI_BASE_URL and
a COMFYUI_WORKFLOW / COMFYUI_WORKFLOW_NODES pair baked from the SAME
translator /v1/images/generations uses, keyed to whichever checkpoint the
slot is actually bound to).

Both blocks are gated on live capability state and explicitly cleared the
moment that state stops being true — never a stale claim of a capability
that isn't really there. A seam is left for a future search-provider
extension (ENABLE_WEB_SEARCH stays off; no search service ships today).

The env file re-renders — and the companion restarts only when the render
actually changed — on capability apply, embed/img slot create or delete, and
every regular component-convergence pass. The Services page's Open WebUI
card shows a wired chip per feature (Chat, Voice, Documents, Images, Web
search), exposed via GET /api/services.

Signed-off-by: thinmintdev <alexander@awideweb.com>

## Verification

`ruff check src tests` clean. Full α unit suite (12808 tests) green apart from
three failures that also fail on a clean `main` checkout
(`tests/providers/test_moonshine_server.py` ×2, needs `ffmpeg`, and
`tests/slots/test_pressure_eviction.py::test_pressure_evict_noop_when_probe_fails`).
Integration β not run locally (needs root + a live `hal0-lemonade.service`); CI covers it.

https://claude.ai/code/session_01UzkumtSWnm3AxixzHG38UG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,26 @@ applying. Add those subsections to a version's section to surface them; see
 
 ## [Unreleased]
 
+### Added
+
+- **Open WebUI is now fully pre-wired**, not just chat + voice: document
+  uploads route through RAG the moment an embed-capable slot is bound
+  (`RAG_EMBEDDING_ENGINE`/`RAG_OPENAI_API_BASE_URL`/`RAG_EMBEDDING_MODEL`
+  point at hal0's own `/v1`), and the image button generates through
+  ComfyUI the moment the `img` slot is bound (`ENABLE_IMAGE_GENERATION`,
+  `COMFYUI_BASE_URL`, and a `COMFYUI_WORKFLOW`/`COMFYUI_WORKFLOW_NODES`
+  pair baked from the SAME translator `/v1/images/generations` uses, keyed
+  to whichever checkpoint the slot is actually bound to). Both blocks are
+  gated on live capability state and explicitly cleared the moment that
+  state stops being true — never a stale claim of a capability that isn't
+  really there. A seam is left for a future search-provider extension
+  (`ENABLE_WEB_SEARCH` stays off; no search service ships today). The env
+  file re-renders — and the companion restarts only when the render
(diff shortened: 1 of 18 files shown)


thinmintdev:
Auto-merge disabled — holding this out of the v1.3.0 cut pending #2256.

The dynamic-override mechanism here is deliberate and correct for
hal0-managed state: an explicit null is what stops a stale capability claim
surviving a re-render. Two side effects need resolving first.

1. `write_openwebui_env`'s merge deletes an operator's hand-set line rather
   than skipping it — `env_vars.pop(key, None)` cannot distinguish a shipped
   default from a value the operator wrote. Meanwhile `_ENV_HEADER` still
   writes "Hand edits are PRESERVED across install/upgrade runs" into the
   file, which is no longer true for the sixteen keys in
   `_DYNAMIC_ENV_KEYS`.

2. `_search_provider_lookup()` returns `None` unconditionally, so the five
   web-search keys are nulled on every render on every box, and the
   companion restarts because the file changed. Pointing Open WebUI at a
   self-hosted SearXNG is an ordinary configuration, and today it is
   silently reverted on `hal0 update`, `hal0 app converge`, the boot
   convergence pass, a capability apply, and on `embed`/`img` slot create or
   delete.

Nothing is wrong with the RAG and image-generation wiring, which is the bulk
of this PR. Re-enable auto-merge once #2256 is addressed, or split the
web-search block out and land the rest.

